PR #18, opened 2017-08-09
A simple Dockerfile for Syft

TheOriginalAlex:
We can use this Dockerfile as a base image to extend other projects off of.


iamtrask:
@anoff or @kevinahuber could I get your eyes on this guy?

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM alpine:edge
+RUN ["apk", "add", "--no-cache", "python3", "python3-dev", "musl-dev", "linux-headers", "g++", "gmp-dev", "mpfr-dev", "mpc1-dev"]
```
kevinahuber: What are all of the `dev` versions for in this?

TheOriginalAlex: They're necessary to install/build numpy and scipy.

PR #144, opened 2017-08-15
Add quick note on how to run tests

swaroopch:


samsontmr:
Perhaps add a note about which directory to run the command in?

swaroopch:
It can be run from the top-level directory, so I wasn't sure how to word that. 

Should I say "cd PySyft; pytest"?

iamtrask:
that works for me

swaroopch:
@iamtrask Added a commit to add `cd PySyft`

Issue #142, opened 2017-08-15
addmm tests failing


swaroopch:
As of bb1e78015a51de36f58ce5beb74b8f0c120aa131 :

```
$ pytest
===================================================================== test session starts =====================================================================
platform darwin -- Python 3.6.2, pytest-3.2.1, py-1.4.34, pluggy-0.4.0
rootdir: /Users/swaroop/personal/openmined/PySyft, inifile:
collected 33 items                                                                                                                                             

tests/test_math.py .......
tests/test_tensor.py ......................FFFF

========================================================================== FAILURES ===========================================================================
______________________________________________________________________ addmm.testaddmm1d ______________________________________________________________________

self = <PySyft.tests.test_tensor.addmm testMethod=testaddmm1d>

    def testaddmm1d(self):
        t1=TensorBase(np.array([1,2,3]))
        t2=TensorBase(np.array([2,3,4]))
        mat=TensorBase(np.array([5]))
        out=t1.addmm(t2,mat,beta=2,alpha=2)
>       self.assertTrue(np.array_equal(out.data,[50]))
E       AssertionError: False is not true

tests/test_tensor.py:127: AssertionError
______________________________________________________________________ addmm.testaddmm2d ______________________________________________________________________

self = <PySyft.tests.test_tensor.addmm testMethod=testaddmm2d>

    def testaddmm2d(self):
        t1=TensorBase(np.array([[1,2],[1,2]]))
        t2=TensorBase(np.array([[1,2],[1,2]]))
        mat=TensorBase(np.array([[2,3],[3,4]]))
        out=t1.addmm(t2,mat,beta=2,alpha=2)
>       self.assertTrue(np.array_equal(out.data,[[10,18],[12,20]]))
E       AssertionError: False is not true

tests/test_tensor.py:134: AssertionError
_____________________________________________________________________ addmm.testaddmm_1d ______________________________________________________________________

self = <PySyft.tests.test_tensor.addmm testMethod=testaddmm_1d>

    def testaddmm_1d(self):
        t1=TensorBase(np.array([1,2,3]))
        t2=TensorBase(np.array([2,3,4]))
        mat=TensorBase(np.array([5]))
        t1.addmm_(t2,mat,beta=2,alpha=2)
>       self.assertTrue(np.array_equal(t1.data,[50]))
E       AssertionError: False is not true

tests/test_tensor.py:141: AssertionError
_____________________________________________________________________ addmm.testaddmm_2d ______________________________________________________________________

[...]

PR #145, opened 2017-08-16
Add Cumprod & erros fix

radicalrafi:
I fixed few errors that @samsontmr & I discussed I also added cumulative product over axis I also noticed the test for addbm failed but since @bartimaeus12 has already noticed the last issue #142 I'll leave it to him .


erfannoury:
My comments on the previous PR still apply here.

PR #147, opened 2017-08-16
Implement index operator

### LINKED ISSUE #128: Support indexing of TensorBase objects
Indexing in Python is an extremely useful operation supported by major libraries like Numpy and Pandas. Therefore, we would like our TensorBase objects to support this operation as well. This [StackOverflow answer](https://stackoverflow.com/a/509295) contains a good explanation of how it works.

Since we are wrapping `numpy`'s `ndarray` in our TensorBase classes, a way of implementing it would be to pipe the operations through to `numpy`'s implementation.

**Acceptance criteria**
* Must support reverse indexing (e.g. `[-1]`).
* Contains unit tests demonstrating indexing behavior.

swaroopch:
Please let me know if I'm on the right track here :)

Fixes #128

### DIFF
--- a/syft/tensor.py
+++ b/syft/tensor.py
@@ -138,6 +138,13 @@ class TensorBase(object):
         self.data /= tensor.data
         return self
 
+    def __getitem__(self, position):
+        """Get value at a specific index."""
+        if self.encrypted:
+            return NotImplemented
+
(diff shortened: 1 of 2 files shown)


swaroopch:
@iamtrask Makes sense, added a commit.

PR #152, opened 2017-08-16
added openmined slack badge to README 💯

cypherai:


anoff:
I'd rather put this on in the docs repository. Or should we add it to each repo?

cypherai:
I agree with you that having this badge in the docs first make sense. However the good side of having it on each repo is that it would facilitate access to our slack for people arriving from Github.

swaroopch:
But ideally, we should be pointing people to the quickstart guide so that they know what slack channels to join ( see https://github.com/OpenMined/Docs/pull/35 ) instead of landing them directly to slack without guidance.

Also don't new folks need to get an invite before visiting the slack site?

iamtrask:
Yeah @anoff I think you both are quite right. It should be in both.

PR #168, opened 2017-08-21
Implemented exp() and frac()

### LINKED ISSUE #41: Implement Default exp Functionality for Base Tensor Type
**User Story A:** As a Data Scientist using Syft's Base Tensor type, we want to implement a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete, exp() should return a new tensor and exp_() should perform the operation inline. For a reference on the operation this performs check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation.

**Acceptance Criteria:**

### LINKED ISSUE #47: Implement Default frac Functionality for Base Tensor Type
**User Story A:** As a Data Scientist using Syft's Base Tensor type, we want to implement a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete, frac() should return a new tensor and frac_() should perform the operation inline. For a reference on the operation this performs check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation.

**Acceptance Criteria:**

bartimaeus12:
fixes #41 
fixes #47


iamtrask:
@bartimaeus12 FYI, there's a mild chance these tests will fail if someone has an inconsistent version of numpy. In the future let's go for a range when working with very specific floating point numbers (aka... "correct" is within ±0.00001 of a specific number instead of being exactly that). Sometimes different versions can round differently.

bartimaeus12:
@iamtrask thats why i used np.allclose it checks wheter theyare approximately close

PR #171, opened 2017-08-22
Implemented default numpy functionality for BaseTensor Object#79

sajalsubodh22:
1. added function numpy() in tensor.py that casts BaseTensor type to np.ndarray()
2. added unittest for numpy with class name numpyTests in test_tensor.py


bartimaeus12:
@sajalsubodh22  your identation in tensor.py is wrong.(line 433)

sajalsubodh22:
whats the issue now? i am not able to understand
@bartimaeus12

sajalsubodh22:
@iamtrask @bartimaeus12 Please review. I'll make the changes as you suggest.

erfannoury:
I'm against the `numpy` for the function. I can see situations that this will cause conflicts with the main numpy package. I think `to_numpy` would be a better name.

sajalsubodh22:
There were already many implementations of already existing functions like abs() etc. Still i'll correct it.

sajalsubodh22:
@iamtrask what issues are coming up on merging numpy_func to develop??

Issue #174, opened 2017-08-22
Remove `jupyter` dependency


anoff:
The `requirements.txt` lists jupyter as a dependency for `pySyft`. Not sure if this is historic but I would really appreciate it to be removed.

Having to bootstrap the `pySyft` binary in mine.js means all those dependencies need to be installed and jupyter seems to be a bigger one (requiring other system/python libs).

/cc @OpenMined/team-pysyft

PR #175, opened 2017-08-22
Removed jupyter from requirements.txt

TheOriginalAlex:
@anoff

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,6 @@ pyRserve
 numpy
 phe
 line_profiler
-jupyter
 clint
 pytest-flake8
 pytest


anoff:
I can't approve because I have no clue how this repo works, haha but LGTM 😜

cypherai:
Not sure but this could potentially impact the `make run` command https://github.com/OpenMined/PySyft/blob/develop/Makefile#L7 ?

We could add `apk add jupyter` to the docker command before launching the notebook?

iamtrask:
Can someone explain why this is a good idea?

swaroopch:
I think the issue is https://github.com/OpenMined/PySyft/issues/174 but it's still unclear to me, @anoff can you please chime in?

anoff:
@swaroopch @iamtrask yes it is based on that issue. The problem is that `mine.js` needs the `pySyft` command line version to run actual training on the model. However jupyter notebooks need a lot of dependencies that need to be baked into the `mine.js` image only to support the jupyter dependency of pySyft. But pySyft (the library) has no ties to jupyter.

This should be split imho following KISS/separation of concerns principle. Unless pySyft is not considered the library but rather the notebook, then we need a name for the lib :)

cypherai:
@anoff is right, we definitely want to remove jupyter from PySyft depencies. The user who want to run PySyft's notebooks with docker just have to do `make run` (we'll need to to update the command accordingly).

PR #176, opened 2017-08-22
Fix for an indentation error on ceil

EdwardDixon:
The ceiling fell in?


EdwardDixon:
Tiny, tiny issue, but it breaks the intro notebook!

EdwardDixon:
Recent changes on the developer branch fixed the same issue.

PR #184, opened 2017-08-24
Implemented numpy functionality for BaseTensor object#79

sajalsubodh22:
1 . added to_numpy() function in syft/tensor.py
2 . added class numpyTests() in tests/test_tensor.py


sajalsubodh22:
@iamtrask  please review this PR. The previous #171  had many issues so i felt it should be closed.

PR #200, opened 2017-09-01
change Paillier to paillier to be consistant with PEP8 and fix case-s…

### LINKED ISSUE #194: Import error in “Paillier Homomorphic Encryption Example”
The very first line in the file notebook is
`from syft.he.Paillier import KeyPair`
I get an import error. Changing “Paillier” to “paillier” solved the problem. 
Python import has some quirks around case sensitivity on OSX which should be considered. In general “Modules should have short, all-lowercase names. ... ”
https://www.python.org/dev/peps/pep-0008/#package-and-module-names

siarez:
…ensitivity bug
fixes #194


siarez:
Fails CI with this message
screen shot 2017-08-31 at 11 36 28 pm

My local tests (i.e. `pytest && pytest --flake8`) are passed

Issue #125, opened 2017-08-13
Failed to install Syft on MacOs


kidinamoto01:
Hello,
I followed the instructions to install PySyft.
But I failed to get all the dependencies after running 
`pip install -r requirements.txt`

the Error messages are the following:
`Collecting singledispatch (from tornado>=4->notebook->jupyter->-r requirements.txt (line 5))
  Downloading singledispatch-3.4.0.3-py2.py3-none-any.whl
Requirement already satisfied: certifi in /Library/Python/2.7/site-packages (from tornado>=4->notebook->jupyter->-r requirements.txt (line 5))
Collecting backports_abc>=0.4 (from tornado>=4->notebook->jupyter->-r requirements.txt (line 5))
  Downloading backports_abc-0.5-py2.py3-none-any.whl
Requirement already satisfied: functools32; python_version == "2.7" in /Library/Python/2.7/site-packages (from jsonschema!=2.5.0,>=2.4->nbformat->nbconvert->jupyter->-r requirements.txt (line 5))
Collecting webencodings (from html5lib>=0.99999999->bleach->nbconvert->jupyter->-r requirements.txt (line 5))
  Downloading webencodings-0.5.1-py2.py3-none-any.whl
Installing collected packages: gmpy2, phe, decorator, ipython-genutils, traitlets, scandir, pathlib2, wcwidth, prompt-toolkit, pickleshare, appnope, IPython, line-profiler, mistune, jupyter-core, nbformat, configparser, entrypoints, pandocfilters, testpath, webencodings, html5lib, bleach, nbconvert, singledispatch, backports-abc, tornado, python-dateutil, pyzmq, jupyter-client, ipykernel, terminado, notebook, jupyter-console, widgetsnbextension, ipywidgets, qtconsole, jupyter, args, clint
  Running setup.py install for gmpy2 ... error
    Complete output from command /usr/bin/python -u -c "import setuptools, tokenize;__file__='/private/tmp/pip-build-rPx3Nd/gmpy2/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-RaN8cm-record/install-record.txt --single-version-externally-managed --compile:
    running install
    running build
    running build_ext
    building 'gmpy2' extension
    creating build
    creating build/temp.macosx-10.11-intel-2.7
    creating build/temp.macosx-10.11-intel-2.7/src
[...]

PR #203, opened 2017-09-01
[docs] Install base libraries first

swaroopch:
Follow-up to https://github.com/OpenMined/PySyft/issues/125#issuecomment-326004427


kevinahuber:
@swaroopch Good catch! I would make it a link rather than text URL, and not branch specific. 

So
```
Install [base libraries](https://github.com/OpenMined/PySonar#base-libraries)
```
Install [base libraries](https://github.com/OpenMined/PySonar#base-libraries)

PR #212, opened 2017-09-03
feat(size for tensorbase)

aradhyamathur:
- size function
- test for size function


aradhyamathur:
with reference to issue https://github.com/OpenMined/PySyft/issues/96

PR #216, opened 2017-09-05
cumprod added to TensorBase class

### LINKED ISSUE #34: Implement Default cumprod Functionality for Base Tensor Type
**User Story A:** As a Data Scientist using Syft's Base Tensor type, we want to implement a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete, cumprod() should return a new tensor. For a reference on the operation this performs check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation.

**Acceptance Criteria:**

siarez:
Closes #34


siarez:
@iamtrask I'm not sure what to do about the conflicts. Should I resolve them and “commit merge”?

PR #219, opened 2017-09-05
added fmod & fmod_ + generic slicing for tensorbase class  v1

radicalrafi:
After doing some research on slice and objects it seems that overload __getitem ([] operator) can be done using a simple hack since we are using numpy as underlying library for .data member we can pass it directly and wrap it around a TensorBase object when returning it .
This needs to be heavily looked at @swaroopch  @iamtrask @cereallarceny


swaroopch:
@radicalrafi The `__getitem__` change looks good to me 👍

iamtrask:
@radicalrafi looks like some travis CI issues?

PlasmaByte:
Damn, added this myself and just noticed your pull request. Foolish me, still getting the hang of the workflow and git in general.
Did you solve your issues?

PR #223, opened 2017-09-08
Implemented default mean functionality on BaseTensor type #63

sajalsubodh22:
1. implemented mean() function in syft/tensor.py
2. implemented class meanTests() in tests/test_tensor.py


iamtrask:
Will merge after CircleCL and such re-runs (i had to handle merge conflicts with another PR)

PR #239, opened 2017-09-18
feat(mv functionality)

aradhyamathur:
- mv implementation
- test for mv

### DIFF
--- a/syft/tensor.py
+++ b/syft/tensor.py
@@ -1141,15 +1141,20 @@ class TensorBase(object):
     def deserialize(b):
         return pickle.loads(b)
 
+    def mv(self, tensorvector):
+        if self.encrypted:
+            raise NotImplemented
+        return mv(self, tensorvector)
+
 
 def mv(tensormat, tensorvector):
     """ matrix and vector multiplication """
     if tensormat.encrypted or tensorvector.encrypted:
         raise NotImplemented
-    elif not tensorvector.data.shape[0] == 1:
+    elif not len(tensorvector.data.shape) == 1:
         raise ValueError('Vector dimensions not correct {}'.format(tensorvector.data.shape))
-    elif tensorvector.data.shape[1] != tensormat.data.shape[0]:
+    elif tensorvector.data.shape[0] != tensormat.data.shape[1]:
         raise ValueError('vector dimensions {} not  \
             compatible with matrix {} '.format(tensorvector.data.shape, tensormat.data.shape))
     else:
-        return TensorBase(np.matmul(tensorvector.data, tensormat.data))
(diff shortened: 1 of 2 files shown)


aradhyamathur:
with reference to https://github.com/OpenMined/PySyft/issues/69

siarez:
@aradhyamathur Did you mean to change the .gitignore file as well? Or is it a mistake?

aradhyamathur:
@siarez yes i changed it since i switched to pycharm

siarez:
For local files that are specific to your development env. (e.g. files generated by your editor), the best practice is to use the git exclude file. check this out: [Git Exclude](https://help.github.com/articles/ignoring-files/#explicit-repository-excludes)

aradhyamathur:
@siarez corrected it. Removed the .gitignore

siarez:
The way you have implemented `mv`, it is not a method of the TensorBase class. In this case, should it can be implemented in math.py instead? 
@iamtrask what do you think?

aradhyamathur:
I checked Pytorch's documentation, they have implemented it as a part of torch and not Tensor class. So I thought it will be better to follow similar convention. @siarez @iamtrask

iamtrask:
When in doubt, follow PyTorch's conventions :)

siarez:
In PyTorch most function are present in both `torch` and `torch.Tensors` (like `mv` here). We haven't followed that convention here. 
Are we in for a major change? How should things look like, if we were to make that change?

aradhyamathur:
@siarez thanks for bringing that up. I have added the implementation to TensorBase as well I guess we are up to date with it now.

siarez:
@aradhyamathur oh yes! that looks good to me. Now I'm wondering, should we do that for all other methods in TensorBase? @iamtrask what do you think?

iamtrask:
@siarez I think the main question there is based on whether we need to do that to support PyTorch's autograd.

iamtrask:
Example: http://pytorch.org/tutorials/beginner/examples_autograd/two_layer_net_autograd.html

aradhyamathur:
The main role is played by the autograd functions and variable only it seems. Here they have actually explained how custom autograd functions can be created http://pytorch.org/tutorials/beginner/pytorch_with_examples.html#pytorch-defining-new-autograd-functions
The rest of tensor operations don't seem to be affected.

siarez:
I spend sometime this morning looking at PyTorch's autograd trying to figure out how it relates to this. I couldn't figure it out :) Many of their functions leads to their C library and I didn't dig deep enough.
I guess we can leave everything else as is. We'll cross that bridge when we get there.

Issue #11, opened 2017-08-07
Add Additive/Multiplicative Depth Tracking


iamtrask:
NOTE: This ticket requires minimal work on [ISSUE 10](https://github.com/OpenMined/Syft/issues/10) to have been already started. Specifically, all work on this ticket should be over the generic Tensor type, the specs of which are described within that issue.

**Description:** Homomorphic Encryption has a fundamental limitation in that after a certain number of additions and multiplications, noise builds up in the cyphertext and endangers the decrypt-ability of the cyphertext. In order to make our tensors easy to use, we need to provide tracking around noise growth by tracking the number of additions and multiplications involved in creating each particular value in any given tensor. This tracking needs to either automatically bootstrap (remove noise) or flag a warning (error) if the noise is about to render the cyphertext undecryptable. Furthermore, as each encryption technique has slightly different needs in this area, our tensor object needs to implement this functionality in a way that is agnostic to the encryption technique.

Acceptance Criteria:

1. For the Paillier, FV, (and YASHE if it's up in time) algorithms, be able to predict when a number is about to become undecryptable with > 99.99999% recall given arbitrary parameterization of the encryption algorithm. The "how" is not descriptive (be creative!) as long as its efficient. Unit Tests should demonstrate this level of accuracy.
2. Implement the logic of (1) in such a way that the user is not aware of it (abstracted behind the operations)
3. Implement the logic of (1) in such a way that it is agnostic of the HE algorithm (including changing the required functions in the encryption algorithm's abstract interface.
4. Implement 1 of 3 settings as options for when a number is about to become undecryptable:
    a. the number is locally bootstrapped (make this the default if available)
    b. a call is made to an outside server (which you should also build in Syft) where the number is decrypted and then re-encrypted with the same key (make this the first backup after the default)
    c. an error is thrown (as a last resort)
5. All functionality should work given arbitrary encryption parameters (Increasing key sizes can increase the max noise tolerated regardless of the algorithm being used)
6. Complete benchmarking test suite for all abstract tensor operations so that their speed can be automatically reported by running the script with a given encryption algorithm. In this way, we should be able to see the tradeoff between speed and stability (bigger keys == slower but larger multiplicative depth). This should allow us to find the optimal tradeoff between the speed of any operation (which gets faster if you bootstrap more often) and the frequency of bootstrapping (which takes time).

mortendahl:
[...]

PR #245, opened 2017-09-21
Added additive and multiplicative depth tracking #11

dipanshunagar:
This can be used to predict to predict when a number is about to become undecryptable. (#11)
@MtDersvan @iamtrask

### DIFF
--- a/syft/he/paillier/basic.py
+++ b/syft/he/paillier/basic.py
@@ -33,7 +33,9 @@ class PaillierTensor(TensorBase):
         if(type(tensor) == TensorBase):
             tensor = PaillierTensor(self.public_key, tensor.data)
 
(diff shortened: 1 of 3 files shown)


iamtrask:
@dipanshunagar epic!!!

iamtrask:
(if you'll forgive the mild scope creep ;) )

dipanshunagar:
Should I call it just depth then? Could it be more descriptive?

MtDersvan:
Awesome! 
What about other operations like `-` and `/`? I think these need to be tracked, too. (e.g. simple 'y = y/5' is overflowing already after `_div_depth` >≈ 16)
Also, `-` doesn't work (https://github.com/OpenMined/PySyft/pull/271) as of right now.

PR #246, opened 2017-09-21
Update math.py

gautam1858:
adding relu function


iamtrask:
looks like there are some failed tests... @gautam1858 would you be willing to fix those so we can merge? Definitely want to get this piece in soon. :)

PR #248, opened 2017-09-22
Adding docker launch/run to README.md

korymath:
Adds explicitness and clarity to setup for Docker users.


swaroopch:
We could mention 'open -a docker' as well

PR #253, opened 2017-09-25
masked_select method and tests added for the TensorBase class

siarez:
Adds the final feature to issue #60 and closes it

### DIFF
--- a/syft/tensor.py
+++ b/syft/tensor.py
@@ -1251,6 +1251,12 @@ class TensorBase(object):
         self.data[indices] = value
         return self
 
+    def masked_select(self, mask):
+        """
+        See :func:`tensor.masked_select`
+        """
+        return masked_select(self, mask)
+
     def eq(self, t):
         """Returns a new Tensor having boolean True values where an element of the calling tensor is equal to the second Tensor, False otherwise.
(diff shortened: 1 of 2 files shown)


syft/tensor.py
```diff
@@ -1278,3 +1284,21 @@ def mv(tensormat, tensorvector):
...
+
+    :param tensor: Input tensor
+    :param mask: The binary mask (non-zero is treated as true)
+    :return: 1D output tensor
+    """
+    mask = _ensure_tensorbase(mask)
+    tensor = _ensure_tensorbase(tensor)
+    if tensor.encrypted or mask.encrypted:
+        raise NotImplemented
+    mask_broadcasted = np.broadcast_to(mask.data, tensor.data.shape)
```
aiorla: Broadcasting functionality is a bit different than the PyTorch version because this line makes broadcasting tensor => mask fail. Without this step I think it works as intended.

siarez: Can you clarify what you mean?

aiorla: Tensor [1.0] & Mask [1.0,1.0,1.0] are broadcastable but fail in this PR.

siarez: Oh! good catch. I will work on a fix.

PR #254, opened 2017-09-26
Implemented BaseTensor.{le,ge,lt} and inplace versions

### LINKED ISSUE #49: Implement Default ge Functionality for Base Tensor Type
**User Story A:** As a Data Scientist using Syft's Base Tensor type, we want to implement a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete, ge() should return a new tensor and ge_() should perform the operation inline. For a reference on the operation this performs check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation.

**Acceptance Criteria:**

### LINKED ISSUE #56: Implement Default le Functionality for Base Tensor Type
**User Story A:** As a Data Scientist using Syft's Base Tensor type, we want to implement a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete, le() should return a new tensor and le_() should perform the operation inline. For a reference on the operation this performs check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation.

**Acceptance Criteria:**

### LINKED ISSUE #59: Implement Default lt Functionality for Base Tensor Type
**User Story A:** As a Data Scientist using Syft's Base Tensor type, we want to implement a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete, lt() should return a new tensor and lt_() should perform the operation inline. For a reference on the operation this performs check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation.

**Acceptance Criteria:**

baldassarreFe:
- Implemented BaseTensor.{le, ge, lt} and their inplace versions.
- Updated Basetensor.gt() and BaseTensor.gt_() to reflect the new changes.
- Tests for all four inequalities.

Closes #49, closes #56, closes #59

### DIFF
--- a/syft/tensor.py
+++ b/syft/tensor.py
@@ -725,19 +725,64 @@ class TensorBase(object):
         else:
             return [TensorBase(x) for x in np.array_split(self.data, n, dim)]
 
-    def gt(self, t):
+    def gt(self, other):
         """Returns a new Tensor having boolean True values where an element of the calling tensor is greater than the second Tensor, False otherwise.
         The second Tensor can be a number or a tensor whose shape is broadcastable with the calling Tensor."""
-        if self.encrypted:
+        if self.encrypted or (isinstance(other, TensorBase) and other.encrypted):
             return NotImplemented
-        return TensorBase(np.greater(self.data, _ensure_tensorbase(t).data))
+        return TensorBase(np.greater(self.data, _ensure_tensorbase(other).data))
 
-    def gt_(self, t):
+    def gt_(self, other):
         """Writes in-place, boolean True values where an element of the calling tensor is greater than the second Tensor, False otherwise.
         The second Tensor can be a number or a tensor whose shape is broadcastable with the calling Tensor."""
-        if self.encrypted:
+        if self.encrypted or (isinstance(other, TensorBase) and other.encrypted):
             return NotImplemented
-        self.data = np.greater(self.data, _ensure_tensorbase(t).data)
+        self.data = np.greater(self.data, _ensure_tensorbase(other).data)
+        return self
+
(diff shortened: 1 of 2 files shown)


aiorla:
Two small comments:
* Maybe `eq` and `ne` can follow the same structure. (this can be done in another PR)
* What do you think about this pattern? (it avoids double checking `other` class and gives a better error)
```python
other = _ensure_tensorbase(other)
if self.encrypted or other.encrypted:
    return NotImplemented
return TensorBase(np.xx(self.data, other.data))
```

baldassarreFe:
Definitely a nicer way of doing that check! I'll push a new commit in a few minutes!

As for `eq` and `ne`, I agree they can be updated too, but I was waiting for two reasons:
- I didn't want to mess around too much on the first PR (I had already updated the tests for `gt()`, that was not on the list of issues)
- I was waiting for a design decision on the matter "Should we have the equality and inequality logic inside the instance methods or should the instances simply delegate those calls to static methods à la PyTorch?", because in that case we'd move that code somewhere else

Also, a general thought about how things are implemented now. I think that if the goal is to provide an API as similar as possible to PyTorch, we should evaluate how to make our encrypted tensors inherit from PyTorch._BaseTensor without breaking any contract. In this way, we could get rid of the Numpy dependency and delegate all unencrypted operations to PyTorch. The sooner this is done, the smaller the amount of code we'll have to rewrite to reflect the changes.

aiorla:
I am also new here so I'll defer to more experienced members on the design questions. :sweat_smile:
I thought PyTorch was not wanted as a dependence, only as an implementation inspiration. (but I've also read that we will try using its autograd so I may be wrong)

PR #257, opened 2017-09-28
Implemented default BaseTensor.mm functionality.

### LINKED ISSUE #66: Implement Default mm Functionality for Base Tensor Type
**User Story A:** As a Data Scientist using Syft's Base Tensor type, we want to implement a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete, mm() should return a new tensor. For a reference on the operation this performs check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation.

**Acceptance Criteria:**

bharathgs:
- Implemented mm() functionality.
- closes #66


gautam1858:
Hi, there are some blank spaces because of which you build is getting failed which is a style guide build error, even I had the same issue, please check for missing whitespace/blank space errors in your build

PR #260, opened 2017-09-29
Implemented Diag Functionality for Base Tensor Type with tests

tusharsoni08:
@iamtrask please review it.


bharathgs:
The function should return a new tensor. Please refer to #36 . Also, the function should be in math.py.

bharathgs:
1. @tusharsoni08, quick question:  why is this returning a np.ndarray?
2. @iamtrask should this not be in syft/math.py?
3. and I think it's not doing the right thing, as in I cannot control which diagonal to choose, as in [this](http://pytorch.org/docs/master/torch.html#torch.diag)

PR #264, opened 2017-09-30
BMM functionality - issue #28

AkashGutha:
issue #28 

Features looking at:
1. if any of the tensors is encrypted return NotImplemented error.
2. Unit tests to check shape returned tensor shape.
3. Unit tests to check the tensor value returned.
4. Inline documentation should take inspiration from pytorch but not copy it (which currently i did).

### DIFF
--- a/syft/__init__.py
+++ b/syft/__init__.py
@@ -15,7 +15,7 @@ s += str(mpc)
 
 s += str(equal) + str(TensorBase) + str(cumprod) + str(cumsum) + str(ceil)
 s += str(dot) + str(matmul) + str(addmm) + str(addcmul) + str(addcdiv)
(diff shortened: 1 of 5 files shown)


AkashGutha:
this is ready to go. 👯‍♂️

AkashGutha:
@bharathgs that fixed them all.

AkashGutha:
@bharathgs the way we are merging seems a bit odd. 

ideally, you would want all the commits to be there. so that you can go through each change in the future to get a better understanding of what's really going on.

bharathgs:
We follow the standard git workflow, which you can refer [here](https://github.com/OpenMined/Docs/blob/master/tutorials/github_contribution_workflow_tutorial.md#squashing-commits)

AkashGutha:
That makes sense. 👍

PR #265, opened 2017-09-30
Changing Test function names from camelCase to snake_case

AkashGutha:
This PR is to change camel case function test names to use _ 
This increases readability.

Currently finished changing tests in test_math.py

### DIFF
--- a/tests/test_integrations_locally.py
+++ b/tests/test_integrations_locally.py
@@ -9,7 +9,7 @@ from syft.he.paillier import KeyPair, PaillierTensor
 
 class PySonarNotebooks(unittest.TestCase):
 
-    def model_Training_DemoNotebook(self):
+    def model_training_demo_notebook(self):
         """If this test fails, you probably broke the demo notebook located at
         PySonar/notebooks/Sonar - Decentralized Model Training Simulation
         (local blockchain).ipynb """
@@ -23,7 +23,7 @@ class PySonarNotebooks(unittest.TestCase):
 
 class PySyftNotebooks(unittest.TestCase):
 
-    def paillier_HEExampleNotebook(self):
+    def paillier_HE_example_notebook(self):
         """If this test fails, you probably broke the demo notebook located at
         PySyft/notebooks/Syft - Paillier Homomorphic Encryption Example.ipynb
         """
@@ -60,7 +60,7 @@ class PySyftNotebooks(unittest.TestCase):
         out7 = prikey.decrypt(y2)
         self.assertEqual(out7, np.array([.5, .5, .5, .5, .5]))
 
-    def paillier_LinearClassifierNotebook(self):
+    def test_paillier_linear_classifier_notebook(self):
         """If this test fails, you probably broke the demo notebook located at
         PySyft/notebooks/Syft - Paillier Homomorphic Encryption Example.ipynb
         """
--- a/tests/test_integrations_locally.py
+++ b/tests/test_integrations_locally.py
@@ -9,7 +9,7 @@ from syft.he.paillier import KeyPair, PaillierTensor
 
 class PySonarNotebooks(unittest.TestCase):
 
-    def modelTrainingDemoNotebook(self):
+    def model_Training_DemoNotebook(self):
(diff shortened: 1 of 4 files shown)


AkashGutha:
Yea. that would be great.

tests/test_math.py
```diff
@@ -73,33 +73,33 @@ def testSigmoid(self):
...
+    def test_mat_mul_2d_int(self):
         t1 = TensorBase(np.array([[1, 2],
                                   [3, 4]]))
         t2 = TensorBase(np.array([[5, 6],
                                   [7, 8]]))
         self.assertTrue(syft.equal(syft.matmul(t1, t2), [[19, 22],
                                                          [43, 50]]))
 
-    def testMatmul2DFloat(self):
+    def test_mat_mul_2d_float(self):
```
aiorla: :+1: to the idea. Just a tiny comment, I think the MatmulTests should be `test_matmul_X` and not `test_mat_mul_X`.

AkashGutha: yes. I agree 👍

baldassarreFe:
I agree on the style change 👍
I've been following the previous style for consistency but I think we should conform to the generally accepted Python guidelines, it will help people jumping on board.

tests/test_integrations_locally.py
```diff
@@ -9,7 +9,7 @@
 
 class PySonarNotebooks(unittest.TestCase):
 
-    def modelTrainingDemoNotebook(self):
+    def model_Training_DemoNotebook(self):
```
kevinahuber: why not `model_training_demo_notebook`?

tests/test_integrations_locally.py
```diff
@@ -23,7 +23,7 @@ def modelTrainingDemoNotebook(self):
 
 class PySyftNotebooks(unittest.TestCase):
 
-    def paillierHEExampleNotebook(self):
+    def paillier_HEExampleNotebook(self):
```
kevinahuber: same, `paillier_he_example_notebook`

kevinahuber:
I _believe_ we can enforce this by having `pep8` run on the test files themselves, I'll play around with that today.

tests/test_integrations_locally.py
AkashGutha: I didn't check. I thought it was a file name. so i left it in caps.

tests/test_integrations_locally.py
AkashGutha: same. I'll check and redo these

tests/test_integrations_locally.py
```diff
@@ -23,7 +23,7 @@ def model_Training_DemoNotebook(self):
 
 class PySyftNotebooks(unittest.TestCase):
 
-    def paillier_HEExampleNotebook(self):
+    def paillier_HE_example_notebook(self):
```
AkashGutha: I'm leaving HE capitalized since it is an abbreviation. 
share any other thoughts on this

iamtrask:
I assume we'll keep waiting on merging this until all the test files have been changed?

AkashGutha:
yes. I'll change the rest too. THen we can merge everything at once.

tests/test_tensor.py
```diff
@@ -561,6 +561,8 @@ def test_chunk_same_size(self):
 
 
 class inequalityTest(unittest.TestCase):
+    # this is an ovverriden function from unittest.Testcase.
```
AkashGutha: Suggest a better way of writing this warning.

should we be doing something like 
```WARNING : message```

File: tests/test_tensor.py [hunk 6]
```diff
@@ -360,7 +360,7 @@ def test_baddbmm_(self):
 
 
 class PermuteTests(unittest.TestCase):
-    def test_3d(self):
+    def dest3d(self):
```
AkashGutha: didn't really understand. why changing name didn't work here.
but reverting to the original name worked. so I'm going to leave it as it is.

AkashGutha:
Almost. Done. we can move this once we have it reviewed and fix any remaining issues

Issue #100, opened 2017-08-10
Implement Default stride Functionality for Base Tensor Type


iamtrask:
**User Story A:** As a Data Scientist using Syft's Base Tensor type, I want to leverage a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete, stride() should return a tuple. For a reference on the operation this performs check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation.

**Acceptance Criteria:**
- If the Base Tensor type's attribute "encrypted" is set to True, it should return a NotImplemented error.
- a unit test demonstrating the correct operation on the Base Tensor type implemented over int and float Tensors.
- inline documentation in the python code. For inspiration on inline documentation, please check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation for this operator.

Ujjwal-9:
I will do it.

Ujjwal-9:
@iamtrask according to pytorch docs stride() returns a tuple not a tensor.
you can check [here](http://pytorch.org/docs/master/tensors.html#torch.Tensor.stride)

iamtrask:
ahhh yes!

iamtrask:
let's follow that convention :)

Ujjwal-9:
@iamtrask then could you edit the issue :smile:

iamtrask:
done ;)

vbankov:
i would like to pick this one up. @Ujjwal-9 do you have any update?

Ujjwal-9:
@vbankov its done. PR is in review process

vbankov:
@Ujjwal-9 no worries :), feel free to copy any of my tests if you'd like

Issue #256, opened 2017-09-27
PySyft design conventions discussion


siarez:
I created this issue as a discussion board about design. So please chip in you views.
A lot of great code has come in recently, but I think we lack a solid design philosophy about what code should go where. And that has resulted in inconsistent code organization, and could get worse if not addressed. For example:
1. There are some functions in syft/math.py that are not implemented in syft/tensor.py(i.e. [cumsum](https://github.com/OpenMined/PySyft/blob/master/syft/math.py#L121) ) and vice versa (i.e. [rsqrt](https://github.com/OpenMined/PySyft/blob/master/syft/tensor.py#L616) ).
2. There are methods in syft/math.py that are not necessarily math operations (i.e. [transpose and unsqueeze](https://github.com/OpenMined/PySyft/blob/master/syft/math.py#L309))
3. There are functions in tensor.py like [mv](https://github.com/OpenMined/PySyft/blob/master/syft/tensor.py#L1329) which feel like they belong to math.py but are implemented in tensor.py

I think of some things that have bearing on how we define our conventions are:
1. How the future of encrypted tensors will look like? What method do they inherit from TensorBase? I expect methods like transpose, reshape, basically anything without math operation will work for encrypted tensors as well, but don't know much about cryptography. So it would be great to hear from someone with experience there.
2. Also should encrypted tensors implement their own math functions? If yes, do they get their own math.py files? 
3. If we want to use PyTorch's Autograd, what bearing does that have on our design?
4. I think following PyTorch's conventions are good, but we should also think about how they mix with our encrypted tensors.

I think after we come to a consensus about design, we should:
1. Update our documentation to be more clear for contributors
2. Update current issues so not to confuse new contributors. I know it was a cause of confusion for me.

Please share your thoughts.

iamtrask:
1. I agree that this will generally be the case in the short run. The only exception will be for a thing called "packing" where instead of encrypting each scalar individually, and entire vector is encrypted as a unit. While this makes it more challenging to index into specific values, it can increase performance significantly in the right setting.

2. This is a great question. In many cases, yes they should. This is particularly true for cases wherein there are different ways to accomplish the same thing. For example, when multiplying a vector by a scalar, one can either encrypt the scalar and perform multiplication or perform multiplication with the scalar decrypted. The latter is much faster. This kind of logic we want to bake into the Encrypted tensor (and is likely to be slightly different for each variant).

[...]

Issue #252, opened 2017-09-25
Inline Documentation Consistency


iamtrask:
User Story: As a regular Contributor to OpenMined's PySyft repository, I want documentation to follow a consistent format to ease my use of the platform. Presently, this is not the case.

Acceptance Criteria:
• a defined template for inline documentation
• modifying PySyft repositories to conform to this template (particularly math.py and tensor.py which are highly irregular)

CT83:
Assuming Documentation like that of equal(TensorBase,TensorBase) in tensor.py is expected. I am on  it !

drewvolpe:
@CT83 I would recommend we use an existing style for docstrings. The two most popular ones are Google style and Numpy style. Both are well-supported by Sphinx and tools like PyLint and Jedi. Given we're using Numpy in various parts of the code, I would vote for NumPy style.

http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

iamtrask:
@CT83 very excited about this!!!!

CT83:
I don't know if you guys know, but you guys were featured here! https://youtu.be/HAC6sqq7_-U?t=7m13s 
I was interested, excited and wanted to help.

CT83:
I don't really know how this works but I worked on one of the two files and I am done; 
https://github.com/OpenMined/PySyft/pull/273 Let me know if what I did was, what was expected and I will do the math.py File too. If not I am open to Advice  :)

momonala:
Hey! I'm new to Openmined and this seems like an issue I can tackle. How can I help?

CT83:
@iamtrask @drewvolpe https://github.com/OpenMined/PySyft/pull/273 I have completed tensor.py should I start with math.py? in the same way?

bharathgs:
@CT83 *tensor.py* looks great, I think you can start off with *math.py*.

Issue #93, opened 2017-08-10
Implement Default set Functionality for Base Tensor Type


iamtrask:
**User Story A:** As a Data Scientist using Syft's Base Tensor type, I want to leverage a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete, set_() should operate inline. For a reference on the operation this performs check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation.

**Acceptance Criteria:**
- If the Base Tensor type's attribute "encrypted" is set to True, it should return a NotImplemented error.
- a unit test demonstrating the correct operation on the Base Tensor type implemented over int and float Tensors.
- inline documentation in the python code. For inspiration on inline documentation, please check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation for this operator.

sp4ke:
I will work on this one.

sp4ke:
I have a few questions before moving forward. 

What does **should operate inline** mean ?

The [pytorch signature](http://pytorch.org/docs/master/tensors.html#torch.Tensor.set_) for this function is `set_(source=None, storage_offset=0, size=None, stride=None)` and the **source** parameter can be either a Tensor, or a Storage which is custom made by an underlying C extension

However since our tensor's internal storage is an ndarray the source is always going to be a Tensor right ? which means storage_offset and size are useless, or am I missing something ?

sp4ke:
ping @iamtrask

sp4ke:
@iamtrask can you give a bit more details how this method is supposed to behave ? There are many ways to implement this and I'm not sure if the signature needs to be the same as PyTorch

PR #267, opened 2017-09-30
Changed weights initialization & added stride functionality #100

Ujjwal-9:
Initialization of all weights to zero mathematically leads to either the neuron values are zero (for multi layers) or the delta would be zero.


Ujjwal-9:
@iamtrask pls look into this PR.

syft/tensor.py
```diff
@@ -1372,6 +1374,30 @@ def mm(self, tensor2):
 
         return syft.mm(self, tensor2)
 
+    def stride(self, dim=None):
+        """Returns the jump necessary to go from one element to the next one in the specified dimension dim."""
```
AkashGutha: Please make the documentation consistent with #252

you can take a look at this link 
https://github.com/OpenMined/PySyft/blob/4e2aa9e34df985f4d8e0703d900eaa2558fd413a/syft/math.py#L259

Ujjwal-9:
The normal vs uniform init seem to be rather unclear. But i have changed my initialization to uniform. 
@trask look into it.

syft/tensor.py
```diff
@@ -1372,6 +1374,30 @@ def mm(self, tensor2):
...
+            out = tuple(out)
+            return out
+        if dim is None:
+            out = calculation(ndim, shp)
+            return out
+        elif ((dim >= 0) and (dim < ndim)) or ((dim < 0) and (dim >= -ndim)):
+            out = calculation(ndim, shp)
+            return out[dim]
+        else:
+            raise RuntimeError("dimension out of range of {}D tensor".format(ndim))
```
AkashGutha: Runtime error!! I guess not.
We can do a simple `Error` or better `ValueError` which suits this situation.
Look into these. `RuntimeError` is definitely not the one we should be using.

AkashGutha:
This PR fixes #100 
adding #100 in the title doesn't actually create reference adding it here will do.

syft/tensor.py
Ujjwal-9: Pytorch used Runtime error for this issue.

syft/tensor.py
```diff
@@ -1397,6 +1399,35 @@ def mm(self, tensor2):
 
         return syft.mm(self, tensor2)
 
+    def stride(self, dim=None):
+        """
+        Returns the jump necessary to go from one element to the next one in the specified dimension dim.
+
+        :param dim: dimension
```
AkashGutha: This still hasn't resonated with the expected style of documentation.
use Parameters and Results and seperators
https://github.com/OpenMined/PySyft/blob/4e2aa9e34df985f4d8e0703d900eaa2558fd413a/syft/math.py#L259

aiorla:
Thank you @Ujjwal-9! :+1: 

First of all I'd like to say that I prefer different PR's for unrelated changes to improve its reviewing.
* Weights initialization
    *  I'm with you that having zero-initialization as the default can be sometimes a bit worrying.
    * Not sure about this normal vs uniform. Commonly, e.g. [PyTorch](http://pytorch.org/docs/master/_modules/torch/nn/modules/linear.html#Linear) or [keras](https://keras.io/initializers/), the weights are initialized with a normal/uniform with a std/limit dependent on the number of features. I'd recommend following again PyTorch example here (although it's not that important).
* `stride` functionality
    * I think that your function does not provide the same functionality than PyTorch (where only a tuple is returned and no parameter "dim" is needed). If your version is a function that does what PyTorch does + extra functionalities with the optional parameter `dim` please tell me and I'll look it more carefully.

Ujjwal-9:
@aiorla if you look at definition of stride in pytorch source code you will find that they provide `dim` parameter. Actually dim is not necessary parameter. checkout stride() definition [here](https://github.com/pytorch/pytorch/blob/312e0ce3ba1fa74854a1fc82aa1b7c24baeb4006/torch/lib/TH/generic/THTensor.c#L28)
You can also try this code:
```
import torch
a = torch.Tensor([[1,2,3,4,5],[6,7,8,9,10]])
a.stride()
a.stride(0)
a.stride(1)
a.stride(-1)
a.stride(-2)
```

aiorla:
It looks like PyTorch [documentation](http://pytorch.org/docs/master/tensors.html#torch.Tensor.stride) is incomplete then, sorry @Ujjwal-9... :sweat_smile:
I'll give it another look later then (if nobody does it before me). :+1:

Ujjwal-9:
@aiorla I checked Numpy implementation but it gives different results. 
And sorry i didn't write unit test. will write it asap. :sweat_smile:

aiorla:
@Ujjwal-9 numpy implementation gives the strides in bytes, so a division by the element size is needed. Does it have more differences? (I haven't thoroughly checked)

sp4ke:
@Ujjwal-9 @aiorla   I am currently working on the storage features from #93 and it requires the `strides` features following PyTorch signature. Since it is already implemented here, should I rebase over this PR ?

sp4ke:
Forgot one question on the `strides` implementation, why not reuse the underlying strides from the ndarray in `self.data` ?

aiorla:
Hi @sp4ke! I talked last week with @Ujjwal-9 on Slack and he told me that he was going to use `strides()` & add the unit test but he might have been busy.

I'm not sure what you should do there (if you don't want to wait for the fix). If I were you I'd maybe implement an inlined version of strides inside `set_()` and add a TODO to use `strides()` when available... (but that may not be the "cleanest" way :sweat_smile:)

sp4ke:
Ha ! I will do that at least so I can progress. I think by the time my PR would be approved strides will be available. 

By the way do you know any document that describes the "architecture" and API of the pysyft library in regard to PyTorch and Numpy. Right now it seems a bit messy, some methods reuse the internal ndarray features and some others re-implement them

bharathgs:
@sp4ke as of now, I think there's no architecture as such. however, there are a few of these inconsistencies which are being discussed in issue #256 which needs to be addressed.

bharathgs:
@Ujjwal-9 any updates on this?

Ujjwal-9:
@bharathgs its complete now you can review it.
changes:
1.  Changed my custom stride function to numpy implementation
2.  Added Tests
3.  Changed weights initialization in linear.py

syft/tensor.py
```diff
@@ -3331,6 +3331,31 @@ def renorm_(self, p, dim, maxnorm):
...
+        Returns the jump necessary to go from one element to the next one in the specified dimension dim.
+
+        Parameters
+        ----------
+        dim : dimension
+            The first operand in the stride operation
+
+        Returns
+        -------
+        Tuple
```
bharathgs: A tuple is returned when there's no Argument, If not an int is returned.
so maybe you can change the doc to reflect that?

Ujjwal-9:
Made the changes

Issue #258, opened 2017-09-28
Well-defined Docker images and Make targets


baldassarreFe:
At present, there is a bit of confusion between the Docker image and the Makefile. In short: 
- the regular `Dockerfile` installs dev dependencies
- the `Development-Dockerfile` installs all that the regular does (note: the install commands are duplicated, we are not using the FROM command), plus Jupyter
- the `run` target in the Makefile relies on the openmined/pysyft image from DockerHub, but once the container is up, it tries to start a notebook server and fails
- the `custom` target does the exact same things as the `run`, with the difference that it accepts an image as a parameter, so that one can build his own dev image based on `Development-Dockerfile` (with Jupyter) and make it work

From my point of view (close-to-zero experience with this kind of deployments), we should have two images on DockerHub:
- **openmined/pysyft** that ships without dev dependencies (e.g. no gmp-dev and mpfr-dev) and with a working installation of PySyft, based either on a PyPi package or simply installed via `git checkout tags/working-release && python3 setup.py install` (this is to develop some application _with_ PySyft)
- **openmined/pysyft-dev** that has all the additional dependencies and uses the `src` folder as source for the PySyft modules (this is to develop PySyft itself)

Regarding the Makefile, I suggest keeping the test and install targets docker agnostic, meaning that if they are run on the local machine they use what's available locally and if they are used inside the container they use the container:
```
install: 
    python3 setup.py install

test: 
   pytest && pytest --flake8

image = openmined/pysyft
docker: 
    docker run --it $(image) ...

image-dev = openmined/pysyft-dev
docker-dev: 
    docker run --it $(image-dev) ... jupyter notebook
```

Examples:
- run tests locally: `make install && make test`
- run tests in docker `make docker-dev make test`
- build a different image and use it with docker: `docker build -f MyDockerfile -t MyImage && make docker image=MyImage`

Provided that this structure will need refinement and additional details, does it make sense in general? 
Can someone with experience suggest improvements?

swaroopch:
Thanks for explaining the problem so clearly. I agree with most of your recommendations, one nit is to have simpler commands than `make docker-dev make test` which is confusing. We can follow different terminology like we do in https://github.com/OpenMined/PyYashe/blob/master/Makefile or we can use a Makefile rule such as:

(untested)
```makefile
docker_%:
  docker run -i -t openmined/pyyashe make $(shell echo $@ | sed s/^v//) $(MAKEFLAGS)"
```

[...]

PR #270, opened 2017-09-30
Update of make targets, dockerfiles and readme

baldassarreFe:
This is a proposal for how the make targets, the docker files and the readme could look like.

The new versions of those files are inspired by #258, and try to separate the concept of building/testing/installing PySyft from the location (local/docker) where this is done.

It's not perfect in many ways, please comment and improve ;)


baldassarreFe:
Thanks for the review, I agree on all three points: 
1. I didn't try installing on windows and I don't know the problems and the solutions, maybe we should summon someone from Slack to add his experience 
2. There is a Notebooks session after the Setup, for now it simply tells where to find the notebooks, but it could be expanded with a list of notebooks and their content
3. I can move the relevant section in its own file, but then wouldn't it be better to have a "curated list of resources" in the Docs repository and link to that? 

Call for help: 
- Can someone from CI set up the automatic Docker build of pysyft and pysyft-dev from the two dockerfiles on master? 
- Can someone with python packaging experience comment on using `setup.py install` vs `setup.py develop` to have the installed package linked to the sources in the local folder?

Dockerfile
```diff
@@ -1,10 +1,13 @@
...
+  g++ linux-headers \
+  musl-dev gmp-dev mpfr-dev mpc1-dev \
+  ca-certificates openblas-dev \
+  make gfortran
 
 WORKDIR /PySyft
-RUN ["pip3", "install", "-r", "requirements.txt"]
-COPY . /PySyft
-RUN ["python3", "setup.py", "install"]
+COPY . .
```
swaroopch: what does this do?

README.md
```diff
@@ -5,86 +5,100 @@
...
 
-## Setup
+## Setup with Docker
 
-### Prerequisites
+The fastest way to get started is to use the pre-assembled Docker image (works on all major operating systems). Alternatively, one can manually set up all the dependencies and develop locally.
+
+### Get Docker
+
+Install Docker following the instructions on the [website](https://www.docker.com/). (For macOS users with [Homebrew](https://brew.sh/) installed, use `brew cask install docker`)
```
swaroopch: `brew cask install docker; open -a docker`

Makefile
```diff
@@ -1,8 +1,21 @@
...
+install:
+	python3 setup.py install
+
+test: install
+	pytest && pytest --flake8
+
+notebook:
+	jupyter notebook --allow-root --ip=0.0.0.0
+
+dockerfile = Dockerfile
```
swaroopch: does this get overridden by make command parameters?

README.md
```diff
@@ -5,86 +5,100 @@
...
-### Installation
+Now, start a container based on [openmined/pysyft](https://hub.docker.com/r/openmined/pysyft/):
+```sh
+make docker-run
+```
 
-The recommended method is using Docker (works on all major operating systems).
+If you want to use Jupyter and try the provided notebooks, create a development image:
+```sh
+make docker-build dockerfile=Development-Dockerfile
```
swaroopch: certainly seems better than what we have currently!

sp4ke:
The issue with docker environment was discussed on slack on it seems better to get rid of the local installation concept. I'm working on a PR that will make it much easier for newcomers to start developing right away just after installing Docker. I will link this issue on my PR

Dockerfile
baldassarreFe: That copies the contents of the current folder (basically the git repo itself) from the docker build context to the image.

README.md
baldassarreFe: Sure, I work on Linux so I don't know these tricks.

Makefile
baldassarreFe: Yes, variables in the makefile can be overridden, let me google [this](https://www.gnu.org/software/make/manual/html_node/Overriding.html) for you

baldassarreFe:
@sp4ke why a new PR instead of contributing here? Anyway, check my comments on your PR ;)

sp4ke:
@baldassarreFe I closed my PR . Once merged it should make things less messy and improvements easier.

iamtrask:
This is looking quite good... although as I'm still a bit new on Docker... do others agree this is ready to Merge?

PR #271, opened 2017-10-01
Changed PaillierTensor `__sub__` to avoid 'AttributeError'

MtDersvan:
Changed `__sub__` analogically to `__add__` to avoid 
```
AttributeError: 'EncryptedNumber' object has no attribute 'data'.
```
If previous behaviour is intended - please feel free to reject the PR.


siarez:
Good catch!

PR #272, opened 2017-10-01
Feature/tensor  lerp

chadlohrli:


iamtrask:
also some style errors :)

iamtrask:
@bharathgs does it not do the lerp similarly to pytorch? it looks quite similar.

bharathgs:
@iamtrask  `start` & `end` -> tensors, `weight` -> scalar, the current code does not operate on tensors and also check the inputs that this is taking in.

alberduris:
Now there's another PR with these features implemented.

https://github.com/OpenMined/PySyft/pull/308

PR #273, opened 2017-10-01
#252 Inline Documentation Consistency 

CT83:
This is one of my first pull requests. So Please feel free to tell me where I am wrong; I would love to get some feedback

I followed the NumPy Style on inline documentation. http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

It's a bit rough around the edges but let me know if I am on the right track

PR #275, opened 2017-10-01
Implements Default half Functionality #52

bharathgs:


AkashGutha:
is there no inplace version for this one?

bharathgs:
In line with [this](http://pytorch.org/docs/master/tensors.html#torch.Tensor.half) and #52 . We do not have an *in_place* version of `half()`.

aiorla:
I like it @bharathgs :+1:. Could you change the inline doc to be like #273?

bharathgs:
@aiorla thanks for pointing out that. We should adhere to this strictly.

syft/tensor.py
```diff
@@ -1348,7 +1348,16 @@ def mm(self, tensor2):
         return syft.mm(self, tensor2)
 
     def half(self):
-        """cast the tensor to half-precision float type."""
+                """
```
AkashGutha: spacing needs dome fix here.

AkashGutha: ignore this. The issue was fixed. My lazy eye missed the last commit.

syft/tensor.py
```diff
@@ -1353,6 +1353,23 @@ def mm(self, tensor2):
 
         return syft.mm(self, tensor2)
 
+    def half(self):
+        """
+        casts the tensor to half-precision float type
+
+        ----------
```
AkashGutha: There are more separators than required. please check that

AkashGutha: this thing is also in the numpy example. not sure why.

syft/tensor.py
```diff
@@ -1397,6 +1397,23 @@ def mm(self, tensor2):
...
```
siarez: To adhere to the docstring format in #252, it should say 
```
Parameters
----------
```
even when there are no parameters

PR #282, opened 2017-10-03
Add minibatching in "learn" API

meetvora:
The PR attempts to add minibatch in learn API and makes use of Openmined's Capsule to allow for secure re-encryption of the weights at defined intervals.


meetvora:
Hi!
I'm trying to update the travis.yml to get the capsule service running.

Issue #12, opened 2017-08-07
Abstract Fixed Point Precision


iamtrask:
**Description:** All of our Homomorphic Encryption algorithms operate on Integers under the hood. However, we want to be able to perform math on numbers of varying precision (most popularly, Floats). Thus, we need someone to put together a generic Fixed Point Precision wrapper that can provide this functionality around encryption algorithms that are Integers such that the following types are supported:

- Byte (8-bit float)
- Half (16-bit float)
- Float (32-bit float)
- Double (64-bit float)
- Arbitrary (n-bit float)

- Char (8-bit int)
- Short (16-bit int)
- Int (32-bit int)
- Long (64-bit int)
- Arbitrary (n-bit int)

Acceptance Criteria:
- All types above are supported assuming 32-bit integer as input.
- All basic operations are supported (+-*/ and inline versions).
- Exponents can be decrypted
- Unit Tests show that it works with all types above given Integers underneath
- Unit Tests show that it works in the encrypted domain (using Paillier is fine)

samsontmr:
Split this into sub-tasks for each type?

iamtrask:
Maybe, however the best way to build it is probably to build the "Arbitrary" one first and then just instantiate the rest of them using specific configs of the arbitrary?

siarez:
How do you guys generally see this implemented? I was imagining each encrypted tensor have its own `self.precision`. Then when a new Tensor is created from an array then numbers in the array are multiplied by `10**self.precision`  and divided by `10**self.precision` after decryption. For example, the `encrypt` method in [pailier/keys.py](https://github.com/OpenMined/PySyft/blob/master/syft/he/paillier/keys.py#L67)  would get a new line `x *= 10 ** precision`:
```
    def encrypt(self, x, same_type=False, precision=4):
        """Encrypts x. X can be either an encrypted int or a numpy
        vector/matrix/tensor."""
        x *= 10 ** precision
        ...
        return self.pk.encrypt(x)
```
Then all math operation for `PailierTensor` should be modified to take precision into account. Does this sound right? or is there a better way to do this?

iamtrask:
Man, great question. I've spent a couple hours yesterday working through a simple version of this just for the Multi-party-computation prototype and I learned a few things about how we might like this to work. First, I found it's important to separate two pieces of configuration.

[...]

PR #288, opened 2017-10-04
Feature/paillier fixed point functionality  fp for paillier tensors

siarez:
…Modifies operation to work with the new `FixedPoint` class.

This party addresses issue #12. By changing the underlying datatype of `PaillierTensor` to `FixedPoint` and letting it accept different fixed point configurations upon instantiation. The goal is for all other Paillier tensor types (i.e. `PaillierFloatTensor`, `PaillierLongTensor`, etc) to derive from `PaillierTensor`

@iamtrask Let me know if this implementation (so far) is what you had in mind, and what needs to change. Definitely let me know if you think this is totally the wrong direction :)

### DIFF
--- a/syft/he/paillier/FixedPoint.py
+++ b/syft/he/paillier/FixedPoint.py
@@ -101,19 +101,16 @@ class FXfamily(object):
         self.scale = 1 << n_bits
         self._roundup = 1 << (n_bits - 1)
 
-        try:
-            thresh = 1 << (n_bits + n_intbits - 1)
-
-            def validate(scaledval):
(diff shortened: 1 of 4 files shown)


iamtrask:
Hey! this is definitely the right direction. Do you think you could also create several "pre-set" extensions of Fixed Point for the common sizes (Float, Int, Long, Double, Binary, etc.) and then let people have different kinds of PaillierTensor objects?

siarez:
Great! 
Yes will do! I wanted to share what I have so far, before going down this path too far :)

PR #290, opened 2017-10-04
only docker needed as dependency 

sp4ke:
As a user who wants to test or contribute on the project, I only need to install Docker as a dependency. 
All requirements are included in the Docker images.

Development workflow for a new user is as follows:

- Pull repo
- Install Docker
- Run `make run` starts a notebook using `pytest-dev:edge` image and local notebook source.
- Run `make test` starts the test suite on the local source code 

Docker image details:
- If the image is run without parameters it will automatically launch jupyter notebook on the root repo directory.
- There are two docker images:
  - Dockerfile --> pysyft
  - Development-Dockerfile --> pysyft-dev
- `pysyft-dev` inherits from `pysyft` image and includes extra development requirements 

- To compile a local docker image run `make create-custom-image`
- To use the local image run `make run-custom`

Closes  #270

### DIFF
--- a/Development-Dockerfile
+++ b/Development-Dockerfile
@@ -1,11 +1,23 @@
-FROM alpine:edge
-RUN ["apk", "add", "--no-cache", "python3", "python3-dev", "musl-dev", "linux-headers", "g++", "gmp-dev", "mpfr-dev", "mpc1-dev", "ca-certificates", "openblas-dev", "gfortran"]
+FROM openmined/pysyft:edge
 
-RUN ["mkdir", "/PySyft"]
-COPY requirements.txt /PySyft
+RUN apk add --no-cache git
 
-WORKDIR /PySyft
-RUN ["pip3", "install", "-r", "requirements.txt"]
-COPY . /PySyft
-RUN ["python3", "setup.py", "install"]
-RUN ["pip3", "install", "jupyter"]
+
+ENV PYSYFT_DIR /src/PySyft
+ENV CAPSULE_DIR /src/Capsule
+
+
+## Capsule dependency
+
+RUN mkdir $CAPSULE_DIR
+WORKDIR $CAPSULE_DIR
+RUN git clone https://github.com/OpenMined/Capsule.git .
+RUN pip3 install -r requirements.txt
+RUN python3 setup.py install
+
+## PySyft Dependency
+WORKDIR $PYSYFT_DIR
+
+EXPOSE 8888
+
+CMD ["jupyter", "notebook", "--ip=0.0.0.0", "--allow-root"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,21 @@
 FROM alpine:edge
-RUN ["apk", "add", "--no-cache", "python3", "python3-dev", "musl-dev", "linux-headers", "g++", "gmp-dev", "mpfr-dev", "mpc1-dev", "ca-certificates", "openblas-dev", "gfortran"]
+RUN apk add --no-cache python3 python3-dev \
+                    musl-dev linux-headers g++ \
+                    gmp-dev mpfr-dev mpc1-dev \
+                    ca-certificates openblas-dev gfortran
 
-RUN ["mkdir", "/PySyft"]
-COPY requirements.txt /PySyft
+RUN mkdir /src
 
-WORKDIR /PySyft
-RUN ["pip3", "install", "-r", "requirements.txt"]
-COPY . /PySyft
-RUN ["python3", "setup.py", "install"]
+ENV PYSYFT_DIR /src/PySyft
+
+## PySyft Dependency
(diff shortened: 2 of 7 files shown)


bharathgs:
This looks great, makes the installation process a lot easier for new-comers.
but just a few thoughts, so pysyft is inspired by pytorch, and if you look at the pytorch installation process [here](https://github.com/pytorch/pytorch#installation) as explained in their *README*
It has clear instructions for docker/binaries/source. 

I think with this updated docker installation process, we should also keep the source/binary installation process explanation as well. however, we should probably make it much clearer, possibly by writing a note with the common installation problems that people might face.(mostly with windows ? ) but we clearly point out that docker is the **recommended** method of installation. what do you guys think? @sp4ke @iamtrask @swaroopch

sp4ke:
@bharathgs I added an **Advanced** section in the readme for those who want to install from source locally. Most users will end up using the docker way since it's the most straightforward way to get everything running especially on Windows. For those who want to install locally they will generally feel comfortable reading the Advanced section. 

As for the common issues, I'm not sure the README.md is the right place for it. We could either create a Wiki or some "common-issues.md" file and redirect to it from the readme.

bharathgs:
@sp4ke I totally agree with you, I think your point on having a *common-issues.md* is a good idea. probably, this should be extended to other projects, as some people are having issues with those as well.
so having a shared *common-issues.md* document is great.
and what about [this](https://github.com/OpenMined/PySyft#for-anaconda-users) ? (this has been smooth process so far) can't we keep this in the main *README*?

baldassarreFe:
Hi, looks like we are working on the same issue!

A general comment first: I bet you checked #270 already, so why not adding to that discussion rather than opening another PR yourself? From my experience, this only leads to confusion and double work.

I agree with the effort of limiting the onboarding struggle as much as possible, but I also believe that we should not force people to use Docker to check our work. For non-experienced users, the overhead of installing _also_ Docker can be enough to make them give up.<br/>
I think the best way is to conceptually split the Makefile in two sections:
```
# Platform independent commands, doesn't matter if you run them locally or inside a docker image
install: ...
develop: ...
test: ...
start notebook: ...

# Docker-related stuff
build prod image: ...
run prod image and open a shell: ...
build dev image: ...
run dev image and open a shell: ...
```

The flow for a new user is:
1. Where do I want to work?
  - **Docker** -> pull the image and run it
  - **Locally** -> install the libraries I need
2. What do I want to do?
  - **notebooks** ...
  - **test** ...
  - the targets in the makefile are the same, everything just works, regardless of where they are run

Also, as people pointed out, it's not ok to have a bulky installation for people that want to import the libary as a dependency, so `sklearn` and `jupyter` can not be in the requirements (it's also referenced from `setup.py`), there was an additional requirements file for that purpose.

Finally, have you checked the discussion on slack regarding the difference between `setup.py install` and `setup.py develop`? Because the former makes a copy of the current directory in your python package folder, preventing the people who want to work on the source to see their changes "live". The latter instead uses symlinks, and people can develop and test their changes.

sp4ke:
@baldassarreFe Regarding your issue agree sorry I discovered it a bit late after working on it. 

> I agree with the effort of limiting the onboarding struggle as much as possible, but I also believe that we should not force people to use Docker to check our work. For non-experienced users, the overhead of installing also Docker can be enough to make them give up.

Yes but realistically many people will join from different OS platforms and just want to have a quick and running environment asap. Plus for non-experienced users it's quite the opposite,  Docker will generally be the easiest and fastest since you have only "one" dependency, compared to the many to install in local. 

So describing both Docker and Local installation in the main readme.md makes it much more confusing for the majority especially that many people who are joining are just interested in have a running environment for testing or developing. On the other hand for local installations, many have heterogeneous python environments/platforms, for example I personally don't use anaconda but PyEnv on Arch Linux, so even the required Libraries described are not named the same way in Arch. If we're going to cover the local installation path on the README it's going to get quite complicated.

I see PySyft users will mainly have two paths: 
- using as library 
- or development 

I'm starting to think best would be to have a development.md and keep the library as default in the readme.md, this way would could keep updating the development.d as needed without breaking changes to the library


> I think the best way is to conceptually split the Makefile in two sections

Agree 

> Also, as people pointed out, it's not ok to have a bulky installation for people that want to import the libary as a dependency, so sklearn and jupyter can not be in the requirements (it's also referenced from setup.py), there was an additional requirements file for that purpose.

Indeed totally makes sense, jupyter and sklearn will only be installed in the Docker image then.

> Finally, have you checked the discussion on slack regarding the difference between setup.py install and setup.py develop? Because the former makes a copy of the current directory in your python package folder, preventing the people who want to work on the source to see their changes "live". The latter instead uses symlinks, and people can develop and test their changes.

No missed that discussion I will update the docker file accordingly

sp4ke:
Closing this issue and continuing discussion in #258

PR #296, opened 2017-10-06
Implementation of index methods in TensorBase #54

### LINKED ISSUE #54: Implement Default index Functionality for Base Tensor Type
**User Story A:** As a Data Scientist using Syft's Base Tensor type, we want to implement a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete, index() and index_select() should return a new tensor and index_add_(), index_copy_(), and index_fill_() should perform the operation inline. For a reference on the operation this performs check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation.

**Acceptance Criteria:**

samuxiii:
Resolve #54  
New methods:  index(), index_add_(),  index_copy_(), index_fill_()

### DIFF
--- a/syft/tensor.py
+++ b/syft/tensor.py
@@ -2848,20 +2848,18 @@ class TensorBase(object):
         self.data = np.remainder(self.data, divisor)
         return self
 
-    def index(self, position):
+    def index(self, m):
         """
         Returns a new Tensor with the element selected by position
 
-        :param index: integer position
+        :param m: integer index or slice
         :return: tensor of selected indices
         """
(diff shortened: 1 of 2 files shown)


samuxiii:
Ok, I've added slices into the index as a possible type of parameter. But.. I don't understand the comment about index_add_, I've almost used the same pytorch example as my unit test.
Please, could you be more explicit? Thanks

bharathgs:
@samuxiii  ahh, my bad.did not see the `index_add_` properly. It looks good.

Issue #78, opened 2017-08-10
Implement Default numel Functionality for Base Tensor Type


iamtrask:
**User Story A:** As a Data Scientist using Syft's Base Tensor type, I want to leverage a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete, numel() should return a new tensor. For a reference on the operation this performs check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation.

**Acceptance Criteria:**
- If the Base Tensor type's attribute "encrypted" is set to True, it should return a NotImplemented error.
- a unit test demonstrating the correct operation on the Base Tensor type implemented over int and float Tensors.
- inline documentation in the python code. For inspiration on inline documentation, please check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation for this operator.

baldassarreFe:
Shouldn't it be possible to compute the number of elements in a tensor even if it is encrypted?
For instance, in [this notebook](https://github.com/OpenMined/PySyft/blob/master/notebooks/Syft%20-%20Paillier%20Encrypted%20Linear%20Classification.ipynb) one can easily count the number of `[e]`s in the encrypted tensors.

Also: [`torch.numel()`](http://pytorch.org/docs/master/torch.html#torch.numel) returns an integer, while the user story says `numel()` should return a tensor. Which one should PySyft implementation follow?

iamtrask:
@baldassarreFe all your points are correct :)

alberduris:
I would like to work on that, cause it seems to be pretty simple. Anyway, I'm not very confident yet, so I'll try but if anyone wants to work on that too, please, feel free to do it :)

alberduris:
Now I have some doubts for @iamtrask, @baldassarreFe or anyone that could help.

If, as @baldassarreFe has suggested, `numel()` will return an integer, and the post is `Returns the total number of elements in the Tensor.` What's the point of implementing `numel` if we already have `nelements`  and `size`?

`size: Size of tensor`
`Returns the total number of elements in the tensor.`

If it's just for having the operation named `numel`, as PyTorch, shouldn't we implement the count for encrypted tensors in `size` and `nelement` too as @baldassarreFe has also suggested?

joewie:
`nelement` is just an alias for `numel` in PyTorch, so I suppose we have both to conform with PyTorch's methods.

[...]

PR #300, opened 2017-10-07
Implement Default numel Functionality for Base Tensor Type #78

alberduris:
Added numel method, modified nelement and size.


joewie:
Disclaimer: I'm not a reviewer.

Why keep the `if...else` statements for `numel`, `nelement`, and `size`?

Also, there are unit tests for `size` and `nelement` in `class DimTests` that you might want to remove since you wrote more comprehensive ones.

alberduris:
@joewie 

> Why keep the if...else statements for numel, nelement, and size?

I kept the `if...else` statements because I'm not sure if it's possible that in the future, the distinction between encrypted and not encrypted will be necessary (I suppose no but not sure), so to maintain the structure that I found.

> Also, there are unit tests for size and nelement in class DimTests that you might want to remove since you wrote more comprehensive ones.

Okay, I'll take a look then. Thanks!

Anyway, the Travis CI build failed and I don't know exactly what to do to solve it.

alberduris:
I know that this PR is marked with the red cross but I think it should be reviewed, because the red cross is just because of the CodeCoverage and as @ibrahimsharaf pointed out in Slack:
> a coverage report is generated now and any new pull request shouldn't give codecov errors
So if it's an old PR you can ignore the codecov failure (edited)
but it won't fail in future PRs

ping @iamtrask, reviewers

ibrahimsharaf:
More details regarding codecov failure: https://docs.codecov.io/docs/error-reference#section-missing-base-commit

PR #302, opened 2017-10-07
Request: Fixing Some Methods in Tensor.py

nahuakang:
After communicating with @bharathgs about `tensor.py`, I fixed a bug in `sqrt()` method which should return a `TensorBase` but is returning Numpy `np.sqrt(self.data)`.

Going through `tensor.py`, I also found out that some in-place methods do not return anything. The corresponding methods in PyTorch usually return `self`, therefore I adjusted the documentation for some methods and added `return self` when appropriate. The corresponding unit tests in `test_tensor.py` are also fixed when necessary.

Please let me know if these adjustments are appropriate or not and I can keep improving the pull request. Thank you for your time and help!

List of methods adjusted:
`sqrt`: return `Tensor` instead of Numpy
`sqrt_`: return `self`
`abs`: return `Tensor`
`abs_`: return `self`
`zero_`: return `self` instead of `self.data`
`transpose_`: return `self`; raise `ValueError` instead of just `print`
`t_`: return `self`
`unsqueeze_`: return `self`; raise `ValueError` instead of just `print`
`rsqrt_`: return `self`
`reciprocal_`: return `self`
`resize_`: return `self` instead of just `print`
`resize_as_`: return `self`

### DIFF
--- a/syft/tensor.py
+++ b/syft/tensor.py
@@ -480,12 +480,13 @@ class TensorBase(object):
 
         Returns
         -------
-        Tensor Data
+        TensorBase:
+            Output Tensor
         """
         if self.encrypted:
             return NotImplemented
         self.data = np.absolute(self.data)
-        return self.data
+        return self
 
     def nelement(self):
         """Returns the total number of elements in the tensor."""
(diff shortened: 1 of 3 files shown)


syft/tensor.py
```diff
@@ -523,7 +524,7 @@ def sqrt(self):
         """
         if self.encrypted:
             return NotImplemented
-        return np.sqrt(self.data)
+        return TensorBase(np.sqrt(self.data))
```
bharathgs: need not be a  TensorBase object, this is just an alias for `numel` .

PR #304, opened 2017-10-08
Add Codecov for code coverage to Travis

ibrahimsharaf:
Using Code Coverage to determine how much code is being tested.


requirements.txt
```diff
@@ -4,6 +4,7 @@ scipy
 phe
 line_profiler
 clint
+coverage
```
baldassarreFe: Hey, I think `line_profiler`, `clint`, `coverage`, `pytest` and `pytest-flake8` should be moved to `test-requirements`. While `sklearn` should go all the way to `dev-requirements`

ibrahimsharaf: Where are the dev-requirements?

baldassarreFe: In `dev-requirements.txt`, but maybe you have to pull from `master` because it's been added quite recently by #270

ibrahimsharaf: @baldassarreFe done.

Issue #307, opened 2017-10-08
Add a docker-test rule in the Makefile


sp4ke:
Many people who use Docker do not develop inside the container but just use it as an isolated environment to run tests and compile code.

In that regard, running tests inside a spawned container should be available as a Makefile rule like `make docker-test`.

Right now, if I want to run tests I need to first launch the container with `docker-run` then execute `make test` inside. I do not use the notebook for developing so the the Makefile should not assume I am running a container with `docker-run`. 

Newcomers who will choose Docker as development environment and use their local editor on their machine will not be able to automate the tests through a shortcut without more knowledge of Docker. For example many editors like vim understand **Makefile** syntax and allow easy execution of make rules. 

Waiting for some feedback before proposing a PR

PR #309, opened 2017-10-08
make test: reinstall pip dependencies only when they change

sp4ke:
`make test` should not reinstall requirements every time it's executed, it's a waste of time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,3 +118,6 @@ lastfailed
 
 # code coverage reports
 .coverage
+
+# makefile dummy files
+.test_last_installed
--- a/Makefile
+++ b/Makefile
@@ -11,11 +11,13 @@ develop:
 	pip3 install -r dev-requirements.txt
 	python3 setup.py develop
 
-test:
+.test_last_installed: test-requirements.txt
+	pip3 install -r test-requirements.txt && touch .test_last_installed
+
+test: .test_last_installed
 	@# Remove pyc files to avoid conflict if tests are run locally
 	@# and inside container at the same time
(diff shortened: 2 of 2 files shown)


baldassarreFe:
Well, pip is smart enough to notice that the dependencies are already installed, it will only add the missing ones if any. And assuming people use a one time only docker container it's good to have that command there

sp4ke:
> Well, pip is smart enough to notice that the dependencies are already installed, it will only add the missing ones if any. 

Unfortunately it will still go through all the requirement packages and compare with local installed ones even if the file didn't change, and this process will get slower as we add new requirements. 

> And assuming people use a one time only docker container it's good to have that command there.

That would make sense if there was a `make docker-test` see #307  that spawns a new container for each test. The way the Makefile is organized now one assumes a long running container, so the pip command is executed each time for no reason after the first usage. Besides, the rule will work as before anyway, it will just make the next tests faster.

> Personal opinion: I'm not a fan of overly complicated Makefiles, I think that a new user should be able to just gaze at the Makefile and understand what's going on. In this case, we can easily avoid having a hierarchy of 3 targets referencing each other to achieve something that is already built into pip

The goal of a Makefile is to speedup repetitive tasks and decouple actions from their dependencies to avoid wasting time. The whole point of a Makefile and Unix philosophy in general is to do one task and do it well. In this case the test rule should only focus on the tests and output test related text and not the extra pip actions if there is no need for it. I can add comments on the Makefile to explain this rule.

sp4ke:
closing the pr since there was no feedback from other members.

PR #310, opened 2017-10-09
capsule preinstalled in pysyft-dev image

nishanthprakash:


dockerfiles/Dockerfile.dev
```diff
@@ -4,6 +4,17 @@
 
 FROM pysyft-base:local
 
+RUN apk update
```
baldassarreFe: Using update (apt or apk) in Docker files is highly discouraged, you can check on their guidelines for a full explanation, but in short it has to do with the fact that Docker caches the intermediate containers simply based on the instruction string. Can you put together those commands using `&&`?

TheOriginalAlex: It also adds extra bloat to the image.  I'd just recommend doing `apk add --initdb --no-cache git redis`.

nishanthprakash: Done.

PR #312, opened 2017-10-10
handle slices on 1 dim tensors, resolves #311

### LINKED ISSUE #311: 1dim TensorBase slice returns ndarray instead of TensorBase
Currently a slice of TensorBase returns an ndarray if `dim() == 1`. This is not the behavior expected in PyTorch

Example pytorch session:
```python
t1 = Tensor([1,2,3,4])
t1[1:]
 2
 3
 4
[torch.FloatTensor of size 3]
```

Same with TensorBase
```python
t2 = TensorBase([1,2,3,4])
t2[1:]
array([2, 3, 4])
```

see https://github.com/OpenMined/PySyft/commit/7fb592dcd4ede5b0d0cc6bdc50b98a0445ad77c0#diff-796d7f13fb1460eef9cffdccebd38faeR148

Is this a bug or expected ?

sp4ke:


samuxiii:
Hi,

Could you move the lines where you test the [] indexing form to _test_indexing_? You place a couple of lines in this _test_index_ which is only to verify index() not [] form.

sp4ke:
@samuxiii Ok. I make an other PR as I couldn't push a new commit on this one.

Issue #70, opened 2017-08-09
Implement Default narrow Functionality for Base Tensor Type


iamtrask:
**User Story A:** As a Data Scientist using Syft's Base Tensor type, I want to leverage a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete, narrow() should return a new tensor. For a reference on the operation this performs check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation.

**Acceptance Criteria:**
- If the Base Tensor type's attribute "encrypted" is set to True, it should return a NotImplemented error.
- a unit test demonstrating the correct operation on the Base Tensor type implemented over int and float Tensors.
- inline documentation in the python code. For inspiration on inline documentation, please check out [PyTorch](http://pytorch.org/docs/master/tensors.html)'s documentation for this operator.

sakchhams:
I would like to aim for this. Seems doable, any algorithmic preferences?

nishanthprakash:
I see that this issue has no update for 7+ days; I would like to pick it up if no one else is working on it now.

nishanthprakash:
I have submitted PR #313, that resolves this. It would be great if this could be reviewed.

PR #313, opened 2017-10-10
Implement Default narrow Functionality for Base Tensor Type #70

nishanthprakash:
This is supposed to address #70

### DIFF
--- a/syft/tensor.py
+++ b/syft/tensor.py
@@ -2984,6 +2984,33 @@ class TensorBase(object):
             raise NotImplemented
         return mv(self, tensorvector)
 
+    def narrow(self, dim, start, length):
+        """
+        Returns a new tensor that is a narrowed version of this tensor.
+        The dimension ``dim`` is narrowed from ``start`` to ``start`` + ``length``.
+
+        :param dim: dimension along which to narrow
+        :param start: starting dimension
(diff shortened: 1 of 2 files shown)


syft/tensor.py
```diff
@@ -2994,6 +2994,33 @@ def mv(self, tensorvector):
...
+        """
+        dim = dim if dim >= 0 else dim + self.dim()
+        if self.encrypted:
+            raise NotImplemented
+        if not isinstance(dim, int):
+            raise TypeError("The value of dimension must be integer")
+        if not isinstance(start, int):
+            raise TypeError("The value of start must be integer")
+        if not isinstance(length, int):
+            raise TypeError("The value of length must be integer")
```
bharathgs: why not something like this: 

    if not isinstance(dim, int) or not isinstance(start, int) or not isinstance(length, int):
        raise TypeError(f"narrow received an invalid combination of arguments - got {dim.__class__.__name__},
        {start.__class__.__name__},{length.__class__.__name__},but expected (int dim, int start, int length")

so that if you pass multiple invalid arguments you can point them out clearly, instead of the current code, where you cant do that.

nishanthprakash: Agree, that would be much better; I've updated the PR.

PR #318, opened 2017-10-10
Use shortcircuit logic to optimize function call

MarcoROG:
Using shortcircuit logic prevents a call to np.allclose which would be unnecessary when the shapes don't match.

This also prevents a ValueError: operands could not be broadcast together with shapes when comparing tensors with different shapes.


MarcoROG:
We're also missing some tests regarding shape mismatch, I'll add them soon

PR #320, opened 2017-10-11
Feature/paillier fixed point functionality

siarez:
This partially addresses issue #12 
This is my third solution for having a more abstract fixed-point implementation.
My previous solutions involved creating our own fixed-point data type. But that was awkward given that `phe` module has its own internal fixed-point representation. 
The next solution was to  have our own fixed-point data type, and only use "raw" encrypting functions of `phe`, but that seemed like a waste. Because `phe` already has alot of fixed-point functionality built-in which we had to replicate for our own data type.

The only problem with `phe`'s fixed-point implementation was that we can not control the precision through its API (It is calculated automatically).

This is the third solution that I came up with. It basically replaces some built-in functions in `phe` module (without altering the code in the module itself) to allow us to explicitly control fixed-point precision encoding. The major change in this pull request is at line 140 of `he/paillier/keys.py`.


iamtrask:
@siarez looks like we're good except for some new unit tests

siarez:
@iamtrask thanks for checking out the code. I did spend a lot of time on this, but now I'm a little suspicious that this is not the right direction. Mainly because, `phe` module has **real** floating-point built-in, and this code modifies that so we can have custom fixed-point tensors. It basically allows us to create PaillierTensors with a fixed number of digits to the right of the decimal point.
 
I don't understand how this is a step forward. Going from floating-point to fixed-point seems like a backward step to me. You can also read my last comment on the issue this is addressing (#12) and share your thoughts.

iamtrask:
@siarez I think this is the right way forward. We're not going to be using PHE for everything... but we're going to need Fixed Point precision in all of our other tensors. A solution that gives us generic support (and perhaps more importantly _consistent_ support) is a huge win!

siarez:
My other question is: Why we need to implement an abstract fixed-point datatype when our end goal is to support floating-point datatypes? Why not implement an abstract floating-point type from the get go?

iamtrask:
As I understand it, it's pretty tough to do when you don't know the actual value of the number (because it's encrypted). In short, how do you know when to move the decimal point and by how much?

PR #323, opened 2017-10-11
Implement default set functionality resolves #93

### LINKED ISSUE #93: Implement Default set Functionality for Base Tensor Type

sp4ke:
Resolves #93 

The methods added on TensorBase are:
- `storage_offset`
- `set_`


sp4ke:
@bharathgs Can you give me a few hints ? Since PyTorch uses a custom Storage I struggled to understand what was the desired behavior for TensorBase

bharathgs:
@sp4ke  
1. so [this (https://github.com/pytorch/pytorch/blob/9ef39a50ee16d4944240ac5cfa6489e3dd294c5f/torch/csrc/generic/methods/Tensor.cwrap#L194) and [this](https://apaszke.github.io/torch-internals.html) might be of interest to you. 

2. so with the current code I noticed that if I do something like:  

    a = TensorBase(np.array([1,2,3,4]))
    a.set_()

it will not change anything,(although I haven't tried your code, I think I'm right) but  the same operation in torch:  

    a = torch.tensor([1,2,3,4])
    a.set_() 

will result in `torch.FloatTensor with no dimension`. you should probably check that first link.

sp4ke:
@iamtrask I'll try to finish it this week

bharathgs:
@sp4ke any updates on this one?

sp4ke:
Yes still in progress just didn't have enough time last week. I'll try to
make an update this weekend.

On Nov 3, 2017 4:11 AM, "bharath g s" <notifications@github.com> wrote:

@sp4ke <https://github.com/sp4ke> any updates on this one?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/OpenMined/PySyft/pull/323#issuecomment-341613946>, or mute
the thread

sp4ke:
With regard to #212 which implements size, why is not respecting the [PyTorch api](http://pytorch.org/docs/master/tensors.html#torch.Tensor.size) ?  PyTorch `size()` returns a tuple, shouldn't we do the same and have an other method if a flat size number is needed ?

I know this is not directly related to #212 but since I have it also on my PR and many of my tests rely on it I wanted to make sure this is what is expected.

bharathgs:
1. Syft does return a tuple  I suppose because it's basically doing `np.array.shape`
(on other note, you might want to consider removing stride, as we already have it implemented)

sp4ke:
@bharathgs Not sure what's wrong with the code coverage ?

PR #324, opened 2017-10-11
Added GPU API based on PyOpenCL to work cross vendors

OceanCoding:
Heyho, first pull request on Github, I hopefully did everything right.

So, I added a API to support GPU computation. There exists and usage_example.py that explains everything you need to know.

The basic idea is to create easy access to GPU computation with as few lines of code possible and as less data transfer between CPU and GPU as possible.
Each class and method should have a doc-string and there are plenty of comments explaining my thought process.

Please keep in mind, that this is only a working prototype and I'm looking to improve the code bit by bit, to for example support cross GPU computation. I have already added one working kernel you can perform on the GPU which is vector addition and am currently working on other functions, such as Rectifier-Functions, Matrix-Multiplications (well, everything you need for a feedforward NN).

Best Regards
Raph

P.S.: If you have further questions, feel free to contact me on Slack (same username as here).


iamtrask:
@kevinahuber this travis-ci seems to be hanging. - is it possible for travis to even build this code?

OceanCoding:
So, do I understand Travis right: The build failed because it couldn't install PyOpenCL??

iamtrask:
@OceanCoding looks that way - perhaps we should ignore it in the build? - @kevinahuber any suggestions?

bharathgs:
@OceanCoding @iamtrask  I found [this](https://gist.github.com/patrickmmartin/e1313dde7b908e8d009f2a13c3cd164b) and [this](https://github.com/travis-ci/travis-ci/issues/8512), not sure if it's of any help.

OceanCoding:
I have really no clue. It seems to be working on most of the machines :/

iamtrask:
Let's get this guy into the new Unity backend!!!

Issue #292, opened 2017-10-05
Multiparty Computation Prototype - Secret Sharing Part 1


iamtrask:
Recently, we implemented a first prototype Multiparty Computation tensor here: https://github.com/OpenMined/PySyft/blob/master/notebooks/MPC%20Tensor%20Proof%20of%20Concept%20Demo.ipynb

This was based off of the following post from our OpenMined contributor @mortendahl:
http://mortendahl.github.io/2017/04/17/private-deep-learning-with-mpc/

@mortendahl also wrote an excellent followup post with a more advanced MPC scheme. As a next step in our PySyft library, we'd like to have a NEW MPCTensor prototype based on this more advanced scheme. The full tutorial for this scheme can be found here.
http://mortendahl.github.io/2017/06/04/secret-sharing-part1/

Acceptance Criteria:
- everything in the first MPC Demo Notebook can be done using the new technique
- Initialize unit testing suite around that functionality.

sakchhams:
seems pretty interesting, I'd like to give it a try. A word of advice?

withai:
I would like to collaborate on this issue.

mortendahl:
hey guys, sorry for being a bit slow these days. don't do anything yet, there's a much better scheme coming out soon, hopefully a draft within the next day or two! :)

mortendahl:
hi all, 

just pushed a draft of the SPDZ scheme i was talking about [here](http://mortendahl.github.io/2017/09/19/private-image-analysis-with-mpc/) -- it's still a WIP but the basics of the protocol should be intelligible

the big advantage of SPDZ over the shamir schemes is that a lot of the processing can be pushed to an offline phase, or alternatively an independent "crypto provider". the disadvantage is that it cannot handle any of the computing parties disappearing.

iamtrask:
👍 This is awesome!!! Great work sir!!! Let's definitely go for this scheme instead!!

iamtrask:
@withai and @sakchhams - any questions on the approach above? Are you still interested in working on it?

sakchhams:
Yes, I want to work on this.

withai:
Yes, I would like to work on it.

iamtrask:
Excellent! @sakchhams could you create a branch for this project so we can get started?

iamtrask:
Have you guys been able to get the code in @mortendahl 's post to run?

mortendahl:
thanks a lot @iamtrask! :)

[...]

PR #325, opened 2017-10-11
Mpc tensor spdz

withai:
Multiparty Computation Prototype - Secret Sharing Part 1 #292


iamtrask:
This might be the biggest PR we've merged yet... should we collaborate a bit on the style/test issues?

iamtrask:
oh actually... @withai can we plan to merge to master on this one? i think that's why it appears to be so massive

withai:
@iamtrask Sure, we shall collaborate

PR #326, opened 2017-10-12
Mpc prototype

iamtrask:


iamtrask:
fixes https://github.com/OpenMined/PySyft/issues/292

iamtrask:
looks like we need some unit tests

PR #327, opened 2017-10-12
First nonlinearity

iamtrask:


iamtrask:
@bharathgs good catch... my bad on that one

PR #329, opened 2017-10-13
add make targets for redis and capsule

nishanthprakash:


iamtrask:
@bharathgs could i get your thoughts on this guy post-update?

bharathgs:
The Makefile has `FLASK_APP=/usr/bin/Capsule/capsule/local_server.py flask run &` but should it not be `../capsule_zmq/local_server.py` ?

nishanthprakash:
I think the django and flask capsule servers have been removed. I'll update this accordingly soon; sorry about the delay.

bharathgs:
@nishanthprakash could you also change [this](https://github.com/OpenMined/PySyft/blob/9bdf747c5b037c527743cae728ce562bc6eb64a8/Makefile#L19) to `pytest && flake8 --ignore=E501`

nishanthprakash:
Done. I notice that in the test target, some scripts still use Django (so do some of the demo notebooks).

PR #336, opened 2017-10-15
Polynomial approximation and activations

MarcoROG:
This PR provides base classes for NN activation layers.
It also provides a PolynomialActivation implementation.

The PolyFunction class represents a polynomial function along with its derivative.
It can either be built by specifying the polynomial coefficients or by approximating functions.


iamtrask:
@MarcoROG it looks like there are still some style errors - can you take a look at those?

MarcoROG:
Indeed, this was a pre-push so that @bharathgs had access to the source.
I'm still missing some documentation and a lot of unit tests which will probably come later this week.

iamtrask:
@MarcoROG groovy... very much looking forward to merging this one! - gonna be an epic increase in the value of OpenMined.

iamtrask:
Let's get this guy into the new Unity backend!

PR #343, opened 2017-10-16
Prevent flake8 from running twice

kevinahuber:


bharathgs:
fixed it, thanks for pointing it out.

Issue #350, opened 2017-10-18
Organize class methods and functions 


tblazina:
Currently there is no order of how methods appear, for example, within the TensorBase class and in math.py. There is no strict convention for organizing methods and functions but a general pattern to follow would be: 

<pre>
class SomeClass(object):
    def __magic_methods__(self):
        "magic methods first, usually in alphabetical order"
    def _private_method(self):
        "worker methods next, also in alphabetical order"
    def a_method(self):
        "then normal methods, also in alphabetical order"`

OceanCoding:
Totally agree with you there! And I will implement this for my Code as well.

I also see a problem with class organization and how they interact with each other, because often the docs aren't sufficient explaining code interactions. I suggest a small diagram or a flow chart to get ppl to get our code faster and of course also easier!

tblazina:
I've gone through all the current code and organized it in PR #351, and it might be a good thing to add also to the style guidelines for the project so we can try to stay consistent throughout.  Also totally agree with you, there should be a relatively easy to understand logic behind the class organization. We use draw.io quite frequently at my work for such things, so this could be a good tool for such a task.

OceanCoding:
Sounds like a good idea, Im getting to it on my code!

PR #351, opened 2017-10-18
Reorganize method classes and functions throughout entire pysyft project

tblazina:
There was previously no logic for how class methods were organized.
This will organize class methods with the following logic:

class SomeClass(object):
    def __magic_methods__(self):
        "magic methods first, usually in alphabetical order"
    def _private_method(self):
        "worker methods next, also in alphabetical order"
    def a_method(self):
        "then normal methods, also in alphabetical order"


tblazina:
This closes #350

tblazina:
good catch! I've moved it in. thanks a lot.

tblazina:
@iamtrask  sure thing. I'll try to fix them this evening!

syft/he/paillier/basic.py
```diff
@@ -119,17 +130,85 @@ def sum(self, dim=None):
             op = self.data.sum(axis=dim)
             return PaillierTensor(self.public_key, op, False, fixed_point_conf=self.fixed_point_conf)
 
+
+class Float():
```
bharathgs: Can you check this, again? this seems like an extra, I don't see it [here](https://github.com/OpenMined/PySyft/blob/master/syft/he/paillier/basic.py)

tblazina: hmm weird, I thought I had just pulled down from master. Let me take that out.

PR #355, opened 2017-10-19
update numel in math.py

raul-jr3:


syft/math.py
```diff
@@ -768,9 +768,9 @@ def numel(tensor):
         total number of elements in the input Tensor
     """
     if tensor.encrypted:
-        return tensor.data.size
-    else:
-        return tensor.data.size
+        return NotImplemented
+
+    return tensor.data.size
```
bharathgs: As per issue #78  discussion, I don't see why we need that `If` part in the first case at all since we can easily count the `[e]` as mentioned over there. In accord with that discussion, I would say it's safe to just remove that `if` entirely. and just return `tensor.data.size`. (I assume this was kept since it was unclear on what we should do, earlier)

raul-jr3: Removed :)

PR #356, opened 2017-10-19
Implement default stride functionality for TensorBase

vbankov:
Hi, this resolves issue #100.

I found out that NumPy and Torch have slightly different implementations of the stride functionality and I adjusted TensorBase to be in line with Torch's implementation. :smiley: happy times


Ujjwal-9:
@vbankov i have already submitted the PR for the same

bharathgs:
@vbankov it seems you might have missed PR #267 which addresses the same issue.
perhaps, you might be interested in working on other open issues? (we have quite a lot of them)

Issue #365, opened 2017-10-22
Naming convention on Test files


alberduris:
#### Motivation: 

Still there are some inconsistencies in the naming of Classes and Functions on Test files. Reach a unified naming convention is needed.

#### Acceptance Criteria: 

All test files must follow the next convention:

 - Class: Camel case with suffix "Tests". For example: **FooBarTests**
 - Functions: Lowercase separated by "_" and "test" prefix. For example: **'test_name_of_test'**

alberduris:
Working on it 🆙

alberduris:
I've encountered this, so I'm not changing that function name to follow convention as an exception (I don't know why is the exception)

@ test_tensor.py:416
```
class PermuteTests(unittest.TestCase):
    # WARNING: don't change the name of this function
    def dest3d(self):
        t = TensorBase(np.ones((2, 3, 5)))
        tdash = t.permute((2, 0, 1))
        self.assertTrue(tdash.data.shape == [5, 2, 3])
        self.assertTrue(t.data.shape == [2, 3, 5])
```

PR #366, opened 2017-10-22
Naming convention on test files solved #365

alberduris:
#### Reference Issue #365

Test files follow naming convention


iamtrask:
@alberduris can you check the test errors?

alberduris:
@iamtrask I've been checking them and for the moment I couldn't figure out what's going on. It is assumed that tests were running correctly before the modification, I've change nothing but the names and now they failed. Tomorrow I'll try to solve it again.

PR #367, opened 2017-10-23
Implement Default multinomial Functionality for Base Tensor Type

borkjt9:

### DIFF
--- a/syft/tensor.py
+++ b/syft/tensor.py
@@ -1930,6 +1930,40 @@ class TensorBase(object):
         self.data = np.random.binomial(1, p.data)
         return self
 
+    def multinomial(self, p, num_samples,  replacement=False):
+        """
+        Returns Tensor with random numbers from the Multinomial Distribution.
+
+        Returns Tensor with random numbers
+        from a multinomial distribution with probability
+        specified by p(arr_like), number of draws specified by num_samples,
+        and whether to replace the draws specified by replacement.
+
+        The p Tensor should be a tensor containing probabilities to
+        be used for drawing the multinomial random number.
+        The values of p do not need to sum to one (in which case we use the values as weights),
+        but must be non-negative and have a non-zero sum.
+
+        Parameters
+        ----------
+        p: TensorBase
+            Weights for the multinomial distribution
+        num_samples: Int
+            Number of samples to be drawn. If replacement is false, this must be lower than the length of p.
+        replacement: bool, optional
+            Whether to draw with replacement or not
+
+        Returns
+        -------
+        Output Tensor
+        """
+        if self.encrypted:
+            return NotImplemented
+        p = _ensure_tensorbase(p)
+        p = p / p.sum()
+        return TensorBase(np.random.choice(len(p), num_samples, replacement, p.data))
+
+
     def uniform_(self, low=0, high=1):
         """
         Fills the tensor in-place with numbers sampled unifromly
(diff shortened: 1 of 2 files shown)


bharathgs:
@iamtrask @siarez 

1. Shouldn't this function be in _math.py_?
2. this is not quite similar to what torch is doing ([`torch.multinomial`](http://pytorch.org/docs/master/torch.html#torch.multinomial)).
3. in this implementation if I do `some_tensor.multinomial(4, True)` I will get an error(as I'm not passing `p`) and I have to do `some_tensor.multinomial(some_tensor, 4, True)` or `some_tensor.multinomial(p, 4, True)`(check torch's note for this). This is not right, why do I have to pass `some_tensor` argument again, If I'm already doing `some_tensor.multinomial`?

note: as an example consider `some_tensor = TensorBase(np.array([0,10,3,0]))` & if you want to explictily specify p, take it as `p = [0.1,0.2,0.3,0.4]`

borkjt9:
@bharathgs 
In response to your questions, 

1. I placed this function in tensor.py to be consistent with the locations of other probability distribution functions: uniform, bernoulli, geometric, etc.
2. Can you provide an example where the results of torch.multinomial & this function return a different result?
3. In response to your first statement, the probabilities array input in torch.multinomial is not an optional field, which is why an error is returned if you do not include it here. In response to your latter statement, can you explain what does not seem correct, perhaps with an example (and include the results your getting)?

Thanks.

bharathgs:
@borkjt9 

1. hmm, that is fine although we've been discussing on this very issue #256 (on what goes where)
2. It was not about the results.
3. sure here you [go](https://imgur.com/a/dU0LB) do you notice the difference in the way the arguments are passed in torch and syft?

siarez:
@bharathgs You are right. This needs to move to math.py so we can use it as `syft.multinomial`. Because it takes `p`/`input` like `torch.multinomial` does. Then we should have a `TansorBase.multinomial` where it takes `self` as `p`/`input`

borkjt9:
The PyTorch documentation states that the multinomial function should have the ability to use separate weights as a parameter. However, I agree with @bharathgs that if one does not supply the weights, the function does use the elements of the base Tensor as weights.

Would it not be better to create a second function, rather than replace the first function, called multinomial_() that takes self as the input parameter? This would provide greater flexibility, as well as conform to the overall naming convention for functions in Syft.

siarez:
@borkjt9 underscore after a function's name implies an "inplace" operation where the changes are made to the tensor itself. I don't think this is what we are going for in this case.
In PyTorch's convention we usually have`torch.some_function(tensor, arg, arg ...)` and `some_tensor.some_function(self, arg, arg ...)` and sometime also `some_tensor.some_function_(self, arg, arg ...)` . In the last case the value of `some_tensor` itself is changed by `some_function`

borkjt9:
@siarez ok understood. It still seems clear though from PyTorch's documentation that the user may supply the weights to be used in the multinomial distribution if desired. 

Perhaps we should have p be an optional parameter where if it is not supplied, it defaults to using the elements of the tensor?

siarez:
@borkjt9 In PyTorch we have both `torch.multinomial` and `Tensor.multinomial`. If we want to follow that convention in Syft, we should have syft.multinomial and TensorBase.multinomial. Please look at code in my pull request above.

PR #373, opened 2017-10-24
Added Swish Activation Function 🐍

### LINKED ISSUE #372: Implement Default swish Functionality for Base Tensor Type
#### User story: 

As a Data Scientist using Syft's Base Tensor type, I want to leverage a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete, swish() should return a new tensor, but swish_() should operate inline.
#### Acceptance Criteria: 

 - [x] If the Base Tensor type's attribute "encrypted" is set to True, it should return a NotImplemented error.
 - [x] a unit test demonstrating the correct operation on the Base Tensor type implemented over int and float Tensors.
#### Notes:
Swish activation (mathematically, f(x)=x*sigmoid(x)) function often performs better than ReLUs.

[Paper](https://arxiv.org/pdf/1710.05941) on Swish Function recently published by a team at Google.

Ujjwal-9:
#### Reference Issue
Fixed #372 

Swish Function often performs better than ReLu.

#### Changes.
- Implemented Swish() in math.py which returns a tensor.
- Implemented Swish_() in tensor.py which operates inline.
- Added Tests for the same.


Ujjwal-9:
@iamtrask, @bharathgs please look into this PR. 🙂

iamtrask:
there are some very strange merge conflicts...

bharathgs:
@Ujjwal-9 It would be better If we have activation functions within `syft.nn`.  as such I've closed this PR for now. However, we can implement this within `syft.nn`. I will update you regarding this issue.
(I suppose you implemented it here based on the RELU. which was implemented in the math module as well. noticed it a while ago. PR #380  will remove it as well)

PR #375, opened 2017-10-27
Initial BV API

### LINKED ISSUE #317: Porting HE libraries completely to Python
The motivation behind this idea is to further the development of HE libraries viz. PyBV, PyAono, and potentially PyFV. Since more people are comfortable with Python and numpy, more people will be able to contribute to this project, which is currently the need of the hour. Moreover, this could serve as a mockup for faster CPU and GPU implementations. This will also solve the high maintenance issues that are persistent with the current Cython implementation.
All in all, I urge everyone who's interested in homomorphic encryption to contribute to building this library. I'll positively roll out the initial version soon and make sure that it mirrors the interface provided by syft.he.paillier.

P.S. We're not restricted to just numpy, but the implementation has to be in Python. We're starting off with numpy, but in the near future, it could be that numpy doesn't have the functionality we need, in which case, we can use multiple libraries.

deevashwer:
#### Reference Issue
Fixes #317 

The API has been re-written predominantly in Python, and will mainly serve as a mockup for faster implementations.

#### TODO
  * Ring addition function from PyPARI API has to be implemented purely in Python, for a faster 
implementation.
  * The interface file between PARI and Python needs to be changed, to allow better type conversion. The current interface file is just for the sake of completion.
  * Support for Full Homomorphism, along with development of PyPARI accordingly.


iamtrask:
This is awesome! Also in speaking with @deevashwer we need to have travis skip this section because the code is automatically generated - and does not conform to the style guide.

PR #391, opened 2017-10-31
Split tensor

### LINKED ISSUE #383: Implement Default split Functionality for Base Tensor Type.
#### User story: 
As a Data Scientist using Syft's Base Tensor type, I want to leverage a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete `split` should return a new tensor. For a reference on the operation these perform check out [PyTorch's](http://pytorch.org/docs/master/torch.html#torch.split) documentation.

#### Acceptance Criteria: 

  - [x] If the Base Tensor type's attribute "encrypted" is set to True, it should return a NotImplemented error.
- [ ] corresponding unit tests demonstrating the correct operation on the Base Tensor type implemented over int and float Tensors.
 - [ ] inline documentation as described over [here](https://github.com/OpenMined/PySyft/blob/85bc68e81a2f4bfc0f0bf6c4252b88d6d7b54004/syft/math.py#L5).

divyanshj16:
This fixes #383 

This add split functionality to default tensor type.

#### Explain your changes.
In syft.math it add a split function which uses numpy.array_split() for splitting the array into given axis into `split size` size of each chunk. The function returns a list containing divided sub-arrays. I had to some initial calculations to match the numpy api.

### DIFF
--- a/syft/tensor.py
+++ b/syft/tensor.py
@@ -3613,12 +3613,9 @@ class TensorBase(object):
 
         Parameters
         ----------
-        tensor: TensorBase
-            array to be divided into sub-arrays.
 
         split_size: int
-            the array will be divided into `split_size` equal arrays along `axis`. If split
-            cannot be done equally the last array will be smaller along the given axis.
+            size of single chunk
 
         axis: int, optional
             The axis along which to split, default is 0.
--- a/syft/tensor.py
+++ b/syft/tensor.py
@@ -3601,3 +3601,35 @@ class TensorBase(object):
 
         self.data.fill(0)
         return self
+
+    def split_(self, split_size, axis=0):
+        """
+        Splits the tensor into multiple equally sized chunks (if possible).
+
+        Last chunk will be smaller if the tensor size along a given axis
+        is not divisible by `split_size`.
+
+        Returns a list of the split tensors
+
+        Parameters
+        ----------
+        tensor: TensorBase
+            array to be divided into sub-arrays.
+
+        split_size: int
+            the array will be divided into `split_size` equal arrays along `axis`. If split
+            cannot be done equally the last array will be smaller along the given axis.
+
+        axis: int, optional
+            The axis along which to split, default is 0.
+
+        Returns
+        -------
+        list: list of divided arrays
+        """
+        if self.encrypted:
+            return NotImplemented
+
+        return syft.math.split(self, split_size, axis)
+
+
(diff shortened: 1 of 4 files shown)


syft/tensor.py
```diff
@@ -3601,3 +3601,33 @@ def zero_(self):
...
+        Splits the tensor into multiple equally sized chunks (if possible).
+
+        Last chunk will be smaller if the tensor size along a given axis
+        is not divisible by `split_size`.
+
+        Returns a list of the split tensors
+
+        Parameters
+        ----------
+        tensor: TensorBase
```
bharathgs: change this inline doc (`tensor`?)

bharathgs:
looks like few test are missing.

divyanshj16:
Earlier I confused `split_size` variable to be the number of splits. But In the new commit, I changed this!
Actually, split size is chunk size.
If there is an array [0,1,2,3,4,5,6,7,8,9] then if `split_size` is 3 then the array will get divided into subarrays of size 3 and last one with size one as [[0,1,2],[3,4,5],[6,7,8],[9]].
In my earlier implementation the above array would have divided into 3 subarrays [[0,1,2,3],[4,5,6,7],[8,9]]. This has been resolved by this last commit.

bharathgs:
your code does not produce the results that you mention in the comment above.
also, please try out the code before a PR.

divyanshj16:
I'll do the changes! I did not understand your second point, `2. in place version of this function is not necessary, however, a split method is(in tensor module)`

bharathgs:
change `split_` -> `split`

divyanshj16:
There is already a `split` method in `tensor.py` at line 3041 and `SplitTests` are also there in `test_tensor.py` at line 977. Its tests are not exhaustive. Also, they are not implemented as PyTorch's implementation. 
This was the reason I had to name my test class as `SplitTests2` and my function as `split_`. 
My functions are working good in accordance with the PyTorch's implementation. See the image below.

image

I'll add tests in test_math for the split function and will create a PR!
Should I  remove the earlier implementation(i.e implementation on the specified lines)?

bharathgs:
ahh, now I see, I missed that. yes remove the earlier implementation

divyanshj16:
Naming conventions are not specific as `axis`(numpy's) and `dim`(pytorch's) are interchangeably used in the whole code. What should I use?

divyanshj16:
I have used `axis` for naming the axis of the tensor. However, we should raise an issue to make it consistent.

PR #392, opened 2017-11-02
Add cross functionality for Base Tensor Type

### LINKED ISSUE #382: Implement Default cross Functionality for Base Tensor Type.
#### User story: 
As a Data Scientist using Syft's Base Tensor type, I want to leverage a default method for computing operations on a Tensor of arbitrary type. For this ticket to be complete `cross` should return a new tensor. For a reference on the operation these perform check out [PyTorch's](http://pytorch.org/docs/master/torch.html#torch.cross) documentation.

#### Acceptance Criteria:

 - [ ] If the Base Tensor type's attribute "encrypted" is set to True, it should return a NotImplemented error.
 - [ ] corresponding unit tests demonstrating the correct operation on the Base Tensor type implemented over int and float Tensors.
 - [ ] inline documentation as described over [here](https://github.com/OpenMined/PySyft/blob/85bc68e81a2f4bfc0f0bf6c4252b88d6d7b54004/syft/math.py#L5).

sriranganathan:
#### Reference Issue
- Fixes #382

Adds default cross functionality for base tensor type

#### Explain your changes.
- [X] Implement cross functionality using numpy.cross
- [X] If the Base Tensor type's attribute "encrypted" is set to True, it should return a NotImplemented error.
- [X] corresponding unit tests demonstrating the correct operation on the Base Tensor type implemented over int and float Tensors.
- [X] Inline documentation.

### DIFF
--- a/syft/math.py
+++ b/syft/math.py
@@ -20,7 +20,7 @@ __all__ = [
     'cumprod', 'cumsum', 'ceil', 'dot', 'floor', 'matmul', 'addmm', 'addcmul',
     'addcdiv', 'addmv', 'bmm', 'addbmm', 'baddbmm', 'sigmoid', 'unsqueeze',
     'sin', 'sinh', 'cos', 'cosh', 'tan', 'tanh', 'zeros', 'ones', 'rand',
-    'randn', 'mm', 'fmod', 'diag', 'lerp', 'renorm', 'numel'
+    'randn', 'mm', 'fmod', 'diag', 'lerp', 'renorm', 'numel', 'cross'
 ]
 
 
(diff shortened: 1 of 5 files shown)


syft/math.py
```diff
@@ -1008,3 +1008,43 @@ def split(tensor, split_size, axis=0):
     list_ = np.array_split(tensor.data, split_according, axis)
 
     return list(map(lambda x: TensorBase(x), list_))
+
+
+def cross(input, other, dim=-1):
```
bharathgs: could you add an encryption [check](https://github.com/OpenMined/PySyft/blob/34af4f778c2b2f1a16a5a8aa505c37b7d9b18009/syft/math.py#L330) for the inputs that are passed?

bharathgs:
Also, update the *__init__.py* as well.

syft/math.py
sriranganathan: Have made the requested changes and updated the PR.

PR #398, opened 2017-11-15
Added comments

gonesbuyo:
I thought that it would be great to have the introductory PySyft Jupyter tutorials commented for the newcomers to understand better what's happening, and since I was getting started with PySyft I decided to accept the task. It sure can be improved but I think this can be of help as a starting point.


iamtrask:
Great job!!!

PR #401, opened 2017-11-19
Add scalar_multiplication, and improve windows install

Floev:
- Also, wrapped no_params_func into params_func (during merge)
- Windows users often don't have ability to run .sh files without additional tweaks. Bat files should be easier?

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,10 +90,10 @@ python3 setup.py develop
 ```
 
 #### Anaconda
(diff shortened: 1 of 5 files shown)


bharathgs:
1. Cygwin should make it possible to run bash scripts on windows.
2. `.bat` files do not work on Linux.

Floev:
Thanks for the comment @bharathgs .  I'll look to add an .sh version, as well.

Issue #633, opened 2017-11-24
Integration Test Prototype


iamtrask:
### User Story
As a developer of PySyft functionality that links to the Unity backend, I want the ability to create Python integration tests which can evaluate whether the client is operating with correct behavior.

### Acceptance Criteria
- [ ] An example integration test evaluating (https://github.com/OpenMined/PySyft/pull/631)

alexanderConstantinescu:
I wouldn't mind picking this issue up as I am imagining that it can be done using separate Docker containers with communicating services. 

How are the tests being executed today / how do you imagine it should be in the future? Integrated into an automatic build and release cycle ? Or only ad hoc execution by each contributor?

iamtrask:
@alexanderConstantinescu and @bendecoste  - do we feel we can close this now?

bendecoste:
#144 isn't merged yet but i'm into it

bendecoste:
err .. it is merged I was thinking of the CI one .. but yeah i'm down to close

alexanderConstantinescu:
I guess you can close this, looking at what has been mentioned/referenced here.

PR #634, opened 2017-11-24
added `zero_()` to `FloatTensor`

bendecoste:
Zeros a tensor on the remote unity server
closes https://github.com/OpenMined/PySyft/issues/630


bendecoste:
>  an integration test in PySyft demonstrating the correct CPU and GPU operation implemented over a FloatTensor while connected to a Unity backend

This isn't done.  I believe it depends on https://github.com/OpenMined/PySyft/issues/633 ?

PR #648, opened 2017-11-28
Feature/add tanh

bendecoste:


syft/syft.py
```diff
@@ -75,16 +75,22 @@ def __repr__(self):
...
-      
+
     def abs_(self):
         return self.no_params_func("abs_")
 
     def neg(self):
         return self.no_params_func("neg")
 
+    def tanh(self):
+        return self.no_params_func("tanh",return_response=False)
```
bendecoste: seems like there are a few different naming conventions going on in this file ... e.g. `__function__(params)` vs `function()` ... let me know if I messed it up!

Floev:
__add__() gives you the + operator. Question is whether we want to provide add () as well

PR #657, opened 2017-11-30
Chore/clean refresh

bendecoste:


bendecoste:
true reconnect logic would be best but I thought it would be a good idea to add this because it's still a better feedback loop than what is on master.  I can update the google dock instructions whenever merged.

alberduris:
@bendecoste 

> I can update the google dock instructions whenever merged.

If you are referring to [this](https://docs.google.com/document/d/1WRd7gGLFN0Awtf86AICYIHtg3gfFWLBa5wYTthsB3i0/edit) Google Doc now it's better to update the [tutorial for it on OpenMined](https://github.com/OpenMined/OpenMined/blob/master/tutorials/HowToAddFunctionToFloatTensor.md)

We should edit the doc to just point to that tutorial I think

bendecoste:
> If you are referring to this Google Doc now it's better to update the tutorial for it on OpenMined

Ok I will open a PR in there.  Didn't know about that one 👍 

> We should edit the doc to just point to that tutorial I think

also 👍

PR #658, opened 2017-11-30
added trunc function

bendecoste:
closes https://github.com/OpenMined/PySyft/issues/619
closes https://github.com/OpenMined/PySyft/issues/620


bendecoste:
There is some discussion in slack re how to do inline/non-inline functions that may change this implementation

PR #660, opened 2017-12-01
Added sinh and sinh_

justin1121:
Also noticed tanh was deleted added that back


iamtrask:
Flying through those functions!!!

PR #675, opened 2017-12-04
Added sign_(), the inline version of sign() as per issue #581.

moodlep:
Hi there, this is related to a PR in OpenMined with the same name for the same issue #581. Thanks.


diegoalejogm:
@bharathgs The sign_ function is not implemented in Unity yet

bharathgs:
@diegoalejogm ahh, noticed it now, there's a PR for it in openmined/openmined.

PR #701, opened 2017-12-14
Messaging and typed values (Syft side)

bosr:
(Lot of text, but it's better to read it)

Pursuing the idea in https://github.com/OpenMined/OpenMined/issues/166, I suggest that we upgrade the string messages returned by Unity to JSON messages like for instance

    {'ok': True, 'objectType': 'tensorID', 'id': 34}

You can read about this in the corresponding PR on the Unity side (https://github.com/OpenMined/OpenMined/pull/215)

In this PR, there are 3 closely related things. I really wanted a set of smaller PR, but I could not find my way, so I did go through all of it this past week.

The topics are
- reorganize methods in sections such as 'tensor operations', 'memory management', 'dim reduction', 'arithmetic ops', 'properties access'
- more importantly those sections use methods with clearer semantics (detailed below, so that you don't have to infer it from the code)
- and even more importantly, the strings used to transfer from Unity don't leak up to those method calls: everything is Python types 

The reordering was necessary to eliminate duplicates, and make sense out of the different category of operations. I saw @emregoynugur did the same in csharp code independently ;)

1. `_op(self, name, params)` returns a `FloatTensor` or a `float` (only in the case of `trace()`). The semantics are: "computes new values: means no inline, no properties reading". Corresponding operations are for instance `round`, `__add__`, `trace`

it is typically called like this
```
    def view_as(self, x):
        assert isinstance(x, FloatTensor)
        return self._op("view_as", x.id)
        # return self.params_func("view_as", [x.id], return_response=True)
```

2. `_op_inline(self, name, params)` returns actually nothing. But the underlying call to Unity had received a json OK status (or an json error message) so it raises an exception in case of failure. 

It is typically called like this:

```
    def view_as_(self, x):
        assert isinstance(x, FloatTensor)
        # self.params_func("view_as_", [x.id], return_response=False)
        self._op_inline("view_as_", x.id)
        return self
```

so it's obvious what is returned, and there is a guarantee that silence means success.

All inline methods use this call.

3. `_manage(self, name)` returns nothing, no parameters can be given. Technically, it is very close to `_op_inline` that true, but no computation is conducted.

Example call

```
    def cpu(self):
        self._manage("cpu")
        return self
        # return self.no_params_func("cpu")   <- left temporarily for clarity
```

so it's obvious what is returned, and silence = success also. Methods that use `_manage` are `cpu()`, `gpu()`, `delete_tensor()`

4. `_get_prop(self, name, params)` returns the Python typed equivalent of the Unity property (bool, float, list[int], etc). The semantics is no computation is conducted, only reading.

```
    def autograd(self, setter=None):
        if setter is None:
            return self._get_prop("autograd")

        assert isinstance(setter, bool)
        self._set_prop("autograd", setter)
        return self

        # if setter is None:
        #     if self.get("autograd") == "1":
        #         return True
        #     return False
        # else:
        #     if setter:
        #         out = self._set_prop("autograd", True)
        #         # out = self.set("autograd", ["1"])
        #     else:
        #         out = self._set_prop("autograd", False)
        #         # out = self.set("autograd", ["0"])

        #     if (out == "1" and setter) or (out == "0" and not setter):
        #         return self
        #     return False

```

5. `_set_prop` is the same semantics for writing a value. It returns nothing, as for inline ops, but the semantics are 'no computation'. Example call above in autograd.

6. `_build_arithmetic_operation(self, x, name)`

Example call:

```
    def __add__(self, x):
        name, param = self._build_arithmetic_op(x, "add")
        return self._op(name, param)
        # return self.arithmetic_operation(x, "add", False)
```

I hope there is more code reuse and clearer semantics also in this call: "this is a computation, with a special operation name (I did *not* change the naming convention of arithmetic operations), it returns a tensor"

I'm eager to get your comments. I have ran the nice "Test Tensor operations" notebook without any problem and ran several experiments myself, but there may be some bugs.

Last thing, as on the Unity side, I temporarily left the commented previous calls for better readability in case you read the code.

### DIFF
--- a/syft/syft.py
+++ b/syft/syft.py
@@ -683,7 +683,7 @@ class FloatTensor():
             result = np.array(response['data']).astype('float').reshape(shape)
             # result = np.fromstring(strValue, sep=',').astype('float').reshape(objectShape)
 
-        elif objectType in ['string', 'int', 'float', 'bool']:
+        elif objectType in ['string', 'int', 'float', 'bool', 'intList']:
             result = response['value']
 
         elif objectType == 'tensorID':
--- a/syft/syft.py
+++ b/syft/syft.py
@@ -703,6 +703,9 @@ class FloatTensor():
 
 
     def _op(self, name, params=None):
+        """
+        Returns a FloatTensor, a float or a list as a result of pure function application (no side-effect).
+        """
         if params and not isinstance(params, list):
             params = [params]
         retval, error = self._parse(self._remote_execute(name, params), typeassert=[FloatTensor, float])
@@ -711,6 +714,11 @@ class FloatTensor():
         return retval
 
     def _op_inline(self, name, params=None):
+        """
+        Inline variant of `op`: executes operation `name` and returns nothing.
+
+        An exception is raised in case the operation failed remotely.
+        """
         if params and not isinstance(params, list):
             params = [params]
         # if not name.endswith("_"):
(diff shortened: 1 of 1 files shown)


bosr:
tldr; there are no API changes, all methods remain the same, only internal mechanics/routing and messaging with Unity have changes

bosr:
(rebased)

bosr:
Submitting that PR helped me realize how I could actually break it in smaller ones! Cool. That was my original intent but I could not find how.

I will submit those PRs
- one PR that reorganizes the api methods in sections. Nothing more than moving methods around.
- the next one makes the api methods uses the 6 internal calls presented above, but each of them simply ends up as a [no_]params_func call. ThIs should be a large set of tiny diffs easy to understand.
- one that introduces the json messaging and parsing, where the 6 calls get hooked to it

I hope it makes sense!

PR #710, opened 2017-12-16
Cleaned up ignored files

clamytoe:


iamtrask:
Thank you sir! Great housekeeping PR!

PR #712, opened 2017-12-19
Inline Documentation

CT83:
Here it is! @iamtrask  Inline Documentation was added. 

I am not done  with all of the functions in https://github.com/OpenMined/PySyft/blob/master/syft/tensor.py but I have created a pull request so you guys can give me the green light to continue, or if there are something you would like worded differently or some word spelled differently.

I followed the NumPy Style on inline documentation. http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html same as the Deprecated-PySyft. #273  

So just give a thumbs up and I will complete and polish this PR ASAP. 👍

### DIFF
--- a/syft/tensor.py
+++ b/syft/tensor.py
@@ -58,32 +58,123 @@ class FloatTensor():
         return self.no_params_func("acos_")
 
     def addmm_(self, x, y):
-
+        """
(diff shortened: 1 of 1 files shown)


iamtrask:
HUGE fan of this PR!!!

PR #713, opened 2017-12-19
started upgrading autograd

iamtrask:
I'm working on upgrading autograd to re-use tensors from the previous dynamic graph when performing the same operations instead of always creating new tensors


iamtrask:
For PySyft, this mostly means improving our logging so that I can debug autograd as i'm implementing it

PR #714, opened 2017-12-19
Inline Documentation - 2

CT83:
This is the completed PR of Inline Documentation.

NumPy Style on inline documentation. http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html same as the Deprecated-PySyft. #273 has been followed.
For future Documentation and Function implementations, same pattern could be followed.
Most of the definitions were picked up from PyTorch and modified.

PS
Some functions including autograd(),no_params_func() and arithmetic_operation() have not been documented as it would be best for someone more experienced to document them due to their extensive use case.

### DIFF
--- a/syft/tensor.py
+++ b/syft/tensor.py
@@ -724,7 +724,7 @@ class FloatTensor():
         ----------
         Returns
         -------
-        Integer
+        int
             int with value of size
         """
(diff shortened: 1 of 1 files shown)


iamtrask:
Woohoo!!!

PR #740, opened 2018-01-05
Reducing default epsilon for RMSProp and Adam to fit in float

kabirkhan:
# Description

1e-8  = 0.00000001 which is too many digits for a float so the value of epsilon becomes 0
which can result in nan errors with RMSProp and Adam.
Not sure if there's an easy way to use doubles in this scenario so setting epsilon to 1e-6 for now

## Type of change


iamtrask:
Great find!

PR #748, opened 2018-01-06
implement model.evaluate() function

daviddemeij:
# Description
To make it easier to evaluate the performance of different models, it's nice to have a model.evaluate() function that returns certain metrics and loss.

- Implemented model.evaluate() that returns metrics you provide (this can be custom metrics as well).
- Return the test loss

Dependent upon: https://github.com/OpenMined/OpenMined/pull/316 !

## Type of change

# How Has This Been Tested?
Tested in the Keras notebook.

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/syft/interfaces/keras/models/sequential.py
+++ b/syft/interfaces/keras/models/sequential.py
@@ -59,4 +59,7 @@ class Sequential(object):
                        log_interval=1)
 		return final_loss
 
+	def evaluate(self, x_test, y_test, verbose=False):
+		return self.syft.evaluate(x_test, y_test, verbose=verbose)
+
 
--- a/syft/interfaces/keras/models/sequential.py
+++ b/syft/interfaces/keras/models/sequential.py
@@ -1,4 +1,3 @@
-import syft
 import syft.nn as nn
 import sys
 from syft.interfaces.keras.layers import Log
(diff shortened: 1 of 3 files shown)


iamtrask:
I'm a little nervous about my merge... @daviddemeij could you take a quick look to make sure it's right?

daviddemeij:
@iamtrask I solved the merge conflict, I think it should be good now! I will also make a PR to update the keras notebook using model.evaluate()

bharathgs:
@daviddemeij `delete_after_use` I think we have to come up with some other approach for this.for now, it's ok I suppose. but it would be nice if the user has to not worry about this.

daviddemeij:
@bharath yes I agree I think there is a fix if we change the hookgraph in Unity (in FloatTensor.Graph.cs) to actually recreate a tendor if it doesn't exist anymore instead of giving that error. Also we should remove a tensor from children if it is deleted. 

I was working on this but couldn't quite get it to work. Maybe I  someone with a better understanding of the hookgraph function can help me with it.

PR #1131, opened 2018-01-13
Fix requirements and update conda install scripts

kynan:
The dependency `zmq` should be `pyzmq`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 numpy
-zmq
+pyzmq
 jupyter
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,3 +1,3 @@
 numpy
-zmq
+pyzmq
```
gavinuhma: are you sure this is necessary? I've never had a problem with this

gavinuhma: or I guess the package name in conda is different?

PR #1303, opened 2018-01-24
store jobs so you can come back later

bendecoste:
now, you can submit a job, disconnect from jupyter, and check on the
result a year later.

Example usage can be found in `notebooks/grid/Grid Example.ipynb` in


bendecoste:
image

PR #1349, opened 2018-06-20
refactored sending and recieving code as well as added composite mess…

robert-wagner:
…ages

# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* CPU:
* GPU:
* PySyft Version:
* Unity Version:

# Checklist:

### DIFF
--- a/syft/core/workers.py
+++ b/syft/core/workers.py
@@ -60,8 +60,9 @@ class BaseWorker(object):
           print events to stdout.
 
     """
+
     def __init__(self,  hook=None, id=0, is_client_worker=False, objects={},
-                 tmp_objects={}, known_workers={}, verbose=True):
+                 tmp_objects={}, known_workers={}, verbose=True, queue_size=0):
 
         # This is a reference to the hook object which overloaded
         # the underlying deep learning framework
@@ -108,6 +109,10 @@ class BaseWorker(object):
         # A flag for whether or not to print events to stdout.
         self.verbose = verbose
 
+        # A list for storing messages to be sent as well as the max size of the list
+        self.message_queue = []
+        self.queue_size = queue_size
+
     def send_msg(self, message, message_type, recipient):
         """Sends a string message to another worker with message_type information
         indicating how the message should be processed.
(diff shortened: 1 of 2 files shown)


syft/core/workers.py
```diff
@@ -122,13 +127,58 @@ def send_msg(self, message, message_type, recipient):
...
+    def compile_composite_message(self):
+        """
+        Returns a composite message in a dictionary from the message queue. Evenatually will take a recipient id.
+
+        * **out (dict)** dictionary containing the message queue compiled as a composite message
+        """
+
+        message_wrapper = {}
+
+        message_wrapper['message'] = {message_number:message for message_number,message in ennumerate(self.message_queue)}
```
iamtrask: Looks like there's a small typo here "ennumerate" -> "enumerate"

syft/core/workers.py
```diff
@@ -138,7 +188,35 @@ def receive_msg(self, message_wrapper):
           of object types. However, the object is typically only used during testing or
           local development with :class:`VirtualWorker` workers.
         """
-        raise NotImplementedError
+
+        if(is_binary):
```
jvmancuso: How does the remote worker know ahead of time if it's binary?  This seems like a good place for try/except

robert-wagner: Not sure. Copy pasted from socket worker. Everything is guaranteed to be binary. I think we can remove that

syft/core/workers.py
robert-wagner: Will fix

syft/core/workers.py
```diff
@@ -60,8 +60,9 @@ class BaseWorker(object):
           print events to stdout.
 
     """
+
     def __init__(self,  hook=None, id=0, is_client_worker=False, objects={},
-                 tmp_objects={}, known_workers={}, verbose=True):
+                 tmp_objects={}, known_workers={}, verbose=True, queue_size=0):
```
jvmancuso: I'm not sure we want to initialize workers with queue_size.  There are times where sending commands in batches is helpful for speed, and there are times where I want to be interactively querying remote workers, likely all in the same script.  The way it's written now, I'd have to run two client workers to be able to do that.  We spoke a bit about including this as a parameter in the context manager, so should we be putting it in TorchHook?

robert-wagner: If you look at the behavior of send message, if queue size is 0 ( the default) messages are sent exactly as they were before. Torch hook also needs a parameter for this. Thank you for reminding me

jvmancuso: But if this is static in the worker, wouldn't I have to reinitialize the worker or create a new worker to change the queue size?

robert-wagner: One planned future improvement is to allow for the modification of that code at runtime with a context manager/setter which changes the size

iamtrask: Could it be dynamic with the number of owners of a tensor? It seems highly likely there will be tensors with varying numbers of owners.

robert-wagner: Will add issue to make this parameter able to be a callable to take care of all of these options. In practice for the beginning it will either be size 0 or infinity

PR #1378, opened 2018-07-02
fixed_precision updated

alhparsa:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/syft/core/hooks/torch/hook.py
+++ b/syft/core/hooks/torch/hook.py
@@ -66,16 +66,16 @@ class TorchHook(BaseHook):
         super().__init__()
 
         self.local_worker = local_worker
-        if(self.local_worker is None):
+        if (self.local_worker is None):
 
             # Every TorchHook instance should have a local worker which is responsible for
             # interfacing with other workers. The worker interface is what allows the Torch
             # specific code in TorchHook to be agnostic to the means by which workers communicate
             # (such as peer-to-peer, sockets, through local ports, or all within the same process)
 
-            if(hasattr(torch, 'local_worker')):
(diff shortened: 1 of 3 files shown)


syft/core/hooks/torch/hook.py
```diff
@@ -211,18 +216,194 @@ def function_router(*args, **kwargs):
...
+
+        tensor_type.fixed_prec_sub = fixed_prec_sub
+
+        def fixed_prec_div(self, other):
+            if hasattr(self, 'fixed_precision') and hasattr(self, 'fixed_precision'):
+
+
+                if self.precision < other.precision:
+                    out = torch.LongTensor.old_div(self.old_mul(10 ** (other.precision - self.precision)), other)
+                    out.precision = 0
```
iamtrask: Is it accurate for out.precision to always be 0 here? Seems like we would want it to be either the precision of self or other, no?

alhparsa: since we have the same precision for both variables, out.precision should be 0 (10^0 = 1)
here is an example:
x = torch.FloatTensor([1,2,3,4,5]).set_precision(4)
y = torch.FloatTensor([1,2,3,4,5]).set_precision(3)
x = [10000,20000,30000,40000,50000]
y = [1000,2000,3000,4000,5000]

what the function does is first set the precision of y equal to x, so y is now [10000,20000,30000,40000,50000]. 

since they both have the same precision, dividing x by y gives us [1,1,1,1,1]. therefore, the precision of output is always 0 regardless of the either variable's precision, because of the conversion. 

hope that makes sense :)

PR #1388, opened 2018-07-05
added whoami to SocketWorker and removed redundant paramter from Base…

iamtrask:
added whoami to SocketWorker and removed redundant paramter from BaseWorker.register_object

# Checklist:


syft/core/hooks/torch/hook.py
```diff
@@ -749,7 +748,7 @@ def get_(self, reduce=lambda x: x[0]):
             x, request_obj_cleanup_method = _out
 
             hook_self.local_worker.register_object(
-                hook_self.local_worker, x, id=x.id)
+                x, id=x.id)
```
robert-wagner: Style: put this on one line or have one line for each parameter

syft/core/hooks/torch/hook.py
```diff
@@ -761,8 +760,7 @@ def get_(self, reduce=lambda x: x[0]):
                 if (x.grad is not None):
                     self.grad = x.grad
 
-            self = hook_self.local_worker.register_object(hook_self.local_worker,
-                                                          _os,
+            self = hook_self.local_worker.register_object(_os,
```
robert-wagner: Style: put on one line

syft/core/workers/socket.py
```diff
@@ -124,6 +125,9 @@ def __init__(self,  hook=None, hostname='localhost', port=8110, max_connections=
             else:
                 print("Ready!")
 
+    def whoami(self):
+        return json.dumps({"hostname": self.hostname, "port": self.port, "id": self.id})
```
robert-wagner: This is really useful for sharing your id with other worker. Do we want to have the ability to call this on any known worker so that workers can share the id of workers that they know with other workers they know? Or more specifically the client can tell the workers who the other workers are

syft/core/hooks/torch/hook.py
```diff
@@ -710,8 +710,7 @@ def send_(self, workers):
             else:
                 zeroed = self.old_set_(tensorvar_type(0))
 
-            self = hook_self.local_worker.register_object(hook_self.local_worker,
-                                                          obj=zeroed,
+            self = hook_self.local_worker.register_object(obj=zeroed,
                                                           id=self.id, owners=workers,
```
robert-wagner: style: one param per line would have missed owners here

syft/core/workers/socket.py
iamtrask: Yeah! Seems like we should. I'll add a default to BaseWorker.

iamtrask:
Hey @robert-wagner - I also added the ability to send/receive pointers to other workers. This will be useful for federated learning initiated by the data owners (client) instead of the server (model owner) wherein the client wants to send a pointer to a tensor that it owns without sending the tensor itself.

iamtrask:
Also just stumbled onto how we can do unit testing for SocketWorker objects. Will add an example

PR #1394, opened 2018-07-07
[WIP] - Hooking Refactor - moving towards Torch 1.0 and simpler, more general extensibility

iamtrask:
# Description

In this refactor, I'm looking to create several new classes which represent new tensor types: AbstractTensor, LocalTensor, PointerTensor, FixedPrecisionTensor, and TorchTensor. The current PyTorch hooking code simply appends large numbers of attributes to standard Torch objects, such that a pointer to a tensor is the same object as the tensor used to be itself... with merely the is_pointer attribute set to True. As the number of tensor types grows, this makes the hooking logic a rather large blob of code that is becoming disorganized.

The goal is for all extended functionality to be organized in classes according to its function.

LocalTensor: a class that implements all functionality for the local computation of tensors
PointerTensor: a class that implements routing functionality to tensor that exists on another machine
FixedPrecisionTensor: a class that modifies high level commands into those necessary for fixed precision operation
TorchTensor: a class which contains all custom logic to be applied to torch tensor objects.

Furthermore, instead of having to implement every permutation of pointer, fixed precision, local state into the hooking functionality, this new class base system will let us "daisy-chain" arbitrary numbers of these abstractions together. For example:

PointerTensor(localhost) -> PointerTensor(remote) -> FixedPrecision -> Pointer(another_remote) -> LocalTensor -> LongTensor 

could be a perfectly legitimate abstraction tree allowing high level (localhost) to make calls to a tensor it thinks is owned by "remote" but is actually owned by "another_remote" and operates with fixed precision.

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

This has not yet been tested as it is a work in progress.

**Test Configuration**:
* CPU:
* GPU:
* PySyft Version:
* Unity Version:

# Checklist:


syft/core/frameworks/torch/hook.py
```diff
@@ -0,0 +1,298 @@
...
+            results = list()
+            if(call_native):
+                for child in self.children:
+                    results.append(getattr(child, "native_"+attr)(*args, **kwargs))
+            else:
+                for child in self.children:
+                    results.append(getattr(child, attr)(*args, **kwargs))
+
+            for result in results:
+                if(type(result) in torch.tensorvar_types and (not hasattr(result, 'owner'))):
```
LaRiffle: What is the role of `and (not hasattr(result, 'owner'))):` ?
Isn't it `owner` -> `owners` ?
Assumingly, every element of type in `torch.tensor_types` should have an attribute owners thanks to the hook, right?

iamtrask: Ah - yes - I'm going to change tensors from having multiple owners to having single owners, where owner is less about who has control of the tensor and more about which Worker object is managing it.

syft/core/frameworks/torch/hook.py
```diff
@@ -14,23 +14,21 @@
...
 
-            if (hasattr(torch, 'local_worker')):
-                self.local_worker = torch.local_worker
-                if (verbose):
-                    print("Torch seems to already have a local_worker object... \
-                          using that one instead...")
-            else:
-                self.local_worker = workers.VirtualWorker(
-                    hook=self, is_client_worker=is_client, queue_size=queue_size)
+            self.local_worker = workers.VirtualWorker(hook=self, is_client_worker=is_client, queue_size=queue_size)
```
iamtrask: what was the thought behind removing this part?

test/tensor_serde_test.py
```diff
@@ -232,14 +232,15 @@ def test_send_and_get_tensor(self):
 
         x.send(bob, ptr_id=1234)
 
-        # the id of x should have changed to that of the pointer
-        assert x.id == 1234
+        # I just changed this
```
iamtrask: Yup - good call

syft/core/frameworks/torch/hook.py
LaRiffle: It's just that as for now the local worker is declared to be an attribute of hook and not of torch anymore, so ` -            if (hasattr(torch, 'local_worker')): ` will never be True.
Therefore, I added another torch attribute to keep track of instanciation of torchhook. Also, I tried to see in what extend we could instanciate several instances of torchhook (what parts should run several times, or only once), for us to run all the tests at once, but it revealed tricky so I dropped this topic. Nvm, could be an issue or a todo

syft/core/frameworks/torch/tensor.py
```diff
@@ -364,9 +364,18 @@ def get(self, deregister_ptr=True, update_ptr_wrapper=True):
         # still exist referencing this pointer will simply call local data instead
         # of sending messages elsewhere, or a closer pointer
         if update_ptr_wrapper:
-            # TODO: modifying self is maybe not a good practice, but it's the only way I found
-            # to ensure that x.get() which shows a tensor is in fact of x.dim() = 1 and not 0
             self.child = syft_tensor
```
iamtrask: An alternative here would be to also forward __str__ (like all other methods) to the FloatTensor that's actually holding data. This would be more performant than calling self.set_ as well.

syft/core/frameworks/torch/tensor.py
```diff
@@ -364,9 +364,18 @@ def get(self, deregister_ptr=True, update_ptr_wrapper=True):
...
+            # x = Float(...)
+            # x.send(...)
+            # x2 = x.get()
+            # We  have x2: [no dim]->[_Local]->[Float()]
+            # Whereas we expect x2: [Float()]
+            # So we use the .set_() method, to change the storage of [no dim]
+            if not isinstance(syft_tensor, sy._PointerTensor) \
+              and syft_tensor.child is not None \
+              and syft_tensor.child.dim() > 0:
+                self.set_(syft_tensor.child)
```
iamtrask: As mentioned above, this could be un-necessarily expensive. The wrapper can stay as a dim0 tensor and just forward all calls to the actual data tensor. We don't have to do it that way... just throwing it out as an option to consider.

syft/core/utils.py
```diff
@@ -1,5 +1,377 @@
...
+
+    if len(locations) > 1:
+        raise Exception('All pointers should point to the same worker')
+    if len(owners) > 1:
+        raise Exception('All pointers should share the same owner.')
+
+    return command, locations, owners
+
+
+def convert_local_syft_to_torch(obj, reverse=False, worker=None):
```
robert-wagner: It's very unclear here what reverse does. Could we rename it to be more clear/ add a doc string which explains what it does

syft/core/utils.py
```diff
@@ -1,5 +1,377 @@
...
+    # Case of basic types and slice
+    if isinstance(obj, (int, float, str, slice)) or obj is None:
```
robert-wagner: This design pattern makes this function very hard to test/ understand what it is doing. Is there a way that we can modify this to avoid all of the different if/else cases

syft/core/utils.py
```diff
@@ -1,5 +1,377 @@
...
+        return obj
+    # Tensors and Variable
+    elif isinstance(obj, tuple(torch.tensorvar_types)):
+        return obj.create_local_tensor(worker) if reverse else obj
+    # sy._SyftTensor (Pointer, Local)
+    elif issubclass(obj.__class__, sy._SyftTensor):
```
robert-wagner: The issubcalls in the middle of all the isinstances feels dirty

syft/core/utils.py
```diff
@@ -1,5 +1,377 @@
...
+        return all(rep)
+    elif isinstance(obj, slice):
+        return True
+    else:
+        logging.warning('Obj is not tensorvar', obj)
+        assert False
+
+
+def assert_has_only_syft_tensors(obj):
+    if isinstance(obj, (int, float, str)):
```
robert-wagner: same as above with all of the if/else constructs

syft/core/utils.py
LaRiffle: Yes sure! Basically if you do reverse=True, `convert_local_syft_to_torch` becomes `convert_local_torch_to_syft`. I will add a doc string, or remove it, for I'm not sure we need it after all

syft/core/utils.py
LaRiffle: Yes, we need a proper function to tell us "Oh yes this is a SyftTensor", or a Tensor/Variable

syft/core/utils.py
LaRiffle: Yes you're right, I was a bit fed up with copy/pasting the whole If/elif/else thing, but in anycase I think this function will be removed.
Oh yes good advice! I will change this in my raising Exception

PR #1395, opened 2018-07-10
Handling more sophisticated args and kwargs when calling torch func

LaRiffle:
# Description

The goal is to provide a fully working demo of federated learning on MNIST.
As for now, it is not possible to handle convolution (`F.conv2d`, `nn.conv2d`), because the args and kwargs required get lost during the hook.
Hence, we provide a stronger handling of arks and kwargs in the transmission process of commands to workers, with a special focus on the conservation of the types not natively supported in JSON.

See the **working example on MNIST** : https://colab.research.google.com/drive/1DxjiUjLUenFaVBrz3QMuJ665Z2HsfmgA

## Type of change

- [X] This change requires a documentation update

# How Has This Been Tested?

Unit tests added on convolution

**Test Configuration**:
Colab config

# Checklist:

### DIFF
--- a/syft/core/utils.py
+++ b/syft/core/utils.py
@@ -1,10 +1,11 @@
 """Framework agnostic static utility functions."""
-import functools
-
 import json
 import re
-import torch
 import types
+import functools
+import logging
+
+import torch
 
 class PythonEncoder():
     """
@@ -67,12 +68,11 @@ class PythonEncoder():
             }
         # Generator (transformed to list)
         elif isinstance(obj, types.GeneratorType):
-            print('Generator args can\'t be transmitted')
+            logging.warning("Generator args can't be transmitted")
             return []
         # Else log the error
         else:
-            print('Unknown type', type(obj))
-            return None
+            raise ValueError('Unhandled type', type(obj))
 
 class PythonJSONDecoder(json.JSONDecoder):
     """
--- a/syft/core/utils.py
+++ b/syft/core/utils.py
@@ -1,6 +1,97 @@
 """Framework agnostic static utility functions."""
 import functools
 
+import json
+import re
+import torch
+
+class PythonEncoder():
+    """
+        Encode python and torch objects to be JSON-able
+        Torch objects are replaced by an id.
+        Note that a python object is returned, not JSON.
+    """
+    def __init__(self, retrieve_tensorvar=False):
+        self.retrieve_tensorvar = retrieve_tensorvar
+        self.found_tensorvar = []
+        self.tensorvar_types = tuple([torch.autograd.Variable,
+                                     torch.nn.Parameter,
+                                     torch.FloatTensor,
+                                     torch.DoubleTensor,
+                                     torch.HalfTensor,
+                                     torch.ByteTensor,
+                                     torch.CharTensor,
+                                     torch.ShortTensor,
+                                     torch.IntTensor,
+                                     torch.LongTensor])
+
+    def encode(self, obj, retrieve_tensorvar=None):
+        if retrieve_tensorvar is not None:
+            self.retrieve_tensorvar = retrieve_tensorvar
+        if self.retrieve_tensorvar:
+            return (self.python_encode(obj), self.found_tensorvar)
+        else:
+            return self.python_encode(obj)
+
+    def python_encode(self, obj):
+        if isinstance(obj, (int, float, str)) or obj is None:
+            return obj
+        elif isinstance(obj, self.tensorvar_types):
+            if self.retrieve_tensorvar:
+                self.found_tensorvar.append(obj)
+            key = '__'+type(obj).__name__+'__'
+            return { key: '_fl.{}'.format(obj.id) }
+        elif isinstance(obj, (tuple, set, bytearray)):
+            key = '__'+type(obj).__name__+'__'
+            return {key:[self.python_encode(i) for i in obj]}
+        elif isinstance(obj, list):
+            return [self.python_encode(i) for i in obj]
+        elif isinstance(obj, dict):
+            return {
+                k: self.python_encode(v)
+                for k, v in obj.items()
+            }
+        else:
+            try:
+                return { '__eval__': str(obj) }
+            except:
+                print('Unknown type', type(obj))
+            return None
+
+class PythonJSONDecoder(json.JSONDecoder):
+    """
+        Decode JSON and replace python types when need
+        Retrieve Torch objects replaced by their id
+    """
+    def __init__(self, worker, *args, **kwargs):
+        super(PythonJSONDecoder, self).__init__(*args,
+            object_hook=self.custom_obj_hook, **kwargs)
+        self.worker = worker
+        self.tensorvar_types = tuple([torch.autograd.Variable,
+                                     torch.nn.Parameter,
+                                     torch.FloatTensor,
+                                     torch.DoubleTensor,
+                                     torch.HalfTensor,
+                                     torch.ByteTensor,
+                                     torch.CharTensor,
+                                     torch.ShortTensor,
+                                     torch.IntTensor,
+                                     torch.LongTensor])
+
+    def custom_obj_hook(self, dct):
+        pat = re.compile('__(.+)__')
+        for key, obj in dct.items():
+            try:
+                obj_type = pat.search(key).group(1)
+                if obj_type in map(lambda x: x.__name__, self.tensorvar_types):
+                    pattern_var = re.compile('_fl.(.*)')
+                    id = int(pattern_var.search(obj).group(1))
+                    return self.worker.get_obj(id)
+                elif obj_type in ('tuple', 'set', 'bytearray', 'eval'):
+                    return eval(obj_type)(obj)
+            except AttributeError:
+                pass
+        return dct
 
 def map_tuple(hook, args, func):
     if hook:
--- a/syft/core/utils.py
+++ b/syft/core/utils.py
@@ -8,7 +8,7 @@ import torch
 class PythonEncoder():
     """
         Encode python and torch objects to be JSON-able
-        Torch objects are replaced by an id.
+        In particular, (hooked) Torch objects are replaced by their id.
         Note that a python object is returned, not JSON.
     """
     def __init__(self, retrieve_tensorvar=False):
@@ -26,6 +26,10 @@ class PythonEncoder():
                                      torch.LongTensor])
 
     def encode(self, obj, retrieve_tensorvar=None):
+        """
+            Performs encoding, and retrieves if requested all the tensors and
+            Variables found
+        """
         if retrieve_tensorvar is not None:
             self.retrieve_tensorvar = retrieve_tensorvar
         if self.retrieve_tensorvar:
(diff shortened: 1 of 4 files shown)


syft/core/utils.py
```diff
@@ -1,6 +1,97 @@
 """Framework agnostic static utility functions."""
 import functools
 
+import json
+import re
+import torch
+
+class PythonEncoder():
```
iamtrask: golden - love this

iamtrask: one thing we want to ensure is that people can't accidentally inject their own dependencies or arbitrary python code by sending something special through JSON.

iamtrask: (not saying your code doesn't... just as a heads up while you're working)

LaRiffle: Thank you for your feedback !
I'm trying to find a case where there would be a security problem, I have this one:
```
encoder = utils.PythonEncoder(retrieve_tensorvar=True)
decoder = utils.PythonJSONDecoder(alice)
enc = [{'__eval__': 'any([exec(\'self.leak = [str(obj) for k, obj in self.worker._objects.items()]\')])'}]
enc = json.dumps(enc)
dec = decoder.decode(enc)
print(decoder.leak)
```

But as this example points out, it can get quite tricky to detect malicious functions that are hidden behind *acceptable* ones.

LaRiffle: Well, the basic functions I would like to preserve are the iterables `enumerate, filter, list_iterator, map, range, zip` and also, `slice` (many other functions such as `any` are not useful as they should be evaluated directly so they can't be jsonfied). So what I can do it is specifically consider them, but then all other functions will be discarded.

LaRiffle: Yes I agree this is not optimal. Either we could make a special case for "slice" and say ok for this one we do "eval" but only this one, or we figure an other way to json-ify it and we don't need to use "eval" for now.

iamtrask: My heart wants to merge this - but i'm just fearful we'll forget about the eval() and it'll come back to bite us later. Do we think we have a fair workaround? (@jvmancuso )

LaRiffle: Is it ok if I hardcode that eval can only be used with `slice` ?

LaRiffle: (I can do it even more strongly than in my last commit 037d65)

syft/core/utils.py
```diff
@@ -1,6 +1,97 @@
...
+    def custom_obj_hook(self, dct):
+        pat = re.compile('__(.+)__')
+        for key, obj in dct.items():
+            try:
+                obj_type = pat.search(key).group(1)
+                if obj_type in map(lambda x: x.__name__, self.tensorvar_types):
+                    pattern_var = re.compile('_fl.(.*)')
+                    id = int(pattern_var.search(obj).group(1))
+                    return self.worker.get_obj(id)
+                elif obj_type in ('tuple', 'set', 'bytearray', 'eval'):
```
jvmancuso: This line is a big problem -- eval on eval can for sure execute arbitrary code

syft/core/utils.py
robert-wagner: We should not call eval ever. It's a security risk

robert-wagner: We should not call eval ever. It's a security risk

iamtrask: I think we can have a workaround. Basically... whatever code you would "eval"... cache that python object in a local dictionary where the key is the string version.

aka... if you wanted to call "print"

guard = {}
guard['print'] = print

....

def receive_python_code_to_execute(function, params):
      return guard[code](params)

receive_python_code_to_execute("print", "hello world")

iamtrask: see how that can allow us to execute arbitrary python code without ever calling exec?

jvmancuso: Originally I'd used `eval` on all torch commands because, after thinking it through, there was no way to inject arbitrary code through it if you check to make sure everything's a legit command under the `torch` module.  (This is because you'd have to sneak that code into their PyTorch install, which means you've already got the ability to execute arbitrary code.)

jvmancuso: But it _is_ slow, so even if you can guarantee security, using a lookup is preferable .

jvmancuso: And also I'd add that the plan was alway to remove eval when we could, and it was originally only in there for convenience while prototyping :grin:

jvmancuso: Also, this could all be handled by switching from JSON to something more extensible like protobuf or some other format more suitable for byte streaming.

LaRiffle: Good idea ! I'll get rid of `eval` using your suggestions then.
Regarding protobuf or equivalent, sure it could be better, though for now I just wanted to fix the things with the args/kwargs so I tried to do as close as possible to the initial implementation. But this is definitefy something to explore.

LaRiffle: I just came across this, which is part of a unit test
```
model = nn.Linear(2, 1)
p = model.parameters() # is a generator
opt = optim.SGD(params=p, lr=0.1)
```
`optim.SGD` calls at a certain point `torch.is_tensor(params)` which we have hooked. The Json encoding implies that either we discard the generator (which we don't want even if in this case it would work), either we iterate on the `params` generator which is empty at the time `is_tensor` has returned, so the next lines of `optim.SGD` trigger an error.
```
param_groups = list(params)
if len(param_groups) == 0:
     raise ValueError("optimizer got an empty parameter list") 
```
We could have call `opt = optim.SGD(params=list(p), lr=0.1)` at the beginning but this is painful for users.
Do you have some suggestion on how we could deal with this ?

jvmancuso: Can you send the full stacktrace?  I'm thinking a specialized hook for `is_tensor` is in order, although I'm not 100% on the error you're describing.

syft/core/utils.py
LaRiffle: Yes sure ! But this is not the last commit, it is the previous one. On the last commit, there is a strong focus on limiting the things we eval

syft/core/utils.py
LaRiffle: It's a short one, that's why I took time to figure out what's under the hood
```
  File "/Users/ryffel/Documents/Code/PySyft/test/torch_test.py", line 328, in test_federated_learning
    opt = optim.SGD(params=model.parameters(), lr=0.1)
  File "/anaconda3/lib/python3.6/site-packages/torch/optim/sgd.py", line 57, in __init__
    super(SGD, self).__init__(params, defaults)
  File "/anaconda3/lib/python3.6/site-packages/torch/optim/optimizer.py", line 34, in __init__
    raise ValueError("optimizer got an empty parameter list")
ValueError: optimizer got an empty parameter list

```

jvmancuso: >We could have called opt = optim.SGD(params=list(p), lr=0.1) at the beginning but this is painful for users.
Do you have some suggestion on how we could deal with this ?

I don't think there's a clean way to do this. It's strange that is_tensor is called on a whole generator and not something like 
`torch.is_tensor(p) for p in params`.

One possibility is to provide a special hook for is_tensor that checks for a generator arg and casts the generator to a list or something else we can handle before sending off the RPC, although that still seems like a dirty solution.

LaRiffle: Yes, as I've been thinking about it, maybe it's beyond the scope of this PR.
But we have a real issue with generators, because we can think of many examples where they could be used and where we would be blocked just as in the last example, because the real problem is that `params` is provided to `is_tensor()` but there is no return, and the code doesn't expect `is_tensor()` to iterate/exploit `params`. So even if we cast the generator args of `is_tensor()`in a list, we will empty `params` for the functions using it after `is_tensor()`.
And in this case it's not very well implemented in pytorch because at the end params is indeed casted into a list, but a bit to late.

jvmancuso:
FYI this should close #1396 when done

syft/core/utils.py
jvmancuso: In general, we can worry about generators containing tensors/variables on a case by case basis, since right now I don't think there are any instances of places where we've overloaded functions that are called on those.

I'm still not sure why `params` comes back empty.  `is_tensor` just checks for isinstance(x, torch.Tensor), which should return False, so that line should just continue.  We're initializing the Optimizer on the client, so that should be an iterable of Parameters or pointers to Parameters -- it should never be empty.  Do you have a specific test that's failing?

CC @iamtrask wasn't the optimizer working during one of your twitch streams?  is this a new bug?

LaRiffle: Actually,  maybe this is not a desirable behaviour, but at a certain point the call to `is_tensor` is serialized, so `params`is iterated over. Hence, when `is_tensor` performs his check `isinstance(x, torch.Tensor)` it already has a list.
This behaviour is linked to the fact that we inspect args and kwargs to retrieve all the tensors. So we inspect the generator, namely we iterated over it, and therefore it is then empty.

LaRiffle: I suspect there was previously a bug when transmitting the arguments, leading to `torch.is_tensor(params)` returning `False` as expected when calling the optimizer, but because it was given empty or incorrect arguments

jvmancuso: That sounds likely.  Can we verify it?  If that's the case, there should be a way to restart the generator, or better yet to fix up the code so that we avoid iterating over it before it's cast to list in `Optimizer.__init__`.

jvmancuso: It seems like the build has been passing, so I think if this were a problem, the unit test you were showing would have broken.

LaRiffle: I'm trying to find an appropriate test.
Which unit test are you referring to ?

jvmancuso: This guy: https://github.com/OpenMined/PySyft/blob/a5b462eba1533974e79682ffb8d7d071a3f6dda8/test/torch_test.py#L459

LaRiffle: Ahah I'm completely confused now! I can't find any example relevant. Still searching

bartimaeus12: that test fails from time to time. trask said that the random numbers havent been seeded.

LaRiffle: Yes @bartimaeus12, also one could extend the number of iterations (originally 3) to lower the probability of fail (I've done this in this PR)

LaRiffle:
Apparently I made the comments disappear by rebasing my fork on the latest updates of master.
However, the `generator` issue we have mentioned should perhaps not be considered in this PR and be handled separately, with a proper unit test etc.
Please tell me if you have any second thoughts on the changes, otherwise I'm done :)

syft/core/utils.py
jvmancuso: Why don't we just use a seed? There's a function called torch.manual_seed for that

LaRiffle: Yes right, but I'm not familiar with Pytorch, wasn't my test, so you know, I took the easy way^^ 
(Btw I run successfully 100/100 tests of test_federated_learning with this dirty fix, compared to 97/100 before, but yeah it's dirty !)

jvmancuso: Hmm, I suppose if we're testing for overall loss reduction, that's an okay fix.  (I'd be happier if it had a 1e-6 probability of failing but oh well...)  I do think that ultimately we want several subtests for this test, as when it fails we'd want to know why it failed and right now it could pretty much be anything in the stack.  So we'd want a test for sending modules, a test for remote execution of a model, a test for doing `backward`, a test for Optimizer.step on a remote model.  I'll put an issue up.

LaRiffle: Yes this would be more appropriate !

jvmancuso: #1398 🙂

PR #1399, opened 2018-07-13
Seeded test_federated_learning and added test for remote optim.step()

### LINKED ISSUE #1398: Seed `test_federated_learning` and spin out specific unit tests
## Issue
I'm proposing that we (1) seed `test_federated_learning`, and (2) spin several smaller tests out of `test_federated_learning`.  I believe we'd want a test for sending modules, a test for remote execution of a model, a test for doing `backward`, and a test for calling Optimizer.step on a remote model.  If any of these tests exist already, they can be ignored.

## Current Behavior

The federated learning test creates a model and some data, sends the data/model out to virtual workers, performs a few iterations of backprop, then checks to see if the last loss is less than the first loss.  One problem with this is that it requires high numbers of iterations to be confident that it'll pass.  In general, there will always be a small but non-zero chance that it'll fail, and tests should be deterministic.  This can be fixed by seeding at the beginning of the test, and instead of checking to see if the loss has decreased, checking for exact loss values on the second iteration.  This makes the test deterministic and also avoids extra compute time during builds.

Additionally, in those cases where it doesn't currently fail by chance, we don't have any insight as to why it failed -- it's testing too much at once.  Ideally, when it fails, there should be other tests that cover some of the same operations that are more specific, so that we know why it failed without having to dig into it and try to reproduce the failure.

bartimaeus12:
# Description
Seeded test_federated_learning and added unittest for remote optim.step()

Fixes #1398

## Type of change

Added unittests.

# How Has This Been Tested?

All tests pass.

# Checklist:

### DIFF
--- a/test/torch_test.py
+++ b/test/torch_test.py
@@ -488,8 +488,39 @@ class TestTorchVariable(TestCase):
         assert model.grad.id in remote._objects
         assert model.grad.data.id in remote._objects
 
+    def test_remote_optim_step(self):
+
+        hook = TorchHook(verbose=False)
+        local = hook.local_worker
+        local.verbose = False
+        remote = VirtualWorker(id=1, hook=hook, verbose=False)
+        local.add_worker(remote)
+
+        data = Var(torch.FloatTensor([[0, 0], [0, 1], [1, 0], [1, 1]])).send(remote)
+        target = Var(torch.FloatTensor([[0], [0], [1], [1]])).send(remote)
+
+        param = []
+
+        for i in range(2):
+            torch.manual_seed(42)
+            model = nn.Linear(2, 1)
+            opt = optim.SGD(params=model.parameters(), lr=0.1)
+            model.send_(remote)
+            model.zero_grad()
+            pred = model(data)
+            loss = ((pred - target) ** 2).sum()
+            loss.backward()
+            opt.step()
+            for i in model.parameters():
+                param.append(i.get())
+            model.reset_parameters()
+
+        assert (param[0] == param[2]).all()
+        assert (param[1] == param[3]).all()
+
     def test_federated_learning(self):
 
+        torch.manual_seed(42)
         hook = TorchHook(verbose=False)
         me = hook.local_worker
         me.verbose = False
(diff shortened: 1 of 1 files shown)


test/torch_test.py
```diff
@@ -488,8 +488,39 @@ def test_send_var_with_gradient(self):
...
+            opt = optim.SGD(params=model.parameters(), lr=0.1)
+            model.send_(remote)
+            model.zero_grad()
+            pred = model(data)
+            loss = ((pred - target) ** 2).sum()
+            loss.backward()
+            opt.step()
+            for i in model.parameters():
+                param.append(i.get())
+            model.reset_parameters()
```
jvmancuso: Hey, I think the point of this test is to make sure that the parameters change after doing all this. In particular, unless I'm missing something, I think it's best to remove this line and make sure that the parameters from the second iteration _don't_ equal those from the first.

jvmancuso: Other than that, looks good!

jvmancuso:
Oh, also, can you make sure that instead of checking that the federated learning test lowers the loss, it equals a specific value? That should be possible after seeding, and it's a more deterministic test.

test/torch_test.py
bartimaeus12: Actually i was testing that two models with same initial weights have the same updates.
but now i have changed it as you have specified.

bartimaeus12:
All changes done.

test/torch_test.py
```diff
@@ -488,8 +488,42 @@ def test_send_var_with_gradient(self):
...
+        loss = ((pred - target) ** 2).sum()
+        loss.backward()
+        opt.step()
+
+        model.get_()
+        for i in model.parameters():
+            param.append(i[:])
+
+        assert not ((param[0] == param[2]).all())
+        assert not ((param[1] == param[3]).all())
```
iamtrask: Could these assertions be specific values as well as opposed to simply checking for equality. It could break in all sorts of terrible ways (param[2] could equal -inf, NaN, or None for example) and this test would still pass. Requiring it to equal a specific value, however, makes the test much more precise

iamtrask:
Excellent work!!

PR #1401, opened 2018-07-15
[pointers_are_tensors] Add move function (with pb of extra registering..)

LaRiffle:
# Description

I added the previously discussed move() function which moves the end chain Tensor to another worker: `self->alice->obj [worker] => self->alice->worker->obj`
This functionality differs of send() which would perform: `self->alice->obj [worker] => self->worker->alice->obj`

However there is still an issue with undesired Tensors being created  (see `Forward or Move tensors.ipynb`), which is due to the ability of `self.child` to create arbitrarily tensors, I don't know how we could handle this at the moment.

## Type of change

# How Has This Been Tested?

No tests so far, just use `Forward or Move tensors.ipynb`

### DIFF
--- a/examples/experimental/Forward or Move tensors.ipynb
+++ b/examples/experimental/Forward or Move tensors.ipynb
@@ -37,7 +37,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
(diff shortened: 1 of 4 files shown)


iamtrask:
Excellent work sir! Will investigate the issue you reported. Extraneous tensor registration has been a persistent issue in this project.

PR #1409, opened 2018-07-19
WebSocket for workers (work in progress)

alhparsa:
This pull request's focus is to replace the existing socket worker with Websocket workers so it can be used for syft.js.


iamtrask:
This is looking really nice! Really coming together!

iamtrask:
Fantastic work and really fast too!

PR #1413, opened 2018-07-21
Context manager for TorchHook is back!

jvmancuso:
Previously, the `__enter__` and `__exit__` methods were deleted from TorchHook.  They're back, and they're better than ever!  In particular, they no longer assume that torch was unhooked before the TorchHook was initialized.  In addition, I wrapped all the various hooking logic from `__init__` into it's own function, preparing for an API change where you have to imperatively establish the hooks instead of having them happen whenever you instantiate TorchHook.

### DIFF
--- a/syft/core/hooks/base.py
+++ b/syft/core/hooks/base.py
@@ -7,10 +7,10 @@ class BaseHook(object):
     def __init__(self):
         pass
 
-    # TODO: decorate with abstractmethod after TorchHook is extended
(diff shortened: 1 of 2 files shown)


iamtrask:
OOOOHHHHH YYYEEEAAAHHH!!!!!

iamtrask:
Odds of this breaking any of our existing notebook demos?

PR #1431, opened 2018-07-29
Minor Tweaks in docs theme and better looking docs on phones

kamathhrishi:


iamtrask:
Excellent work @kamathhrishi !!!

PR #1432, opened 2018-07-30
[Pointers are tensors] Playing with PlusIsMinus Tensor : full support on Torch Tensors

LaRiffle:
# Description
Enable support on functions and methods on local and remote tensor non-trivial chains (ie including a PlusIsMinusTensor) 

The process chain is a bit changed:
The idea that we should always send a wrapper on even a partial chain is relativised: we still do this when sending commands / objects to another worker, because it is useful for a worker receiving an object for instance to have directly the wrapper available (just as it is convenient wor the local_worker, i.e. us); but within a local chain (ie Float > _PlusIsMinus > _Pointer), the intermediate nodes don't need the wrapper to execute correctly, and this enable to more easily overload methods as well, for the head element (of e.g. command['self']) is of the same type than the class in which handle_call is performed.

## Type of change

# How Has This Been Tested?

Unittest

# Checklist:

### DIFF
--- a/syft/core/frameworks/torch/hook.py
+++ b/syft/core/frameworks/torch/hook.py
@@ -400,7 +400,7 @@ class TorchHook(object):
 
     def _execute_call(hook_self, attr, self, *args, **kwargs):
         """
-        Forwar the call to the local_worker
+        Forward the call to the local_worker
         """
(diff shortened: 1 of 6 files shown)


iamtrask:
WOOHOO!!!

PR #1458, opened 2018-08-29
Mpc tensor

robert-wagner:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* CPU:
* GPU:
* PySyft Version:
* Unity Version:

# Checklist:


syft/core/frameworks/torch/tensor.py
```diff
@@ -529,6 +531,79 @@ def abs(self):
...
+         super().__init__(child=None, parent=parent, torch_type=torch_type, owner=owner, id=id,
+                         skip_register=skip_register)
+         self.pointer_tensor_dict = pointer_tensor_dict
+    
+    @classmethod
+    def handle_call(cls, syft_command, owner):
+        commands = torch_utils.split_to_pointer_commands(syft_command)
+        result_dict = {worker_id:_PointerTensor.handle_call(commands[worker_id], owner) for worker_id in commands}
+        #TODO: @trask @theo could you take a look at this if you have better ideas on how to get these parameters
+        return _GeneralizedPointerTensor(result_dict, None, None, id=None, owner=owner, skip_register=False)
```
iamtrask: The attributes for the new _GeneralizedPointerTensor should be either created or pulled from result_dict. 

parent - Normally, the "parent" would come from the high level torch object wrapping the result that's returned to you, but since you'll have multiple, I suppose the only thing to do is to pick one at random. 

torch_type - you should be able to get this from any of the answers in result_dict, since they should all be identical (optional: a check to ensure that they are in fact identical)

id - this should be initialized by _GeneralizedPointerTensor automatically when it calls super().__init__ on line 596. Might be good to check that it is though.

skip_register - i think setting this to False makes sense... (which you have done)

iamtrask: @LaRiffle feel free to comment

LaRiffle: Just one point regarding the "parent", the usual case would be, as far as I can imagine
```
x = sy.FloatTensor([..])
x = sy._MPCTensor().on(x)
x.send([alice, bob])
```
And _MPCTensor would use under the hood the _GeneralizedPointerTensor. Since the parent of _MPCTensor is clear, _GeneralizedPointerTensor should inherit the same one I think.

iamtrask: Agreed - although I think that in the specific line i was referring to... a method is being called generating a new _GPT class which doesn't necessarily have a parent yet.

robert-wagner:
It should. Will close and recreate

PR #1468, opened 2018-09-05
WIP - search/query functionality on custom tensor IDs

iamtrask:


examples/toy/SocketWorker Client.ipynb
```diff
@@ -181,7 +182,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 10,
+   "execution_count": 9,
```
robert-wagner: Could you rerun these so that each cell is run once in order?

syft/core/workers/socket.py
```diff
@@ -162,6 +163,20 @@ def listen(self, num_messages=-1):
             finally:
                 connection.close()
 
+    def search(self, query='#boston'):
```
robert-wagner: I dont think having a default for query is good. Put the default in the demo

test/torch_test.py
```diff
@@ -192,6 +192,7 @@ def test_plus_is_minus_backward_remote(self):
         x.child = x.child.child
 
         target = sy._PlusIsMinusTensor().on(torch.FloatTensor([1, 1]))
+        # target = torch.FloatTensor([1, 1])
```
robert-wagner: delete this or uncomment it out

Issue #1509, opened 2018-09-19
Implement a new SecureNN Tensor


iamtrask:
In this issue, we seek to have a new PyTorch multi-party computation tensor leveraging the protocol from [SecureNN](https://eprint.iacr.org/2018/442.pdf). The goal of this project is to give fast support for comparisons which we can use to bootstrap higher level functionality (such as ReLU). The new tensor should extend all functionality which already exists in MPCTensor (which is hopefully renamed SPDZTensor by now), and simply add new functionality for fast comparisons.

PR #1533, opened 2018-09-20
WIP - SecureNN Integration

iamtrask:
This pull request is for the new SecureNN functionality which will equip our MPC system with the ability to perform private comparisons, as documented in https://github.com/OpenMined/PySyft/issues/1509


iamtrask:
I merged the changes we've made from Master into the branch. Note that because multiplication is slightly different for snn (additional mask), I split it out into two multiplication methods... one being the normal spdz multiplication and the other for the snn version.

PR #1537, opened 2018-09-21
[WIP] Federated learning with tensorflow demo

### LINKED ISSUE #1522: Federated Learning with Tensorflow Demo
### User Story:
Add a TensorFlow demo using PySyft's virtual workers and the custom federated optimizer.

## Expected Behavior

Have a demo showing usage of the custom federated optimizer and the PySyft's workers together.

### Relevant Literature or Examples (for features)
Something similar to [this](https://github.com/coMindOrg/federated-averaging-tutorials/blob/master/advanced_federated_classifier.py) example.

ceballosiker:
Federated learning demo with VirtualWorkers and tf estimator API

# Description

Added tensorflow demo with tensorflow's estimator API and VirtualWorkers.

Fixes   #1522 

# Checklist:

### DIFF
--- a/examples/other/experimental/Tensorflow Federated Learning/virtual_worker_tensorflow.py
+++ b/examples/other/experimental/Tensorflow Federated Learning/virtual_worker_tensorflow.py
@@ -1,0 +1,105 @@
+# TensorFlow and tf.keras
+import tensorflow as tf
+from tensorflow import keras
+
+# Import pysyft
+import syft as sy
+
+# Helper libraries
+import numpy as np
+import os
+
+BATCH_SIZE = 32
+EPOCHS = 5
+
+x = sy.array([1]) # Why do I need this in order for local_worker to exist?
+
+alice = sy.VirtualWorker(id="alice") # If I get here without calling sy.array first it raises an error
+bob = sy.VirtualWorker(id="bob")
+
+workers = [bob, alice]
+
+fashion_mnist = keras.datasets.fashion_mnist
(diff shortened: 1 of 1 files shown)


ceballosiker:
Still not sure if the estimator API is the best choice for this, but it is worth trying as it simplifies a lot the usage of tensorflow.

Thoughts:
* Main problem right now is figuring out how to combine the tensorflow training pipeline with the architecture of the virtualworkers. We can't directly send the model to another worker as we do with pytorch, so it is hard to see how the distributed training happens in a single script.
* I think it will be best if we just train them independently and use the overloaded numpy to send arrays between workers. Similar to what I did [here](https://github.com/coMindOrg/federated-averaging-tutorials/blob/master/federated-MPI/mpi_basic_classifier.py) using MPI

iamtrask:
I really like the numpy overloading idea - would this still allow tensorflow's internal API to trigger the sending/receiving?

iamtrask:
So I think we could actually use VirtualWorkers to setup the distributed graph. As I understand it, there's a considerable amount of coordination required to get everyone more or less "on the same page" with the graph training on the right ports, with the right dataset, and the right algorithm. We could just use the VirtualWorkers interface to coordinate the initialization of the distributed graph and then let TF do the heavy lifting after that.

In the long run, we could even extend the VirtualWorkers to be able to create secure port forward connections so that all of TF's communications over the wire would be encrypted

iamtrask:
Any luck @ceballosiker?

iamtrask:
happy to reopen this if necessary but it seems that this PR is dormant

Issue #1542, opened 2018-09-21
Tensors shouldn't store information about commands


LaRiffle:
## Context
The _PlusIsMinusTensor has been created as a canonical example of what can be done by extending the SyftTensor.However, it does not make sense that we modify the behaviour of commands from an argument.

This underlines that the architecture is very much argument oriented, which is excellent when we send tensor or transform them in MPCTensor, but rather unefficient and complicated when we want to execute command on them.

Idea: using contexts defining workers involved and mpc. PR #1543 https://github.com/OpenMined/PySyft/pull/1543 gives extended explanation of a possible solution.

A typical example would be:
```
x = sy.FloatTensor([1.2])

#x.send(bob) #<- This is what you don't need to do

with sy.session(bob):
    # Here all the commands and ensors are sent to bob
    z = torch.add(x, x)

#z.get() or x.get() #<- This is what you don't need to do

print(z) # This is FloatTensor
```

LaRiffle:
^^^^ I just added an example of what I meant by context.

iamtrask:
I assume we'll be leaving in .send() and .get() as backups? There are some cases when it will be very challenging to use the indented case (such as when you're working with tensors in multiple places at once or pointers to pointers.

LaRiffle:
Exactly, .send() and .get() will be kept. Sending a tensor to bob before opening a context session with bob for example means that when the session ends, that tensor still is at bob (example: the data when doing training over epochs).
If you have two remote datasets at bob and alice as it is the case in most of our examples, you would probably iterate on each batch, and depending of whom stores it, you would open the appropriate remote session.

PR #1543, opened 2018-09-22
[WIP] Context session

LaRiffle:
References Issue #1542

## Description

We introduce the notion of Session. A session is a global object which holds information about the context in which  computations should be performed. For example, it can say that the computations should be sent to worker Bob, or parallized across workers, secured using MPC, logged, etc.). As for now, we only support sending computation to a single worker.

This notion is introduced to move the context information which were previously enclosed in syft chains. For example, the PlusIsMinusTensor (a tensor informing that all `add` op touching it should become `sub`) could presumably be more clear as a PlusIsMinusSession, same for a LogSession in some contexts.

The other interesting benefits are that tensors involved in a remote session will be sent for the duration of this session et fetch back at the end automatically, and all remote pointers created in this processed are get() and destroyed after the session has ended (the pointers existing before the session are not altered), which makes echo to the issue of stale remote data.

This notion of Syft sessions complements the notion of Syft tensors on the notion of contexts of computation. Syft Tensor are still supposed to represent tensor transformation such as sending, or sharing for instance, while sessions represent transformations on computations.

### Example 1 (working)
```
x = sy.FloatTensor([1.2])

#x.send(bob) #<- This is what you don't need to do

with sy.session(bob):
    # Here all the commands and ensors are sent to bob
    z = torch.add(x, x)

#z.get() or x.get() #<- This is what you don't need to do

print(z) # This is FloatTensor
```
Output:
```
INFO:Session started with bob, (secure: False)
INFO:bob:Executing cmd torch.add
INFO:Closing session

 2.4000
[syft.core.frameworks.torch.tensor.FloatTensor of size 1]
```

### Example 2 (working)

```
x = sy.FloatTensor([1.1])
x.fix_precision()
#x = x.eval() <- If you don't eval, no pb the fixed_precision will be set only in the session (try uncommenting)
chain_print(x)

with sy.session(bob):
    # Here all the commands and ensors are sent to bob
    z = torch.add(x, x)
    
chain_print(z)
print(z)
```

### Example 3 (dream)

```
# workers connected are (alice, bob, charlie) who compute things, 
# and health_corp who has awesome ML models

data = Dataset('my-3D-heart-image')

with sy.MultiPartySession(parties=(bob, alice, charlie)):
    model = health_corp.search('#heart-disease-model')  # model is searched at health_corp
    # which returns a MPCTensor shared between (bob, alice, charlie)

    pred = model(data) # data is automatically is shared with MPC, and the result 
    # pred is also a MPCTensor
    
    # End of session: try to fetch model and pred: succeeds depending of the 
    # privacy of the tensor.
    # For instance, model.get() fails as I don't own the model, but pred owner 
    # is the data owner, it's me, so pred.get() succeeds.
    
print(pred) # OMG I have no heart disease!
```

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


iamtrask:
This is a very cool initiative!

robert-wagner:
I think this is a really cool idea which has popped up a couple times. It makes a lot of sense to me that the user should only have to add 1 to their code to have it function in mpc setup. I think we should strive to have it be as close as possible to being 
```
with mpc(workers):
    pytorch_as_if_everything_is_local
```

iamtrask:
Something worth mentioning - I would strongly endorse avoiding the "with" session based logic as a hard requirement. It can be convenient if you're just doing a simple 2-party manipulation, but it is very likely we will be doing vastly more complex MPC interactions.

A good example here is a MPC capable pytorch optimizer - in this case there are tensors shared between a wide variety of different n-person sets (might be just 2... might be more). A session would be very frustrating way of doing this (it would be taking the most beautiful aspect of PyTorch and making it Tensorflowy... which is bad)

iamtrask:
This is why I prefer the explicit .send() and .get()... with all the code between automatically working. It's the same idea as .session() except that I can do other operations IN BETWEEN session operations. For example
`x = sy.FloatTensor([1.1])
x.fix_precision()
#x = x.eval() <- If you don't eval, no pb the fixed_precision will be set only in the session (try uncommenting)
chain_print(x)

with sy.session(bob):
    # Here all the commands and ensors are sent to bob
    z = torch.add(x, x)
    
    # what if i need to do something with z here which interacts with someone other than bob?
    # but ultimately comes back to bob?.... if I just... end the session... presumably the digital assets
    # will be returned to my context (maybe i shouldn't have permission to do this)? .. when in fact
    # what i actually want is for the assets bob has to be sent directly to another worker

chain_print(z)
print(z)`


with

iamtrask:
Yeah - I think the use case where session breaks down is when you need to coordinate arbitrary sending and receiving of tensors that aren't the local machine. Aka - sometimes the local python worker won't actually be sending or receiving any shares... but they still need to coordinate how other parties send and receive shares. .session() is a very constraining way to do this because it assumes that some type of ownership will be consistent for the whole session. However, in SecureNN and Secure Aggregation (for Federated Learning) alone we have already seen that this is not the case.

iamtrask:
It seems that we're going to address this issue instead in the PyTorch 1.0 refactor - thus actually working from this PR would likely be quite challenging.

PR #1568, opened 2018-09-27
Sigmoid

robert-wagner:
This does not work because when multiplying tensors by themselves we get a ('All arguments should share the same child type.', [<class 'syft.core.frameworks.torch.tensor.LongTensor'>, <class 'syft.core.frameworks.torch.tensor._MPCTensor'>]) error. I have no clue how to debug this in the current pysyft. @iamtrask @LaRiffle If you could take a look at it that would be great. This is the last step


examples/other/experimental/new spdz demo.ipynb
```diff
@@ -43,10 +43,29 @@
   {
    "cell_type": "code",
    "execution_count": 3,
+<<<<<<< HEAD:examples/other/experimental/new spdz demo.ipynb
```
jvmancuso: looks like a merge conflict in your ipynb 😬

iamtrask:
Should I keep this open?

robert-wagner:
Changes have been implented elsewhere

Issue #1605, opened 2018-10-08
GREAT FIRST PROJECT: Make PySyft FAST!!!!


iamtrask:
This issue is probably the BEST beginner issue because it doesn't require any knowledge outside of basic python. The goal is just to take this demo

https://colab.sandbox.google.com/drive/17upxCYJmJ6Zoxv0KjiJ1ZbchlJybsfhs#scrollTo=BXp7uO9wi1qo

and make it as fast as possible by modifying the PySyft codebase (READ: it's not necessarily about modifying the demo code itself... more about making the library generically faster so that this demo runs faster)

This is also a **group** project - meaning you can feel free to jump on at any time. **Just comment with your interest below!!!**

Also - make sure you join the #team_pysyft channel in our slack [slack.openmined.org](http://slack.openmined.org)

Edit: If you're new to optimizing python code - [here's a good tutorial](http://mbatchkarov.github.io/2014/07/14/profiling-python/)

MikeThomz:
I am new here but interested.

iamtrask:
This is the perfect issue for new! 

Step 1) see if you can get that colab notebook running :)
Step 2) see if you can get the same demo running on your local machine (install pysyft and such)
Step 3) start trying to make the code faster!

iamtrask:
Fixed the link!

iamtrask:
The main pieces I expect to be the bottlenecks are:

1) anything involving python string computation / comparisons
2) send/receiving the same message multiple times (such as calls to my_var.get_shape())
3) message size (this is a big one)

But this is certainly far from exhaustive.

iamtrask:
Another big one is "garbage collection" - at present for remote tensors, we never delete them (even if you delete the pointer to them locally). This means that PyTorch doesn't get to re-use memory objects (which is a big part of its optimization) and instead has to create new objects all the time (this also blows up RAM)

iamtrask:
Another optimization is when we repeatedly send the same commands to a worker over and over again... For example

```
for epoch in range(n_epochs)

     # all my neural network code
```

We should explore ways to not have to repeatedly send the same operations over and over again when most of our training code is highly iterative (as above)

uhvardhan:
> Another optimization is when we repeatedly send the same commands to a worker over and over again... For example
> 
> ```
> for epoch in range(n_epochs)
> 
>      # all my neural network code
> ```
[...]

PR #1606, opened 2018-10-09
#1605 replaced for loop with list comprehension and map

amit-rastogi:
Performance improvement changes - replaced for loop with list comprehension and map


iamtrask:
First contribution to https://github.com/OpenMined/PySyft/issues/1605!!!

PR #1611, opened 2018-10-09
Speed up 10%

LaRiffle:


iamtrask:
looks like there are a couple style errors.

after you run "black syft" from the root directory and commit we can merge this guy!

PR #1612, opened 2018-10-10
Fixes #1592, should replace all modulos (%) with `torch.fmod`

### LINKED ISSUE #1592: URGENT: remove all % operators and replace with torch.fmod()
@channel - BUG IN PYTORCH 0.3.1

The modulus operator doesn't always work.
https://github.com/pytorch/pytorch/issues/1164

torch.fmod() works correctly
torch.remainder() and the % sign do NOT work.

We need to swap out all uses of % for torch.fmod() pronto

ionlights:
As spec'd in #1592, all modulos (%) need to be replaced with `torch.fmod` when seeking remainder on `torch.Tensor`s.

I believe I've covered these, but I'm not 100% sure on testing these, please advise. 🙂


iamtrask:
Just merged with master - if Travis passes then we'll be golden!

ionlights:
Will make the last edit to have it pass Travis.

iamtrask:
Sounds good! Looking forward to it! :)

syft/core/frameworks/torch/tensor.py
```diff
@@ -1559,95 +1559,88 @@ def __gt__(self, other):
...
-            gp_response = (
-                (
-                    self.child
-                    * 10 ** (2 * other.precision_fractional - self.precision_fractional)
-                )
-                / other.child
-            ) % self.field
-        return gp_response
+        if (not hasattr(other, 'precision_fractional')):
+            other = other.fix_precision(precision_fractional = self.precision_fractional)
```
alhparsa: It's just an indentation error, fix the indentations on lines `1632`, `1635`, `1637` and `1640`. That should fix the build error.

iamtrask:
It looks like there are some errors in the implementation (unit tests aren't passing) - think you can take a look, @ionlights ?

PR #1615, opened 2018-10-10
depricated precommit in favor of black

robert-wagner:


robert-wagner:
Actually scratch this, I am going to add black to the precommit rather than removing it entirely for those who want to use it (which might just be me until I get the makefile working)

Issue #1602, opened 2018-10-08
Base Worker Command Statistics


iamtrask:
BaseWorker classes have a bottleneck send_msg() function which all commands are sent through. We would like to be able to optionally track statistics about commands being sent. In particular, we would like to optionally be able to track how many messages are sent to-from particular workers, of a particular message type, and track statistics like "completion time". 

This project is about implementing a first version of this kind of statistical tracking which should greatly help us with optimizing the system as well as understanding general performance characteristics and constraints.

replomancer:
I'd like to do this.

iamtrask:
It's all yours!

kamathhrishi:
I don't think this has been completed yet right?

replomancer:
Well, this issue is about "first version of (...) statistical tracking". I've written something for start and my PR has been accepted. I don't know if people use it and what they think about it. There's always room for improvement, but I found it hard to do better without more explicit requirements.

I would close this issue and create new ones if there are more concrete ideas for statistics-related features.

kamathhrishi:
Alright then.

iamtrask:
Sounds good!

iamtrask:
One good follow-on project might be to report statistics to the notebook as an option (right now it only writes to a file). Another might be more statistics relating to message size, number of messages, etc.

PR #1630, opened 2018-10-14
[WIP] Add profiling in BaseWorker.send_msg

replomancer:
Hello,

Here's what I have for now (#1602, first issue) and I need some feedback.
The method calls are profiled and basic info is stored in a log file.

I don't know how exactly you'd like to use the stats and what would be
most helpful.
Also I test my code by running PySyft tests (python3 setup.py test) which
may be giving me a wrong idea about how the stats will look in "real use"
(e.g. I don't see any "numpy_cmd" type messages being sent during the tests).

1. I believe instead of logging individual calls like in this code,
   it would be better to have grouping and summaries. I don't know
   what grouping would be useful to you and when to start/stop collecting
   data for total stats. I could create functions to start and stop profiling
   with the idea to call them in a repl or notebook, but I don't know
   if it's good for you.
   Some total statistics can already be gathered from the log with
   grep, cut, wc etc. e.g.
   `less send_msg_profiling.log  | cut -f2,3 | sort |  uniq -c`
   One thing to remember is to rm or mv the log file whenever old entries
   become irrelevant.

2. I don't know what parts of message are relevant for the stats/logs.
   I currently save (message_type:mode) or (message_type:integer_val)
   when the message is not a dict e.g. "req_obj:81697078720". I think
   "req_obj:int" would be better(?).

3. The completion time is currently shown in milliseconds and I wonder if this
   is the right time unit. I see mostly fractions of milliseconds, but
   I imagine in "real use" the times will be much higher.

4. I have put a PROFILE_MODE flag in profiling.py file. It should probably
   go somewhere else if it's supposed to be manually set.

5. Is there an easy way to have something more similar to "real use"
   than what I'm doing (running "python3 setup.py test")?


replomancer:
Here's a sample log line:
`2018-10-14 11:45:51     me->alice       obj:acquire     0.17 ms total calls: 90`
Columns are separated with tabs.
I think this would be a better format:
`2018-10-14 19:46:17     From: me        To: alice       type: obj       0.31 ms total calls: 90`
Sender and recipient are in separate columns (separated by tabs) so filtering is easier. Message modes and integer values are ignored. 

This is from the issue description:
> In particular, we would like to optionally be able to track how many messages are sent to-from particular workers,
`grep "To: bob" send_msg_profiling.log | wc -l`
`grep "From: me" send_msg_profiling.log | wc -l`

> of a particular message type,
` grep "type: req_obj" send_msg_profiling.log | wc -l`

> and track statistics like "completion time".
Individual calls have completion times recorded. Stats like average, max, min... for a given message type require using cut and sort or awk (ouch!).

iamtrask:
Great PR! This is a really nice feature! 

With regard to your ideas for future work and a "real use" case, in my head it's something like "i have a notebook which trains a neural net and I want to be able to figure out which calls are the most costly so that perhaps I can change my code to make it faster." Is that enough context for future work?

Issue #1505, opened 2018-09-19
Logging tensor


robert-wagner:
We want a tensor which rather than doing any operations simply writes the operation to a file

stephenjfox:
I'd love to give this one a shot. I've been silently keeping up with the community since its inception and I think I can do this.

@all okay with this?

robert-wagner:
Go for it! If you have any questions I'll be happy to help!

stephenjfox:
Seeing as @robert-wagner has taken off with #1632, I'm going to make those tests pass instead.

PR #1632, opened 2018-10-14
Log tensor

robert-wagner:
Completes #1505


iamtrask:
Hey @robert-wagner - looks like the build doesn't pass when it's merged into Master. Perhaps if you merged master yourself then you could inspect the issue?

https://travis-ci.org/OpenMined/PySyft/builds/441376656?utm_source=github_status&utm_medium=notification

Issue #1631, opened 2018-10-14
Get rid of relative imports


robert-wagner:
Currently in our code base there are many relative imports. Relative imports can cause problems if code is moved around. As such we should switch every relative import to an absolute import

llealgt:
is this an open issue?i can take it if so

iamtrask:
It was sorted out in 1631

PR #1634, opened 2018-10-14
fixed the relative imports issue #1631

Wattabak:
#1631 all previous relative imports are now absolute


robert-wagner:
Could you rebase this on master?

PR #1640, opened 2018-10-19
Fix regexp for black exclude option in pyproject.toml

replomancer:
Whitespaces in regexp inside pyproject.toml shouldn't be there. Maybe now travis will build fine?


replomancer:
OK, I thought the PR which adds pandas to requirements has already been merged. This change only fixes black.

iamtrask:
Excellent work!

PR #1643, opened 2018-10-24
Wrong import path fixed.

Nivek92:
import syft.core fails as the torch folder is in the frameworks subfolder


iamtrask:
Thank you @Nivek92!

PR #1644, opened 2018-10-27
Update MNIST Example.ipynb

peter9711:
fix an import problem caused by TorchHook which has been moved into frameworks.torch


iamtrask:
Well done!

PR #1650, opened 2018-10-30
Fix flake 8

robert-wagner:


iamtrask:
Merging because it seems that flake8 is fixed - but there are unit test errors.

PR #1651, opened 2018-10-30
added coverage

robert-wagner:


replomancer:
Hey,
This branch is 63 commits behind master. On master directories excluded by black are in pyproject.toml (not in .travis.yml).

I think you removed torch from requirements by accident?

iamtrask:
The Torch 1 buildout is going to be a bit strange from a dependency management perspective - we're sortof starting from scratch as we're changing so much - so it's ok to ignore the 63 commits behind part.

Issue #1642, opened 2018-10-23
WebSocketWorker example code doesn't work


replomancer:
The example code from WebSocketWorker docstring breaks. The first line imports TorchHook from a no longer existing module:
```>>> from syft.core.hooks import TorchHook```
The hook is available in syft now. When this is fixed, after a few steps a nonexistent function utils.PythonJSONDecoder is called on the server side. I don't know what else needs to be updated.

There are no tests for this Worker.

replomancer:
I'd like to fix this Worker and add a test or two.

Ogofo:
Are you making any progress on this issue and do you need help?

replomancer:
I see a lot of broken code in workers, fixing this is not a "fun issue" and may take a little more time.

If you're interested, take a look at [workers_test.py](https://github.com/OpenMined/PySyft/blob/4036adcac0ea20d702d5f9eee314a188b2e297d2/test/core/workers_test.py). There's only one test. Try executing example code from [VirtualWorker's docstring](https://github.com/OpenMined/PySyft/blob/4036adcac0ea20d702d5f9eee314a188b2e297d2/syft/core/workers/virtual.py#L45-L72). It brakes. Adding any worker tests or fixing VirtualWorker would be nice. If you still want to help with workers, maybe try there, so we don't get in each other's way or spend time both doing exactly the same thing.

Or you can take a look at "good first issues" here: https://github.com/OpenMined/PySyft/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

Ogofo:
Do you have a branch that you are actively working on or a list of broken things that need to be repaired? That would help us do not get in each other's way. 

A potential issue I found is if the WebSocketWorker is initialized with `is_client_worker=True` which is the default, he will not register any received objects in either `_objects` or `_tmp_objects`. This is because  `set_obj(self, remote_key, value, force=False, tmp=False)`of BaseWorker only allows client worker to register temporary objects, but the `tmp` variable is not set to `True`.

replomancer:
Yes, I got a little stuck in this issue. We're talking with Ogofo and Nivek92 on Slack. I think this weekend I'll create a pull request with what I have: a correctly failing test and some small incomplete fixes. Or maybe it will be in my repo only, because I think I have no permissions to create a PR to a new branch here and I don't want this to be a pull request to master.

mari-linhares:
Should this issue be closed?

replomancer:
[...]

PR #1653, opened 2018-11-03
[WIP] Fix websocketworker

replomancer:
I've created this pull request to bugfix branch, because I want to easily share code with Ogofo and Nivek92 who expressed interest in issue #1642. This code is not ready for master. Can someone create a separate branch for this maybe? I will then set it in this PR.

I believe the reason why there are so few tests is because this TorchHook solution (possibly combined with the way we write/run tests) makes adding new tests much harder for people. Running tests produce lots of  `Torch was already hooked... skipping hooking process` and `Replacing old worker which could cause unexpected behavior` warnings. In my case I'm adding a correctly failing test (TestWebSocketWorker) and when I run it with `python3 setup.py test` I see different error messages than when I test the same code manually (in a repl) or with
**`python -m unittest -v test.core.workers_test.TestWebSocketWorker`**

I understand the TorchHook solution is waiting for improvement. Meanwhile I tried to find a fix/workaround for tests. I experimented with setUp/tearDown and modules reloading, but wasn't successful.

So far I've put almost all effort into the test and I don't know how hard the complete fix may be.

### DIFF
--- a/syft/core/workers/websocket.py
+++ b/syft/core/workers/websocket.py
@@ -2,7 +2,7 @@ import websockets
 import asyncio
 import json
 
-from syft.core import utils
+from syft.core.frameworks.encode import decode
 from syft.core.workers import BaseWorker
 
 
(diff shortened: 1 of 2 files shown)


iamtrask:
Well done! Sorry I was a bit slow to close this one!

iamtrask:
Re: the warnings. We should probably just elminate the warnings (or only set them to go off when verbose==True)

Issue #1666, opened 2018-11-09
Add serde logic for torch.Tensor


iamtrask:
This issue should extend the current logic in syft/deser.py to include support for serializing/deserializing torch.Tensor objects. 

- [ ] add logic (if necessary) for torch.Tensor serialization and deserialization
- [ ] add unit test demonstrating proper serde of torch.Tensor objects
- [ ] add docstrings for all methods describing their functionality including an example of how to serialize/deserialize it.

Ogofo:
I'll be working on this one.

Ogofo:
Review the merged PR. 

 [x] add logic (if necessary) for torch.Tensor serialization and deserialization
 [x] add unit test demonstrating proper serde of torch.Tensor objects
 [x] add docstrings for all methods describing their functionality including an example of how to serialize/deserialize it.

iamtrask:
Excellent work!

PR #1680, opened 2018-11-14
Issue #1666

Ogofo:
Added serialization and deserialization of torch.Tensor objects. 

Changed the _simplifier to return a list instead of a tuple, as it looks like the `msgpack` is encoding a tuple as a list. This is throwing errors in the `_details` method. 

Renamed the methods `compress`  and `decompress` to `_compress` and `_decompress` as these are private methods.

### DIFF
--- a/test/test_serde.py
+++ b/test/test_serde.py
@@ -1,21 +1,24 @@
-from syft.serde import _simplify
+from syft.serde import _simplify, serialize, deserialize
 from unittest import TestCase
+from torch import tensor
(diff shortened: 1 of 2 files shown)


test/test_serde.py
```diff
@@ -1,21 +1,24 @@
-from syft.serde import _simplify
+from syft.serde import _simplify, serialize, deserialize
```
robert-wagner: have each import on separate lines

Issue #1664, opened 2018-11-09
Add serde logic for range


iamtrask:
This issue should extend the current logic in syft/deser.py to include support for serializing/deserializing range objects. 

- [ ] add logic (if necessary) for range serde
- [ ] add unit test demonstrating proper range of ints
- [ ] add docstrings for all methods describing their functionality including an example of how to serialize it.

Nivek92:
Hey I will work on that issue

PR #1681, opened 2018-11-15
Solved: Add serde logic for range #1664

Nivek92:
Added the logic (if necessary) for range serde.

Added a unit test

Added add docstrings for all methods describing their functionality including an example of how to serialize it


iamtrask:
A very excellent PR - statically typed, well documented, clean code, and conforms to conventions. Well done!

Issue #1670, opened 2018-11-09
Add serde logic for np.ndarray


iamtrask:
Part of Epic: https://github.com/OpenMined/PySyft/issues/1656

This issue should extend the current logic in syft/deser.py to include support for serializing/deserializing np.ndarray objects. 

- [ ] add logic (if necessary) for np.ndarray serialization and deserialization
- [ ] add unit test demonstrating proper serde of np.ndarray objects
- [ ] add docstrings for all methods describing their functionality including an example of how to serialize/deserialize it.

robert-wagner:
Great work on this issue!

PR #1684, opened 2018-11-15
Solved: Add serde logic for np.ndarray #1670

Nivek92:
added logic for np.ndarray serialization and deserialization
 added unit test demonstrating proper serde of np.ndarray objects
 added docstrings for all methods describing their functionality including an example of how to serialize/deserialize it.


iamtrask:
Also - well done! Very clean, well documented code!

Issue #1685, opened 2018-11-15
Fix travis ci build process


Nivek92:
The is an error in the paths for the exclude command

Ogofo:
There is another issue within the Travis CI run, I would suggest to use this ticket to resolve the issues and get the Travis build running. 

Current issue is: 

```
$ coverage run python setup.py test
No file to run: 'python'
The command "coverage run python setup.py test" exited with 1.
```

Ogofo:
The previous can be resolved by changing the line to `coverage run setup.py test`.

Currently the build would still faile because:

1. The test coverage is not 100% yet. 
2. The `bash scripts/execute_notebooks.sh` will fail, because the script does not yet exist on the `torch_1` branch

Nivek92:
Associated 

Get 100% test coverage #1703
Make tests pass #1705

iamtrask:
Woot woot!!!

PR #1687, opened 2018-11-15
Solution for #1685

Ogofo:
Fixed travis.yml to correctly ignore `.egg/` and `build/` directory while checking code style using `black`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ before_script:
     # stop the build if there are Python syntax errors or undefined
     # names
     - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
-    - black --check --verbose --exclude=/.eggs/,/build/ .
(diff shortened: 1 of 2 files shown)


robert-wagner:
Waiting for Travis to pass then will nerge

robert-wagner:
So the issue isn't actually with build. The issue is with serde (build is the root directory in this case)

Ogofo:
I Think the changes are still necessary as you can see in #1683, the `serde.py` is correctly formatted but it fails due to .egg not being excluded. The `serde.py` is not correctly formatted in the torch_1 Head currently, the changes in 1c293c7 should still be necessary. Let's see if makes the build working.

iamtrask:
Hmm - survey says no dice yet - looks like there's still a code coverage issue. I don't see any obvious indicators as to where though

PR #1692, opened 2018-11-16
In `handle_call` for the "end_get" case need to call get() on a _SyftTensor

antonrd:
Currently it's called on a `dict` object and calling `end_get` in tutorial 3 generates an error because `dict` does not have a `get()` method with no arguments.

Adds a test that demonstrates a scenario in which the code currently fails.

### DIFF
--- a/syft/core/frameworks/torch/tensor.py
+++ b/syft/core/frameworks/torch/tensor.py
@@ -1028,7 +1028,8 @@ class _PointerTensor(_SyftTensor):
         self_ = tensor_command["self"] if has_self else None
 
         if attr == "end_get":
-            response = self_.get()
+            # For end_get use the tensor in the original Syft command
+            response = syft_command["self"].get()
         else:
             command, locations, owners = torch_utils.compile_command(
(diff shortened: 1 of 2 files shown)


syft/core/frameworks/torch/tensor.py
```diff
@@ -1028,7 +1028,8 @@ def handle_call(cls, syft_command, owner):
         self_ = tensor_command["self"] if has_self else None
 
         if attr == "end_get":
-            response = self_.get()
+            # For end_get use the tensor in the original Syft command
+            response = syft_command["self"].get()
```
antonrd: I realize this may not be the best solution but given my limited knowledge of the codebase this is what I came up with. Will be happy to modify, but at least I think we have an issue that needs resolving here, one way or another.

iamtrask: I literally tried to solve this yesterday and if your solution works it's VASTLY superior! Well done! (will check in a bit)

iamtrask:
Well done!

antonrd:
Nice, thanks!

PR #1693, opened 2018-11-16
Replaced pickle for ndarray

Nivek92:
Removed pickle and used numpy functions instead so we don't need to worry about code injection in pickle and because it's faster than pickle


iamtrask:
Well done! Thank you for jumping on this so proactively!

Issue #1696, opened 2018-11-16
Create_pointer for torch tensors


LaRiffle:
This is a child issue of https://github.com/OpenMined/PySyft/issues/1695. Please check this issue to see if you have anything you need.

Initialize a pointer to this tensor with .create_pointer().
- [x] Instanciate a pointer tensor that inherits from an Abstract / Syft Tensor
- [x] Define appropriate attr (torch_type, location, id_at_location, owner, skip_register)

Nivek92:
This is finished isn't it?

LaRiffle:
Yes it is, the PR #1706 does only need a better coverage in tests
I'll do this when I find some time, any contrib welcome :)

PR #1706, opened 2018-11-19
Create pointer #1696

LaRiffle:
Add a hook to overload Tensor
Add a PointerTensor
Add a create_pointer method

TODO:
- [ ] Work tensor repr/str
- [ ] Add tests
- [ ] Add proper docstrings to all Classes, Methods, and Functions
- [ ] Add static types to methods and function parameters

Please tell me what points are not ok with torch1 approach, otherwise I'll be tempted to copy paste most of the 0.3.1 work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ before_script:
   # https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds
   - flake8 . --count --exit-zero --statistics
 script:
-  - coverage run --omit=*/venv/*,setup.py setup.py test
+  - coverage run --omit=*/venv/*,setup.py,.eggs/* setup.py test
   - coverage report --fail-under 95 -m
   # - bash scripts/execute_notebooks.sh
   # - if grep -q "Traceback (most recent call last):" nb_to_md.txt; then false; else true; fi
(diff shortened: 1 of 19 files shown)


Nivek92:
Hey @LaRiffle  from my understanding the "Fix type confusion in serde.py" is incorrect. The returns are tuple (exactly 3 elements each) but are returned as python lists instead of python tuple because of the serialization which can't handle python tuple.

The type annotations describe the "collection" type not the actual python return type I think. I may be wrong though

LaRiffle:
Hum yes I get your point. I can revert sure, although my PyCharm is a bit bothered by this typing issue. But I'm no python typing expert, so :)

Nivek92:
"I'm no python typing expert" - me neither so I would suggest we stick with the approach that doesn't cause PyCharm problems because I just assume PyCharm knows better :D 

If in the future we find out that the other way is correct we can still change it

iamtrask:
Note: I just added the last two bullets to the list at the top (referring to docstrings and static typing we'd want to have before merging)

LaRiffle:
In response to #1707, here is a version of what could be moving attributes out of torch. This is not final at all, but I would like to have our opinion on that solution.

Nivek92:
Nice work @LaRiffle I like it much better this way. So much cleaner.

iamtrask:
I noticed you removed the work-in-progress label. Are you ready to merge?

LaRiffle:
Merge create pointer - solve Issue #1696 - pass tests

PR #1717, opened 2018-11-21
First tests added

Nivek92:
Some firsts tests added for TorchTensor and the get_method.

Some functionality is still missing to make the tests pass.

### DIFF
--- a/test/tensors/test_native.py
+++ b/test/tensors/test_native.py
@@ -1,5 +1,4 @@
 from syft.frameworks.torch.hook import TorchHook
-from syft.frameworks.torch.tensors import TorchTensor
 from syft.frameworks.torch.tensors import PointerTensor
 from syft.workers.virtual import VirtualWorker
 from unittest import TestCase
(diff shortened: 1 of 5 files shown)


Nivek92:
I would like to get some feedback if the tests make sense.

If so I would try to work on making them pass

test/tensors/test_native.py
```diff
@@ -0,0 +1,63 @@
...
+from syft.frameworks.torch.tensors import PointerTensor
+from syft.workers.virtual import VirtualWorker
+from unittest import TestCase
+import torch
+
+
+class TestTorchTensor(TestCase):
+    def test_owner_default(self):
+        hook = TorchHook()
+        tensor = TorchTensor()
```
iamtrask: this should probably be replaced with a test which initializes a torch tensor the normal way.

x = torch.Tensor([1,2,3,4,5])

or

x = torch.zeros(5)

something like that.

test/tensors/test_native.py
```diff
@@ -0,0 +1,63 @@
...
+
+        assert tensor.owner == hook.local_worker
+
+    def test_create_pointer(self):
```
iamtrask: this one looks good except for the tensor initialization (see above) which should also be done the normal torch way.

test/tensors/test_native.py
```diff
@@ -0,0 +1,63 @@
...
+        hook = TorchHook()
+        tensor = TorchTensor()
+        bob = VirtualWorker()
```
iamtrask: for our other unit tests previously we were able to initialize one set of workers at the top and then use them for all tests (made them run a bit faster)

test/tensors/test_native.py
```diff
@@ -0,0 +1,63 @@
...
+
+        bob = VirtualWorker()
+
+        tensor = torch.rand(5, 3)
+        tensor.id = 1
+
+        pointer = tensor.send(bob)
+
+        assert type(pointer) == PointerTensor
+        assert bob.get_obj(1) == tensor
```
iamtrask: nicely done!

test/tensors/test_native.py
```diff
@@ -0,0 +1,63 @@
...
+
+        bob = VirtualWorker()
+
+        tensor = torch.rand(5, 3)
+        tensor.id = 1
+
+        pointer = tensor.send(bob)
+
+        assert type(pointer) == PointerTensor
+        assert pointer.get() == tensor
```
iamtrask: nicely done!

test/tensors/test_native.py
robert-wagner: A good way to do this would be to create a conftest.py and put all the worker initialization in there. That way we could use them across all tests

test/tensors/test_native.py
robert-wagner: A good way to do this would be to create a conftest.py and put all the worker initialization in there. That way we could use them across all tests

robert-wagner:
Hi, should coverage and tests be failing on this pr or is that something that needs to be looked into?

Nivek92:
the test fail, as some functionality is still missing. as soon as that is added the test should pass (hopefully)

so in regards to tests it should be fine

iamtrask:
Sorry I've been a little slow on this one - was a bit busy getting ready for (And then traveling to) NeurIPS. I'm a bit reluctant to merge failing tests as it makes it hard to merge other code because we have to remember (in our minds) which tests are supposed to be failing. Perhaps we can comment the failing tests out for now instead?

robert-wagner:
Should the base of this project be changed to torch_1 since the pointer tensor pr was merged?

Nivek92:
did that already @robert-wagner

PR #1728, opened 2018-11-26
Fix typos in introduction

mari-linhares:


robert-wagner:
Hi, could you run `black syft` on your branch. It will fix the error that travis is throwing. Alternatively if you run `make install_hooks` it will automatically run for every commit

Issue #1673, opened 2018-11-09
Add serde logic for slice


iamtrask:
Part of Epic: https://github.com/OpenMined/PySyft/issues/1656

This issue should extend the current logic in syft/deser.py to include support for serializing/deserializing slice objects. 

- [ ] add logic (if necessary) for slice serialization and deserialization
- [ ] add unit test demonstrating proper serde of slice objects
- [ ] add docstrings for all methods describing their functionality including an example of how to serialize/deserialize it.

Nivek92:
@iamtrask Like ellipsis I don't actually understand why would we want to serde a slice object?

Would you not just send the slice itself instead of the slice object?

Ogofo:
@iamtrask can you close this, please? :) Thanks

PR #1731, opened 2018-11-26
Solving #1673 - adding logic to serde python slice objects.

Ogofo:
- Solving #1637 
- Adding _compressed_ test to all serde tests

### DIFF
--- a/syft/serde.py
+++ b/syft/serde.py
@@ -405,6 +405,46 @@ def _detail_ndarray(arr_representation: Tuple[bin, List[int], str]) -> numpy.nda
 
     return res
 
+#   slice
+
+
+def _simplify_slice(my_slice: slice) -> Tuple[int, int, int]:
+    """
+    This function creates a list that represents a slice.
+
+    Args:
+        slice: a python slice
+
+    Returns:
+        list: a list holding the start, stop and step values
+
+    Usage:
+
+        slice_representation = _simplify_slice(slice(1,2,3))
+
+    """
+    return [my_slice.start, my_slice.stop, my_slice.step]
+
+
+def _detail_slice(my_slice: Tuple[int, int, int]) -> slice:
+    """
+    This function extracts the start, stop and step from a list.
+
+    Args:
+        list: a list defining the slice parameters [start, stop, step]
+
+    Returns:
+        range: a range object
+
+    Usage:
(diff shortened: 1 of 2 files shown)


syft/serde.py
```diff
@@ -405,6 +405,46 @@ def _detail_ndarray(arr_representation: Tuple[bin, List[int], str]) -> numpy.nda
 
     return res
 
+#   slice
+
+
+def _simplify_slice(my_slice: slice) -> Tuple[int, int, int]:
```
robert-wagner: This function does not return a tuple. It returns a list

Ogofo: I copied that one from line 317 

`def _simplify_range(my_range: range) -> Tuple[int, int, int]:`

Probably should adjust that line as well. 

```suggestion
def _simplify_slice(my_slice: slice) -> List:
```

Nivek92: That's the same discussion @LaRiffle and I had. It returns a PYTHON List but the data structure is actually a tuple because it has exactly 3 elements. you don't want to get a list with 4 elements but a List with 3 integers in it.

That's why the return type is described as a Tuple of ints

Though we actually were not sure if the type annotations should describe the actual python return type or the actual data structure that is returned. I prefer data structure so you can catch if someone is returning nonsense.

Your thoughts? @robert-wagner @Ogofo

robert-wagner: Theoretically it should be both. If we don't have it correspond to the actual python return type, then we loose any potential benefit of automated type checking. But if we don't have it correspond to the theoretical data structure then we don't get help understanding what is actually going on. We could change this function to return a tuple pretty easily which would satisfy both

robert-wagner: Theoretically it should be both. If we don't have it correspond to the actual python return type, then we loose any potential benefit of automated type checking. But if we don't have it correspond to the theoretical data structure then we don't get help understanding what is actually going on. We could change this function to return a tuple pretty easily which would satisfy both

Ogofo: Sure we can return a tuple. But we have to be aware that `msgpack`  is binarizing the tuple as a list. So we have to change the input `def _detail_slice(my_slice: Tuple[int, int, int]) -> slice:`  to `def _detail_slice(my_slice: List) -> slice:`. Think we should make this clear within the docstring.

Issue #1704, opened 2018-11-17
Lz4 Compressed binary is short than original


Nivek92:
While working on another issue I noticed, that the compressed binary can be longer than the original which doesn't make sense.

As such I would suggest to only use the compressed version if it's shorter than the original. But then we would need to make sure decompression is only used when compression was really applied.

amit-rastogi:
I am picking this up

amit-rastogi:
@Nivek92 Yes I am looking into adding both the compression schemes as options into the library as per my discussion with @iamtrask yesterday.

amit-rastogi:
@Nivek92 @iamtrask So as I understand there would be following tasks as part of this issue-
- add "zstd" as an option in the library
- add the functionality to decide whether to compress the input or not. Add corresponding logic in decompression as well.
-add unit tests

Nivek92:
@amit-rastogi sounds good to me. I edited the original issue text to reflect these tasks

PR #1734, opened 2018-11-28
Guarantee Lz4 Compressed binary is shorter than original #1704

amit-rastogi:
Change details-
1. Added support for 'zstd' compression algorithm in the library. Now both 'lz4' and 'zstd' algorithms can be used.
2. Handled the scenario when compression algorithm output is greater than the input stream in serialize\deserialize apis. Added a 1-byte header to the serialized output to indicate output is compressed or not.
3. Added 2 unit test cases for zstd algorithm for testing compression\decompression apis and serialize\deserialize apis.
4. Added zstd dependency in requirements.txt
5. Executed unit test cases using "python setup.py test". All tests passed including the newly added unit tests.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,7 @@ coverage
 ipykernel
 msgpack
 lz4
+zstd
 pre-commit
 sphinx_rtd_theme
 websockets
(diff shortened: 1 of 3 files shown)


robert-wagner:
Hi, it looks like there are some merge conflicts as well as some formatting issue. Could you run `black syft` to resolve the formatting errors (you can have this run automatically for every commit with `make install_hooks`. The merge conflict should be solveable by re pulling in the base branch

amit-rastogi:
Thanks @robert-wagner. I have taken care of the issues. Checks are passing now.

PR #1736, opened 2018-11-29
Fix the search function accordingly to ex/torch/toy sockets workers

LaRiffle:
Fix:
https://github.com/OpenMined/PySyft/blob/master/examples/torch/toy/SocketWorker%20Server.ipynb
https://github.com/OpenMined/PySyft/blob/master/examples/torch/toy/SocketWorker%20Client.ipynb

Which underlines that worker.search(#dataset)  is broken


miranthajayatilake:
Thanks for this @LaRiffle !

PR #1737, opened 2018-12-03
Federated Learning with Differential Privacy on Remote Machine

miranthajayatilake:
When effectively implementing Federated Learning in real world applications, it is important to build an architeture where data on a remote server could be accessed in order to train and test the model.

In the previous examples this is implemented using virtual clients. This example attempts to implement it over websockets.


iamtrask:
Well done!

PR #1747, opened 2018-12-06
[WIP] Torch refactor hook

Ogofo:
Suggestion to Issue 1746

### DIFF
--- a/syft/frameworks/torch/hook.py
+++ b/syft/frameworks/torch/hook.py
@@ -5,7 +5,7 @@ import logging
 import types
 import syft
 from syft import workers
-from syft.frameworks.torch.tensors import PointerTensor, AbstractTensor, TorchTensor
+from syft.frameworks.torch.tensors import TorchTensor
 from syft.frameworks.torch.torch_attributes import TorchAttributes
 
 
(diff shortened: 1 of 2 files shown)


syft/frameworks/torch/hook.py
```diff
@@ -5,7 +5,7 @@
 import types
 import syft
 from syft import workers
```
robert-wagner: Do these on separate lines to avoid merge conflicts

syft/frameworks/torch/hook.py
```diff
@@ -187,11 +185,14 @@ def _which_methods_should_we_auto_overload(self, tensor_type):
 
         to_overload = []
 
-        for attr in dir(tensor_type):
+        for attr in dir(TorchTensor):
```
robert-wagner: I don't think this change is correct. Not all tensor types have the same attributes as torch tensor

syft/frameworks/torch/hook.py
Ogofo: I don't really understand this :D I only see separated lines.

syft/frameworks/torch/hook.py
Ogofo: resolved in ddf8533726553bf7e53b95281d9013e402b7b546

syft/frameworks/torch/hook.py
iamtrask: Yeah I'm not sure either

robert-wagner: Sorry about this, meant the line after. Doing these comments on mobile is difficult

Issue #1746, opened 2018-12-06
[Discussion] Hook to only overload functions that are overloaded within the native tensor?


Ogofo:
Regarding the last merge of https://github.com/OpenMined/PySyft/pull/1742 some torch functionality is not possible anymore as the following example of  @LaRiffle shows:

```
import syft as sy
import torch
from syft.frameworks.torch.hook import TorchHook
hook = TorchHook(torch)
torch.Tensor([1, 2])
```

This is an issue within the hook, where all torch methods are replaced by a method that only returns `None` (see `hook.py/_assign_methods_to_use_child`). Currently the `hook.py/_get_overloaded_method()` is deprecated due to the changes of #1742.

I would like to discuss whether we need to overload all torch methods by default or whether we should overload methods that are specified within the pysyft native TorchTensor.

I created the following PR to demonstrate possible changes #1747. 

On the other hand, this shows us, that our test-cases are not enough and do not guarantee torch stability after hooking. Please see my other issue opened on this topic #1735 and join the discussion on how we could improve this.

iamtrask:
As mentioned in the PR you made - I intentionally removed functionality which was copy-pasted from the 0.3.1 implementation which we intend to re-write (lest we start making progress based on old code). 

However, I _think_ the issue you specifically flagged above (the code example you wrote down) works now. 

Should we chat more or should I close this?

mari-linhares:
@iamtrask can we close this?

mari-linhares:
closing it :), feel free to re-open this if needed.

PR #1752, opened 2018-12-06
Torch 1749 serde posthook

Ogofo:
This is based on #1746. The test should be generally fine, but on current `torch_1`the test will fail as the torch functionality is broken.

Can be merged based on approval of #1746.

### DIFF
--- a/syft/frameworks/torch/hook.py
+++ b/syft/frameworks/torch/hook.py
@@ -131,8 +131,6 @@ class TorchHook:
         # Rename native functions
         self._rename_native_functions(tensor_type)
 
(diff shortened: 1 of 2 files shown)


iamtrask:
Seems like tests... passed?

LaRiffle:
@iamtrask There was an old version of `.travis.yml` without the coverage requirement (which currently is 95%)

Ogofo:
Somehow the PR does not look very clean to me anymore due to so many changes and branches. I will close this one and open up a new one that only represents the actual changes to test_serde.py.

Issue #1740, opened 2018-12-04
Reduce duplication in serde test logic


Nivek92:
At the moment there are two different test cases for uncompressed/compressed serialization even though they only differ by the flag for compress & compressed.

amit-rastogi:
Hey @Nivek92 I'm picking this up.

Update the following test cases to reduce redundancy-
test_compress_decompress_lz4
test_compress_decompress_zstd
test_compressed_serde_lz4
test_compressed_serde_zstd

Nivek92:
There is still duplicate code though e.g.
compressed = _compress(original)
decompressed = _decompress(compressed)

compressed = _compress(original, "zstd")
decompressed = _decompress(compressed, "zstd")

these are 4 lines while we could also do

for attr in ["lz4","zstd"]:
  compressed = _compress(original, attr)
  decompressed = _decompress(compressed, attr)

Ogofo:
Regarding the open PR #1753, I suggest to refactor the test to use `@pytest.mark.parametrize()` annotations to the multiple test cases. Before implementing this, we have to resolve the issue that comes with class methods that inherit `unittest.TestCase`. Unit tests are not allowed the have arguments passed to a method. 

I'm not sure if there is a deeper meaning to those classes other than grouping the test cases. I would suggest to let the classes inherit from `object` if one wants to keep the grouping. Otherwise, I would suggest to 

(@robert-wagner mentioning you here as you reviewed the specific PR)

amit-rastogi:
Do we need to add pytest in requirements.txt?

PR #1753, opened 2018-12-06
Reduce duplication in serde test logic #1740

amit-rastogi:
Updated the following test cases to reduce redundancy-

Merged the following test cases into a single test case - test_compress_decompress
test_compress_decompress_lz4
test_compress_decompress_zstd

Merged the following test cases into a single test case - test_compressed_serde
test_compressed_serde_lz4
test_compressed_serde_zstd

Ran black on test folder
Ran unit tests

### DIFF
--- a/test/test_serde.py
+++ b/test/test_serde.py
@@ -140,29 +140,24 @@ class TestSerde(TestCase):
 
         assert numpy.array_equal(arr, arr_serialized_deserialized)
 
-    def test_compress_decompress_lz4(self):
+    def test_compress_decompress(self):
         original = msgpack.dumps([1, 2, 3])
         compressed = _compress(original)
         decompressed = _decompress(compressed)
         assert type(compressed) == bytes
         assert original == decompressed
 
-    def test_compress_decompress_zstd(self):
-        original = msgpack.dumps([1, 2, 3])
         compressed = _compress(original, "zstd")
         decompressed = _decompress(compressed, "zstd")
         assert type(compressed) == bytes
         assert original == decompressed
 
-    def test_compressed_serde_lz4(self):
+    def test_compressed_serde(self):
         arr = numpy.random.random((100, 100))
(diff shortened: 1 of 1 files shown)


robert-wagner:
Thank you for this pr but I fundamentally disagree with this change. Tests are supposed to help identify where problems come into existence in the code. By combining these two types of tests we make it harder to identify if the problem is in the general compress method or just in one of the compression schemes. This increased difficulty in identifying where the problem is increases the cognitive burden of debugging later on

Nivek92:
I see your point but I would argue code duplication is bad in tests also. One would then have to first need a test if the method works and second a test to test all compress methods

iamtrask:
Seems like best of both worlds here is to have a general "setup" and "teardown" method which is called in separate unit tests for each compression type.

Ogofo:
@robert-wagner I suggest using something like https://github.com/OpenMined/PySyft/pull/1752/commits/1499769da312aa449722d3d90d046ad9a5206102 where I set parameters and just for-looped through them. I'm not sure if this is the most elegant way, but once a test fails the argument values are clearly displayed so you see where the problem is.

Nivek92:
After thinking about it, I actually have a suggestion that would combine both. 

Why not just do something like

def createTest(name):
    def func(*args):
        ....
    func.__name__ = name
    return func

for arg in [arg1, arg2]:
  createTest(f"test_func_{arg1}")

that way you have different tests for each argument but in seperate test. 

No code duplication, Seperate Tests = best of both worlds

robert-wagner:
There is also parameterize in pytest which may be the best of both worlds. Ie we have test compression but we run it with multiple sets of parameters for the compression

```@pytest.mark.parametrize("method", ["","zstd")
def test_compress(method)
```

Ogofo:
Because of the approval of #1762, I think this is obsolete. Thank you @amit-rastogi for the PR :) 

If this one should stay open message me.

PR #1774, opened 2018-12-12
First Draft for a guide on how to contribute to pysyft.

Ogofo:
This was created by me and @Nivek92. We propose this as the first draft of an entry point for new contributors and defining rules for current contributors. This way we can formalize the rules for contributions and aim for a clean, readable and well-designed code base.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,0 +1,105 @@
+# Contributors Guidelines to PySyft
+
+# Getting Started
+
+## Slack
+A great first place to join the Community is the Slack channel http://slack.openmined.org.
+
+## Issues
+On https://github.com/OpenMined/PySyft/issues you can find all open Issues. You can find a detailed explanation on how to work with issues below under [Issue Allocation](#Issue-Allocation).
+
+
+# Setup
+
+## Forking a Repository
+To contribute to PySyft you will need to fork the OpenMind/PySyft repository.
+Then you can work risk-free on your fork.
+
+You will just need to fork once. After that you can call `git fetch upstream` and `git pull 'branch-name'` before you do your local changes to get the remote changes and be up-to-date
+
+
+## Setting up Pre-Commit Hook
+PySyft uses the python package `pre-commit` to make sure the correct formatting (black & flake) is applied.
+
+You can install it via `pip install -r requirements.txt` or directly doing `pip install pre-commit`
+
+Then you just need to call `pre-commit install`
+
+## Syncing a Forked Repository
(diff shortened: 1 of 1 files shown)


CONTRIBUTORS.md
```diff
@@ -0,0 +1,105 @@
...
+
+If you want to work on an open issue, please post a comment telling that you will work on that issue, we will assign you ass the assignee then.
+
+**Caution**: We try our best to keep the assignee up-to-date but as we are all humans with our own schedule delays are possible, so make sure to check the comments once before you start working on an issue even when noone is assigned to it.
+
+
+## Writing Test Cases
+Always make sure to create the necessary tests and keep test coverage at 100%. You can always ask for help in slack or via github if you don't feel confidant about your tests.
+
+We aim to have a 100% test coverage, and the Travis CI will fail if the coverage is below this value. You can evaluate your coverage using the following commands.
```
jvmancuso: how was the 100% test coverage chosen?  from what I've heard, it's not a great measure of test suite quality/robustness (a la Goodhart's law), and as such many projects go with something in the 70-90% range

robert-wagner: The idea behind 100% test coverage is that we want to be intentional about what we are not testing. We could use pragma: no cover for some functions if we feel they don't need a test or can't be tested but that should be a well thought out choice

robert-wagner: The idea behind 100% test coverage is that we want to be intentional about what we are not testing. We could use pragma: no cover for some functions if we feel they don't need a test or can't be tested but that should be a well thought out choice

CONTRIBUTORS.md
```diff
@@ -0,0 +1,105 @@
...
+
+You documentation should not describe the obvious, but explain what's the intention behind the code and how you tried to realize your intention.
+
+You should also document non self-explanatory code fragments e.g. complicated for-loops. Again please do not just describe what each line is doing but also explain the idea behind the code fragment and why you decided to use that exact solution.
+
+#### Imports
+For better merge compatibility each import is within a separate line. Mulitple imports from one package are written in one line each.
+
+Example:
+```
```
robert-wagner: Maybe also put something about relative imports

Ogofo: Sure! Good idea, do you have a suggestion what should be added?

CONTRIBUTORS.md
LaRiffle: 100% is not possible nor practical, but we can start with a high barrier like 90% - 95% and lower than if needed?

PR #1779, opened 2018-12-16
Fix #1760

### LINKED ISSUE #1760: More descriptive error (or warning) when workers can't see each other?
In the Part 4 of the tutorial available at examples/tutorials/ there is this block of code:

```
bob.add_workers([alice, secure_worker])
alice.add_workers([bob, secure_worker])
secure_worker.add_workers([alice, bob])
```

If you remove these lines on cell 6 while trying to perform a move operation you see the following error:

```
/syft-0.1.0-py3.6.egg/syft/core/frameworks/torch/tensor.py in register_pointer(self)
   1015     def register_pointer(self):
   1016         worker = self.owner
-> 1017         location = self.location.id
   1018         id_at_location = self.id_at_location
   1019         # Add the remote address

AttributeError: 'str' object has no attribute 'id'
```

Which is not a very descriptive message and does not indicate the real problem.

The same error message shows on Part 3 if you remove these  lines:

```
# making sure that bob/alice know about each other
bob.add_worker(alice)
alice.add_worker(bob)
```

And try to run something like:
```
x = sy.FloatTensor([1,2,3,4,5]).send(bob).send(alice)
```

---

I think a more descriptive error message can help users debug this simple mistake, would be happy to add this.

Cheers!

mari-linhares:


test/torch_test.py
```diff
@@ -359,6 +359,16 @@ def test_send_get_tensor(self):
         # because .get_() was called, x should no longer be in the remote worker's objects dict
         assert ptr_id not in bob._objects
 
+    def test_send_pointer_to_unknow_worker(self):
```
robert-wagner: Unknow->unknown

syft/core/workers/base.py
```diff
@@ -607,10 +607,11 @@ def get_worker(self, id_or_worker):
             if id_or_worker in self._known_workers:
                 return self._known_workers[id_or_worker]
             else:
-                logging.warning(
-                    "Worker", self.id, "couldnt recognize worker", id_or_worker
+                raise TypeError(
```
robert-wagner: Should this be a type error?

mari-linhares: To be honest, I'm not particularly convinced of this as well. In some places a "`Exception` is raised, but I don't think this is the best approach. Maybe raise a `RuntimeWarning`  (https://docs.python.org/2/library/exceptions.html#exceptions.RuntimeWarning) ?

test/torch_test.py
mari-linhares: done. thanks!

syft/core/workers/base.py
robert-wagner: Pure Exceptions should never be raised as we could always be more specific. A runtime warning makes sense to me

mari-linhares: Done.

PR #1780, opened 2018-12-20
WIP: Serde Logic for PointerTensor

Nivek92:


syft/workers/base.py
```diff
@@ -128,7 +128,7 @@ def request_obj(self, obj_id, location):
 
     # SECTION: Manage the workers network
 
-    def get_worker(self, id_or_worker):
+    def get_worker(self, id_or_worker, fail_hard=False):
```
Ogofo: Documentation for `fail_hard` argument is missing.

test/torch/tensors/test_tensor.py
```diff
@@ -0,0 +1,44 @@
+from unittest import TestCase
+import torch
+
+
+import syft
+from syft.frameworks.torch.tensors import TorchTensor, PointerTensor
+
+
+class TestNative(TestCase):
```
Ogofo: I think we switched `TestCase` to object. Pytest does not require `TestCase`, but `TestCase` is disallowing the use of parametrizing methods.

test/torch/tensors/test_tensor.py
```diff
@@ -0,0 +1,44 @@
...
+    def test_init(self):
+        tensor_extension = TorchTensor()
+        assert tensor_extension.id is None
+        assert tensor_extension.owner is None
+
+
+class TestPointer(TestCase):
```
Ogofo: `TestCase`: see above

syft/serde.py
```diff
@@ -40,6 +40,10 @@
 import io
 import numpy
 import zstd
+import syft
```
Ogofo: Is this import really needed?

Nivek92:
I noticed, that in this version you have to register all workers to "me" aka the local_worker otherwise we do not get a virtual worker during deserialization but a string with the id of the worker (e.g. Pointer.location will be "Alice" instead of the VirtualWorker with id Alice) 

I am not sure if that is the desired/expected behavior.

syft/serde.py
Nivek92: Yeah because I want the local_worker and thus call `syft.local_worker.get_worker(val_str, fail_hard=True)`

iamtrask:
I attempted a merge for you which is now in this branch https://github.com/OpenMined/PySyft/tree/Nivek92-fallback

PR #1794, opened 2018-12-24
docs refactoring hook.py

R-Stefano:


iamtrask:
Need to run "black" to autoformat

R-Stefano:
There were an incongruence in variables. The parameter was compressScheme while it was called compress_scheme at line 95

robert-wagner:
It looks like you need to rerun black

PR #1795, opened 2018-12-24
Refactored tests to use pytest style and only create hooks and workers once

robert-wagner:


robert-wagner:
Will work on resolving merge conflicts. The majority of this was written before @iamtrask 's pr today

robert-wagner:
Discarding this because it would be faster to redo the pr than fix these merge conflicts. Yikes

Issue #1701, opened 2018-11-16
Initial Garbage Collection Support for PointerTensor 


iamtrask:
This is a child Issue of https://github.com/OpenMined/PySyft/issues/1695. Be sure to check the parent issue to ensure this issue has all dependencies required to finish it.

In this issue, we want to make it so that when a PointerTensor goes out of scope, the tensor that it is pointing to also gets deallocated. In particular, we want to VERY robustly test this functionality (with all tensor types) so that any later code doesn't break this (i.e., by accidentally creating circular dependentices, etc.)

- [x] add .del() command which a worker can receive and correctly use to deallocate/deregister a tensor
- [x] override \__del__ for PointerTensor objects so that they send the delete command
- [x] extensively test this functionality to ensure it works in a variety of ways (both for explicit "del x" and for automatic garbage collection)

Other Acceptance Criteria
- [x] all methods have robust doc strings
- [x] EXTENSIVE tests demonstrating both automatic and explicit garbage collection working correctly for ALL tensor types (Float, Int, etc.) 

Known Dependencies: torch.Tensor must already have .send() and PointerTensor must have .serialize() implemented.

mari-linhares:
I'm planning to have a look at this.

iamtrask:
Me too - with the PR I just merged a few minutes ago we should be more or less ready.

PR #1796, opened 2018-12-24
Initial Garbage Collection Support

iamtrask:
This PR does several things:
- Fixes https://github.com/OpenMined/PySyft/issues/1701
- moves the MSG_TYPE codes from the top of base.py to it's own codes.py file (to eliminate circular dependency issue)
- removed objects.py which wasn't being used


test/torch/tensors/test_tensor.py
```diff
@@ -42,3 +43,84 @@ def test_create_pointer(self):
         x = torch.Tensor([1, 2])
         x.create_pointer()
         x.create_pointer(location=self.james)
+
+    def test_explicit_garbage_collect_pointer(self):
```
robert-wagner: We should add a test for garbage collection of a pointer to a pointer

syft/codes.py
```diff
@@ -0,0 +1,7 @@
+class MSGTYPE(object):
```
mari-linhares: Can you add a comment describing the file, and a brief comment defining the class?

syft/frameworks/torch/tensors/pointer.py
```diff
@@ -1,4 +1,5 @@
 from .abstract import AbstractTensor
+from syft.codes import MSGTYPE
```
mari-linhares: `from syft import codes` is preferable, but since the imports in other files do not follow this I suggestcreate a bug and fix it later.

test/torch/tensors/test_tensor.py
iamtrask: FYI: @robert-wagner and @mari-linhares: I'm working on getting GC for pointer->pointer->tensor working. It's been a lot more trouble than I expected.

iamtrask: (I also didn't work on the 25th/26th of Dec... shameful I know)

robert-wagner:
@mari-linhares test objects is removed in another already submitted pr.

mari-linhares:
> @mari-linhares test objects is removed in another already submitted pr.

Oh, great :).

test/torch/tensors/test_tensor.py
robert-wagner: What problems are you having with pointer to pointer?

robert-wagner: What problems are you having with pointer to pointer?

LaRiffle:
Something which is not clear to me is how `del x_ptr` actually calls `x_ptr.child.__del__` (as x_ptr is an instance of torch.tensor). Where does this happen ?

Additionally I experience strange behaviours on notebooks like:
__Cell 0__
```
# Imports and setup
```
__Cell 1__
```
x = torch.Tensor([1, 2, 3, 4])
x_ptr = x.send(bob)
x_ptr.child # <-- try with and without that line
```
__Cell 2__
```
del x_ptr
```
If you keep the last line of Cell 1, *the del has no effect*. If you remove, it will work.

iamtrask:
Closed because #1805 is integrating this work

PR #1797, opened 2018-12-24
Refactored tests to use pytest style and only create hooks and workers once

robert-wagner:


iamtrask:
Hey @robert-wagner - i did my best to merge things based on what I know of your changes - would appreciate a quick readthrough to make sure it's right though.

Issue #1506, opened 2018-09-19
Create a Polynomial Tensor


iamtrask:
Create an abstraction such that any other tensor can leverage polynomial approximations of various non-linear functions by either extending this tensor (or by putting it in the chain hierarchy) or both. This initial functionality should include approximations for sigmoid, tanh, relu, exp, and log.

kamathhrishi:
Do we really need a separate tensor for this or shall we just make it a part of SPDZ?

jvmancuso:
I'd defer to @iamtrask, but I imagine this could be done in SPDZ

kamathhrishi:
This should be closed ?

PR #1801, opened 2018-12-27
Polynomial Tensor

kamathhrishi:
Skeleton code for Polynomial tensor. Not sure of the type of data to be passed and returned and if series should be encrypted. 

Suggest further improvements and additions. 

Should tests be written for this? 

In reference to issue #1506 (Create a Polynomial Tensor)

### DIFF
--- a/syft/core/frameworks/torch/__init__.py
+++ b/syft/core/frameworks/torch/__init__.py
@@ -10,6 +10,7 @@ from syft.core.frameworks.torch.tensor import (
     _PointerTensor,
     _SPDZTensor,
     _SNNTensor,
+    _PolynomialTensor
 )
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


iamtrask:
This is a great start! I think we should definitely have a few tests for this guy before merging! Preferably something that includes tests for both Tensors and Variables. Bonus points for showing it working with SNNTensor.

I'd also be a huge fan of adding a notebook in the tutorials showing how to use this with SNNTensor.

iamtrask:
The .child interface is a bit easier for Torch 1.0 now that the last PR has been merged - I do recommend moving your new code to the Torch 1.0 branch if you're interested.

kamathhrishi:
WIll re-write functionality for Syft 1.0

Issue #1729, opened 2018-11-26
Removing the head torch tensor wrapper


LaRiffle:
This is a sub issue of  #1657 questioning the design of tensor chains.

If I understand correctly, in "chain" structure of torch1 Pysyft, local tensor chains will not loop which is great, but we will have a empty wrapper at the head.

For example if I want to locally log operations of a tensor with a LogTensor, I'll have:

    [Empty Torch Tensor]
      v
    [LogTensor]
      v
    [Torch Tensor]

Similarly, when I send a pointer I'll have:

    [Empty Torch Tensor]
      v
    [PointerTensor]

This means that we're handling a bunch of empty torch tensors as wrappers. This adds not only complexity as it makes the chain longer but also adds some overhead related to allocating those empty tensors (even if we could argue that we could have a store of pre-allocated wrappers, at the expense of extra code complexity)

Is it really necessary to have this empty top wrapper? It seems like just removing it wouldn't change anything in the code behavior. Additionally, it's not likely that exposing a non torch tensor as a head object to the user would be a problem, as long as he _thinks_ that it's one.

robert-wagner:
The main reason for having an empty torch tensor at the head is for compatibility with the built in torch modules. They break if you pass them a non torch tensor. Ideally, if we could subclass torch tensor, we would not need these header tensors but that functionality is not yet available in PyTorch.

LaRiffle:
Uh. Do you have an example of torch module which would break? Usually we also hook them as well

robert-wagner:
I unfortunately dont know one of off the top of my head. Ill see if I can set up an minimal failing example

Ogofo:
I'm just wondering if the tensor should be the one holding the send method. Wouldn't a worker be more suited to actually send objects to another target worker? I'd like to understand why you chose this way of designing the workflow. I would expect something like this:

```
my_worker = Worker()
tensor = torch.Tensor()
my_worker.register_worker(remote_worker_address)

tensor_ptr = my_worker.send(tensor)
```

If the design stays this way, I imagine having a wrapper at the top of the chain, that replaces a tensor with a pointer after send is called, might be the most fail-proof method. Otherwise I could imagine the following scenario:

```
x = torch.Tensor()
x_ptr = x.send(bob)

//modifying and using a tensor that changed his owner
x = x*x
```

Ogofo:
Yeah your totally correct. I think what we are trying to achieve is creating a transfer method. 

[...]

Issue #1698, opened 2018-11-16
Adding .get() method to PointerTensor


iamtrask:
This is a child Issue of https://github.com/OpenMined/PySyft/issues/1695. Be sure to check the parent issue to ensure this issue has all dependencies required to finish it.

In this issue, you should add the method .get() to PointerTensor. This method should send a command to the remote worker to serialize it and send it to the local server. 

- [ ] ensure .get() works as described above
- [ ] .get() has correct docstrings (and any other functions created while creating .send() )
- [ ] .get() has several unit tests. In particular it should test getting tensors of ALL types (Float, Int, etc.)

Known Dependencies: torch.Tensor must already have .create_pointer() and .serialize() implemented.

Issue #1697, opened 2018-11-16
Adding .send() to torch.Tensor()


iamtrask:
This is a child Issue of https://github.com/OpenMined/PySyft/issues/1695. Be sure to check the parent issue to ensure this issue has all dependencies required to finish it.

In this issue, you should add the method .send() to torch.Tensor. This tensor should call .create_pointer() (to create a pointer), then serialize the tensor, and send the message to a remote worker. Thus, this method will accept two parameters: self and location. It will return the pointer object.

This issue relies on two Sub Issues to be accomplished first (in order):

- [ ] register_obj() added to BaseWorker  #1699
- [ ] receive_obj() added to BaseWorker

Final Acceptance Criteria:
- [ ] ensure .send() works as described above
- [ ] .send() has correct docstrings (and any other functions created while creating .send() )
- [ ] .send() has several unit tests. In particular it should test sending tensors of ALL types (Float, Int, etc.)

Known Dependencies: torch.Tensor must already have .create_pointer() and .serialize() implemented.

LaRiffle:
One point I remember discussing is whether we should rather ask the remote worker the storage address, instead of forcing him to use a given one, which could be already be in use.
Any thoughts?

Ogofo:
I think this is a good idea. I think it would make the most sense, to send the object to the `storage_worker` and let him respond with a `storage_id`. That way we can create a PointerTensor with `id_at_location=storage_id`.

LaRiffle:
Yes I agree, but one drawback is that the operation needs to be synchronous and wait for the response.

Ogofo:
True. That could slow down the process. 

Another approach could be to just store an internal reference to the remote Tensor within the PointerTensor. The Remote Worker would register the Tensor with his own chosen ID and create a mapping between the PointerTensor ID and his own remote ID.

To clarify the example: 

```
# local worker
ptr_on_local_worker = tensor.send()

# remote worker
remote_worker.receive(tensor, ptr_id)

def receive(self, tensor, ptr_id):
    id_on_remote_worker = self.register(tensor)
    mapping.add(ptr_id, id_on_remote_worker)
```

LaRiffle:
Yes, basically have an internal storage id which is a function `f` : `id_on_remote_worker = f(<worker>, <ptr_id>)`  should be fine !

Issue #1658, opened 2018-11-09
Pytorch 1.0 Garbage Collection


robert-wagner:
This is an Epic which is a sub-project under the project Torch 1.0 - the specs for which you can find here (#1587) - and the KanBan board for which you can find [here](https://github.com/orgs/OpenMined/projects/8?add_cards_query=is%3Aopen).

All work is happening in the [torch_1](https://github.com/OpenMined/PySyft/tree/torch_1) branch.

# Description

We should have a Pool of wrappers which we can re-use.

iamtrask:
I _think_ we're good to go based on https://github.com/OpenMined/PySyft/pull/1805 - @LaRiffle ?

LaRiffle:
I think this isn't too bad however I'd like some more tests to be performed or questions to be answered:
- have a GC test file
- How new tensors types like the LoggingTensor (will) implement __del__ and GC ?

LaRiffle:
Ok I think #1829 Solves this.

PR #1805, opened 2019-01-02
Torch 1 and (not) headless tensor chains - send/get - GC - ptr.method() - etc.

LaRiffle:
Continue addressing #1657
- Improve ownership on send / get tensors #1697 #1698
- Add GC from #1796* to solve #1701 (see also #1658)
- Add hooking for torch methods on pointers, following up on #1729

PR was first intended to explore chains without torch wrappers. This solution is simpler to write but makes PySyft less flexible for libraries using PyTorch. We're currently putting back the wrapper while keeping the features developed in the first setting.

### Typical examples
```
import syft as sy
import torch
hook = sy.TorchHook(torch)
me = hook.local_worker
```
**Fast remote op on pointers**
```
x = torch.Tensor([1, 2, 3, 4])
x2 = torch.Tensor([4, 3, 2, 1])
x_ptr = x.send(bob)
x2_ptr = x2.send(bob)
```
(Run it twice to see the speed-up)
```
%%time
y_ptr = x_ptr + x2_ptr
```
**torch module functional**
(Run it twice to see the speed-up)
```
%%time
F.relu(x_ptr)
```
**Garbage collection**
```
# Clean registries
bob._objects = {}
alice._objects = {}
x = torch.Tensor([1, 2, 3, 4])
x_ptr = x.send(bob)
x_ptr_ptr = x_ptr.send(alice)
print(bob._objects, alice._objects)
del x_ptr_ptr
print(bob._objects, alice._objects)
```

**Todo checklist:**
- [x] Handling duplicate send from same tensor, and chained send (`x.send(bob).send(alice)`)
- [x] Tests for the method hooking
- [x] module Functional hooking (like `torch.nn.functional.relu(ptr1)`)

**For next PRs**
 - Change function name with integer when sending messages
- module Functional hooking (like `torch.nn.Linear(2, 3)(ptr1)`)
- Functional hooking (like `torch.add(ptr1, ptr2)`)

### DIFF
--- a/syft/exceptions.py
+++ b/syft/exceptions.py
@@ -1,6 +1,18 @@
 """Specific Pysyft exceptions."""
 
 
+class PointerFoundError(BaseException):
+    """Exception raised for errors in the input.
+
+    Attributes:
+        expression -- input expression in which the error occurred
+        message -- explanation of the error
+    """
+
+    def __init__(self, pointer):
+        self.pointer = pointer
+
+
 class WorkerNotFoundException(Exception):
     """Raised when a non-existent worker is requested."""
 
--- a/syft/exceptions.py
+++ b/syft/exceptions.py
@@ -1,7 +1,7 @@
 """Specific Pysyft exceptions."""
 
 
-class PointerFoundError(BaseException):
+class RemoteTensorFoundError(BaseException):
     """Exception raised for errors in the input.
 
     Attributes:
--- a/syft/frameworks/torch/hook.py
+++ b/syft/frameworks/torch/hook.py
@@ -252,8 +252,8 @@ class TorchHook:
                 # structure is the same as the args object, with 1 where there was
                 # (torch or syft) tensors and 0 when not (ex: number, str, ...)
                 rule = build_rule(args)
-                # Build a function with this rule to efficiently replace syft tensors (but not pointer)
-                # with their child in the args objects
+                # Build a function with this rule to efficiently replace syft tensors
+                # (but not pointer) with their child in the args objects
                 args_hook_function = build_args_hook(args, rule)
                 # Store this utility function in the registry
                 hook_self.args_hook_for_overloaded_attr[attr] = args_hook_function
@@ -463,17 +463,18 @@ class TorchHook:
                 added to it, typically just torch.Tensor.
         """
 
-        @property
-        def location(self):
-            return self.child.location
-
-        tensor_type.location = location
-
-        @property
-        def id_at_location(self):
-            return self.child.id_at_location
-
-        tensor_type.id_at_location = id_at_location
+        # Not needed in the headless setting
+        # @property
+        # def location(self):
+        #     return self.child.location
+        #
+        # tensor_type.location = location
+        #
+        # @property
+        # def id_at_location(self):
+        #     return self.child.id_at_location
+        #
+        # tensor_type.id_at_location = id_at_location
 
         @property
         def id(self):
@@ -541,22 +542,23 @@ class TorchHook:
 
         return to_overload
 
-    def _rename_native_functions(self, tensor_type: type):
-        """Renames functions that are not auto overloaded as native functions.
-
-        Args:
-            tensor_type: The type of tensor whose native methods are getting
-                renamed. Typically just torch.Tensor.
-        """
-        for attr in self.to_auto_overload[tensor_type]:
-
-            lit = getattr(tensor_type, attr)
-
-            # if we haven't already overloaded this function
-            if f"native_{attr}" not in dir(tensor_type):
-                setattr(tensor_type, f"native_{attr}", lit)
-
-            setattr(tensor_type, attr, None)
+    # Not needed at the moment
+    # def _rename_native_functions(self, tensor_type: type):
+    #     """Renames functions that are not auto overloaded as native functions.
+    #
+    #     Args:
+    #         tensor_type: The type of tensor whose native methods are getting
+    #             renamed. Typically just torch.Tensor.
+    #     """
+    #     for attr in self.to_auto_overload[tensor_type]:
+    #
+    #         lit = getattr(tensor_type, attr)
+    #
+    #         # if we haven't already overloaded this function
+    #         if f"native_{attr}" not in dir(tensor_type):
+    #             setattr(tensor_type, f"native_{attr}", lit)
+    #
+    #         setattr(tensor_type, attr, None)
 
     @staticmethod
     def _add_methods_from__torch_tensor(tensor_type: type, syft_type: type):
--- a/syft/frameworks/torch/hook.py
+++ b/syft/frameworks/torch/hook.py
@@ -6,7 +6,7 @@ import syft
 from ... import workers
 
 from ...workers import BaseWorker
-from .tensors import TorchTensor
+from .tensors import TorchTensor, PointerTensor
 from .torch_attributes import TorchAttributes
 from .tensors.abstract import initialize_tensor
 
@@ -102,6 +102,9 @@ class TorchHook:
         self.to_auto_overload = {}
 
         self._hook_native_tensor(torch.Tensor, TorchTensor)
+
+        # Overload auto overloaded with Torch methods
+        self._add_methods_from__torch_tensor(PointerTensor, TorchTensor)
 
         # Add the local_worker to syft so that it can be found if the hook is
         # called several times
--- a/syft/frameworks/torch/hook.py
+++ b/syft/frameworks/torch/hook.py
@@ -400,6 +400,18 @@ class TorchHook:
         """
 
         @property
+        def location(self):
+            return self.child.location
+
+        tensor_type.location = location
+
+        @property
+        def id_at_location(self):
+            return self.child.id_at_location
+
+        tensor_type.id_at_location = id_at_location
+
+        @property
         def id(self):
             if not hasattr(self, "_id"):
                 self._id = int(10e10 * random.random())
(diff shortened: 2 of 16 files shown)


robert-wagner:
How does this work with torch.nn

LaRiffle:
@robert-wagner As for now I've provided a hook only for methods on pointers.
In the hook.py code you can see experimental code for hooking methods on other syft tensors (where you need to forward to the child), I'm working on this to have it clear and robust.*
`torch.nn` (as well as `torch` and `torch.nn.functional`)  is slightly different as it concerns functions. It has to be hooked to support execution on native torch tensors, pointer tensors and syft tensors. This means that contrary to methods we don't know what is the nature of the args until the call is made.
In the previous version of Pysyft we did hook `torch.nn.functional` and relied on the fact that `torch.nn` almost always rely on `torch.nn.functional`. As you and @iamtrask mentioned, this time we will also need to hook `torch.nn` otherwise you need to give args that are torch tensors (which is not the case as we remove the torch heads). I believe this doesn't add code complexity, and I can work on an example today or tomorrow to show this. It will use * and will make it clearer.

LaRiffle:
@robert-wagner I've added a hook on the functions of `torch.nn.functional` to see how it would work. It's a bit more complicated with `torch.nn` because we need to use Parameters which we don't support atm

robert-wagner:
At the current time this pull request does not work with the nn module. A minimal working counter example is as follows
```
import torch
import syft as sy
hook = sy.TorchHook(torch)
alice = sy.VirtualWorker(hook=hook)
x = torch.Tensor([1,2,3,4])
x_ptr = x.send(alice)
y = torch.nn.functional.relu(x_ptr)
```
The above code throws the error `TypeError: relu(): argument 'input' (position 1) must be Tensor, not PointerTensor `
Until this error is resolved I am opposed to merging headless tensors

syft/exceptions.py
```diff
@@ -1,6 +1,18 @@
 """Specific Pysyft exceptions."""
 
 
+class PointerFoundError(BaseException):
```
robert-wagner: This error name is unclear

syft/frameworks/torch/hook.py
```diff
@@ -223,32 +352,70 @@ def two_fold(lambdas, args):
...
-                args_hook_function = build_args_hook(args, rule)
-                hook_self.args_hook_for_overloaded_attr[attr] = args_hook_function
-
-            hook_args = hook_self.args_hook_for_overloaded_attr[attr]
-            new_args = hook_args(args)
-            return attr(*new_args)
-
         return overloaded_attr
 
+    # def _hook_syft_tensor(self, tensor_type: type, syft_type: type):
```
robert-wagner: Should this commented out code be deleted or uncommented

syft/frameworks/torch/hook.py
```diff
@@ -296,17 +463,18 @@ def _hook_properties(tensor_type: type):
...
-            return self.child.location
-
-        tensor_type.location = location
-
-        @property
-        def id_at_location(self):
-            return self.child.id_at_location
-
-        tensor_type.id_at_location = id_at_location
+        # Not needed in the headless setting
```
robert-wagner: Since is not needed in the headless setting should this be deleted

syft/frameworks/torch/hook.py
```diff
@@ -374,22 +542,23 @@ def _which_methods_should_we_auto_overload(self, tensor_type: type):
...
-        for attr in self.to_auto_overload[tensor_type]:
-
-            lit = getattr(tensor_type, attr)
-
-            # if we haven't already overloaded this function
-            if f"native_{attr}" not in dir(tensor_type):
-                setattr(tensor_type, f"native_{attr}", lit)
-
-            setattr(tensor_type, attr, None)
+    # Not needed at the moment
```
robert-wagner: Since this is not needed should it be deleted

syft/frameworks/torch/tensors/native.py
```diff
@@ -93,6 +94,8 @@ def create_pointer(
             ptr_id: A string or integer parameter to specify the id of the pointer
                 in case you wish to set it manually for any special reason.
                 Otherwise, it will be set randomly.
+            garbage_location: If true (default), delete the remote tensor when the
```
robert-wagner: What situations would we not want to delete the tensor when the pointer is deleted?
I can definitely see the case where there are multiple clients that want to access the same data on a worker but would that data be sent from a client or simply added on worker startup? I also think this parameter should be renamed to something like garbage_collect_data

syft/frameworks/torch/tensors/pointer.py
```diff
@@ -62,12 +63,15 @@ def __init__(
                 different from the location parameter that specifies where the
                 pointer points to.
             id: An optional string or integer id of the PointerTensor.
+            garbage_location: If true (default), delete the remote tensor when the
```
robert-wagner: See above comments on this parameter

syft/frameworks/torch/torch_attributes.py
```diff
@@ -31,7 +31,10 @@ def __init__(self, torch: ModuleType, hook: ModuleType) -> None:
         self.hook = hook
 
         # List modules that we will hook
-        self.torch_modules = {"torch": torch, "torch.nn.functional": torch.nn.functional}
+        self.torch_modules = {
+            "torch": torch,
+            "torch.nn.functional": torch.nn.functional,
+        }  # "torch.nn": torch.nn
```
robert-wagner: Comment for torch.nn should be deleted

syft/workers/base.py
```diff
@@ -2,10 +2,10 @@
 import random
 
 from abc import abstractmethod
+import syft
```
robert-wagner: nit: import syft as sy?

LaRiffle:
> At the current time this pull request does not work with the nn module. A minimal working counter example is as follows
> 
> ```
> import torch
> import syft as sy
> hook = sy.TorchHook(torch)
> alice = sy.VirtualWorker(hook=hook)
> x = torch.Tensor([1,2,3,4])
> x_ptr = x.send(alice)
> y = torch.nn.functional.relu(x_ptr)
> ```

This is due to an error in `_detail_pointer_tensor` wher the worker.id is supposed to be a str:
`alice = sy.VirtualWorker(hook=hook, id="alice")` will fix your example until I get the function corrected.

syft/frameworks/torch/tensors/native.py
LaRiffle: This is because of the way we send data.
In the previous version of pysyft we had a `private` attribute in the serialisation process to desambiguate two scenarios:
- *local* send a tensor to *bob* which must receive the tensor and not a pointer to it (send with `private = false`)
- *local* performs an op on a pointer at *bob*, and must receive a pointer, so bob send back the result with `private = true` (and in practice then the tensor is serialized without its data)

What we do instead is now for the case non-private to just pickle/ser the tensor and send it, and for the private case to create locally the pointer and send it (see base.py execute_command), partly because the built-in method for ser torch tensor keep all the data.

But of course if *bob* creates a pointer to his own result, and sends it back to *local*, then this pointer gets garbage collected and the result is deleted, which we want to avoid. That's why we add a property `dont_garbage_collect_target` (which is not serialized and sent to local)

syft/exceptions.py
LaRiffle: I'll rename it

syft/frameworks/torch/tensors/native.py
LaRiffle: Agreed on the `garbage_collect_data` name

LaRiffle:
I would be a bit more optimistic @robert-wagner 
One example maybe is torch.nn.Linear (see https://pytorch.org/docs/stable/_modules/torch/nn/modules/linear.html#Linear)
You'll "send" the module, namely all the parameters and then call it on a pointer:
```
x_ptr = torch.Tensor([1, 2]).send(bob)
lin = torch.nn.Linear(2, 2)
lin_ptr = lin.send(bob)
lin_ptr(x_ptr)
```
The `forward` is as follows:
```
def forward(self, input):
         return F.linear(input, self.weight, self.bias)
```
Which means that the input is not checked against and just `F.linear` has to be hooked.

This is just one example, but I think it works

robert-wagner:
My concern with other modules in the ecosystem has less to due with the nn module since at the very least we can hook it. My concern comes more into play when we think about other packages such as pyro. At several instances in the pyro package there are is instance checks for torch.Tensor which affect the logic of the package. Do we want to have the pysyft play nicely with these other ecosystem packages out of the box or do we want to have to hook all of them ahead of time?

LaRiffle:
>  My concern comes more into play when we think about other packages such as **pyro**. At several instances in the pyro package there are is instance checks for torch.Tensor which affect the logic of the package. Do we want to have the pysyft play nicely with these other ecosystem packages out of the box or do we want to have to hook all of them ahead of time?

Ok I see your point, again if you have an example it would definitely be great, otherwise I can try to imagine.
I think that there is a trade-off at this point:
**PRO** Faster computation as chains are shorter and less tensors have to be allocated (unfortunately I don't have numbers)
**CON** Need to specify in advance the scope of modules you want to use (at the very least with a `hook.add_hook_for(pyro)`)
**PRO** The code is simpler to write (see garbage collection)
**CON** `.send()` can't be inplace and this is painful
I can't think at others but please feel free to add them.
As I said this PR is more experimental as I wanted to see how far we can go _in practice_ with pointers. The code wouldn't not change that much if we keep heads.

iamtrask:
Btw - huge applause on the branch name - you win branch name of the year!

PR #1807, opened 2019-01-05
Add pointer tensor tutorial for torch 1

robert-wagner:
This tutorial replicates the functionality of the pointer tensor tutorial for the torch 1 branch. It is the start of having tutorials for torch_1


robert-wagner:
Hey, Thanks for the heads up @LaRiffle . I based this off of one of the examples in the torch directory but if definitely would be better to base these off of the core tutorial series. I am going to discard this pr and resubmit basing it off of the core tutorial series

PR #1808, opened 2019-01-06
[WIP] Async operations

### LINKED ISSUE #1659: PyTorch 1.0 Automatic Command Batching / Async
This is an Epic which is a sub-project under the project Torch 1.0 - the specs for which you can find here (#1587) - and the KanBan board for which you can find [here](https://github.com/orgs/OpenMined/projects/8?add_cards_query=is%3Aopen).

All work is happening in the [torch_1](https://github.com/OpenMined/PySyft/tree/torch_1) branch.

# Description:

Let's say we have a series of commands to remote tensors:
```
c = a + b
y = a + c
```
We want all of the JSON for this command to be sent (in order) asynchronously to the worker. The only way this is possible is by having the id of method results be identified by the client. In other words, when running "c = a + b" the client specifies the correct id of "c" so that it can correctly specify the following command "y = a + c" using the correct id for "c".

The way we would like to implement this so that it occurs efficiently is a try/except method. In this case, when y = a + c is sent to be executed, it will error out with a "key not found" error because c will likely not (yet) exist. At this point, we want to save the command into a key-value "command queue" where the key is the id of "c" (the one that doesn't exist yet) and the value is the serialized command instructions. This queue would be checked whenever a new tensor is registered into worker._objects. If the tensor being registered is blocking a command, then after registration that command would be released for execution.

# Sub Tickets

- [ ] #1809 
- [ ] #1810 
- [ ] #1811

mari-linhares:
Fixes #1659

- [x] Add async decorator implementation
- [ ] Make send operations async (c = a + b)  - see Issue #1809 for details
- [ ] Raise error when y = a + c is executed
- [ ] Turn off async mode - #1810
- [ ] Save y = a + b in key-value (c.id, serialized_command) command queue - #1811
- [ ] If new tensor is registered into worker._objects
- [ ] Check if tensor.id is blocking some operation, and release blocked commands

### DIFF
--- a/syft/async_call.py
+++ b/syft/async_call.py
@@ -6,7 +6,7 @@ from .exceptions import AsyncOperationNotFinalizedError
 from .exceptions import ThreadTimeOutError
 
 
-class AsyncCall(object):
+class AsyncCallWrapper(object):
     """Asynchronous call definition.
 
     Runs `self.fnc` asynchronously using `threading`.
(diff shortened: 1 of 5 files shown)


mari-linhares:
I'll close this PR for now, I've been searching more about async operations in Python and I think threading is not the way to go.

Threading is clean and easily readable. But the operational systems control when it's executed not us, and can be especially difficult to deal with shared data structures (structure would need to be locked).

Asyncio looks like a better option: async / await are builtin since python 3.5 and asyncio is an std library.

iamtrask:
Great call! Thank you for doing the due diligence research on this - sounds like we dodged a bullet.

LaRiffle:
Very interested  about what can come out with asyncio!

jvmancuso:
Asyncio definitely would be convenient as a standard library, but I'd
consider taking a look at dask distributed and/or grpc before making a
final decision.

Issue #1657, opened 2018-11-09
PyTorch 1.0 Tensor Abstractions


robert-wagner:
This is an Epic which is a sub-project under the project Torch 1.0 - the specs for which you can find here (#1587) - and the KanBan board for which you can find [here](https://github.com/orgs/OpenMined/projects/8?add_cards_query=is%3Aopen).

All work is happening in the [torch_1](https://github.com/OpenMined/PySyft/tree/torch_1) branch.

# Description

We want to build a tensor abstraction which doesn't use LocalTensor at the bottom because we don't need it anymore. Tensor chains will NOT be circular anymore! We will just call .wrap() on the top tensor whenever we need to.

We will create a tensor abstraction layer which allows one to extend any functionality in Torch tensors. Note: experiment with allowing methods to go direct as opposed to passing them through handle_call.

We will automatically generate all functionality for the PointerTensor class.

- [x] See Initial Pointer Tensor Support #1695 

Deliverables Include:

- [ ] Jupyter Notebook Tutorial for "how to create a new tensor type". This should include basic overriding as well as how to do tensors which have other tensors as attributes. It should also include the ability to serialize/deserialize the new tensor such that it can be sent and received across the network (and it works with PointerTensor).

PR #1812, opened 2019-01-09
Towards Tensor Abstraction Functionality

iamtrask:
In this PR, I'm making some simplification/code cleanup towards #1657. 

In particular, I'm working to replace (nearly) all try/except functionality with explicit if/else. This is because when try/except is used as if/else logic... it's like if/else except you don't know what the conditional logic is from reading the code. Furthermore, it starts running the code in the try...  until things break... which wastes runtime running code when we actually want to run what's in the except clause.

I also added message logging to BaseWorker (off by default)

While this PR is not end functionality, it can be merged as-is in an effort for all of us to stay up-to-date.


syft/frameworks/torch/hook.py
```diff
@@ -273,19 +275,21 @@ def overloaded_attr(_self, *args, **kwargs):
...
-            try:
+            # has_child = hasattr(_self, "child")
+            # if(has_child):
+            #     has_pointer_child = isinstance(_self.child, syft.frameworks.torch.tensors.PointerTensor)
+            # else:
+            #     has_pointer_child = False
+
+            # TODO: change if statement to "if has_pointer_child"
+
+            if not isinstance(_self, syft.frameworks.torch.tensors.PointerTensor):
```
LaRiffle: Nice trick for methods! Why did you comment the code inspecting the child? Isn't it precisely what we have to do as _self will always be of type torch.Tensor ?

syft/frameworks/torch/hook.py
```diff
@@ -273,19 +275,21 @@ def overloaded_attr(_self, *args, **kwargs):
...
                 # Transform the args
-                try:
-                    # Load the utility function to transform the args
-                    hook_args = hook_self.args_hook_for_overloaded_attr[attr]
-                    # Try running it
-                    new_self, new_args = hook_args((_self, args))
-                except IndexError:  # Update the function in cas of an error
```
LaRiffle: There is a subtleness here: some functions accept different signatures (especially something like f(*args), think of print) but the `hook_args` function is designed to work for a single signature: so it can break (with low probabilty) in which case we need to update it with the new signature conveyed by `args` using `build_hook_args_function`. Or we should store all versions instead of updating.

syft/frameworks/torch/hook.py
iamtrask: Yes - but it broke some other things. In the spirit of "incremental changes" I wanted to merge pieces of the fix earlier.

syft/frameworks/torch/hook.py
iamtrask: It seems like hook_Self.args_hook_for_overloaded_attr is being checked on line 276 - thus there's no need to do it twice, correct?

iamtrask: ok - after further discussion - i added this try/except back for now as I understand now why it's used. I also added a unit test showing what breaks if you remove it.

syft/frameworks/torch/hook.py
robert-wagner: Could you add a comment explaining why this is commented out in the code?

PR #1814, opened 2019-01-10
LogTensor and decentralized handling of commands

LaRiffle:
Add a LogTensor and change the way commands are handled:

Add a handle_command method in syft/torch tensor classes for methods:
Move the call handling from the hook to each tensor class to allow for specific behaviour depending of the class.
Add a symmetric architecture to rebuild the chain on top of the response once the chain has been completely un-wrapped (like a forward and backward logic)

Add a LogTensor

```
# build a long chain tensor
wrapper = torch.Tensor()
wrapper.child = LogTensor().on(torch.Tensor([1, 2, 3]))
x = wrapper
print(x)
# operate on it
y = x.add(x)
print(y)
```

TODO:
 - resolve some todos on imports

### DIFF
--- a/syft/frameworks/torch/tensors/__init__.py
+++ b/syft/frameworks/torch/tensors/__init__.py
@@ -1,3 +1,4 @@
+from .log import LogTensor  # noqa: F401
 from .pointer import PointerTensor  # noqa: F401
 from .abstract import AbstractTensor  # noqa: F401
 from .native import TorchTensor  # noqa: F401
--- a/syft/frameworks/torch/tensors/abstract.py
+++ b/syft/frameworks/torch/tensors/abstract.py
@@ -9,6 +9,38 @@ class AbstractTensor(ABC):
     This is the tensor abstraction.
     """
 
+    def __str__(self) -> str:
+        if hasattr(self, "child"):
+            return type(self).__name__ + ">" + self.child.__str__()
+        else:
+            return type(self).__name__
+
+    def __repr__(self) -> str:
+        if hasattr(self, "child"):
+            return type(self).__name__ + ">" + self.child.__repr__()
+        else:
+            return type(self).__name__
+
+    def on(self, tensor, wrap=True):
+        """
+        Add a syft(log) tensor on top of the tensor.
+        :param tensor: the tensor to extend
+        :param wrap: if true, add the syft tensor between the wrapper
+        and the rest of the chain. If false, just add it at the top
+        :return: a syft/torch tensor
+        """
+        if not wrap:
+            self.child = tensor
+            return self
+        else:
+            # if tensor is a wrapper
+            if not hasattr(tensor, "child"):
+                tensor = tensor.wrap()
+
+            self.child = tensor.child
+            tensor.child = self
+            return tensor
+
     def wrap(self):
         """Wraps the class inside torch tensor.
 
--- a/syft/frameworks/torch/tensors/log.py
+++ b/syft/frameworks/torch/tensors/log.py
@@ -23,22 +23,6 @@ class LogTensor(AbstractTensor):
         self.id = id
         self.child = None
 
-    def __str__(self) -> str:
-        if hasattr(self, "child"):
-            return type(self).__name__ + ">" + self.child.__str__()
-        else:
-            return type(self).__name__
-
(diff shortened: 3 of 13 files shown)


syft/frameworks/torch/tensors/__init__.py
```diff
@@ -1,3 +1,4 @@
+from .log import LogTensor  # noqa: F401
```
robert-wagner: We should make these absolute imports (I think we can just delete each of the .)

syft/frameworks/torch/tensors/abstract.py
```diff
@@ -9,6 +9,38 @@ class AbstractTensor(ABC):
     This is the tensor abstraction.
     """
 
+    def __str__(self) -> str:
```
robert-wagner: This should call __repr__ or vice versa to reuse code

syft/frameworks/torch/tensors/native.py
```diff
@@ -21,17 +22,56 @@ class TorchTensor(AbstractTensor):
     """
 
     def __str__(self) -> str:
```
robert-wagner: This functionality is in abstract tensor

syft/frameworks/torch/tensors/pointer.py
```diff
@@ -73,6 +74,28 @@ def __init__(
...
+        # Get info on who needs to send where the command
+        owner = pointer.owner
+        location = pointer.location
+
+        # Send the command
+        response = owner.send_command(location, command)
+
+        return response
+
     def __str__(self):
```
robert-wagner: This functionality is covered by abstract tensor

test/torch/test_hook.py
```diff
@@ -113,16 +113,16 @@ def test_get_pointer_to_pointer_binary_method(self, attr):
...
-    def test_hook_module_functional(self, attr):
-        self.setUp()
-        attr = getattr(F, attr)
-        x = torch.Tensor([1, -1, 3, 4])
-        expected = attr(x)
-        x_ptr = x.send(self.bob)
-        res_ptr = attr(x_ptr)
-        res = res_ptr.get()
-        assert (res == expected).all()
+    # @pytest.mark.parametrize("attr", ["relu", "celu", "elu"])
```
robert-wagner: Why is this commented out?

syft/frameworks/torch/tensors/log.py
```diff
@@ -0,0 +1,61 @@
...
+                the tensor is located.
+            id: An optional string or integer id of the LogTensor.
+        """
+        self.parent = parent
+        self.owner = owner
+        self.id = id
+        self.child = None
+
+    @classmethod
+    def handle_method_command(cls, command):
```
robert-wagner: Do this in the hook

syft/frameworks/torch/tensors/abstract.py
LaRiffle: I agree!

syft/frameworks/torch/tensors/native.py
LaRiffle: Not exactly, the code is a bit different because here you call `native_`

syft/frameworks/torch/tensors/pointer.py
LaRiffle: Again code is a bit different, that's why we overwrite the method `__str__`

test/torch/test_hook.py
LaRiffle: Because it's not working at the moment, as it's function not method and it's a bit more tricky in the new setting, but it will get back as the PR gets more mature.

syft/frameworks/torch/tensors/log.py
LaRiffle: Yup as discussed with Andrew

iamtrask:
Looking great in terms of functionality! I'm inclined to merge it as is at the moment although I think we can improve the addition of custom methods to LogTensor. (eliminate the  handle_method_command method). But in the spirit of keeping things moving I think this is a great piece of work.

syft/frameworks/torch/tensors/__init__.py
iamtrask: Will fix

PR #1821, opened 2019-01-17
Add x.max() for fixp shared tensor

LaRiffle:


iamtrask:
Thank you for adding a test as well, especially given the tight deadline

PR #1826, opened 2019-01-19
Move hook of methods

LaRiffle:
Move hook method from online class routing to pre routing in the hook

Rename LogTensor -> LoggingTensor

Add a decorator to easily over-hook method for syft tensors  …
See add() exemple in LoggingTensor

### DIFF
--- a/syft/frameworks/torch/tensors/log.py
+++ b/syft/frameworks/torch/tensors/log.py
@@ -2,21 +2,21 @@ import syft
 from syft.frameworks.torch.tensors.abstract import AbstractTensor
 
 
-class LogTensor(AbstractTensor):
+class LoggingTensor(AbstractTensor):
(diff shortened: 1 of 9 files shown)


syft/frameworks/torch/tensors/log.py
```diff
@@ -1,74 +1,59 @@
 import syft
 from syft.frameworks.torch.tensors.abstract import AbstractTensor
+from .utils import hook
```
robert-wagner: absolute imports :(

iamtrask:
@LaRiffle - would be great to fix that relative import when you get the chance - but I won't hold the PR for it.

iamtrask:
Really, really great job on this - absolutely love the smoothness of creating new tensor types! It's even better than I had hoped!

PR #1829, opened 2019-01-25
Improve Garbage Collection verifications

LaRiffle:
Reorganise test files and add one specific to GC: test.gc.py

At the end of this file, one can find test for the Logging Test. One is failing:

```
x = torch.Tensor([1, 2])

x_ptr = x.send(bob)
x_log = LoggingTensor().on(x_ptr)

assert x.id in bob._objects

del x_log
assert x.id not in bob._objects
```
However this works:
```
x = torch.Tensor([1, 2])

x_log = LoggingTensor().on(x.send(bob))

assert x.id in bob._objects

del x_log
assert x.id not in bob._objects
```
This is due to x_ptr init, which increases the reference of x_log to 3  (should be 2 if you use `sys.getrefcount(x_log)`).
Is this something that we want as a behavior ?


iamtrask:
This is NOT expected behavior and we should definitely get that REF count back down to 2 using weakref.

LaRiffle:
But then x_ptr should be del as well, no ?

iamtrask:
Yeah - all these are test cases which we want to have support for. It doesn't all have to happen at once, but it's a VERY good idea to get this working before we start adding lots of complex functionality because GC is such a tough thing to add in later.

PR #1834, opened 2019-01-29
Polynomial Tensor Torch_1

kamathhrishi:
Starting Polynomial Tensor PR for Torch_1. Currently a work in progress with this being the first version.


robert-wagner:
Hey @kamathhrishi you want to run `black test` on your code to fix the travis error

syft/frameworks/torch/tensors/Polynomial.py
```diff
@@ -0,0 +1,77 @@
...
+            
+            x: A float value
+           
+            return: 
+               
+            approximation of the log function as a float"""
+
+        return (x) - ((x ** 2) / 2) + ((x ** 3) / 3) - ((x ** 4) / 4)
+
+    def EvalRelative(x_true, x_pred):
```
robert-wagner: is this method a part of the original tensor api? if not what is the purpose of this method

syft/frameworks/torch/tensors/Polynomial.py
```diff
@@ -0,0 +1,77 @@
+class PolynomialTensor:
+
+    """Tensor type which provides polynomial approximation functionalities using Taylor Series expansion
+       since computing exact functions could impose a overhead on computation
+    """
+
+    def sigmoid(self, x: float) -> float:
```
robert-wagner: this should not take another value and should just be called on self. Alternatively this could be made a class method and take a tensor instead of a float

syft/frameworks/torch/tensors/Polynomial.py
```diff
@@ -0,0 +1,77 @@
...
+            
+           x: A float value
+           
+           return: 
+               
+           approximation of the sigmoid function as a float"""
+
+        return (1 / 2) + (x / 4) - (x ** 3 / 48) + (x ** 5 / 480)
+
+    def exp(self, x: float, order: int) -> float:
```
robert-wagner: same as above, this should either operate on self or be a class method. Additionaly should other functions have the option to choose order or should we pick a standard order for the entire class?

syft/frameworks/torch/tensors/Polynomial.py
```diff
@@ -0,0 +1,77 @@
...
+        value = 1
+
+        for i in range(1, order):
+
+            value += (x ** order) / denom
+            denom *= i + 1
+
+        return value
+
+    def tanh(self, x: float) -> float:
```
robert-wagner: same as above as to operating on self or making a class method

syft/frameworks/torch/tensors/Polynomial.py
```diff
@@ -0,0 +1,77 @@
...
+            
+            x: A float value
+           
+            return: 
+               
+            approximation of the tanh function as a float"""
+
+        return (x) + ((x ** 3) / 3) + (((x ** 5) / 120)) + ((x ** 7) / 5040)
+
+    def log(self, x: float) -> float:
```
robert-wagner: same as above as to operating on self or making a class method

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,60 @@
...
+
+class TestPolynomialTensor:
+
+    """ Test cases to ensure working of Polynomial Tensor. The tests under these ensure that the error between actual funcion values and approximations do not deviate too much"""
+
+    def __init__(self):
+
+        self.tensor = PolynomialTensor()
+        # Maximum permissible error as calculated by EvalRelative under PolynomialTensor
+        self.threshold = 0.1
```
robert-wagner: The error of polynomial approximations grows without bound the further you go from zero. It seems a bit missleading to say we have a maximum permissible error when that is not the case

syft/frameworks/torch/tensors/Polynomial.py
```diff
@@ -0,0 +1,77 @@
+class PolynomialTensor:
```
robert-wagner: polynomial tensor should be a subclass of AbstractTensor. Furthermore we probably should have a constructor which takes an order parameter which specifies the order of the polynomial approximation at definition time (with a sensible default of 4 or 5)

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,60 @@
...
+
+import math
+
+class TestPolynomialTensor:
+
+    """ Test cases to ensure working of Polynomial Tensor. The tests under these ensure that the error between actual funcion values and approximations do not deviate too much"""
+
+    def __init__(self):
+
+        self.tensor = PolynomialTensor()
```
robert-wagner: Each test should get its own tensor to avoid side effects

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,60 @@
+import syft 
+from syft.frameworks.torch.tensors.Polynomial import PolynomialTensor
+
+import math
```
robert-wagner: we probably should be comparing to the torch.nn.functional definitions of these functions rather than the math ones as we are trying to preserve feature parity with torch

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def EvalRelative_Test(self, x_true, x_pred):
+
+        assert self.tensor.EvalRelative(15 / 7.5) == 0.5
+        assert self.tensor.EvalRelative(15 / 1.5) == 0.9
+        assert self.tensor.EvalRelative(15 / 15) == 0
+
+    def SigmoidTest(self):
+
+        test_range = 4
```
robert-wagner: why does test range vary between the tests (is there a reason we are testing this function over a smaller range

syft/frameworks/torch/tensors/Polynomial.py
robert-wagner: Looking at it further, it seems to me that this should be a helper function that lives in the test file

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,60 @@
...
+        for i in range(-test_range, test_range, 1):
+
+            assert (
+                self.tensor.EvalRelative(self.tensor.exp(i / 10), math.exp(i / 10))
+                < self.threshold
+            )
+
+    def ExpTest(self):
+
+        test_range = 10
```
robert-wagner: ditto as to test range question from above

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,60 @@
...
+        for i in range(-test_range, test_range, 1):
+
+            assert (
+                self.tensor.EvalRelative(self.tensor.exp(i / 10), math.exp(i / 10))
+                < self.threshold
+            )
+
+    def tanhTest(self):
+
+        test_range = 9
```
robert-wagner: ditto as to test range question from above

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,60 @@
...
+        for i in range(-test_range, test_range, 1):
+
+            assert (
+                self.tensor.EvalRelative(self.tensor.tanh(i / 10), math.tanh(i / 10))
+                < self.threshold
+            )
+
+    def LogTest(self):
+
+        test_range = 4
```
robert-wagner: ditto as to test range question from above

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,60 @@
...
+    def EvalRelative_Test(self, x_true, x_pred):
+
+        assert self.tensor.EvalRelative(15 / 7.5) == 0.5
+        assert self.tensor.EvalRelative(15 / 1.5) == 0.9
+        assert self.tensor.EvalRelative(15 / 15) == 0
+
+    def SigmoidTest(self):
+
+        test_range = 4
+        for i in range(-test_range, test_range, 1):
```
robert-wagner: It might give clearer error messages if we parameterize these rather than putting a for loop in the test

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,60 @@
+import syft 
+from syft.frameworks.torch.tensors.Polynomial import PolynomialTensor
+
+import math
+
+class TestPolynomialTensor:
```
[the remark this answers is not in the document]
kamathhrishi: Could you elaborate more on this issue?

test/torch/tensors/test_poly.py
kamathhrishi: Nope

robert-wagner: Could we test the same range for each then?

PR #1837, opened 2019-01-29
Parameter hooking

iamtrask:
ptr = torch.Tensor([1.0, -1.0, 3.0, 4.0]).send(bob)
print(ptr.child)
p = nn.Parameter(ptr)
p.data = ptr
print(p.data.child) # --> trigger attribute error

This used to fail but now it does not

### DIFF
--- a/syft/frameworks/torch/hook.py
+++ b/syft/frameworks/torch/hook.py
@@ -3,6 +3,8 @@ import re
 import random
 import logging
 import types
+import torch
+import torch.nn as nn
 from functools import wraps
 
 
@@ -117,6 +119,8 @@ class TorchHook:
         # hooks the tensor constuctor function
         self._hook_tensor()
 
+        self._hook_parameters()
+
         self._hook_torch_module()
 
         # Add the local_worker to syft so that it can be found if the hook is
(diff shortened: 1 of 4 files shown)


syft/frameworks/torch/hook.py
```diff
@@ -210,6 +214,61 @@ def _hook_pointer_tensor_methods(self):
...
 
+    def _hook_parameters(self):
+        """
+        This method overrides the torch Parameter class such that
+        it works correctly with our overridden tensor types. The
+        native torch Parameter class kept deleting all of our
+        attributes on our custom tensors, so we wrote our own.
+        """
+
+        class Parameter:
```
robert-wagner: Is there a way we can code gen this? I would hate to see pysyft break because they changed the source of parameter and we didn't manually copy it

syft/workers/base.py
```diff
@@ -376,8 +376,8 @@ def de_register_obj(self, obj, _recurse_torch_objs=True):
 
         if hasattr(obj, "id"):
             self.rm_obj(obj.id)
-        if hasattr(obj, "owner"):
-            del obj.owner
+        if hasattr(obj, "_owner"):
+            del obj._owner
```
robert-wagner: This is a good change

syft/frameworks/torch/hook.py
iamtrask: Not really - because we don't really use any of their code other than __repr__.

syft/workers/base.py
iamtrask: i don't remember making this change

PR #1839, opened 2019-01-29
Remote Parameters and nn modules hooking

LaRiffle:
Change the way Parameter is hooked
and hook module,
Try to have a basic fed learnin example working

I think the PR is ok for now. We're really close to have this FL example (just try running tutorial Part2).


iamtrask:
looks like there's a missing "import copy" in hook.py

Issue #1835, opened 2019-01-29
Test hook.py should be modified to not use class structure


robert-wagner:
We do not need to use the class structure required by unittest with pytest. Doing things in the pytest style of not having tests inside a testing class makes cleaner code and avoids side effects of tests interacting. As such the test_hook.py file should be refactored to the pytest style (see other tests for examples)

alexis-thual:
I could try and do this one if you want to assign me :)

alexis-thual:
Sorry akki, there seem to be some overlap in our works...
I basically refactored all test files, except from `./test/workers/test_worker.py` and `./test/test_serde.py`.

@robert-wagner it seems that I can't add you to the reviewers of the PR I just opened. I left a couple of comments in the PR if you want to read them first!

robert-wagner:
As this work has been done twice, I'm closing this now

PR #1841, opened 2019-02-01
change tests to not use class structure and fix some typos

akki2825:
in reference to https://github.com/OpenMined/PySyft/issues/1835

### DIFF
--- a/test/torch/test_hook.py
+++ b/test/torch/test_hook.py
@@ -1,4 +1,4 @@
-""""Tests relative to verifying the hook process behaves properly."""
+"""Tests relative to verifying the hook process behaves properly."""
 
 import pytest
 import torch
@@ -9,10 +9,10 @@ import random
 import syft
 from syft.exceptions import RemoteTensorFoundError
 from syft.frameworks.torch.tensors import PointerTensor
+from test.conftest import hook
 
 
-def test__init__():
-    hook = syft.TorchHook(torch, verbose=True)
+def test__init__(hook):
     assert torch.torch_hooked
     assert hook.torch.__version__ == torch.__version__
 
@@ -27,9 +27,7 @@ def test_torch_attributes():
     syft.torch._command_guard("torch.add", "torch_modules", get_native=False)
 
 
-def test_worker_registration():
-    hook = syft.TorchHook(torch, verbose=True)
-
+def test_worker_registration(hook):
     me = hook.local_worker
     boris = syft.VirtualWorker(id="boris", hook=hook, is_client_worker=False)
     me.add_workers([boris])
@@ -39,9 +37,7 @@ def test_worker_registration():
     assert boris == worker
 
 
-def test_pointer_found_exception():
-    hook = syft.TorchHook(torch, verbose=True)
-
+def test_pointer_found_exception(hook):
     me = hook.local_worker
 
     instance_id = str(int(10e10 * random.random()))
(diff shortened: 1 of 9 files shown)


test/torch/test_hook.py
```diff
@@ -1,4 +1,4 @@
-"""Tests relative to verifying the hook process behaves properly."""
+""""Tests relative to verifying the hook process behaves properly."""
```
LaRiffle: This is a typo I assume

robert-wagner: There are too many quotes here

test/torch/test_hook.py
```diff
@@ -11,194 +11,227 @@
...
+    syft.torch._command_guard("torch.add", "torch_modules", get_native=False)
+
+
+def test_worker_registration():
+    hook = syft.TorchHook(torch, verbose=True)
+
+    me = hook.local_worker
+    boris = syft.VirtualWorker(id="boris", hook=hook, is_client_worker=False)
+    me.add_workers([boris])
+
```
LaRiffle: I'm not sure this is what we want to do: this involves massive duplicate code that self.setUp() did avoid previously.

LaRiffle: @robert-wagner what code design where you thinking of?

test/torch/test_hook.py
```diff
@@ -11,194 +11,227 @@
...
+    m = MyLayer()
+    out = list(m.parameters())
+    assert len(out) == 1
+    assert out[0] == m.some_params
+
+
+def test_torch_module_hook():
+    """Tests sending and getting back torch nn module like nn.Linear"""
+
+    hook = syft.TorchHook(torch, verbose=True)
```
robert-wagner: In general we can use the hook from conftest and don't need to rehook for each test. This can be done by passing hook as a parameter to each method. Also it should make the tests run faster

test/torch/test_hook.py
robert-wagner: @LaRiffle we don't need to call hook, we should be passing as arguments to the tests the fixtures in conftest as needed

LaRiffle: Perfect that works for me then :)

robert-wagner: The general idea was to reduce code duplicate across the project by moving the functionality of setup to a common fixture

test/torch/test_hook.py
akki2825: Thanks for the feedback. Changes have been made.

test/torch/test_hook.py
```diff
@@ -72,9 +68,7 @@ def test_build_get_child_type():
 
 
 @pytest.mark.parametrize("attr", ["abs"])
-def test_get_pointer_unary_method(attr):
-    hook = syft.TorchHook(torch, verbose=True)
-
+def test_get_pointer_unary_method(attr, hook):
     instance_id = str(int(10e10 * random.random()))
 
     bob = syft.VirtualWorker(id=f"bob{instance_id}", hook=hook, is_client_worker=False)
```
LaRiffle: You can also use the same mecanism to load the workers (see in conftest the workers fixture)
This way it will be even more compact :)

akki2825: Oh yeah. On it!

akki2825: Wondering same can be done to other tests (files) ?

robert-wagner: Alot of them already have been. Feel free to update any that you see

LaRiffle: Would be super cool to have this done in all the files! 💯

akki2825: Will do it! :D

akki2825: `hook` as arg is not required when `workers` is passed. Changes have been made

robert-wagner:
This looks good after it is rebased on torch_1

akki2825:
@robert-wagner merge conflicts have been resolved and rebased!

robert-wagner:
Hey @akki2825 Thank you very much for your work this pr. It looks like #1851 already implemented the same functionality (Looks like some wires got crossed somewhere). As such I am closing this.

PR #1848, opened 2019-02-02
federated learning working but there are still some strange bugs about

iamtrask:


iamtrask:
I had to comment out a few unit tests - strangely all of the failing tests work when i run them in a notebook - but they struggle when in the unit test suite. Very strange.

PR #1865, opened 2019-02-04
Fix tests and improve coverage

LaRiffle:


syft/frameworks/torch/federated/dataloader.py
```diff
@@ -28,22 +28,22 @@ def default_collate(batch):
         elem_type.__module__ == "numpy"
         and elem_type.__name__ != "str_"
         and elem_type.__name__ != "string_"
-    ):
+    ):  # pragma: no cover
```
robert-wagner: why are these all here?

test/torch/federated/test_dataset.py
```diff
@@ -20,13 +20,10 @@ def test_federated_dataset(workers):
...
-        i = 0
-        loss_accum = 0
-        while fd.keep_going():
-            i += 1
-
-            data, target = fd.step()
+    epochs = 2
+    for epoch in range(1, epochs + 1):
+        for batch_idx, (data, target) in enumerate(train_loader):
+            pass
```
robert-wagner: what is this testing? This seems like it purely makes pytest think that that code was run while testing nothing

test/torch/tensors/test_logging.py
```diff
@@ -117,3 +121,16 @@ def test_remote_function_on_log_chain(workers):
...
+def test_print_log_chain():
+    """
+    Test sending and getting back a chain including a logtensor
+    """
+
+    # Build a long chain tensor Wrapper>LoggingTensor>TorchTensor
+    x_tensor = torch.Tensor([1, 2, 3])
+    x = LoggingTensor().on(x_tensor)
+
+    assert isinstance(x.__str__(), str)
```
robert-wagner: we should test that we are actually printing the right string

syft/frameworks/torch/federated/dataloader.py
LaRiffle: Before it is from pytorch - just kept it for generality purposes

test/torch/federated/test_dataset.py
LaRiffle: It is testing a no failing - we're a bit short of time so I didn't add exptensive tests

test/torch/tensors/test_logging.py
LaRiffle: Printing is hard because it changes a lot (we just changed pointer printing), meaning that we need to rewriting the tests very often and copy pasting strings (with ids in it etc) is painful.

syft/frameworks/torch/federated/dataloader.py
robert-wagner: I mean the pragmas

test/torch/federated/test_dataset.py
robert-wagner: Is there a way that we can test this for real?

test/torch/tensors/test_logging.py
robert-wagner: How often do we plan on changing this? It seems to me like something that would not change often

iamtrask: If history is any indicator - once every couple months

test/torch/federated/test_dataset.py
iamtrask: Good question - strictly speaking it's job is to just feed data. I suppose one could check:
- to make sure it feeds all the data once per iteration
- to make sure it doesn't feed remote data from different workers.

but neither are worth blocking this at the moment.

test/torch/tensors/test_logging.py
robert-wagner: lol

test/torch/federated/test_dataset.py
robert-wagner: Valid. We should open an issue for it

syft/frameworks/torch/federated/dataloader.py
LaRiffle: This is to avoid coverage taking this into account. We don't and won't have tests for this part in the middle term

test/torch/federated/test_dataset.py
LaRiffle: 👍

PR #1877, opened 2019-02-05
Debugger help

LaRiffle:
As we'll dive into debug logs, let's have more explicit stack traces --> 
Change hook `overloaded_attr` to explicit names like `overloaded_pointer_method`, `overloaded_syft_method`  for stacktrace clarity


iamtrask:
Good idea - also like the "many_fold" final option.

PR #1879, opened 2019-02-05
[Wont be merged] added intial functionality for matrix multiplication

robert-wagner:


robert-wagner:
After further review, this pull request will not be merged because the logic for generating tripples is not general enough. Just keeping this around for my reference until the new version is written

PR #1887, opened 2019-02-08
Add verbose workers

robert-wagner:
Should this go to dev or master? Doesn't break anything


iamtrask:
Just as a matter of practice -  let's do dev

iamtrask:
seems customary to only update master when we actually change version numbers (which we can totally do regularly... but we want to be intentional about it)

iamtrask:
this will allow people to know that if they keep using the same version of the software nothing changes

robert-wagner:
Sounds good

Issue #1893, opened 2019-02-10
Moving a module to cuda throws errors


robert-wagner:
```
import numpy as np                                                                                                                                         
import torch                                                                                                                                               
import torch.utils.data as data                                                                                                                            
import torch.nn as nn                                                                                                                                      
import torch.nn.functional as F                                                                                                                            
import torch.optim as optim                                                                                                                                
from torchvision import datasets, transforms                                                                                                               
                                                                                                                                                           
import torch                                                                                                                                               
from torch import nn                                                                                                                                       
from torch import optim                                                                                                                                    
from torchvision.datasets.mnist import MNIST                                                                                                               
import pdb                                                                                                                                                 
                                                                                                                                                           
import syft as sy                                                                                                                                          
class Net(nn.Module):                                                                                                                                      
    def __init__(self):                                                                                                                                    
        super(Net, self).__init__()                                                                                                                        
        self.conv1 = nn.Conv2d(1, 20, 5, 1)                                                                                                                
        self.conv2 = nn.Conv2d(20, 50, 5, 1)                                                                                                               
        self.fc1 = nn.Linear(4*4*50, 500)                                                                                                                  
        self.fc2 = nn.Linear(500, 10)                                                                                                                      
                                                                                                                                                           
    def forward(self, x):                                                                                                                                  
        x = F.relu(self.conv1(x))                                                                                                                          
        x = F.max_pool2d(x, 2, 2)                                                                                                                          
        x = F.relu(self.conv2(x))                                                                                                                          
        x = F.max_pool2d(x, 2, 2)                                                                                                                          
        x = x.view(-1, 4*4*50)                                                                                                                             
        x = F.relu(self.fc1(x))                                                                                                                            
        x = self.fc2(x)                                                                                                                                    
        return F.log_softmax(x, dim=1)                                                                                                                     
                                                                                                                                                           
                                                                                                                                                           
hook = sy.TorchHook(torch)                                                                                                                                 
device = torch.device("cuda")                                                                                                                              
model = Net().to(device)                                                                                                                                   
print(model) ```

**Comment by LaRiffle** [comment]
Can you add the stacktrace?
One **good-first-issue** would be in the case you have only cpu, and try to call .to(device), to fix the error which is only due to us not serializing devices. This could be easily handled.

**Comment by jlebensold** [comment]
```
Traceback (most recent call last):
  File "test.py", line 37, in <module>
    model = Net().to(device)                                                                                                                                   
[...]

PR #1894, opened 2019-02-10
Towards a complete example on MNIST with CNN

LaRiffle:
## Real world example of PySyft
The goal of this PR is to have a canonical example of PySyft in a "real world" setting, here MNIST + CNN.
We want to show how little changes there are to move from local pytorch to federated learning with pysyft.

## Roadmap to follow up on this PR

 -  **momentum args in optim SGD** #1896
 -  **use `FederatedDataset` & `FederatedDataLoader` @jonlebensold**
 -  Support .to(device) hooking #1893
 -  Warning in log_softmax -> no hook_kwargs


iamtrask:
For this demo - can we move the MNIST pre-processing code to the sandbox?

jlebensold:
@LaRiffle and I are working through an example.

I think that any MNIST demo needs to have the following design requirements:

1. The orchestrator (federated server) should not know about the individual datasets or tensors on each worker
2. CUDA support
3. An approach to messaging that is protocol independent and supports different implementations of the same code

LaRiffle:
@jlebensold I believe that this MNIST demo is intended for people joining OM with a PyTorch background, so we should be as close as possible from the PyTorch official example to that people understand that FL can be turned on with very few changes.
 So I would rather focus on having a search + FederatedDataset + FederatedLoader stack for the data (very close to the official example)
and also I agree cuda would be cool, but I don't know how hard this is.

iamtrask:
@LaRiffle - ok good point - maybe we can do a followon later which really leans on search as sortof "what deep learning will look like in the future." This would actually make for a killer blogpost too.

test/torch/test_federated_learning.py
```diff
@@ -5,7 +5,7 @@
 
 hook = sy.TorchHook(torch)
 from torch import nn
-from syft import optim
+from torch import optim
```
iamtrask: i can't believe you got this working! this is awesome!

PR #1908, opened 2019-02-14
Fix remote in-place method garbage collection error

### LINKED ISSUE #1896: Garbage Collection issue with in-place methods on tensors
This triggers a KeyError. I suspect unexpected garbage collection.
```
buf = torch.tensor([[1., 2], [4., 2]]).send(bob)
buf.add_(buf)
buf.get()
```

LaRiffle:
Fix #1896

### DIFF
--- a/syft/frameworks/torch/torch_attributes.py
+++ b/syft/frameworks/torch/torch_attributes.py
@@ -240,9 +240,12 @@ class TorchAttributes(object):
         :param method_name: the name for the method
         :return: boolean
         """
-        if method_name not in self.inplace_methods:
-            self.inplace_methods[method_name] = method_name[-1] == "_" and "__" not in method_name
-        return self.inplace_methods[method_name]
+        try:
+            return self.inplace_methods[method_name]
+        except KeyError:
+            is_inplace = method_name[-1] == "_" and "__" not in method_name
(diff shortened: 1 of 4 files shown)


syft/frameworks/torch/torch_attributes.py
```diff
@@ -232,6 +234,16 @@ def get_native_torch_name(attr: str) -> str:
         native_func_name = ".".join(parts)
         return native_func_name
 
+    def is_inplace_method(self, method_name):
+        """
+        Says if a method is inplace or not by test if it ends by _ and is not a __xx__
```
iamtrask: is this going to be called during runtime or during hooking? If during runtime - is there a way we can avoid the string manipulation?

iamtrask: Especially if this method is going to be executed for _every_ command

iamtrask: oh wait - actually i see what's going on.

Oddly I think this is a good place to use a try/catch for speed. That way we don't have to check whether method_name is in self.inplace_methods twice (needlessly)

iamtrask: *gasp!!!*

PR #1911, opened 2019-02-15
Improved __str__, MAJOR notebook bugfix, and a few small conveniences

iamtrask:


syft/frameworks/torch/tensors/interpreters/add.py
```diff
@@ -43,6 +43,16 @@ def __init__(
         self.Q_BITS = Q_BITS
         self.field = (self.BASE ** Q_BITS) - 1  # < 63 bits
 
+    def __repr__(self):
+        return self.__str__()
+
+    def __str__(self):
+        type_name = type(self).__name__
+        out = f"[" f"{type_name}]"
```
jlebensold: @iamtrask is that valid python?

PR #1912, opened 2019-02-15
Polynomial Tensor 

kamathhrishi:
Continuation of #1834.. Currently in Progress.  In reference to #1506.


syft/frameworks/torch/tensors/interpreters/Polynomial.py
```diff
@@ -0,0 +1,54 @@
+from syft.frameworks.torch.tensors.interpreters.abstract import AbstractTensor
+
+
+class PolynomialTensor(AbstractTensor):
+
+    """Tensor type which provides polynomial approximation functionalities using Taylor Series expansion
+       since computing exact functions could impose a overhead on computation
+    """
+
+    def sigmoid(self, x: float) -> float:
```
robert-wagner: So this method should either operate on self or be a class method

syft/frameworks/torch/tensors/interpreters/Polynomial.py
```diff
@@ -0,0 +1,54 @@
...
+            
+           x: A float value
+           
+           return: 
+               
+           approximation of the sigmoid function as a float"""
+
+        return (1 / 2) + (x / 4) - (x ** 3 / 48) + (x ** 5 / 480)
+
+    def exp(self, x: float) -> float:
```
robert-wagner: Same as above, this should operate on self. See add.py for examples for methods like __add__ (or mul as soon as I finish debugging it)

syft/frameworks/torch/tensors/interpreters/Polynomial.py
```diff
@@ -0,0 +1,54 @@
...
+            
+           x: (float) value for which approximation must be found
+           
+           return: 
+               
+           value: (float) approximation of the exponential function """
+
+        return 1 + x + (x ** 2 / 2) + (x ** 3 / 6) + (x ** 4 / 24) + (x ** 5 / 120) + (x ** 6 / 840)
+
+    def tanh(self, x: float) -> float:
```
robert-wagner: Same as above

syft/frameworks/torch/tensors/interpreters/Polynomial.py
```diff
@@ -0,0 +1,54 @@
+from syft.frameworks.torch.tensors.interpreters.abstract import AbstractTensor
+
+
+class PolynomialTensor(AbstractTensor):
```
robert-wagner: This should probably subclass AdditiveSharingTensor

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,102 @@
...
+
+    assert EvalError(three, one) == 50
+    assert EvalError(one, three) == 50
+
+
+def testSigmoid():
+
+    Ptensor = PolynomialTensor()
+
+    x = torch.randn(DATA_SIZE)
```
robert-wagner: Evaluate on deterministic input

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,102 @@
...
+
+    m = torch.nn.Sigmoid()
```
robert-wagner: Don't compare performance to nn.sigmoid, compare it to what the value of the approximation should be. We don't care how good the polynomial approximation is in terms of accuracy, we care how accurately it is implemented

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,102 @@
...
+    ten = m(x)
+
+    assert (EvalError(ten, Ptensor.sigmoid(x))) < DATA_THRESHOLD
+
+
+def testExp():
+
+    Ptensor = PolynomialTensor()
+
+    x = torch.randn(DATA_SIZE)
```
robert-wagner: See above

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,102 @@
...
+
+    ten = torch.exp(x)
+
```
robert-wagner: See above

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,102 @@
...
+
+def testtanh():
+
+    # Ideally the error for any approximation to be under DATA_THRESHOLD. For tanh it could be not be improved beyond 12 , which should be addressed in the future.
+    Tanh_thresh = 12
+
+    Ptensor = PolynomialTensor()
+
+    x = torch.randn(DATA_SIZE)
```
robert-wagner: See above

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,102 @@
...
+
+    ten = torch.tanh(x)
```
robert-wagner: See above

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,102 @@
...
+def testSigmoid():
+
+    Ptensor = PolynomialTensor()
+
+    x = torch.linspace(-3, 3, steps=10)
+    expected = torch.tensor(
+        [-0.1938, 0.0372, 0.1530, 0.2688, 0.4174, 0.5826, 0.7313, 0.8470, 0.9628, 1.1938]
+    )
+
+    assert EvalError(expected, Ptensor.sigmoid(x)) == True
```
robert-wagner: You should just exact compare these

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,102 @@
...
+def testExp():
+
+    Ptensor = PolynomialTensor()
+
+    x = torch.linspace(-3, 3, steps=10)
+    expected = torch.tensor(
+        [0.2179, 0.1224, 0.1905, 0.3679, 0.7165, 1.3956, 2.7179, 5.2813, 10.1764, 19.2679]
+    )
+
+    assert EvalError(expected, Ptensor.exp(x)) == True
```
robert-wagner: Same as above

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,102 @@
...
+            -3.2151e-01,
+            3.2151e-01,
+            7.6790e-01,
+            2.0803e00,
+            3.1835e01,
+            3.3883e02,
+        ]
+    )
+
+    assert EvalError(expected, Ptensor.tanh(x)) == True
```
robert-wagner: Same as above

syft/frameworks/torch/tensors/interpreters/Polynomial.py
```diff
@@ -0,0 +1,57 @@
...
+            
+           x: Torch tensor
+           
+           return: 
+               
+           approximation of the sigmoid function as a torch tensor"""
+
+        return (1 / 2) + (x / 4) - (x ** 3 / 48) + (x ** 5 / 480)
+
+    def exp(x: torch.tensor) -> torch.tensor:
```
robert-wagner: I think these should probably operate on self

syft/frameworks/torch/tensors/interpreters/Polynomial.py
```diff
@@ -0,0 +1,57 @@
+from syft.frameworks.torch.tensors.interpreters.add import AdditiveSharingTensor
+import torch
+
+
+class PolynomialTensor(AdditiveSharingTensor):
```
iamtrask: IMO - PolynomialTensor (PT) should not be a subclass of AdditiveSharingTensor (AST) because while AST will use PT, PT will also be used in an identical way by other tensors (such as Zero Knowledge Tensor). Thus, we want PT to be a tensor on its own (extending AbstractTensor) with a .child which is assumed to be an integer type. The main unit tests can then proceed with IntegerTensor underneath with some light unit testing of compatability with AdditiveSharingTensor

syft/frameworks/torch/tensors/interpreters/Polynomial.py
robert-wagner: Sill agree with this concept that for all of these x should actually be self

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,102 @@
...
+        """
+
+    within_range = torch.abs(x_true - x_pred) < MARGIN
+    within_range = torch.eq(within_range.all(), 1)
+    within_range = within_range.item() == 1
+
+    return within_range
+
+
+def test_EvalError():
```
robert-wagner: We don't need EvalError any more if we are testing exact values

syft/frameworks/torch/tensors/interpreters/Polynomial.py
```diff
@@ -0,0 +1,141 @@
...
+            + (function((x ** 3) / 6))
+            + (function((x ** 4) / (24)))
+            + (function((x ** 5) / (120)))
+            + (function((x ** 6) / (840)))
+            + (function((x ** 7) / (6720)))
+        )
+
+    def exp_inter(self, x, function=torch.tensor):
+
+        raise NotImplementedError
```
robert-wagner: Should these be implemented?

syft/frameworks/torch/tensors/interpreters/Polynomial.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def tanh_inter(self, x, function=torch.tensor):
+
+        raise NotImplementedError
```
robert-wagner: Should these be implemented?

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,92 @@
...
+    expected = torch.tensor([1.2220, 2.9582, 7.1763, 20.3064, 54.4606], dtype=torch.double)
+
+    Ptensor = PolynomialTensor()
+
+    f1 = Ptensor.interpolate((lambda x: np.exp(x)), np.linspace(0, 10, 50))
+
+    assert torch.allclose(expected, torch.tensor(f1(torch.tensor([0, 1, 2, 3, 4]))), 1e-04)
+
+
+testSigmoid()
```
robert-wagner: Why is this here?

test/torch/tensors/test_poly.py
```diff
@@ -0,0 +1,160 @@
...
+    assert torch.allclose(expected, torch.tensor(result, dtype=torch.double), atol=1e-03)
+
+
+def testfixedprecision():
+
+    x_tensor = torch.Tensor([1, 2, 3])
+    x = FixedPrecisionTensor().on(x_tensor)
+
+
+testfixedprecision()
```
robert-wagner: Why is this here?

robert-wagner: Bump

PR #1914, opened 2019-02-18
Add serialization of torch.device to support hook of .to(device)

LaRiffle:
Just a serialization of torch device. Issues with Gpu not solved


iamtrask:
I'm not sure that's possible in travis - we don't have GPU support there.

LaRiffle:
It's not about GPUs at the moment, it's juste about sending a command .to('cpu') over the network

LaRiffle:
@robert-wagner I've added the test!

PR #1897, opened 2019-02-11
Work on fix_precision features

LaRiffle:
This add improved usage for fixed_precision.
I used it to convert models in integers for simulation of Functional Encryption [here](https://github.com/LaRiffle/collateral-learning/blob/master/notebooks/Part%205%20-%20Towards%20real%20FE%20-%20Quadratic%20network%20in%20fixed%20precision.ipynb)

PR #1916, opened 2019-02-18
Mpc multiplication

robert-wagner:
Creating a pr for this to make what I am working on more public. Currently we have multipication working without truncation. I want to make sure that the truncation code is in line with #1897 @LaRiffle Could we set up a time to discuss?


iamtrask:
merged a conflict

LaRiffle:
I think it is now fixed. The major issue you add was `% self.field` because it is not behaving as you would expect: `-3 % 100 == 97 != -3`
I've created an algebric_modulo method which fixes this, but for now I've just removed moduli, they are only and however needed with larger numeric values, so algebric_modulo will be useful

Issue #1838, opened 2019-01-29
Add data to workers on spinup


robert-wagner:
We want to be able to initialize workers with data. In practice we want to pass data as a parameter to the workers. In practice we want that parameter to either take a tensor object, a list of tensor objects or a file pointer to the data.

kamathhrishi:
We should close this issue?

PR #1936, opened 2019-02-20
Adding Data workers on Signup 

kamathhrishi:
Addresses Issue #1838


syft/workers/base.py
```diff
@@ -50,9 +51,11 @@ class BaseWorker(AbstractWorker):
     """
 
     def __init__(
-        self, hook, id=0, known_workers={}, is_client_worker=False, log_msgs=False, verbose=False
+        self, hook, id=0, known_workers={},Data={},is_client_worker=False, log_msgs=False, verbose=False
```
iamtrask: Data should be lower case

syft/workers/base.py
```diff
@@ -50,9 +51,11 @@ class BaseWorker(AbstractWorker):
...
     ):
         """Initializes a BaseWorker."""
+        
+        self.Load_Data(Data)
```
iamtrask: method Load_Data should be lower case

syft/workers/base.py
```diff
@@ -117,6 +122,15 @@ def _recv_msg(self, message):
 
         """
         raise NotImplementedError  # pragma: no cover
+        
+    def Load_Data(self,Data):
```
iamtrask: I think all data will be garbage collected automatically (deleted) after this method runs. Instead each tensor should be registered (i.e., don't use .send(), use register_obj())

PR #1938, opened 2019-02-22
Add draft version of a plan

LaRiffle:
A Plan is a specific kind of worker, to which you give your commands, but which doesn't execute them until you specifically ask a response (through a .get() for instance), and then it execute all the commands in order and return the response. This is a we call a plan.
The idea is to send a "plan" ie all the instructions at once to a remote worker, while you still execute your code line by line like in PyTorch.
It uses the fact that now the worker that sends the command fix the id where the result is stored and therefore and can already create a pointer to a result which doesn't exist so far and go on with instructions. 

Here is how you make and use a plan
```
import syft as sy
import torch as th

hook = sy.TorchHook(th)

#build "locally"
@sy.func2plan()
def plan_a(data):
    return data.abs()

x = th.tensor([-1,2,3])
y = plan_a(x)

#build "remotely"
@sy.func2plan()
def plan_b(data):
    return -data.abs()

plan_b.send(bob)
x_ptr = th.tensor([-1,7,3]).send(bob)
p = plan_b(x_ptr)
print(p.get())

plan_b.get()
plan_b.send(alice)

x_ptr = th.tensor([-1,2,3]).send(alice)
p = plan_b(x_ptr)
print(p.get())


```


iamtrask:
I *think* I fixed tests (they work on my machine) - we'll see if Travis likes it.

LaRiffle:
Nice changes! Now there is still something to think about: the role of the PlanWorker is here to 1. batch the commands and 2. to execute them all at once. Maybe what we really what to do is to have a PlanWorker working as an auxiliary of the local worker (so on the same machine) (doing 1.), and when the plan is ready for execution, to send it to a real remote worker (which will do 2.). Thoughts?

syft/workers/base.py
```diff
@@ -224,12 +230,12 @@ def send(
     ) -> PointerTensor:
         """Sends tensor to the worker(s).
 
-        Send a syft or torch tensor and his child, sub-child, etc (all the
+        Send a syft or torch tensor/object and it's child, sub-child, etc (all the
```
robert-wagner: Good use of gender neutral language

syft/workers/plan.py
```diff
@@ -0,0 +1,261 @@
+from .base import BaseWorker
```
robert-wagner: Absolute import

syft/workers/plan.py
```diff
@@ -0,0 +1,261 @@
...
+                )
+            )
+
+        else:
+            _obj.append(item)
+
+    return _obj
+
+
+class Plan(BaseWorker):
```
robert-wagner: Is there a good reason that plans are a subclass of workers? What functionality do they inherit?

syft/workers/plan.py
```diff
@@ -0,0 +1,261 @@
...
+
+class PlanPointer:
+    def __init__(self, id, location, id_at_location, owner):
+        self.id = id
+        self.location = location
+        self.id_at_location = id_at_location
+        self.owner = owner
+
+
+def replace_ids(obj, change_id, to_id, from_worker, to_worker):
```
robert-wagner: Should this just be a private class method? Is it ever called from outside this file

syft/workers/plan.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+        self.readable_plan = replace_ids(
+            obj=self.readable_plan,
+            change_id=-1,
+            to_id=-1,
+            from_worker=from_worker_id,
+            to_worker=to_worker_id,
+        )
+
+    def __call__(self, *args, **kwargs):
```
robert-wagner: This is a good interface. I like that these are callables as it is very intuitive

syft/workers/plan.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+        Mock send function that only specify that the Plan will have to be sent to location
+        """
+        if self.location is not None:
+            raise NotImplementedError(
+                "Can't send a Plan which already has a location, use .get() before"
+            )
+        else:
+            self.location = location
+        return self
```
robert-wagner: Should send return  a plan pointer?

syft/workers/plan.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def get(self):
+        """
+        Mock get function
```
robert-wagner: Why is this a mock function?

syft/workers/plan.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+        self.replace_worker_ids(self.location.id, self.owner.id)
+        self.location = None
+        self.ptr_plan = None
+        return self
+
+    def _send(self, location):
```
robert-wagner: Why is the real send private but the mock send public?

syft/workers/utils.py
```diff
@@ -0,0 +1,41 @@
...
+import random
+from .plan import Plan
+
+
+def make_plan(plan_blueprint):
+    """For folks who would prefer to not use a decorator, they can use this function"""
+    return func2plan()(plan_blueprint)
+
+
+def func2plan():
```
robert-wagner: You can write code that functions as either a decorator or just a normal function. We should do that here rather than having two functions

test/conftest.py
```diff
@@ -5,26 +5,24 @@
 from syft import TorchHook
 
 
-@pytest.fixture(scope="function", autouse=True)
+@pytest.fixture(scope="session", autouse=True)
```
robert-wagner: I agree with the session tag. I had some tests that were being weird without it. We should keep this in mind

test/conftest.py
```diff
@@ -5,26 +5,24 @@
...
 def workers(hook):
+    # Define 3 virtual workers
     alice = syft.VirtualWorker(id="alice", hook=hook, is_client_worker=False)
     bob = syft.VirtualWorker(id="bob", hook=hook, is_client_worker=False)
     james = syft.VirtualWorker(id="james", hook=hook, is_client_worker=False)
 
-    # TODO: should one set this boolean to true?
-    # It was done previously in self.setUp() from `test_hook.py`
-    # hook.local_worker.is_client_worker = True
+    # bob.add_workers([alice, james])
```
robert-wagner: We should delete these lines

test/workers/test_plan.py
```diff
@@ -0,0 +1,75 @@
...
+    plan_abs.get()
+    plan_abs.send(alice)
+
+    x_ptr = th.tensor([-1, 2, 3]).send(alice)
+    p = plan_abs(x_ptr)
+    x_abs = p.get()
+    assert (x_abs == th.tensor([1, 2, 3])).all()
+
+
+# def test_plan_remote_function(workers):
```
robert-wagner: Why are these commented out

syft/serde.py
```diff
@@ -74,6 +77,9 @@ def serialize(obj: object, compress=True, compress_scheme=LZ4) -> bin:
         compress_scheme (int): the integer code specifying which compression
             scheme to use (see above this method for scheme codes) if
             compress == True.
+        simplified (bool): in some cases we want to pass in data which has
+            already been simplified - in which case we must skip double
+            simplification - which would be bad.... so bad... so... so bad
```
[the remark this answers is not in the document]
LaRiffle: Ahahah Andrew's poetry!

syft/workers/abstract.py
```diff
@@ -1,5 +1,27 @@
...
+    Generate id and store the list of ids generated
+    Can take a pre set list in input and will complete
+    when it's empty.
+    """
+
+    def __init__(self, given_ids=list()):
+        self.given_ids = given_ids
+        self.generated = []
+
+    def pop(self, *args):
```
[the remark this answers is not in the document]
LaRiffle: you're right, I changed the name

syft/workers/base.py
```diff
@@ -67,21 +69,26 @@ def __init__(self, hook, id=0, data={}, is_client_worker=False, log_msgs=False,
...
+        # Declare workers as appropriate
         self._known_workers = {}
         if hook.local_worker is not None:
-            for k, v in hook.local_worker._known_workers.items():
-                if v is not hook.local_worker:
-                    self._known_workers[k] = v
-                    v.add_worker(self)
-            hook.local_worker.add_worker(self)
-        self.add_worker(self)
+            if self.id not in self.hook.local_worker._known_workers:
```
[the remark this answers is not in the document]
LaRiffle: We don't want to add several times know workers -> generate warning and the debug of pytest gets really dirty with all the warning messages

syft/workers/plan.py
LaRiffle: They catch the call to function to build plans which is crucial to the plan functionality

syft/workers/plan.py
LaRiffle: Right!

syft/workers/plan.py
LaRiffle: Because the send is virtual, it is not a necessity but I agree this leads to confusion

syft/workers/plan.py
LaRiffle: The interest is that you can call .send() but actually this doesn't trigger a call to a remote worker, just "a promise" that it will be sent (with _send() ) as soon the plan in called (and built)

test/workers/test_plan.py
LaRiffle: They are no longer valid behaviour we want to test

LaRiffle: I've removed them

test/conftest.py
LaRiffle: done

PR #1942, opened 2019-02-23
Add draft version of AutogradTensor

mcleonard:
Adding AutogradTensor for performing backpropagation. Only function that works right now is addition. An example is shown in the notebook.


mcleonard:
In the last commit I wrote some functionality to generate grad_fn classes from a file called derivatives.yaml using build_gradients.py. It's similar to this file from Pytorch: https://github.com/pytorch/pytorch/blob/master/tools/autograd/derivatives.yaml. Except that I'm generating a Python file. 

This structure should make it easier to construct gradient functions for Tensor methods. I think the next step is to generate methods that can be attached to Autograd.

mcleonard:
Making progress. This time I set it up so that AutogradTensor automatically checks if a method has a gradient function defined in gradients.py. As we add gradients to derivatives.yaml, they'll automatically be available in the autograd system.

Something I need help with is having `.grad` work for Wrappers...

iamtrask:
Thank you for leaving good comments here - makes it really easy to stay updated :)

mcleonard:
Okay! Finally added tests for ~6 gradient functions. 

@iamtrask What's left to do to merge all this?

iamtrask:
See https://github.com/OpenMined/PySyft/pull/2095 for continuation of this PR

Issue #1456, opened 2018-08-28
Type annotations for pysyft


robert-wagner:
# Issue Template

### User Story:
As a Developer of PySyft <br>
I would like to know what the type of various parameters are when looking at the code. Since python now has a typing module, we should add type annotations to track the type of various parameters to ease development

## Expected Behavior

Every method has correct type annotations and running mypy on the repository passes

## Current Behavior

no methods have type annotations

abdulrahmanmazhar:
Hello, I guess I can start working on this if it's not assigned to anyone already 😊

iamtrask:
This is also a great first project - although I suspect it will take the collective action of several! Feel free to submit Pull Requests as you make progress (doesn't have to solve the whole issue in one go)

abdulrahmanmazhar:
Sure, That would be fun to be working with 😊

iamtrask:
Hey @abdulrahmanmazhar - how it's going?

abdulrahmanmazhar:
@iamtrask Hello, Indeed I'm sorry to say that I didn't get much of a progress since I was having some assignments for 2 interviews, But I was considering using "pyannotate" and some manual typing so hopefully I will get some progress and will push it up as a PR very soon, Really sorry about the delay :smile:

iamtrask:
Feel free to do incremental Pull Requests :) - like... even if you only add type in one or two places in one file... just go ahead and PR it.

abdulrahmanmazhar:
Sure and thank you a lot :)

iamtrask:
Hey @abdulrahmanmazhar - are you on OpenMined's slack? If so, what's your username? (i'm @trask). If now - would you be willing to join so we can have a quick chat? (slack.openmined.org)

abdulrahmanmazhar:
Hey @iamtrask indeed no I'm not and sure that would be my pleasure :)

replomancer:
Mypy requires stub files for msgpack, numpy, pandas and setuptools. I wanted to add those, but it's not working for me on the first try. I will continue only if there's no one else working on this already (@abdulrahmanmazhar ?). We need to be careful about the stub licenses.

abdulrahmanmazhar:
@replomancer actually I'm not working on it at the moment so you are more than welcome to start working on it and it's my pleasure to help you as well if you needed something i can do :smile:

replomancer:
OK, I'll take care of the stubs.

replomancer:
[...]

PR #1962, opened 2019-03-03
Type annotation

alaaboudib:
This is my first contribution attempt ever to OpenMined's PySyft. So, I made it so concise with the issue #1456 

I annotated the 'message' type in _recv_msg and _send_msg (in virtual.py) as binary although they are marked as strings in the doc strings in base.py. I thought this might be a typo  since `send_msg` and `recv_msg` use binary message types. So please let me know if I am missing something here.

### DIFF
--- a/syft/workers/base.py
+++ b/syft/workers/base.py
@@ -206,14 +206,7 @@ class BaseWorker(AbstractWorker):
         # Step 1: route message to appropriate function
         response = self._message_router[msg_type](contents)
 
-        # # Step 2: If response in none, set default
-        # TODO: not sure if someone needed this - if this comment
-        # is still here after Feb 15, 2018, please delete these
-        # two lines of (commented out) code.
-        # if response is None:
-        #     response = None
-
-        # Step 3: Serialize the message to simple python objects
+        # Step 2: Serialize the message to simple python objects
         bin_response = serde.serialize(response)
         return bin_response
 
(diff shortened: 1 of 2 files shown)


syft/workers/base.py
```diff
@@ -206,14 +206,7 @@ def recv_msg(self, bin_message):
...
         response = self._message_router[msg_type](contents)
 
-        # # Step 2: If response in none, set default
-        # TODO: not sure if someone needed this - if this comment
-        # is still here after Feb 15, 2018, please delete these
-        # two lines of (commented out) code.
-        # if response is None:
-        #     response = None
-
-        # Step 3: Serialize the message to simple python objects
```
iamtrask: would be cool to still keep this "Step 3:" piece

alaaboudib: Thank you @iamtrask 
I marked it as # Step 2 since the previous # Step 2 was deleted. Should it still appear as # Step 3 then?

iamtrask:
Yeah I think you're right @kamathhrishi

syft/workers/base.py
alaaboudib: So I kept it but as Step 2 not 3, is it ok?

alaaboudib:
Thank you @iamtrask for the merge. @kamathhrishi , i will be working on that.

Cheers!

PR #1964, opened 2019-03-03
Fix error in tutorial 2

tyrinwu:
I found two places to change while running the tutorial today. Please correct me if I am wrong.

+ It seems that `syft` does not have `optim`?

+ `opt.step` would not take `data.shape[0]` as input because `opt.step` should take some closure.
Doc: https://pytorch.org/docs/stable/optim.html#torch.optim.Optimizer.step


tyrinwu:
@LaRiffle np! btw, do we need this on the `dev` branch? I wasn't too familiar with the branching strategy in this repository.

PR #1965, opened 2019-03-03
Fix moving a module to cuda

bhushan23:
new_data and native_param_data are not different device.
moving new_data to cpu before passing to native_param_data's set method

Fixes: #1893


iamtrask:
So while this will technically get rid of the error in #1893 I'm not sure it's the fix we want long term. We really do want to support GPUs as opposed to always pushing everything to the CPU. :)

Issue #1958, opened 2019-03-01
Automatically add workers to known workers


robert-wagner:
Since each worker has access to the hook, we are able to get the workers that the local workers has knowledge of at creation. This allows us to automatically add the new worker on spinup to all workers known workers

alaaboudib:
I am interested in this issue. However, I am new to PySyft, I just followed the 8 tutorials. Can you please explain this issue a bit more? maybe give a concrete example of what you are willing to have as a result?

robert-wagner:
Hey @alaaboudib I didn't see that you were interested in doing this issue before completing it myself this morning. Sorry about that :slightly_smiling_face:

alaaboudib:
No problem @robert-wagner , can you refer me to your PR to learn more about how you did it?

robert-wagner:
Hey @alaaboudib I tagged you in the pr (it is linked above this message)

alaaboudib:
Thank you @robert-wagner

PR #1969, opened 2019-03-04
automatically register workers with each other on creation

robert-wagner:


robert-wagner:
@alaaboudib This is the pr for #1958

PR #1970, opened 2019-03-04
initial draft for websocket worker

robert-wagner:
Hi @iamtrask @jlebensold @LaRiffle. This is a first draft for the websocket worker pr. The user would spin one of these up on each machine they are interested in. I need to figure out a way to get the response back from the remote server when calling receive message. Currently it is being sent but is hard to detect. Any thoughts on this would be appreciated


jlebensold:
@robert-wagner what about having a local queue and a remote queue and having two different async listeners?

PR #1977, opened 2019-03-06
Type annotation

alaaboudib:
Here are some more type annotations. I was not very sure how to annotation Syft Tensors that are not PointerTensors. I ended up using `AbstractTensor`. 

For torch tensors I used `torch.Tensor`.

For annotating self type classes I used strings "BaseWorker" when annotating this type from inside the BaseWorker class. It seems to be the solution for python<=3.6. For python 3.7 and above, there is another solution by importing

`from __future__ import annotations` then self type annotation is supposed to work out of the box

Thanks for the review

### DIFF
--- a/syft/workers/base.py
+++ b/syft/workers/base.py
@@ -11,6 +11,7 @@ from syft.exceptions import WorkerNotFoundException
 from syft.workers import AbstractWorker
 from syft.codes import MSGTYPE
 from typing import Union, List
+import torch
 
 
 class BaseWorker(AbstractWorker):
--- a/syft/workers/base.py
+++ b/syft/workers/base.py
@@ -9,6 +9,7 @@ from syft.frameworks.torch.tensors.interpreters import PointerTensor
 from syft.exceptions import WorkerNotFoundException
 from syft.workers import AbstractWorker
 from syft.codes import MSGTYPE
+from typing import Union, List
 
 
 class BaseWorker(AbstractWorker):
@@ -88,7 +89,7 @@ class BaseWorker(AbstractWorker):
 
     # SECTION: Methods which MUST be overridden by subclasses
     @abstractmethod
-    def _send_msg(self, message, location):
+    def _send_msg(self, message: bin, location: BaseWorker) -> bin:
         """Sends message from one worker to another.
 
         As BaseWorker implies, you should never instantiate this class by
@@ -98,8 +99,8 @@ class BaseWorker(AbstractWorker):
         example to study is VirtualWorker.
 
         Args:
-            message: A string representing the message being sent from one
-                worker to another.
+            message: A binary message to be sent from one worker
+                to another.
             location: A BaseWorker instance that lets you provide the
                 destination to send the message.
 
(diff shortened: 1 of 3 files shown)


alaaboudib:
@LaRiffle @iamtrask 
Do you guys receive automatic notifications for PRs or should I tag reviewers explicitly? 

It seems that I cannot choose reviewers for this PR

robert-wagner:
Hey @alaaboudib , most of the time notifications are sent through either the code channel or through email for new prs but they can sometimes get missed. If you want a specific person to review your pr, it is usually best to either tag them in the pr or ping them on slack. It is possible to assign reviewers but I am not sure the permissions needed for that

syft/workers/base.py
```diff
@@ -105,7 +108,7 @@ def _send_msg(self, message, location):
         raise NotImplementedError  # pragma: no cover
 
     @abstractmethod
-    def _recv_msg(self, message):
+    def _recv_msg(self, message: bin) -> bin:
```
robert-wagner: This method does not return binary data (it raises an exception). For specific implentions (ie virtual worker) it returns either a pointer tensor or None

syft/workers/base.py
```diff
@@ -6,9 +6,12 @@
 import syft as sy
 from syft import serde
 from syft.frameworks.torch.tensors.interpreters import PointerTensor
+from syft.frameworks.torch.tensors.interpreters import AbstractTensor
 from syft.exceptions import WorkerNotFoundException
 from syft.workers import AbstractWorker
 from syft.codes import MSGTYPE
+from typing import Union, List
```
robert-wagner: Could you split these imports on to separate lines

syft/workers/base.py
```diff
@@ -175,7 +180,7 @@ def send_msg(self, msg_type, message, location):
 
         return response
 
-    def recv_msg(self, bin_message):
+    def recv_msg(self, bin_message: bin) -> bin:
```
robert-wagner: Like _recv_msg this does not return binary data

syft/workers/base.py
```diff
@@ -336,15 +346,16 @@ def send_command(self, recipient, message):
 
         return response
 
-    def set_obj(self, obj):
+    def set_obj(self, obj: Union[torch.Tensor, AbstractTensor]) -> None:
```
robert-wagner: This is clever since all of the syft tensor types subclass abstract tensor. Good job

syft/workers/base.py
alaaboudib: @robert-wagner The doc string for this method `recv_msg` (notice no underscore at the beginning of the method name) actually says:

`Returns:
A binary message response
`

Is that an error?

alaaboudib: @robert-wagner The doc string for this method `recv_msg` (notice no underscore at the beginning of the method name) actually says:

`Returns:
A binary message response
`

Is that an error?

alaaboudib: @robert-wagner The doc string for this method `recv_msg`  actually says:

`Returns:
A binary message response
`

Is that an error?

robert-wagner: Nevermind you are correct

syft/workers/base.py
robert-wagner: Ignore this comment

alaaboudib:
@robert-wagner I have taken your comments into account, please check my new update to this PR

iamtrask:
Well done @alaaboudib!!!

PR #1983, opened 2019-03-09
Handling remote commands with complex response; fix #1975

### LINKED ISSUE #1975: Get .max() working on PointerTensor; multiple ptrs in response
We can't return tuples of tensors when calling remote executions. This means that functions like .max() fail on PointerTensor objects.

LaRiffle:
Add a method to handle registration and sending back resp to client when a command is sent.
Support response with multi tensors; like `torch.split(ptr, 2)` which returns a tuple of pointers

fix #1975

### DIFF
--- a/syft/frameworks/torch/hook_args.py
+++ b/syft/frameworks/torch/hook_args.py
@@ -425,80 +425,89 @@ def build_response_hook(response, rules, wrap_type, wrap_args, return_tuple=Fals
     return lambda x: f(lambdas, x)
 
 
-def zero_fold(*a):
+def zero_fold(*a, **k):
     return tuple()
 
 
-def one_fold(return_tuple):
-    def _one_fold(lambdas, args):
-        return lambdas[0](args[0])
+def one_fold(return_tuple, **kwargs):
+    def _one_fold(lambdas, args, **kwargs):
+        return lambdas[0](args[0], **kwargs)
 
     def tuple_one_fold(lambdas, args):
-        return (lambdas[0](args[0]),)
+        return (lambdas[0](args[0], **kwargs),)
 
     return {False: _one_fold, True: tuple_one_fold}[return_tuple]
 
 
-def two_fold(lambdas, args):
-    return lambdas[0](args[0]), lambdas[1](args[1])
+def two_fold(lambdas, args, **kwargs):
+    return lambdas[0](args[0], **kwargs), lambdas[1](args[1], **kwargs)
 
 
-def three_fold(lambdas, args):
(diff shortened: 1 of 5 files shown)


LaRiffle:
> Can we add static type information to method parameters?

What are you thinking about? (which methods?)

iamtrask:
Sorry - i meant "type annotations" which we want for all methods.

syft/frameworks/torch/hook_args.py
```diff
@@ -531,3 +543,146 @@ def number_identity(i):
...
+        register=True,
+        owner=owner,
+        ptr_id=tensor.id,
+        garbage_collect_data=False,
+    )
+    return pointer
+
+
+def build_register_response(
+    response: object, rules: object, return_tuple: bool = False
```
robert-wagner: Is object really the best we can do for these?

syft/frameworks/torch/hook_args.py
```diff
@@ -531,3 +543,146 @@ def number_identity(i):
...
+    response_hook_function = build_register_response(response, rule)
+    return response_hook_function
+
+
+def register_transform_tensor(
+    tensor: object, owner: sy.workers.AbstractWorker = None
+) -> PointerTensor:
+    """
+    Register a tensor and create a pointer that references it
+    :param tensor: the tensor
```
robert-wagner: Switch to other doc style for consistency

syft/frameworks/torch/hook_args.py
```diff
@@ -531,3 +543,146 @@ def number_identity(i):
...
+    # (torch or syft) tensors and 0 when not (ex: number, str, ...)
+    rule = build_rule(response)
+    # Build a function with this rule to efficiently replace syft tensors
+    # (but not pointer) with their child in the args objects
+    response_hook_function = build_register_response(response, rule)
+    return response_hook_function
+
+
+def register_transform_tensor(
+    tensor: object, owner: sy.workers.AbstractWorker = None
```
robert-wagner: I think tensor in this case should be a union of abstract tensor and torch tensor

syft/frameworks/torch/hook_args.py
LaRiffle: Can be tuple, tensor, tuple of int, str, tensor very various

syft/frameworks/torch/hook_args.py
LaRiffle: right

syft/frameworks/torch/hook_args.py
robert-wagner: Sounds good

PR #1993, opened 2019-03-14
A few more type annotations

AlanAboudib:
I just added some few type annotations as I studied the code


syft/frameworks/torch/tensors/interpreters/abstract.py
```diff
@@ -12,7 +12,7 @@ class AbstractTensor(ABC):
 
     is_wrapper = False
 
-    def __init__(self, tags=None, description=None):
+    def __init__(self, tags=None, description: str = None):
```
LaRiffle: You can probably also add a type info for tags like :
```
from typing import List

def my_func(l: List[str]):
    pass
```

syft/frameworks/torch/tensors/interpreters/abstract.py
```diff
@@ -47,7 +47,7 @@ def __len__(self):
         except IndexError:
             return 0
 
-    def on(self, tensor, wrap=True):
+    def on(self, tensor, wrap: bool = True):
```
LaRiffle: same you can for tensor put  ~ `tensor: sy...AbstractTEnsor`

syft/frameworks/torch/tensors/interpreters/multi_pointer.py
```diff
@@ -17,7 +18,7 @@ def __init__(
         shape=None,
         point_to_attr=None,
         tags=None,
-        description=None,
+        description: str = None,
         children=None,
```
LaRiffle: maybe you want to add annotation for all arguments (whe it's not too painful) or for the whole file, this way people continuing the type annotation know they don't have to check everything
https://docs.python.org/3/library/typing.html can be a good resource

syft/frameworks/torch/tensors/interpreters/pointer.py
```diff
@@ -47,7 +47,7 @@ def __init__(
         shape=None,
         point_to_attr=None,
```
LaRiffle: same as above

alaaboudib:
Thank you @LaRiffle , I will work on the changes. One question, If I insert wrong types, will they be detected by the automated check?

LaRiffle:
I don't think so :/

AlanAboudib:
@LaRiffle Here is a PR for type annotations I had started earlier. I completed the functions I had started with. However, I didn't annotation entire files. I need some more code studying to understand all the variable types.

PR #1994, opened 2019-03-14
[Issue 1456] add type annotations and address pylint errors in some files

antonrd:
This is work related to issue #1456  and also adds a few small changes that I uncovered while working on this:
    1) Adds type annotations to a few methods. Not 100% sure that the selected types are the most accurage ones, please advice if there is an issue.
    2) Data argument that gets passed to workers has as default value of {} but seems like in reality it's a list. Also it's not a good practice to use mutalbe object as a default value to a methof so changing it to None and handling this in the actual methot that loads the data.
    3) Addresses some small pylint errors.
    4) Reorders the imports to follow the PEP 8 guidelines.


syft/workers/base.py
```diff
@@ -141,10 +141,10 @@ def load_data(self, data: List[Union[torch.Tensor, AbstractTensor]]) -> None:
 
         """
 
-        for tensor in data:
-
-            self.register_obj(tensor)
-            tensor.owner = self
+        if data:
```
antonrd: Decided to handle a `None` value here and avoid using a {} or [] as default argument value, which is not a good practice overall.

syft/workers/websocket_client.py
```diff
@@ -1,11 +1,25 @@
...
-        self, hook, host, port, id=0, is_client_worker=False, log_msgs=False, verbose=False, data={}
+            self,
+            hook: sy.TorchHook,
+            host: str,
+            port: int,
+            id: Union[int, str] = 0,
+            is_client_worker: bool = False,
+            log_msgs: bool = False,
+            verbose: bool = False,
+            data: List[Union[torch.Tensor, AbstractTensor]] = None
```
antonrd: Assumed this type looking at other places in the code where this argument is passed, let me know if I'm missing something.

syft/workers/websocket_server.py
```diff
@@ -66,7 +71,7 @@ def __init__(
             msg = await websocket.recv()
             await self.broadcast_queue.put(msg)
 
-    async def _producer_handler(self, websocket):
+    async def _producer_handler(self, websocket: websockets.WebSocketCommonProtocol):
```
antonrd: I hope this is the right type to specify here, although looks a bit too specific. Please suggest if there is a better way to handle that.

robert-wagner:
@antonrd As a heads up, Travis is failing because you need to run `black syft`

antonrd:
Yeah, sorry about that, will try to resolve. Currently fighting some weird issue with flake8 when run locally. Once it's resolved I will address the reformatting issues.

syft/workers/websocket_server.py
robert-wagner: I think this is correct

PR #1998, opened 2019-03-16
Serialization strategies

mccorby:
This PR enables different strategies to serialise/deserialise tensors.
Right now the serialisation of the tensors is delegated to Torch but the serialised objects are not compatible with other platforms

This PR includes a strategy to seder the tensors using npy format

### DIFF
--- a/syft/serde.py
+++ b/syft/serde.py
@@ -27,20 +27,20 @@ tuple type. The same is true for all other simplifier/detailer functions.
 
 By default, we serialize using msgpack and compress using lz4.
 """
-import io
-import zstd
 from tempfile import TemporaryFile
 from typing import Collection
 from typing import Dict
 from typing import Tuple
-
-import lz4
-import msgpack
-import numpy
 import torch
+import msgpack
+import lz4
 from lz4 import (  # noqa: F401
     frame,
 )  # needed as otherwise we will get: module 'lz4' has no attribute 'frame'
+import io
+import numpy
+import warnings
+import zstd
 
 import syft
 import syft as sy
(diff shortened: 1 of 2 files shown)


syft/serde.py
```diff
@@ -157,6 +155,74 @@ def deserialize(
     return _detail(worker, simple_objects)
 
 
+def serialize_tensor(tensor) -> bin:
```
robert-wagner: This is clever

syft/serde.py
```diff
@@ -27,35 +27,33 @@
 
 By default, we serialize using msgpack and compress using lz4.
 """
-
+import io
+import zstd
```
robert-wagner: Could you move this import down to below torch

syft/serde.py
```diff
@@ -157,6 +155,74 @@ def deserialize(
...
+    """
+    return torch_tensor_deserializer(tensor_bin)
+
+
+def numpy_tensor_serializer(tensor: torch.Tensor) -> bin:
+    """Strategy to serialize a tensor using numpy npy format.
+    If tensor requires to calculate gradients, it will detached.
+    """
+    if tensor.requires_grad:
+        print(
```
robert-wagner: Rather than printing this, I think we should raise it as a warning ie warnings.warn

examples/tutorials/Part 2 - Intro to Federated Learning.ipynb
```diff
@@ -235,7 +234,7 @@
     "            loss.backward()\n",
     "\n",
     "            # 5) change those weights\n",
-    "            opt.step(data.shape[0])\n",
+    "            opt.step()\n",
```
robert-wagner: Why was this changed?

syft/serde.py
robert-wagner: I think we might want to make this and deserialize tensor private functions as the user should not be calling them

examples/tutorials/Part 2 - Intro to Federated Learning.ipynb
mccorby: Leftover from the fork. Removed

mccorby:
@robert-wagner All checks passed. Addressed your comments

PR #2002, opened 2019-03-20
Add an (experimental) tutorial for how to use Plans

LaRiffle:


syft/__init__.py
```diff
@@ -18,6 +18,7 @@
 from syft.workers import VirtualWorker
 from syft.workers import Plan
 from syft.workers import func2plan
+from syft.workers import meth2plan
```
robert-wagner: Can we change the name to not be meth2plan, more something like method2plan (certain connotations in english)

iamtrask:
This tutorial is some of the best light vacation reading I've ever had. Really really excellent.

syft/__init__.py
LaRiffle: Hahaha yes you're right!

LaRiffle:
> This tutorial is some of the best light vacation reading I've ever had. Really really excellent.

Thank you! :)

PR #2005, opened 2019-03-22
Add support for specific overloading of torch function at the tensor level

LaRiffle:


syft/frameworks/torch/tensors/interpreters/abstract.py
```diff
@@ -1,8 +1,11 @@
 from abc import ABC
-import torch
 import random
-import syft as sy
 import weakref
+import functools
+
+import torch
```
robert-wagner: super nitpicky but can we sort these alphabetically?

syft/frameworks/torch/tensors/interpreters/abstract.py
```diff
@@ -12,18 +15,28 @@ class AbstractTensor(ABC):
 
     is_wrapper = False
 
-    def __init__(self, tags=None, description=None):
+    def __init__(self, id=None, owner=None, tags=[], description=None, child=None, parent=None):
```
robert-wagner: Is parent a thing now? if so what does it do and how does this affect garbage collection, seems like it would create a reference loop

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -307,7 +307,7 @@ def __truediv__(self, shares: dict, divisor):
             field=self.field, crypto_provider=self.crypto_provider
         ).set_shares(divided_shares)
 
-    @hook
+    @overloaded.method
```
robert-wagner: I like this .method call alot

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -110,7 +110,7 @@ def __repr__(self) -> str:
 
             big_repr = False
 
-            if self.tags is not None:
+            if self.tags is not None and len(self.tags) > 0:
```
robert-wagner: Theoretically this could just be and len(self.tags) since if it is len 0 then it evaluates to false

syft/frameworks/torch/tensors/interpreters/pointer.py
```diff
@@ -144,7 +142,7 @@ def __str__(self):
 
         big_str = False
 
-        if self.tags is not None:
+        if self.tags is not None and len(self.tags) > 0:
```
robert-wagner: Same as above. Your call on this, I just think it looks cleaner but I'm weird

syft/frameworks/torch/tensors/interpreters/utils.py
```diff
@@ -1,9 +1,16 @@
 import syft
 
 
-def hook(attr):
+class Object(object):
```
robert-wagner: What does this class do, Also its name should be changed

syft/frameworks/torch/tensors/interpreters/utils.py
```diff
@@ -1,9 +1,16 @@
...
+    pass
+
+
+overloaded = Object()
```
robert-wagner: Could we just define overloaded as a normal class?

syft/frameworks/torch/tensors/interpreters/utils.py
```diff
@@ -23,3 +30,55 @@ def hook_args(self, *args, **kwargs):
         return response
 
     return hook_args
+
+
+overloaded.method = overload_method
+
+
+def overload_function(attr):
```
robert-wagner: I think this file should probably move up a few directories. It doesnt really belong in interpreters and should probably just be in the same directory as the hook

LaRiffle: Yes I agree! Also it should also be in a normal class i got confused using the Object()

syft/frameworks/torch/tensors/interpreters/abstract.py
LaRiffle: It already a thing, it exists but we don't really make use of it
We could discuss removing it

robert-wagner: Sounds good. I will add an issue for it

syft/frameworks/torch/overload_torch.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @staticmethod
+    def overload_function(attr):
+        """
+        hook args and response for functions that hold the @overloaded.function decorator
+        """
+
+        def hook_args(*args, **kwargs):
+
+            # TODO have a better way
```
robert-wagner: Could you clarify this comment

syft/frameworks/torch/tensors/interpreters/abstract.py
```diff
@@ -122,6 +133,65 @@ def get_class_attributes(self):
...
+        # Send it to the appropriate class and get the response
+        response = new_type.handle_func_command(new_command)
+
+        # Put back LoggingTensor on the tensors found in the response
+        response = sy.frameworks.torch.hook_args.hook_response(cmd, response, wrap_type=cls)
+
+        return response
+
+
+def rgetattr(obj, attr, *args):
```
robert-wagner: what does this do?

robert-wagner: comments would be helpful here

robert-wagner: Normally we have been doing the google style docstrings. Please change for consistency

PR #2008, opened 2019-03-25
Replace hard-coded compression by strategy

mccorby:
This PR will help to open the compression/decompression process to workers that do not have access to the configuration used during seder

The way of choosing a compression with this PR is by overriding the function `_apply_compress_scheme`
The function will default to `LZ4` compression.
If no compression is required then use `_apply_compress_scheme = apply_no_compression`
Other compression schemes, as SZTD, can be applied similarly: 
`_apply_compress_scheme = apply_zstd_compression`

## Changes:
* Replace flag determining compression by flag indicating possible compression schemes (including NO_COMPRESSION)
* Added hooks that can be overridden to set the compression scheme (or none)

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/abstract.py
+++ b/syft/frameworks/torch/tensors/interpreters/abstract.py
@@ -90,21 +90,13 @@ class AbstractTensor(ABC):
         # wrapper.child.parent = weakref.ref(wrapper)
         return wrapper
 
-    def serialize(
-        self, compress=True, compress_scheme=0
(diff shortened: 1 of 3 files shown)


mccorby:
> I'm curious - do you think it would make sense to pass the compression code with the message as well? This would allow the receiving party to know how the rest of the message was compressed.
If we end up using some kind of configuration that should make it as I would expect the entire system to agree on the compression used. 
If we like more flexibility then a possibility would be adding such configuration on every message but in my opinion, that would increase the size of the stream being sent without adding much value in most cases.

PR #2019, opened 2019-03-26
Improve sandbox tests.

mari-linhares:


robert-wagner:
Please run black syft
you can set this up to automatically run by running make install_hooks

mari-linhares:
Thanks, Bobby! I ran `pre-commit install` instead of `make install_hooks` and it seems to do the job.

PR #2020, opened 2019-03-27
Changes WebsocketClientWorker#search to communicate with the server over websockets

antonrd:
This PR is a result of a discussion with @robert-wagner where I was wondering what are the ways for two non-virtual players to interact and exchange tensors. I'm not sure if this solution is good enough. It's more like a proposal for how to address the issue that currently the websocket client does not actually talk to the server over websockets when a search for tensors is issued.

The PR also addresses one issue with the serialization module (msgpack) - that regular strings when serialized and deserialized become from the type `bytes` and need to be cast back to string for comparisons to work.

Also, tried to add a test that demonstrates how this search functionality would work now.


syft/workers/websocket_client.py
```diff
@@ -36,6 +38,14 @@ def __init__(
 
         super().__init__(hook, id, data, is_client_worker, log_msgs, verbose)
 
+    def search(self, *query):
```
iamtrask: nice contrib!

syft/workers/base.py
```diff
@@ -714,6 +714,11 @@ def search(self, *query):
         for key, tensor in self._objects.items():
             found_something = True
             for query_item in query:
+                # If deserialization produced a bytes object instead of a string,
```
iamtrask: good catch here - i wonder if we could handle this more generically in serde.py - seems like we should

syft/workers/websocket_client.py
```diff
@@ -36,6 +38,14 @@ def __init__(
...
+        # Prepare a message requesting the websocket server to search among its objects
+        message = (MSGTYPE.SEARCH, query[0])
```
antonrd: I just realized that this should probably send the whole `query`, not just the first element. Which would mean that on the receiving end, after deserializing, this would result in tuple of elements and would need to be passed to `search` with `*` in front. This would require some more changes in base.py, though. Hm...

syft/workers/base.py
antonrd: Actually, yeah, this makes a lot of sense. I need to study a bit the serde module to see how to change it to handle these cases.

syft/workers/websocket_client.py
iamtrask: Agreed

iamtrask: (changes to base.py for this purpose are ok)

antonrd: this is now addressed

syft/workers/base.py
antonrd: Seems like msgpack should support that through a couple of args to the loads/dumps methods as described here: https://github.com/msgpack/msgpack-python#string-and-binary-type
but when I set these args a bunch of tests fails so I'll have to look into them. Also, I notice now that in this file there is already one other instance where we check for a `bytes` object. Looks like someone has already faced this issue.

PR #2021, opened 2019-03-28
Preparation for SecureNN

LaRiffle:
SecureNN uses alot MultiPointer and AdditiveShared Tensor,
I'm doing here many changes to theses tensors to fix them -- so that the secureNN can work properly. It will be done in a second PR.
This is really to address errors we have in MultiPointer and AdditiveShared Tensor, and joint interactions between them.

### DIFF
--- a/syft/frameworks/torch/crypto/spdz.py
+++ b/syft/frameworks/torch/crypto/spdz.py
@@ -14,7 +14,7 @@ def spdz_mul(
             self. Equivalent to calling self.child.
         other_shares: a dictionary <location_id -> PointerTensor) of shares corresponding
             to the tensor being multiplied by self.
-        cypto_provider: an AbstractWorker which is used to generate triples
+        crypto_provider: an AbstractWorker which is used to generate triples
         field: an interger denoting the size of the field
         """
     locations = list(shares.keys())
@@ -22,6 +22,7 @@ def spdz_mul(
     other_shape = other_shares[locations[0]].shape
     triple = crypto_provider.generate_triple(cmd, field, shares_shape, other_shape, locations)
     a, b, c = triple
+    a, b, c = a.child, b.child, c.child
     d = {}
     e = {}
     for location in locations:
--- a/syft/__init__.py
+++ b/syft/__init__.py
@@ -23,6 +23,8 @@ from syft.workers import make_plan
 
 # Import Tensor Types
 from syft.frameworks.torch.tensors.decorators import LoggingTensor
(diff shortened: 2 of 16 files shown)


syft/frameworks/torch/crypto/spdz.py
```diff
@@ -14,14 +14,15 @@ def spdz_mul(
...
-        cypto_provider: an AbstractWorker which is used to generate triples
+        crypto_provider: an AbstractWorker which is used to generate triples
         field: an interger denoting the size of the field
         """
     locations = list(shares.keys())
     shares_shape = shares[locations[0]].shape
     other_shape = other_shares[locations[0]].shape
     triple = crypto_provider.generate_triple(cmd, field, shares_shape, other_shape, locations)
     a, b, c = triple
+    a, b, c = a.child, b.child, c.child
```
robert-wagner: Why are the children added back?

LaRiffle: The way I'm reasoning is that all objects a, b or c should be AdditiveSharedTensor or MultiPointerTensoir, (all without wrappers)
I'm pretty confident this is a good way to proceed, I just made sure I had the good object in hands, but this line should not be there and crypto_provider.generate_triple shoudl return the correct type

robert-wagner: So the setup earlier was that everything in this method was actually a dictionary of pointer tensors. Has that changed?

LaRiffle: Yes I think it was not the best choice -- well it was good to have it run quickly; but it forces to always iterate manually while we have protocols to handle classic operations transparently

LaRiffle: I think we can re-rewrite the spdz protocol making the hypothesis we only handle AdditiveSharedTensor and MultiPointerTensor. That's what I did for snn, but I added functionalities so that they would play together nicely.

LaRiffle: I've just fixed the snn thing (so relu, eq, comp) works, I'm planning on opening another PR with code cleaned and correctly committed

robert-wagner: Sounds good to me. I have an open pr with decompose inplemented. Just need to write tests for it

LaRiffle: Maybe you can compare with the version of decompose implemented here also, it looks like it's also working :)

LaRiffle:
Refactored in #2027

PR #2022, opened 2019-03-28
Fix #1992

### LINKED ISSUE #1992: Urgent: fail gracefully when initializing two VirtualWorker objects with the same id/name
When people are introduced to PySyft, they often fall into the mistake (particularly in Jupyter Notebbooks) of initializing a VirtualWorker with the same ID twice. This causes really strange errors because both workers end up co-existing together. We need to modify the VirtualWorker (or perhaps the BaseWorker) API to make it so that this fails gracefully (aka... that the new worker actually becomes the old worker or all intensive purposes)

Aka - we want this code to function normally

bob = sy.VirtualWorker(hook, "bob")

x = th.tensor([1,2,3,4,5]).send(bob)

bob = sy.VirtualWorker(hook, "bob")

y = th.tensor([1,2,3,4,5]).send(bob)

z = x + y

z.get() # returns [2,4,6,8,10]

We should accomplish this by having both the new "bob" and the old "bob" point to the same underlying _objects dictionary.

mari-linhares:
Fix #1992.

### DIFF
--- a/syft/workers/base.py
+++ b/syft/workers/base.py
@@ -1,4 +1,3 @@
-import copy
 import logging
 import random
 import sys
--- a/syft/workers/base.py
+++ b/syft/workers/base.py
@@ -92,6 +92,9 @@ class BaseWorker(AbstractWorker):
             known_workers = self.hook.local_worker._known_workers
             if self.id in known_workers:
                 if isinstance(known_workers[self.id], type(self)):
+                    # If the a worker with this id already exists and it has the
+                    # same type as the one being created, we copy all the attributes
+                    # of the existing worker to this one.
                     self.__dict__.update(known_workers[self.id].__dict__)
                 else:
                     raise RuntimeError("Worker initialized with the same id and different types.")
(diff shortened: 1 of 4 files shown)


syft/workers/base.py
```diff
@@ -94,8 +84,26 @@ def __init__(
...
+
         self.load_data(data)
 
+        # Declare workers as appropriate
+        self._known_workers = {}
+        if hook.local_worker is not None:
+            known_workers = self.hook.local_worker._known_workers
+            if self.id in known_workers:
+                if isinstance(known_workers[self.id], type(self)):
+                    self.__dict__.update(known_workers[self.id].__dict__)
```
robert-wagner: What exactly is this line doing? Could you add a comment

mari-linhares: If you think more details would be helpful I can try to add a better explanation.

test/conftest.py
```diff
@@ -1,10 +1,27 @@
 import pytest
 import torch
+from multiprocessing import Process
 
 import syft
 from syft import TorchHook
 
 
+@pytest.fixture(scope="session", autouse=True)
```
robert-wagner: This should not be auto used and scope should be function. (Just don't pass parameter s)

PR #2028, opened 2019-04-01
Websocket worker example

midokura-silvia:
This pull request contains an example implementation of a federated training using websockets and federated averaging. The training on the workers is currently run sequentially (no scheduler implemented).


examples/tutorials/Advanced/Websockets - Federated training/README.md
```diff
@@ -0,0 +1,25 @@
...
+The scripts in this folder let you execute a federated training via three websocket connections.
+
+The script start_websocket_servers.py will start the Websocket server workers for Alice, Bob and Charlie.
+```
+$ python start_websocket_servers.py
+```
+
+The training is then started by running the script run_websocket_client.py:
+```
+$ python run_websocket_client.py
```
robert-wagner: Any way that we could create a version of this in a notebook so that people trying to use it can iterate step by step?

examples/tutorials/Advanced/Websockets - Federated training/README.md
```diff
@@ -0,0 +1,25 @@
...
+```
+This script
+ * loads the MNIST dataset,
+ * distributes it onto the three workers
```
robert-wagner: Could we do this step by intitializing each of alice, bob and charlie with the data parameter. That seems to be much more in line with how fl will work in the wild

examples/tutorials/Advanced/Websockets - Federated training/run_websocket_server.py
```diff
@@ -0,0 +1,46 @@
...
+    p.start()
+    return p
+
+
+parser = argparse.ArgumentParser(description="Run websocket server worker.")
+parser.add_argument(
+    "--port", "-p", type=int, default=8777, help="port number of the websocket server worker"
+)
+parser.add_argument("--host", type=str, default="localhost", help="host for the connection")
+parser.add_argument("--id", default="alice", help="name (id) of the websocket server worker")
```
robert-wagner: This probably shouldn't have a default as that can lead to conflicts

examples/tutorials/Advanced/Websockets - Federated training/run_websocket_server.py
```diff
@@ -0,0 +1,46 @@
...
+        server.start()
+
+    p = Process(target=target)
+    p.start()
+    return p
+
+
+parser = argparse.ArgumentParser(description="Run websocket server worker.")
+parser.add_argument(
+    "--port", "-p", type=int, default=8777, help="port number of the websocket server worker"
```
robert-wagner: This should probably not have a default as that can lead to conflicts

syft/frameworks/torch/federated/dataloader.py
```diff
@@ -110,6 +110,50 @@ def stop(self):
...
+        # Assign the worker to invoke
+        self.worker = loader.workers[worker_idx]
+
+        # The function used to stack all samples together
+        self.collate_fn = loader.collate_fn
+
+        # Create a sample iterator for each worker
+        self.sample_iter = iter(loader.batch_samplers[self.worker])
+
+    # def __len__(self):
```
robert-wagner: why is this commented out

syft/frameworks/torch/federated/dataloader.py
```diff
@@ -110,6 +110,50 @@ def stop(self):
         raise StopIteration
 
 
+class _DataLoaderOneWorkerIter(object):
+    r"""Iterates once over the worker's dataset, as specified by its sampler"""
```
robert-wagner: does this need to be a raw string literal?

syft/workers/websocket_client.py
```diff
@@ -32,20 +36,36 @@ def __init__(
...
+        self.ws.send(str(binascii.hexlify(message)))
+        response = binascii.unhexlify(self.ws.recv()[2:-1])
+        return response
+
     def _recv_msg(self, message: bin) -> bin:
         """Forwards a message to the WebsocketServerWorker"""
 
-        self.ws.send(str(binascii.hexlify(message)))
-        response = binascii.unhexlify(self.ws.recv()[2:-1])
+        response = self._receive_action(message)
```
robert-wagner: This will fail or hang if the websocket is not connected. What is the check after this actually doing?

test/torch/federated/test_utils.py
```diff
@@ -0,0 +1,32 @@
...
+import torch
+from syft import TorchHook
+
+from syft.frameworks.torch.federated import utils
+from syft.frameworks.torch import federated
+
+hook = TorchHook(torch)
+
+
+class DataLoaderTestCase(unittest.TestCase):
```
robert-wagner: Format this in pytest style

test/torch/federated/test_utils.py
```diff
@@ -0,0 +1,32 @@
...
+    def setUp(self):
```
robert-wagner: This is handled by the conftest workers

test/torch/federated/test_utils.py
```diff
@@ -0,0 +1,32 @@
...
+
+import torch as th
+import syft as sy
+import torch
+from syft import TorchHook
+
+from syft.frameworks.torch.federated import utils
+from syft.frameworks.torch import federated
+
+hook = TorchHook(torch)
```
robert-wagner: this is handled by the conftest hook

examples/tutorials/Advanced/Websockets - Federated training/README.md
midokura-silvia: I thought of it, but debugging a notebook is more difficult than putting breakpoints in an editor as pycharm. 
The script starting the websocket servers will not go away, as starting the processes from a notebook does not work (see discussion in the slack channel).
I will see what makes sense in a notebook, as it shouldn't be too long neither.

syft/frameworks/torch/federated/dataloader.py
midokura-silvia: That is a left-over of some tests. I will un-comment it.

syft/frameworks/torch/federated/dataloader.py
midokura-silvia: No, I don't think so. I will remove the r-prefix from all the docstrings in the file.

syft/workers/websocket_client.py
midokura-silvia: _receive_action will not fail, if the websocket is not connected. but the response will be empty. 
The check on self.ws.connected has to be after a self.ws.recv() call as otherwise the variable is not updated. 
So the steps are: 1) try to send/receive, 2) check if connection is ok, if not 3) reconnect and send/receive again. 4) check if connection is ok now, if not fail.

test/torch/federated/test_utils.py
midokura-silvia: will do

iamtrask:
Hey @midokura-silvia!! This is some great work! Are you in PySyft's slack? I'd love to have a chat.

examples/tutorials/Advanced/Websockets - Federated training/README.md
midokura-silvia: I added a notebook version of the example.

examples/tutorials/Advanced/Websockets - Federated training/README.md
midokura-silvia: This could be a separate example, where the focus is on setting up the remote data in a realistic way. My suggestion is to do a separate pull request for this.

examples/tutorials/Advanced/Websockets - Federated training/README.md
```diff
@@ -0,0 +1,25 @@
+# Federated learning using websockets - MNIST example
```
robert-wagner: Could you make the name of this folder more command line friendly (ie no spaces

midokura-silvia: Would the name "websockets-example-MNIST" be good?
Shall we change the "Advanced" folder to "advanced" to align it as well?

iamtrask:
Excellent work Silvia! And Also excellent reviewing Bobby!

PR #2037, opened 2019-04-02
added copy and removed precision integral

robert-wagner:


test/torch/tensors/test_precision.py
```diff
@@ -145,3 +145,12 @@ def test_fixed_precision_and_sharing(workers):
 
     y = y.get().float_prec()
     assert (y == torch.tensor([2, 4, 6, 8.0])).all()
+
+
+def test_get_preserves_attirbutes(workers):
```
LaRiffle: there is a typo in func name

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -294,7 +283,16 @@ def handle_func_command(cls, command):
...
+        return FixedPrecisionTensor(
+            owner=self.owner,
+            id=self.id,
+            field=self.field,
+            base=self.base,
+            precision_fractional=self.precision_fractional,
+            kappa=self.kappa,
+            tags=self.tags,
+            description=self.description,
+        ).on(self.child.get())
```
LaRiffle: Maybe you could leverage the `def get_class_attributes(self):` of precision.py

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -49,7 +49,16 @@ def get_class_attributes(self):
...
+        return {
+            "owner": self.owner,
+            "id": self.id,
+            "field": self.field,
+            "base": self.base,
+            "precision_fractional": self.precision_fractional,
+            "kappa": self.kappa,
+            "tags": self.tags,
+            "description": self.description,
+        }
```
LaRiffle: Oh, one thing I've not mentioned, you shouldn't provide owner and id in the get_class_attributes as you don't want tensors results to have the same id than the initial tensor and the owner should be inferred naturally.
Regarding tags and description, you may also want not to add them here.
Basically get_class_attributes is all the attribute of x you want to see in f(x), for any f

LaRiffle: Well for owner I'm not 100% sure maybe you can keep it

robert-wagner: So for .get we want to preserve these attributes. I will modify it to manually add those attributes

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -279,7 +275,13 @@ def get(self):
         """Just a pass through. This is most commonly used when calling .get() on a
         FixedPrecisionTensor which has also been shared."""
         class_attributes = self.get_class_attributes()
-        return FixedPrecisionTensor(**class_attributes).on(self.child.get())
+        return FixedPrecisionTensor(
+            **class_attributes,
+            owner=self.owner,
+            tags=self.tags,
+            description=self.description,
+            id=self.id
```
LaRiffle: Perfect!

PR #2042, opened 2019-04-04
FFN secure evaluation using SPDZ and SecureNN

LaRiffle:
Remaining:
- [x] Argmax on Additive Shared Tensors (on-going)

Will be done later:
- Tutorial with secure evaluation of a feed forward network


test/torch/crypto/test_snn.py
```diff
@@ -20,6 +20,63 @@ def test_xor_implementation(workers):
     assert (w.virtual_get() == w_real).all()
 
 
+def test_private_compare(workers):
+    """
+    Should be  β′ = β ⊕ (x > r).
+    :param workers:
+    :return:
```
robert-wagner: Good style

syft/frameworks/torch/crypto/securenn.py
```diff
@@ -141,73 +172,183 @@ def msb(a_sh, alice, bob):
...
-    x_sh = x.share(bob, alice, crypto_provider=crypto_provider).child
-    # Get last column / value as decompose reverts bits: first one is in last position
-    x_bit_0 = x_bit[..., -1:]
+    x_sh = x.share(bob, alice, field=L - 1, crypto_provider=crypto_provider).child
+    x_bit_0 = x_bit[
+        ..., -1
+    ]  # Get last value as decompose reverts bits: 1st one is in last position
     x_bit_sh_0 = x_bit_0.share(
-        bob, alice, crypto_provider=crypto_provider
+        bob, alice, field=L, crypto_provider=crypto_provider
```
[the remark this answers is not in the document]
LaRiffle: No this is correct according to the paper

syft/frameworks/torch/crypto/securenn.py
```diff
@@ -141,73 +172,183 @@ def msb(a_sh, alice, bob):
...
+        .get()
+        .child
+    )
+    r_shares = r_sh.child
+    alpha = (
+        ((r_shares[workers[0].id] + (r_shares[workers[1].id] * 1).move(workers[0])).get() >= L)
+        .long()
+        .send(*workers)
+        .child
+    )  # FIXME security
```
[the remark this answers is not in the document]
LaRiffle: alpha is known be the local worker and this should 't happen

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -188,15 +214,35 @@ def _getitem_multipointer(self, self_shares, indices_shares):
...
+        """
+        selected_shares = {}
+        for worker, share in self_shares.items():
+            selected_shares[worker] = share[indices]
+
+        return selected_shares
+
     def __getitem__(self, indices):
         tensor_type = type(indices)
+        print(tensor_type)
```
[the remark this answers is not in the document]
LaRiffle: no you're right!

syft/frameworks/torch/crypto/securenn.py
```diff
@@ -7,8 +7,10 @@
 
 import syft as sy
 
+p = 67
```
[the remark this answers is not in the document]
LaRiffle: a variable from the paper

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -571,6 +571,11 @@ def share(self, *owners, field=None, crypto_provider=None):
             .wrap()
         )
 
+    def share_(self, *args, **kwargs):
```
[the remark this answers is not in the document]
LaRiffle: Not yet I think

syft/frameworks/torch/crypto/securenn.py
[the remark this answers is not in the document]
robert-wagner: Could you add a comment for what it represents

syft/frameworks/torch/crypto/securenn.py
```diff
@@ -43,26 +82,28 @@ def private_compare(x, r, BETA, j, alice, bob, crypto_provider):
...
         β′ = β ⊕ (x > r).
     """
+    assert isinstance(x, sy.AdditiveSharingTensor)
+    assert isinstance(r, sy.MultiPointerTensor)
+    assert isinstance(BETA, sy.MultiPointerTensor)
+
+    alice, bob = x.locations
+    crypto_provider = x.crypto_provider
+    p = x.field
+    L = 2 ** Q_BITS  # 2**l
```
robert-wagner: Should this have 1 subtracted from it? L is not a prime field in this case

syft/frameworks/torch/crypto/securenn.py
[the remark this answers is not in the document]
robert-wagner: Could you update the comment to be more specific

syft/frameworks/torch/crypto/securenn.py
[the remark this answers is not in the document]
LaRiffle: I already have :)

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -9,7 +9,7 @@ def __init__(
         self,
         owner=None,
         id=None,
-        field: int = (2 ** 31) - 1,
+        field: int = (2 ** 62) - 1,
```
robert-wagner: This field is not prime

robert-wagner: In mathematics, a finite field or Galois field (so-named in honor of Évariste Galois) is a field that contains a finite number of elements. As with any field, a finite field is a set on which the operations of multiplication, addition, subtraction and division are defined and satisfy certain basic rules. The most common examples of finite fields are given by the integers mod p when p is a prime number.

Is from Wikipedia. From mortens blog the field needs to be a finite field

test/torch/tensors/test_additive_shared.py
```diff
@@ -59,39 +66,118 @@ def test_send_get(workers):
...
     assert (y == (t * t)).all()
 
+    # # with non-default fixed precision
+    # t = torch.tensor([1, 2, 3, 4.0])
+    # x = t.fix_prec(precision_fractional=2).share(bob, alice, crypto_provider=james)
+    # y = (x * x).get().float_prec()
+    #
+    # assert (y == (t * t)).all()
 
-def test_mul_with_no_crypto_provider(workers):
```
robert-wagner: Add a test that the crypto provider is added correctly

syft/frameworks/torch/crypto/securenn.py
LaRiffle: It doesn't need to be prime (this value is set as is in the article)

syft/frameworks/torch/tensors/interpreters/precision.py
LaRiffle: The article of SecureNN claims that additions and multiplication should be done in a field of 2**32 or 2**64 as it allows to use the native modulo implementation fo the int and long types.
The field for the comparison in return (p=67 in the implementation & paper) should be prime

test/torch/tensors/test_additive_shared.py
LaRiffle: As for now, when the crypto provider is not supplied, the local worker is used - it makes the syntax simpler for small use cases

robert-wagner: Agreed. However, we should test that (assert crypto provider is equal to local worker)

syft/frameworks/torch/tensors/interpreters/precision.py
robert-wagner: Sounds good.

LaRiffle: I agree there is a lot of subtleties here which I don't address in terms of fields mainly because I keep allowing negative values - because long can be negative.
There is more work to do there but most of the protocol is now written :)

test/torch/tensors/test_additive_shared.py
LaRiffle: Sure!

syft/frameworks/torch/crypto/securenn.py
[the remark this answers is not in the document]
LaRiffle: Done!

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -573,6 +573,11 @@ def share(self, *owners, field=None, crypto_provider=None):
...
```
robert-wagner: Can you add a comment saying this is inplace(a la share)

syft/frameworks/torch/tensors/interpreters/pointer.py
```diff
@@ -309,3 +309,12 @@ def attr(self, attr_name):
         ).wrap()
         self.__setattr__(attr_name, attr_ptr)
         return attr_ptr
+
+    def share(self, *args, **kwargs):
```
robert-wagner: What does this return?

test/torch/tensors/test_additive_shared.py
```diff
@@ -59,39 +66,118 @@ def test_send_get(workers):
...
+    assert (y == (t * t)).all()
+
+    # with fixed precision
     t = torch.tensor([1, 2, 3, 4.0])
     x = t.fix_prec().share(bob, alice, crypto_provider=james)
     y = (x * x).get().float_prec()
 
     assert (y == (t * t)).all()
 
+    # # with non-default fixed precision
```
robert-wagner: why are these commented out?

test/torch/tensors/test_additive_shared.py
```diff
@@ -187,3 +273,97 @@ def test_comp(workers):
...
+    Just to show that handle_func_command works
+    Even is torch.abs should be hooked to return a correct value
+    """
+    alice, bob, james = workers["alice"], workers["bob"], workers["james"]
+
+    t = torch.tensor([-21]).share(bob, alice, crypto_provider=james).child
+    _ = torch.abs(t).get()
+
+
+def test_mul_with_no_crypto_provider(workers):
```
robert-wagner: this isnt mul should be test_init_with_no_crypto_provider

syft/frameworks/torch/tensors/interpreters/pointer.py
LaRiffle: Allows you to send an order to share remotely: useful for privacy when local worker shoudl'nt be trusted

LaRiffle: * on the way shares are done

test/torch/tensors/test_additive_shared.py
LaRiffle: Oh no reason I just forgot to uncomment it! Thanks :)

test/torch/tensors/test_additive_shared.py
LaRiffle: Correct!

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: Done!

syft/frameworks/torch/hook_args.py
```diff
@@ -209,9 +211,12 @@ def hook_response(attr, response, wrap_type, wrap_args={}, new_self=None):
     if not response_is_tuple:
         response = (response, 1)
 
-    attr_id = "{}@{}.{}".format(attr, wrap_type.__name__, hash(frozenset(wrap_args.items())))
+    hash_wrap_args = hash(frozenset(wrap_args.items()))
+    attr_id = f"{attr}@{wrap_type.__name__}.{response_is_tuple}.{hash_wrap_args}"
 
     try:
+        assert attr not in exclude_functions
```
robert-wagner: Why are these assertions in a try vs just a Boolean if

syft/frameworks/torch/hook_args.py
```diff
@@ -614,6 +619,8 @@ def register_response(
     attr_id = "{}".format(attr)
 
     try:
+        assert attr not in exclude_functions
```
robert-wagner: Same as above

syft/frameworks/torch/hook_args.py
LaRiffle: The try catch not only errors of this assertion but also errors in the two other lines
```
# Load the utility function to transform the args
        response_hook_function = hook_method_response_functions[attr_id]
        # Try running it
        new_response = response_hook_function(response)
```

PR #2043, opened 2019-04-04
Update securenn.py

estefanyd:
added a few type annotations

### DIFF
--- a/syft/frameworks/torch/crypto/securenn.py
+++ b/syft/frameworks/torch/crypto/securenn.py
@@ -11,7 +11,7 @@ import syft as sy
 Q_BITS = 31
 
 
-def decompose(tensor):
+def decompose(tensor: torch.Tensor):
     """decompose a tensor into its binary representation."""
     n_bits = Q_BITS
     powers = torch.arange(n_bits)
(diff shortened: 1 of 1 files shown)


syft/frameworks/torch/crypto/securenn.py
```diff
@@ -11,7 +11,7 @@
 Q_BITS = 31
 
 
-def decompose(tensor):
+def decompose(tensor: torch.Tensor):
```
LaRiffle: Actually it can also be an additiveSharedTensor

syft/frameworks/torch/crypto/securenn.py
```diff
@@ -25,7 +25,7 @@ def decompose(tensor):
     return tensor
 
 
-def flip(x, dim):
+def flip(x: torch.Tensor, dim: int):
```
LaRiffle: same here, x  can also be an additiveSharedTensor

syft/frameworks/torch/crypto/securenn.py
```diff
@@ -118,7 +118,7 @@ def private_compare(x, r, BETA, j, alice, bob, crypto_provider):
     return result_ptr.get()
 
 
-def msb(a_sh, alice, bob):
+def msb(a_sh: torch.Tensor, alice: sy.BaseWorker, bob: sy.BaseWorker):
```
LaRiffle: a_sh is  an additiveSharedTensor (and can't be a torch.Tensor)

PR #2044, opened 2019-04-05
 Add search functionality for plans; Fix #2035

### LINKED ISSUE #2035: Add search functionality for plans
We want data owners to be able to call .search on a worker which has a plan and get a pointer to that plan. In practice this means we need to add tags to plans as well as a way to register them to a worker

mari-linhares:
- [x] Add name, tags and description to plan serialization
- [x] Add tag and describe methods to plan
- [x] Plans now can be executed locally in a transparent way
- [x] Add search functionality for plans; Fix #2035 

Open Questions:

* Currently only non-client workers can execute plans, client workers get errors when trying to run plans since they can't register objects of their own. How do we want to deal with this? This could be a potential pain for beginners.

Example, this case fails since the local worker is a client worker:
```
import syft as sy
import torch as th

hook = sy.TorchHook(th)

@sy.func2plan
def plan_abs(data):
    return data.abs()

x = th.tensor([-1, 2, 3])
x_abs = plan_abs(x)

print('local plan', x_abs)
```
Error:

```
KeyError: 'Tensor "84471588453" not found on worker "me"!!! You just tried to interact with an object ID:84471588453 on worker me which does not exist!!! Use .send() and .get() on all your tensors to make sure they\'reon the same machines. If you think this tensor does exist, check the ._objects dictionaryon the worker and see for yourself!!! The most common reason this error happens is because someone calls.get() on the object\'s pointer without realizing it (which deletes the remote object and sends it to the pointer). Check your code to make sure you haven\'t already called .get() on this pointer!!!'
```

This works just fine:

```
import syft as sy
import torch as th

hook = sy.TorchHook(th, is_client=False)

@sy.func2plan
def plan_abs(data):
    return data.abs()

x = th.tensor([-1, 2, 3])
x_abs = plan_abs(x)

print('local plan', x_abs)
```

Output: `local plan tensor([1, 2, 3])`

* What should happen if a worker wants to run locally a plan that was already built on another worker? Currently the plan is being built again.

* If the plan is executed locally should it return a pointer to the result or the result itself (tensor)? Currently .get() is being called on the pointer, so a local client would get the result itself (as shown in the second code snippet above).


syft/workers/plan.py
```diff
@@ -73,12 +77,66 @@ def method(self: object) -> Plan:
     return method
 
 
-class PlanPointer:
-    def __init__(self, id, location, id_at_location, owner):
-        self.id = id
+class PlanPointer(AbstractTensor):
```
robert-wagner: Why is plan pointer a tensor?

mari-linhares: This was mainly a workaround, to be able to search tensors and plans without distinction in the code. In the [search function at base.py](https://github.com/OpenMined/PySyft/blob/8f9263fa884c2ba7f2442f2108069b592b348e1a/syft/workers/base.py#L753) they create the pointer and wrap it.

Another possibile, and potentially more intuitive, approach would be to treat them differently in the search function and make PlanPointer just a wrapper that points to the Plan.

robert-wagner: One of the things about plans that always stood out for me was that they were callables so plan(tensor) worked. Is that still the case with plan pointers?

mari-linhares: Not really, a plan pointer is not callable.

robert-wagner: In that case what is the intended use of plan pointer, are they what is returned when you call search for a plan, then you need to call get to use it? If that is the case, I think we should just return the plan

mari-linhares: > In that case what is the intended use of plan pointer, are they what is returned when you call search for a plan, then you need to call get to use it

Yes.

>  If that is the case, I think we should just return the plan

sgtm.

LaRiffle: So there is a bit of a confusion about Plan Pointer, the only reason why I had kept them is that they were convenient for serialisation, They don't behave as PointerTensor

mari-linhares: Yep, but I believe we still need some kind of wrapper around the plan to something similar to [what is being done to tensors](https://github.com/OpenMined/PySyft/blob/8f9263fa884c2ba7f2442f2108069b592b348e1a/syft/workers/base.py#L753), right? Or we should update the plan's location/owner on the plan itself?

LaRiffle: That's not strictly needed, we could "update the plan's location/owner on the plan itself?" but overall yes I agree we should find a way so that the Plan sending / and getting is closer to the way tensors are handled, even if for Plan the sending is lazy and doesn't happen for real until you need a command to be executed remotely

mari-linhares: For now, I removed the PlanPointer, and the plan itself is returned.

robert-wagner: With that in mind we might just want to get rid of plan pointer entirely (is it used for anything?)

mari-linhares: I think it's being used for serialization only. I'll try to get rid of it entirely.

mari-linhares: Actually, since this can make this PR a little to big and removing this pointer must be done carefully, I'll just remove my changes, and keep it as a simple wrapper.

PR #2050, opened 2019-04-09
Remove PlanPointer

mari-linhares:


syft/workers/base.py
```diff
@@ -739,29 +741,29 @@ def search(self, *query):
...
 
-                if tensor.tags is not None:
-                    if query_item in tensor.tags:
+                if obj.tags:
+                    if query_item in obj.tags:
                         match = True
 
-                if tensor.description is not None:
-                    if query_item in tensor.description:
+                if obj.description:
```
robert-wagner: If the description is 0 or an empty string this will return incorrectly. The previous check was correct

LaRiffle:
Except the point you mention everything is good for me!

LaRiffle:
@mari-linhares I'll add an option to have several locations at the same time - it's work I should have cleaned and pushed, so if you need it tell me I will make it available faster

syft/workers/base.py
mari-linhares: Thanks for the careful review! Done.

PR #2055, opened 2019-04-10
Federated LSTM Example

kamathhrishi:


examples/tutorials/Advanced/Federated LSTM-Predicting Next Word.ipynb
```diff
@@ -0,0 +1,520 @@
...
+   "metadata": {},
+   "source": [
+    "<h1>Federated LSTM-Predicting Next Word</h1>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This example demonstrates how predicting next word model in PyTorch could be trained using federated learning with PySyft. We distribute the text data to two workers Bob and Alice to whom the model is sent and trained. Upon training the model the trained model is sent back to the owner of the model and used to make predictions.Federated learning applied to predicting next word helps to improve accuracy of predictor while ensuring the exact text of the end users are not known. \n",
```
mari-linhares: Maybe add a link to the Pytorch only tutorial so the reader can take a look if they want to

mari-linhares: .Federated learning applied ->. Federated learning applied (missing space)

examples/tutorials/Advanced/Federated LSTM-Predicting Next Word.ipynb
```diff
@@ -0,0 +1,520 @@
...
+   "execution_count": 4,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "class TextDataset(Dataset):\n",
+    "\n",
+    "    def __init__(self,text,transform=None):\n",
+    "        \n",
```
mari-linhares: Use [google style for docstrings](https://github.com/google/styleguide/blob/gh-pages/pyguide.md)

mari-linhares: Remove spaces between a function definition and docstring.

examples/tutorials/Advanced/Federated LSTM-Predicting Next Word.ipynb
```diff
@@ -0,0 +1,520 @@
...
+    "        \"\"\"arguments:\n",
+    "        \n",
+    "             text (List of Strings): Text corpus \n",
+    "             transform: List of transforms to be performed on the input data\n",
+    "             \n",
+    "        \"\"\"\n",
+    "\n",
+    "        self.text = text\n",
+    "        self.data=[]\n",
```
mari-linhares: always add spaces around equal sign for attribution: self.data=[] -> self.data = []

DanyEle:
I tried running your code and even tried creating my own example with LSTM / RNNs, however when operating on the model pointer and not on the model itself, I always get the same error you also got in your 'Federated LSTM-Predicting Next Word.ipynb' notebook. 

image


When operating on the model.get() or the plain model (i.e.: not the one being pointed to), I don't have any issues though. 

Any ideas why this issue related to the model pointer could happen?

kamathhrishi:
> I tried running your code and even tried creating my own example with LSTM / RNNs, however when operating on the model pointer and not on the model itself, I always get the same error you also got in your 'Federated LSTM-Predicting Next Word.ipynb' notebook when running the forward pass on the LSTM.
> 
> image
> 
> When operating on the model.get() or the plain model (i.e.: not the one being pointed to), I don't have any issues though.
> 
> Any ideas why this issue related to the model pointer could happen?

@DanyEle Thank you for reproducing the issue :) . PySyft currently does not have support for LSTM yet, which @LaRiffle is trying to solve in #1995 and has some issues. If you are interested in solving it you can talk to him about it.

DanyEle:
I think I manged to solve it on my local end: basically you had to create a pointer to the hidden layer and then send the hidden layer to the worker as well, then operate on both the model's pointer and hidden layer's pointer. Now i'm going to try and see if I can also fix your example code as well,

DanyEle:
Well I've been working the whole day on trying to fix my issues on training a RNN in a federated way, but still bumped into the same error I posted above. Even with the partial solution about sending pointers to hidden layers, the backward propagation procedure did not work and the gradient could not be computed. I wonder if RNNs are supported by PySyft?

kamathhrishi:
@DannyEle I don't think we support RNNs too. Haven't tried it yet. But , one way to go about RNNs is to implement them from scratch using linear layers like the charecter level name generator given in Pytorch tutorials. Thank you for trying though :) it will be quite sometime till we get torch LSTM/RNN working. Temporarily it would be great to have the RNN from scratch example , if you would like to tackle that you could go ahead :)

kamathhrishi:
https://pytorch.org/tutorials/intermediate/char_rnn_generation_tutorial.html this is for generating names. There is also a charecter level classifier.

DanyEle:
Thanks for the link, I was actually working on getting a very similar example to work in a federated way, namely: https://pytorch.org/tutorials/intermediate/char_rnn_classification_tutorial.html . This RNN is also based off linear layers, as you suggested. However, I've been having with very little success on getting that working, as I had the very same issue as you had when performing the forward pass.

DanyEle:
Alright, I finally managed to get the training running for a simple RNN. The issue I was running into was related to the fact that not all parameters were being set in the model and I was just trying to access the first one. Should I open a pull request for my example with RNNs, even though it's not in a jupyter notebook, but in a plain Python file or should I put it in a jupyter notebook and then open a pull request?

kamathhrishi:
@DanyEle That's amazing :) A Jupyter notebook in the format of other tutorials under Advanced. Feel free to submit a PR for it.

**Bonus:** It would be great if you could make an example of it trying to predict the next word given a context. This is particularly great because it makes it more intuitive as to why we are doing federated learning :)

DanyEle:
There are some remaining issues related to the weights not being updated and the loss not being properly initialized. I opened an issue about them:  https://github.com/OpenMined/PySyft/issues/2122

PR #2013, opened 2019-03-26
Draft for socket IO server side and Android worker

mccorby:
This PR contains a new socket server worker using `socket.io`
It includes a demo for such server and a notebook to use it in conjunction with clients, in particular, the Android worker being implemented in https://github.com/OpenMined/AndroidWorker

PR #2056, opened 2019-04-10
SocketIO server that allows sync communication with Android

mccorby:
A `socketio` server worker implementation to allow sync-ed communication with other systems.

The server worker is used to communicate with an Android app (see https://github.com/OpenMined/AndroidWorker)

Initially open here https://github.com/OpenMined/PySyft/pull/2013

### DIFF
--- a/syft/workers/socketio_client.py
+++ b/syft/workers/socketio_client.py
@@ -11,10 +11,10 @@ from syft.workers.virtual import VirtualWorker
 
 
 class WebsocketIOClientWorker(VirtualWorker):
-    """A client which will send a message to a WebsocketIOServerWorker when it receives
-    a message from another worker via _recv_msg
-    This client then waits until the server returns with a result or an ACK at which point
-    it finishes the _recv_msg operation
+    """A worker that forwards a message to a SocketIO server and wait for its response.
+
+    This client then waits until the server returns with a result or an ACK at which point it finishes the
+    _recv_msg operation.
     """
 
     def __init__(
@@ -28,6 +28,20 @@ class WebsocketIOClientWorker(VirtualWorker):
         verbose: bool = False,
         data: List[Union[torch.Tensor, AbstractTensor]] = None,
     ):
+        """
+        Args:
+            hook (sy.TorchHook): a normal TorchHook object
+            host (str): the host this client connects to
+            port (int): the port this client connects to
+            id (str or id): the unique id of the worker (string or int)
+            log_msgs (bool): whether or not all messages should be
+                saved locally for later inspection.
+            verbose (bool): a verbose option - will print all messages
+                sent/received to stdout
+            data (dict): any initial tensors the server should be
+                initialized with (such as datasets)
+        """
+
         self.port = port
         self.host = host
         self.uri = f"http://{self.host}:{self.port}"
--- a/syft/workers/socketio_client.py
+++ b/syft/workers/socketio_client.py
@@ -1,0 +1,84 @@
+import time
+from typing import List
+from typing import Union
+
+import socketio
+import torch
+
+import syft as sy
+from syft.frameworks.torch.tensors.interpreters import AbstractTensor
+from syft.workers.virtual import VirtualWorker
+
+
+class WebsocketIOClientWorker(VirtualWorker):
+    """A client which will send a message to a WebsocketIOServerWorker when it receives
+    a message from another worker via _recv_msg
+    This client then waits until the server returns with a result or an ACK at which point
+    it finishes the _recv_msg operation
+    """
+
+    def __init__(
+        self,
+        hook,
+        host: str,
+        port: int,
+        id: Union[int, str] = 0,
+        is_client_worker: bool = False,
+        log_msgs: bool = False,
+        verbose: bool = False,
+        data: List[Union[torch.Tensor, AbstractTensor]] = None,
+    ):
+        self.port = port
+        self.host = host
+        self.uri = f"http://{self.host}:{self.port}"
+
+        self.response_from_client = None
+        self.wait_for_client_event = False
+
+        # Creates the connection with the server
+        self.sio = socketio.Client()
+        super().__init__(hook, id, data, is_client_worker, log_msgs, verbose)
+
+        @self.sio.on("client_ack")
+        def on_client_ack(args):
+            # The server broadcasted an ACK from another client
+            self.response_from_client = "ACK"
+            # Tell the wait_for_client_event to clear up and continue execution
+            self.wait_for_client_event = False
+
+        @self.sio.on("client_send_result")
+        def on_client_result(args):
+            if log_msgs:
+                print("Receiving result from client {}".format(args))
+            # The server broadcasted the results from another client
+            self.response_from_client = args
+            # Tell the wait_for_client_event to clear up and continue execution
+            self.wait_for_client_event = False
+
+    def _send_msg(self, message: bin) -> bin:
+        raise RuntimeError(
+            "_send_msg should never get called on a ",
+            "WebsocketIOClientWorker. Did you accidentally "
+            "make hook.local_worker a WebsocketIOClientWorker?",
+        )
+
+    def _recv_msg(self, message: bin) -> bin:
+        # Sends the message to the server
+        self.sio.emit("message", message)
+        self.wait_for_client_event = True
+        # Wait until the server gets back with a result or an ACK
+        while self.wait_for_client_event:
+            time.sleep(0.1)
+
+        # Return the result
+        if self.response_from_client == "ACK":
+            # Empty result for the serialiser to continue
+            return sy.serde.serialize(b"")
+        return self.response_from_client
+
+    def connect(self):
+        self.sio.connect(self.uri)
+        self.sio.emit("client_id", self.id)
+
+    def disconnect(self):
+        self.sio.disconnect()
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -33,7 +33,7 @@ before_script:
   # https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds
   - flake8 . --count --exit-zero --statistics --select=E,F,W,C90
 script:
-  - coverage run --omit=*/virtualenv/*,setup.py,.eggs/*,*__init__.py,syft/workers/websocket_server.py setup.py test
+  - coverage run --omit=*/virtualenv/*,setup.py,.eggs/*,*__init__.py,syft/workers/websocket_server.py,syft/workers/socketio_* setup.py test
```
robert-wagner: Why do we not care about coverage of these files?

syft/workers/socketio_client.py
```diff
@@ -0,0 +1,83 @@
...
+        # Sends the message to the server
+        self.sio.emit("message", message)
+        self.wait_for_client_event = True
+        # Wait until the server gets back with a result or an ACK
+        while self.wait_for_client_event:
+            time.sleep(0.1)
+
+        # Return the result
+        if self.response_from_client == b"":
+            return sy.serde.serialize(b"")
```
robert-wagner: What exactly is this test doing?

robert-wagner: Also in what cases would this response be returned

syft/workers/socketio_server.py
```diff
@@ -0,0 +1,126 @@
...
+            # Broadcast the result
+            self.socketio.emit("client_send_result", args)
+
+    def run(self):
+        self.socketio.run(self.app, host=self.host, port=self.port, debug=self.log_msgs)
+
+    def _send_msg(self, message: bin) -> bin:
+        raise RuntimeError(
+            "_send_msg should never get called on a ",
+            "WebsocketClientWorker. Did you accidentally "
```
robert-wagner: Server worker

.travis.yml
mccorby: I've been trying to make the tests not hanging when using `socketio` but with no luck. 
A test like this
```suggestion
hook = sy.TorchHook(torch)
app = Flask("test")
socketio = SocketIO(app)
server_worker = WebsocketIOServerWorker(
    hook, "localhost", 5000, app=app, socketio=socketio, log_msgs=True
)

bob = socketio.test_client(app)

def test_client_id():
    android = socketio.test_client(app)
    android.emit("client_id", "android")
    assert len(server_worker.clients) == 1
    android.disconnect()
```

hangs indefinitely

syft/workers/socketio_client.py
mccorby: This represents an empty response. It is happening when the other side (i.e. Android) sends an `ACK`
The deserialisation will not accept something like `"ACK"` thus why it returns an empty response

mccorby: I will make the ACK operation to be `"ACK"` instead of empty up to the point where it's returned to be deserialised

.travis.yml
robert-wagner: I have a feeling the problem is that starting the app hangs. You can get around this by starting it in a separate process then killing the process when everything is done. This is what we do for websocket workers. The easiest way to figure out exactly what is hanging is to just put print statement every line of the test and see where it gets to

syft/workers/socketio_client.py
robert-wagner: Sounds good. That or adding a comment explaining why this is here would work (you should probably add the comment regardless)

.travis.yml
mccorby: I will give it a try. Thanks for the hint

test/workers/test_websocketIOServerWorker.py
```diff
@@ -0,0 +1,40 @@
...
+
+import torch
+
+import syft as sy
+from syft.workers import WebsocketIOServerWorker
+
+
+def _payload(location):
+    x = torch.tensor([10, 20, 30, 40, 50.0])
+    print("About to send x")
```
robert-wagner: Should these print statements be here?

test/workers/test_websocket_worker.py
```diff
@@ -54,4 +57,8 @@ def test_websocket_workers_search(hook, start_proc):
     assert results[0].owner.id == "me"
     assert results[0].location.id == "fed2"
 
+    client_worker.ws.shutdown()
+    client_worker.ws.close()
+    time.sleep(0.1)
     server.terminate()
+    server.join()
```
robert-wagner: what does this join do?

syft/workers/socketio_client.py
```diff
@@ -0,0 +1,84 @@
...
+import socketio
+import torch
+
+import syft as sy
+from syft.frameworks.torch.tensors.interpreters import AbstractTensor
+from syft.workers.virtual import VirtualWorker
+
+
+class WebsocketIOClientWorker(VirtualWorker):
+    """A client which will send a message to a WebsocketIOServerWorker when it receives
```
mari-linhares: Change this to google docstring style. Example: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

syft/workers/socketio_server.py
```diff
@@ -0,0 +1,129 @@
...
+from flask import Flask
+from flask_socketio import SocketIO
+
+import syft as sy
+from syft.frameworks.torch.tensors.interpreters import AbstractTensor
+from syft.workers.virtual import VirtualWorker
+
+
+class WebsocketIOServerWorker(VirtualWorker):
+    """
```
mari-linhares: Same.

syft/workers/socketio_server.py
```diff
@@ -0,0 +1,129 @@
...
+            if args not in self.clients:
+                self.clients.append(args)
+                # If this server has been created with a payload, execute it
+                if self._payload is not None:
+                    print(f"Client {args} connected. Start executing payload")
+                    self._start_payload()
+
+        @self.socketio.on("message")
+        def send_command(args):
+            print("Sending command to whoever is listening {}".format(args))
```
mari-linhares: Same.

test/workers/test_websocketIOClientWorker.py
```diff
@@ -0,0 +1,42 @@
+import time
+
+import socketio
+import torch
+
+from syft.workers import WebsocketIOServerWorker, WebsocketIOClientWorker
+
+
+def create_dummy_client():
```
mari-linhares: Maybe move this to conftest?

syft/workers/socketio_server.py
```diff
@@ -0,0 +1,129 @@
...
+        self._init_job_thread()
+        self.clients = []
+        self.wait_for_client_event = True
+        self._payload = payload
+
+        super().__init__(hook=hook, id=id, data=data, log_msgs=log_msgs, verbose=verbose)
+
+        @self.socketio.on("connect")
+        def on_client_connect():
+            print("New client established connection")
```
mari-linhares: Remove print statement or add a verbose mode to execution.

syft/workers/socketio_server.py
```diff
@@ -0,0 +1,129 @@
...
+
+        @self.socketio.on("client_id")
+        def on_client_id(args):
+            # Register client id and execute the payload for it
+            if args not in self.clients:
+                self.clients.append(args)
+                # If this server has been created with a payload, execute it
+                if self._payload is not None:
+                    print(f"Client {args} connected. Start executing payload")
```
mari-linhares: Same.

syft/workers/socketio_server.py
```diff
@@ -0,0 +1,129 @@
...
+    def _send_msg(self, message: bin) -> bin:
+        raise RuntimeError(
+            "_send_msg should never get called on a ",
+            "WebsocketIOServerWorker. Did you accidentally "
+            "make hook.local_worker a WebsocketIOServerWorker?",
+        )
+
+    def _recv_msg(self, message: bin) -> bin:
+        """Forwards a message to the WebsocketIOClientWorker"""
+        print("_recv_msg in Web socket Server")
```
mari-linhares: Same.

test/workers/test_websocketIOServerWorker.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def test_payload_execution():
+    android = server_worker.socketio.test_client(server_worker.app)
+    android.emit("client_id", "android")
+    time.sleep(0.1)
+    android.emit("client_ack", "Android")
+    time.sleep(0.3)
+    android.emit("client_ack", "Android")
+    time.sleep(0.3)
```
mari-linhares: Could it be 0.1? To avoid making the tests slower.

mccorby: I tried with several values and (unfortunately) this was the minimum I could use to make the test pass
Using any `time.sleep()` in tests is usually a bad idea and we probably need to find another way to improve all those tests that are using it

test/workers/test_websocketIOClientWorker.py
mccorby: The dummy client must connect after the server has been created

test/workers/test_websocket_worker.py
mccorby: Removed. Leftovers of attempts to make the test pass

test/workers/test_websocketIOServerWorker.py
mari-linhares: Agree. I'll create an issue for this.

test/workers/test_websocketIOClientWorker.py
mari-linhares: Cool, sgtm.

PR #2064, opened 2019-04-12
Create fetch_plan in base

mari-linhares:
* Plans are no longer searchable :cry: . This is mainly due to the different behavior when "searching" for plans and searching for tensors.
* Now is possible to fetch a plan based on its id :smile: . This will get a plan from a worker's registry, create a copy of the plan and change the owner to whoever is fetching the plan.

This is the result of a [~~pair~~ triple programming section](https://github.com/OpenMined/PySyft/commit/66f5cc9688f5c74140772766420ba5d034d24c01) with @robert-wagner and @jlebensold.

### DIFF
--- a/syft/workers/base.py
+++ b/syft/workers/base.py
@@ -717,7 +717,7 @@ class BaseWorker(AbstractWorker):
         shape = self.send_msg(MSGTYPE.GET_SHAPE, pointer, location=pointer.location)
         return sy.hook.torch.Size(shape)
 
-    def fetch_plan(self, plan_id: Union[str, int]) -> "Plan":
+    def fetch_plan(self, plan_id: Union[str, int]) -> "Plan":  # noqa: F821
         """Fetchs a copy of a the plan with the given `plan_id` from the worker registry.
 
         Args:
--- a/syft/workers/base.py
+++ b/syft/workers/base.py
@@ -167,7 +167,6 @@ class BaseWorker(AbstractWorker):
                 tensor.owner = self
 
     def send_msg(self, msg_type: int, message: str, location: "BaseWorker") -> object:
-
         """Implements the logic to send messages.
 
         The message is serialized and sent to the specified location. The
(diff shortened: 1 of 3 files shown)


syft/workers/base.py
```diff
@@ -719,15 +719,37 @@ def request_remote_tensor_shape(self, pointer):
...
         return sy.hook.torch.Size(shape)
 
+    def fetch_plan(self, plan_id):
+        """Fetchs a copy of a the plan with the given `plan_id` from the worker registry.
+
+        Args:
+            plan_id: A string indicating the plan id.
+
+        Returns:
+            A plan if a plan with the given `plan_id` exists. Otherwhise returns None.
```
robert-wagner: Can you say that it returns none otherwise

syft/workers/base.py
```diff
@@ -719,15 +719,37 @@ def request_remote_tensor_shape(self, pointer):
         shape = self.send_msg(MSGTYPE.GET_SHAPE, pointer, location=pointer.location)
         return sy.hook.torch.Size(shape)
 
+    def fetch_plan(self, plan_id):
```
robert-wagner: Can you add typing

syft/workers/base.py
```diff
@@ -719,15 +719,37 @@ def request_remote_tensor_shape(self, pointer):
...
+        """Fetchs a copy of a the plan with the given `plan_id` from the worker registry.
+
+        Args:
+            plan_id: A string indicating the plan id.
+
+        Returns:
+            A plan if a plan with the given `plan_id` exists. Otherwhise returns None.
+        """
+        if plan_id in self._objects:
```
jlebensold: do we prefer indentation to [guard](https://blog.startifact.com/posts/none_05_more_on_guarding.html) statements?

mari-linhares: thanks for pointing out this blog! I think we can go with the guard statement. I'll fix it.

PR #2065, opened 2019-04-12
Improve documentation for workers/base.py

mari-linhares:


syft/workers/base.py
```diff
@@ -583,7 +599,7 @@ def get_worker(self, id_or_worker, fail_hard=False):
 
         return id_or_worker
 
-    def add_worker(self, worker):
+    def add_worker(self, worker: "BaseWorker"):
```
robert-wagner: Why is this a string?

syft/workers/base.py
```diff
@@ -348,12 +355,20 @@ def execute_command(self, message):
                 )
                 raise ResponseSignatureError(return_ids.generated)
 
-    def send_command(self, recipient, message, return_ids=None):
+    def send_command(
+        self, recipient: "BaseWorker", message: str, return_ids: str = None
```
robert-wagner: Why is base worker a string

syft/workers/base.py
```diff
@@ -59,7 +59,13 @@ class BaseWorker(AbstractWorker):
     """
 
     def __init__(
-        self, hook, id=0, data=None, is_client_worker=False, log_msgs=False, verbose=False
+        self,
+        hook: "sy.TorchHook",
```
robert-wagner: Why is hook a string

syft/workers/base.py
```diff
@@ -630,7 +646,7 @@ def add_worker(self, worker):
             )
         self._known_workers[worker.id] = worker
 
-    def add_workers(self, workers):
+    def add_workers(self, workers: List["BaseWorker"]):
```
robert-wagner: Why is base worker a string

syft/workers/base.py
LaRiffle: You can mention a type if you are in its own class, that's kind the way of handling it

robert-wagner: Ah okay

syft/workers/base.py
mari-linhares: TorchHook imports BaseWorker. So I can't import hook on BaseWorker. The way to handle this is using [Forward references](https://www.python.org/dev/peps/pep-0484/#forward-references)

Issue #2062, opened 2019-04-12
plan.send() implementation is not intuitive


mari-linhares:
Currently, the send operation for plans is lazy, which in practice means that the plan is not actually sent when you call send.

This behavior is confusing and different from the expected behavior when sending tensors.

Here is an example of a unintended effect that can be very confusing :confused: : https://github.com/OpenMined/PySyft/blob/dev/test/workers/test_plan.py#L137

LaRiffle:
You're right this is a real problem. The lazy behaviour is because we need to build the plan prior to sending it, so we wait until it is executed on some data to sent it, otherwise we should run it on mock data which would very un practical.

One thing easy would be to change the name .send() to something else so that there would be no confusion, like .delegate()
Another possibility would be that A send a plan empty to B and B fills it when the sender A executes it with some pointers to B, with some mechanism to create. If C request plan execution to B before it is built it would raise some error.

jvmancuso:
Just to chime in an opinion without much context -- the ability to take code that's executed dynamically (i.e. without a plan) and turn it into code that's compiled statically (i.e. with a plan) should require as little modification as possible, since I imagine the intended workflow would be to develop code dynamically (for intuition) in preparation to deploy it statically (for speed).  There are past issues where we explored various APIs for this, although they may have never left the Grid project.

jvmancuso:
One of those APIs was a context manager that's sent and executes when you leave it -- could be an interesting option to investigate

mari-linhares:
I feel that we're trying to hard to make the tensor interface work for the plan interface (pointers, search, send, ...) when in practice it doesn't bring that many benefits as we expect.

Having said that I think the first option (send -> delegate (or something else)) could be a cleaner and easier to follow interface than actually trying to explain how send works for plans.

jvmancuso:
I think what I meant is that it would be nice if there was minimal thought for a user when interacting with this abstraction.  With a context manager, one could take code that runs in the normal way (i.e. interactively sending tensors/RPCs), wrap it with a single line, and all of a sudden have it execute as a plan.  So a user can focus on the logic of their code without worrying about batching things up, and then the jump from that to something more optimized is a one-step thought.

What might make a context manager useful at addressing the original comment is that lazy methods can be handled in a way that seems non-lazy.  I'm a bit far removed from the codebase so I'm not sure what this means in practice, but I hope this makes sense conceptually.

iamtrask:
[...]

PR #2068, opened 2019-04-13
Plans Design Doc discussion

mari-linhares:
**this PR should not be merged**

## Summary

There is a lot of discussion going on about how the design and implementation of the Plan abstraction, this draft PR is a single place where we can keep track of all the plans we have for plans :sweat_smile:.

The goal is to avoid multiple changes in the code and interface and come up with an abstraction that works well for the intended use cases.

Ps: if you think there's a better way for us to keep a public log of this discussion feel free to suggest in this thread.

**Relevant PRs and link**:

* Plans drat #1938
* Search for plans #2044 (plans can not be searched anymore)
* fetch_plan #2064 
* [Theo's excelent introduction to Plans](https://github.com/OpenMined/PySyft/blob/dev/examples/tutorials/Part%208%20-%20Introduction%20to%20Plans.ipynb)
* Plans should not be workers #2099 

### Original Definition

A plan is a concept defined in the paper [Towards Federated Learning at Scale: System Design](https://arxiv.org/pdf/1902.01046.pdf)...

> *The server tells the selected devices what computation to run
with an FL plan, a data structure that includes a TensorFlow
graph and instructions for how to execute it.*

> *As already mentioned, FL plans are not specialized to training,
but can also encode evaluation tasks - computing quality
metrics from held out data that wasn’t used for training,
analogous to the validation step in data center training.*

> *An FL plan consists of two parts: one for the device and one
for the server. The device portion of the FL plan contains,
among other things: the TensorFlow graph itself, selection
criteria for training data in the example store, instructions
on how to batch data and how many epochs to run on the
device, labels for the nodes in the graph which represent
certain computations like loading and saving weights, and
so on. The server part contains the aggregation logic, which
is encoded in a similar way. Our libraries automatically split
the part of a provided model’s computation which runs on
device from the part that runs on the server (the aggregation).*

### Pysyft Definition of a Plan

From the first draft of Plans...

> The idea is to send a "plan" ie all the instructions at once to a remote worker, while you still execute your code line by line like in PyTorch.
It uses the fact that now the worker that sends the command fix the id where the result is stored and therefore and can already create a pointer to a result which doesn't exist so far and go on with instructions.

## Plan use cases

* How are plans going to be used? Do we expect plans to be executed remotely or locally?

    * Plans should be built only once. Usually locally by their original owner (the worker who has access to the plans blueprint).
    * After plans are built they can be served to multiple workers.

```python
# run script to start client workers....

## -------------- SERVER  (Scheduler) ------------------

# start a locak worker
me = local_worker()

# model definition
class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.fc1 = nn.Linear(2, 3)
        self.fc2 = nn.Linear(3, 2)

    @sy.method2plan
    def forward(self, x):
        x = F.relu(self.fc1(x))
        x = self.fc2(x)
        return F.log_softmax(x, dim=0)

    @sy.method2plan
    def forward_no_softmax(self, x):
        x = F.relu(self.fc1(x))
        x = self.fc2(x)
        return x

net = Net()
workers = [alice, bob]

def update_model_callback_fn(res):
    net.set_weights(res.model.weights)
    pass

train_config = sy.TrainConfig(model=net, batch_size=16, epochs=1, loss='mse', optimizer='adam')

for worker in workers:
   train_config.send(worker)
   worker.run_training(update_model_callback_fn)
```

## How workers interact with Plans?

There seems to be a little bit of confusion about how plans are defined: are plans workers or tensors? Although plans are BaseWorkers there are a lot of operations performed on Plans named the same aa tensor operations: search, PlanPointers (recently removed #2050), ... but with different behavior.

Here are some ideas of possible changes in the current interface:

#### worker.fetch_plan(plan_id)

#2064 - Already implemented

Instead of searching for plans as we do for plans, a user can fetch a plan from another worker (they will get a copy of the plan).

#### worker.send(plan)

> The interest is that you can call .send() but actually this doesn't trigger a call to a remote worker, just "a promise" that it will be sent (with _send() ) as soon as the plan is called (and built)

This is good because we just build the pĺan when we need to, and potentially this will be a remote operation.

This is bad because is not an intuitive behavior and can be very confusing for the user.

Being discussed #2062.

## Plan serialization

It seems to be done: #2067

## Other questions

* Currently only non-client workers can execute plans, client workers get errors when trying to run plans since they can't register objects of their own. How do we want to deal with this? This could be a potential pain for beginners.

* What should happen if a worker wants to run locally a plan that was already built on another worker? Currently, the plan is being rebuilt.

* If the plan is executed locally should it return a pointer to the result or the result itself (tensor)? Currently .get() is being called on the pointer, so a local client would get the result itself.
* The Pysyft concept of Plan will make it possible to isolate the remote workers from the server to a certain extent. But what degree of separation is the project aiming for? Will the control over the batch selection and all these training details move to the remote workers? Or be part of the plan? - [Silvia](midokura-silvia)

### DIFF
--- a/plans_doc.md
+++ b/plans_doc.md
@@ -1,0 +1,1 @@
+This file should not be merged :smile:!


midokura-silvia:
It would be good if a plan could be completely independent of a worker, i.e. not be a worker itself. It should be like a sort of recipe. And a worker should only depend on a plan and a model to be able to run a training.

LaRiffle:
In the case we want plans to perform a complete training we will need to support if / for statements. This is not possible currently because those can't be caught by our python plan instance.

Looking at the Spark syntax, I wonder if we can try to use special methods on plans to specify logic such as `.foreach(...)` `.if(...)` etc. This would mean keep plans simple and linear but add ability to inter-connect them to create a logical graph of plans.

Here is an example of what it could look like:
```
class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.fc1 = nn.Linear(2, 3)
        self.fc2 = nn.Linear(3, 2)

    @sy.method2plan
    def forward(self, x):
        x = F.relu(self.fc1(x))
        x = self.fc2(x)
        return F.log_softmax(x, dim=0)

    @sy.method2plan
    def forward_no_softmax(self, x):
        x = F.relu(self.fc1(x))
        x = self.fc2(x)
        return x


net = Net()

net.send(device_1)
# net.forward is a plan
net.forward.send(device_1)

# existing
pointer_to_data = device_1.search('input_data')[0]
pointer_to_result = net.forward(pointer_to_data)

# future ?
training = net.forward.foreach(lambda batch: batch)
dataloader = [ptr_data for ptr_data in device_1.search('input_data')]
training(dataloader)

# or just
dataloader = [ptr_data for ptr_data in device_1.search('input_data')]
training = net.forward.foreach(dataloader)

# future ? if == provided
@sy.func2plan
def is_encrypted(data):
    return data.is_encrypted()

plan = (
    net.forward_no_softmax
    .providing(is_encrypted)
    .otherwise(net.forward)
)
```

Any comments?

midokura-silvia:
@LaRiffle 
The main concern I have is that the training is being controlled (started and supervised) by the local wrapper of the remote worker. This makes it difficult to parallelize the execution of a large number of workers. The goal should be to package the model and the plan in the beginning, send it to the remote worker and then wait (do something else) until it comes back with the result.

LaRiffle:
@midokura-silvia 
I hear your concern, it is super important to parallelize and have async modes:
@iamtrask is working on async for all kinds of operations: namely you seed an op, have a pre-reference to the result (already done) but don't wait for the remote execution to run.
There an ongoing PR to send plans to multiple workers (ie .send(alice, bob) ) #2051 

I think this should help a lot!

Regarding the logical graph of plan I explained above, actually it would be sent all at once: if you send a plan, all its logical dependencies should be sent, just like we built it all at once too. You should still handle a single object

mari-linhares:
```
## SERVER  (Scheduler)
class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.fc1 = nn.Linear(2, 3)
        self.fc2 = nn.Linear(3, 2)

    @sy.method2plan
    def forward(self, x):
        x = F.relu(self.fc1(x))
        x = self.fc2(x)
        return F.log_softmax(x, dim=0)

    @sy.method2plan
    def forward_no_softmax(self, x):
        x = F.relu(self.fc1(x))
        x = self.fc2(x)
        return x

net = Net()

net.send(device_1)
net.send(device_2)
# net.forward is a plan
net.forward.send(device_1)
net.forward.send(device_2)
# send also batch_size, epochs, loss

def callback_fn(res):
    # update model...
    pass

for _ in range(epochs):
  device_1.run_training(callback_fn)
  device_2.run_training(callback_fn)


# existing
pointer_to_data = device_1.search('input_data')[0]
pointer_to_result = net.forward(pointer_to_data)

# future ?
training = net.forward.foreach(lambda batch: batch)
dataloader = [ptr_data for ptr_data in device_1.search('input_data')]
training(dataloader)

# or just
dataloader = [ptr_data for ptr_data in device_1.search('input_data')]
training = net.forward.foreach(dataloader)

# future ? if == provided
@sy.func2plan
def is_encrypted(data):
    return data.is_encrypted()

plan = (
    net.forward_no_softmax
    .providing(is_encrypted)
    .otherwise(net.forward)
)

```

iamtrask:
I'm a huge fan of decoupling TrainConfig from Plan specifically - this is great! Let's do it!

LaRiffle:
Yes this is much better, I like it a lot!

mari-linhares:
I think we're all on the same page about plans.

I'm closing this in favor of tutorial 9 - introduction to plans. The tutorials should cover all main use cases of plans.

PR #2073, opened 2019-04-16
Improvements for Grid

iamtrask:
I've made several improvements which are necessary to get Grid working on Flask/Heroku. 

- I fixed the string serde bug (string -> bin)
- added full serialization of virtual workers


syft/serde.py
```diff
@@ -69,7 +69,13 @@
 
 
 # High Level Public Functions (these are the ones you use)
-def serialize(obj: object, simplified=False) -> bin:
+def serialize(
+    obj: object,
+    simplified=False,
+    force_no_compression=False,
```
robert-wagner: Pleas add explanation for what these parameters mean (also what to do when they conflict)

PR #2074, opened 2019-04-17
Hook

tqhdesilva:
- Moved flake8 pre-commit args to .flake8 config.
- added explicit revision to black pre-commit hook
- removed `language_version = python3.6` from black pre-commit hook. This breaks pre-commit hooks unless you have python3.6 installed.  Not sure if there is a good way to require >=python3.6, or if we should update contributing guide to require 3.6 only.


iamtrask:
Excellent contribution!

tqhdesilva:
Not sure if `flake8 . --count --exit-zero --statistics` should be `flake8 . --count --exit-zero --statistics --select=E,F,W,C90`. Is that just supposed to display flake8's default errors as warnings in travis?

tqhdesilva:
@robert-wagner could you take another look?

robert-wagner:
Great work @tqhdesilva !

PR #2084, opened 2019-04-21
 Fixed a timeout issue occuring on raspberry PIs or low-power devices

DanyEle:
When wanting to train a model via websockets on low-power devices with a slow CPU (e.g.:Raspberry PIs or mobile device), a timeout exception occurrs, as shown in the following pictures.

error_broken_pipe_client

issue_pysyft

I fixed the issue by removing the default ping_timeout and close_timeout timers set to 20 seconds for the workers.


kamathhrishi:
@DanyEle Could you run the command `black syft` and commit the change?

DanyEle:
Ran the command `black syft` . Should be good to go.

DanyEle:
Added two more fixes on the server-side, as the server may also time out when handling slow clients.

iamtrask:
OMG Raspberry PI!!! Any chance you'd be willing to write a blogpost tutorial about your experience with this?!!?!

syft/workers/websocket_client.py
```diff
@@ -38,7 +38,8 @@ def __init__(
 
         # creates the connection with the server which gets held open until the
         # WebsocketClientWorker is garbage collected.
-        self.ws = websocket.create_connection(self.uri, max_size=None)
+        # Daniele Gadler: Also avoid the server from timing out on the server-side
```
mari-linhares: @DanyEle can you remove the author name from the line? It's not a code pattern in syft to keep the author's name in a comment.

syft/workers/websocket_client.py
```diff
@@ -38,7 +38,8 @@ def __init__(
...
+        self.ws = websocket.create_connection(self.uri, max_size=None, timeout=9999999)
```
mari-linhares: create a TIMEOUT constant instead of using 9999999

syft/workers/websocket_client.py
```diff
@@ -51,6 +52,7 @@ def search(self, *query):
         return sy.serde.deserialize(response)
 
     def _send_msg(self, message: bin, location) -> bin:
+        self.ws.settimeout(9999999)
```
mari-linhares: use constant

syft/workers/websocket_client.py
```diff
@@ -70,7 +72,8 @@ def _recv_msg(self, message: bin) -> bin:
             logger.warning("Websocket connection closed (worker: %s)", self.id)
             self.ws.shutdown()
             time.sleep(1)
-            self.ws = websocket.create_connection(self.uri, max_size=None)
+            # Daniele Gadler: also avoid timing out on the server-side
+            self.ws = websocket.create_connection(self.uri, max_size=None, timeout=9999999)
```
mari-linhares: use constant

syft/workers/websocket_client.py
```diff
@@ -70,7 +72,8 @@ def _recv_msg(self, message: bin) -> bin:
             logger.warning("Websocket connection closed (worker: %s)", self.id)
             self.ws.shutdown()
             time.sleep(1)
-            self.ws = websocket.create_connection(self.uri, max_size=None)
+            # Daniele Gadler: also avoid timing out on the server-side
```
mari-linhares: remove author's name from comment

syft/workers/websocket_server.py
```diff
@@ -120,6 +120,15 @@ def __init__(
     def start(self):
         """Start the server"""
 
-        start_server = websockets.serve(self._handler, self.host, self.port, max_size=None)
+        # Daniele Gadler: fix the timeout issues at the client side
```
mari-linhares: remove comment

DanyEle:
> OMG Raspberry PI!!! Any chance you'd be willing to write a blogpost tutorial about your experience with this?!!?!

Hey, I would be really glad to do that if the whole training procedure were working. For now, I managed to fix the data distribution part, as the timeout issues occurred there. However, for some weird reason the training stops at the first training iteration. I tried waiting for a few hours yesterday, however nothing happens after the loss in iteration (1) is computed, as you can see:

My main reason to use raspberry PIs is for text processing however, so I am going to try and see if I manage to get them working for that purpose. I saw there is a nice example about that in the examples/advances tutorials: "Federated Word Vectors". I will try and adapt that code to make it work in a federated way.

syft/workers/websocket_client.py
DanyEle: Done!

syft/workers/websocket_server.py
DanyEle: Done

syft/workers/websocket_client.py
DanyEle: Done

syft/workers/websocket_client.py
DanyEle: Done

syft/workers/websocket_client.py
```diff
@@ -15,6 +15,9 @@
 logger = logging.getLogger(__name__)
 
 
+TIMEOUT_INTERVAL = 9999999
```
robert-wagner: what is the default for this if nothing is passed (also could you add a justification for why this was chosen)

syft/workers/websocket_client.py
```diff
@@ -51,6 +55,7 @@ def search(self, *query):
         return sy.serde.deserialize(response)
 
     def _send_msg(self, message: bin, location) -> bin:
+        self.ws.settimeout(TIMEOUT_INTERVAL)
```
robert-wagner: This should not be here (if you see the message below this method should never be called)

syft/workers/websocket_client.py
DanyEle: This value was just picked to avoid sockets from timing out. 9999999 seconds looked like a sufficiently large amount of time to me. If nothing is passed or no timeout is set, then you can refer to the default values listed in the respective documentation for sockets:

https://docs.python.org/3/library/socket.html
If no timeout is supplied, the global default timeout setting returned by getdefaulttimeout() is used.

And for websockets:

https://websockets.readthedocs.io/en/stable/api.html

ping_timeout=20, close_timeout=10 are the default values.

syft/workers/websocket_client.py
DanyEle: You are right. It should be removed: i just added it to address any possible call to websockets.

PR #2085, opened 2019-04-23
TrainConfig draft

mari-linhares:
- [x] TrainConfig draft
- [x] Serialization for TrainConfig
- [x] Tests

### DIFF
--- a/test/federated/test_train_config.py
+++ b/test/federated/test_train_config.py
@@ -46,5 +46,5 @@ def test_send_train_config(hook):
     # Run forward pass and calculate loss
     pointer_to_pred = train_config.forward_plan(pointer_to_data)
     pointer_to_loss = train_config.loss_plan(pointer_to_pred.wrap(), pointer_to_target)
-    pointer_to_loss.get()
-    assert True
+    loss = pointer_to_loss.get()
+    assert loss > 0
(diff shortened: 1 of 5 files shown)


test/federated/test_train_config.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # Get a pointer to federated device data
+    pointer_to_data = federated_device.search("data")[0]
+    pointer_to_target = federated_device.search("target")[0]
+
+    # Run forward pass and calculate loss
+    pointer_to_pred = train_config.forward_plan(pointer_to_data)
+    pointer_to_loss = train_config.loss_plan(pointer_to_pred.wrap(), pointer_to_target)
+    pointer_to_loss.get()
+    assert True
```
jlebensold: can we use a better assertion here? Given that we're calculating a `loss`, can we do an assertLess, e.g. https://github.com/tensorflow/federated/blob/master/tensorflow_federated/python/learning/federated_averaging_test.py#L135

mari-linhares: @jlebensold actually since I'm not doing any training I don't think it makes a lot of sense to use assertLess. But I can check if the value is a scalar. What do you think?

jlebensold: that works too. `assert True` basically just means that the code doesn't crash... but not much else

mari-linhares: agree, done.

PR #2087, opened 2019-04-23
Add WebSocket over TLS

IonesioJunior:
Improve Secure level of data flow between remote workers adding WebSocket over TLS.
Benefits:
- Authenticate component identity by certificates.
- Provides end-to-end encryption

PS: These changes do not modify or render infeasible the old workers' behavior.


syft/workers/websocket_client.py
```diff
@@ -34,11 +36,18 @@ def __init__(
...
-        self.uri = f"ws://{self.host}:{self.port}"
 
         # creates the connection with the server which gets held open until the
         # WebsocketClientWorker is garbage collected.
-        self.ws = websocket.create_connection(self.uri, max_size=None)
+        # Secure flag adds a secure layer applying cryptography and authentication
+        self.uri = f"ws://{self.host}:{self.port}"
+        if secure:
+            self.uri = f"wss://{self.host}:{self.port}"
+            ssl_settings = {"cert_reqs": ssl.CERT_NONE}
```
IonesioJunior: This line should only be used in development environments.
It allows to work with self-signed certificates facilitating the development.

syft/workers/websocket_server.py
```diff
@@ -119,7 +124,15 @@ def __init__(
 
     def start(self):
         """Start the server"""
-
-        start_server = websockets.serve(self._handler, self.host, self.port, max_size=None)
+        # Secure behavior: adds a secure layer applying cryptography and authentication
+        if not (self.cert is None) and not (self.key is None):
```
jlebensold: Overall I think this PR looks ok. This boolean is a bit confusing. since we have `secure` in the top, can we use that down here too and `raise` if the settings aren't right. Reading this I get the sense that a user could inadvertently run the server in an insecure mode without realizing it.

IonesioJunior: The secure flag doesn't exist on the server side, but can be implemented if this makes the idea of ​​a secure server more explicit.

**kwargs at client side**
`
kwargs_websocket = {"host": "localhost",
                    "hook": hook,
                    "verbose": True,
                    "secure": True }
`
**kwargs at server side**
`
kwargs = {
    "id": args.id,
    "host": args.host,
    "port": args.port,
    "hook": hook,
    "verbose": True,
    "cert_path" : "localhost.cert",
    "key_path" : "localhost.key"
}
`


Should I add a secure flag to websocket server too?

iamtrask:
This is a really high value contribution - thank you @IonesioJunior

syft/workers/websocket_server.py
robert-wagner: If the server side has the secure flag can we make it reject all non secure connections?

IonesioJunior: By default, the websocket library already rejects connections that do not implement the pre-established security protocol (if it is established). The security protocol is defined when creating the ssl context `ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)`

syft/workers/websocket_client.py
iamtrask: FYI - over the last little bit I've been working on a Heroku based flask app worker - when the time comes I'd love to pick your brain on how to integrate these - especially the security pieces.

IonesioJunior: I would be happy to help. :smiley:

iamtrask: Excellent!

PR #2095, opened 2019-04-23
Continuing Mcleonard autograd tensor

iamtrask:
This PR is a continuation of https://github.com/OpenMined/PySyft/pull/1942


syft/frameworks/torch/tensors/interpreters/autograd.py
```diff
@@ -0,0 +1,139 @@
...
+from . import gradients
+
+
+def backwards_grad(grad_fn, in_grad=None):
+    back_grad = grad_fn(in_grad)
+    for next_grad_fn, next_grad in zip(grad_fn.next_functions, back_grad):
+        backwards_grad(next_grad_fn, next_grad)
+
+
+class AutogradTensor(AbstractTensor):
```
midokura-silvia: The AutogradTensor seems to be missing a get() function. Could be something like:
```
    def get(self, deregister_ptr: bool = True):
        if self.grad is not None and hasattr(self.grad, 'child'):
            self.grad.get(deregister_ptr)

        self.child.get(deregister_ptr)
        return self
```

test/torch/tensors/test_autograd.py
```diff
@@ -0,0 +1,149 @@
...
+
+    assert (a.grad.get() == a_torch.grad).all()
+    assert (b.grad.get() == b_torch.grad).all()
+    
+    a.get();
+    b.get();
+    c.get();
+
+
+def test_sqrt_backwards(workers):
```
midokura-silvia: test_sqrt_backwards fails with:
E   TypeError: can not serialize 'AutogradTensor' object
msgpack/_packer.pyx:279: TypeError

syft/frameworks/torch/tensors/interpreters/autograd.py
mari-linhares: I've added this in the last commits, thanks Silvia!

midokura-silvia:
The unit test issue is solved in #2143. 
The hook had already stored values for 'Tensor.backward', as backward() is used in the unit test test/syft_torch/test_hook.py::test_RNN_grad_set_backpropagation.

PR #2096, opened 2019-04-24
Added generic pointers and new callable pointers

### LINKED ISSUE #2088: Make Pointer not specific to Tensors
As we start sending more objects than just tensors (so TrainConfig, Plan), we would like to have Pointer be more general and no only associated to Tensors. It wouldn't change much from what it is: have a location and forwards commands, but would be way more elegant!

A Pointer could point at Tensors of course, but also at Plans, at Config, or at anything which is stored in the remote worker resgitry (ie bob._objects)

midokura-silvia:
In order to obtain generic pointers, I introduced an AbstractObject class and derived the GenericPointer class from this class. 

Shall we also move the abstract.py file outside of the tensors.interpreters folder?
This is just a start and will probably need more adaptations. Please comment.

fixes #2088


iamtrask:
This is a really good idea - we already have pointers to all sorts of things which don't have any meaningful inheritance structure. A good first use case for this might be pointers to numpy arrays or some other framework object (numpy arrays are suitably similar to torch tensors but are different enough to test the abstraction)

LaRiffle:
@iamtrask pointers to np array would be a good idea! We had this idea with Sylvia also for the TrainConfig object Marianne is building: we are now sending many things so it's worth having generic pointers

midokura-silvia:
The first usecase of the new pointers is to send a torch.jit.ScriptModule to a worker.

syft/frameworks/torch/pointers/callable_pointer.py
```diff
@@ -0,0 +1,129 @@
...
+            object used to point to on a remote machine.
+
+        TODO: add param get_copy which doesn't destroy remote if true.
+        """
+
+        if self.point_to_attr is not None:
+
+            raise exceptions.CannotRequestTensorAttribute(
+                "You called .get() on a pointer to"
+                " a tensor attribute. This is not yet"
```
mari-linhares: will this used only for tensors? If not rename tensor to object?

syft/frameworks/torch/tensors/interpreters/abstract.py
```diff
@@ -7,9 +7,9 @@
 import weakref
 
 
-class AbstractTensor(ABC):
+class AbstractObject(ABC):
```
mari-linhares: rename abstractTensor -> abstractObject in the docstrings.

syft/workers/base.py
```diff
@@ -324,14 +323,25 @@ def send(
                 local_autograd=local_autograd,
                 preinitialize_grad=preinitialize_grad,
             )
+        elif isinstance(obj, torch.jit.ScriptModule):
```
mari-linhares: I don't think this sould be in this PR.

test/torch/pointers/test_callable_pointer.py
```diff
@@ -0,0 +1,59 @@
+import torch
+from syft.frameworks.torch.pointers import callable_pointer
+
+
+def test_create_callable_pointer(workers):
```
mari-linhares: can you add a test case where you call the callable pointer?

syft/workers/base.py
```diff
@@ -324,14 +323,25 @@ def send(
                 local_autograd=local_autograd,
                 preinitialize_grad=preinitialize_grad,
             )
+        elif isinstance(obj, pointers.ObjectWrapper):
```
mari-linhares: I don't like this checking, considering that we're creating a generic pointer for having more types of pointers in the future this can get ugly :(.

It would be better if the object wrapper was responsible by creating its pointer (obj.create_pointer) since I don't think for an objectwrapper we'll always have a callable_pointer, right?

During the object wrapper creation (considering this will be done internally) we can define the pointer for it.

syft/workers/base.py
```diff
@@ -348,6 +358,8 @@ def execute_command(self, message: tuple) -> PointerTensor:
         command_name = command_name
         # Handle methods
         if _self is not None:
+            if type(_self) == int:
```
mari-linhares: Is this really needed? If so we need to find a better way to deal with it.

midokura-silvia: This is needed to provide an id of an object, whose method will be called. This is the easiest way to reference an object at the location. And the only thing known by the pointer.

syft/__init__.py
```diff
@@ -27,8 +27,8 @@
 from syft.frameworks.torch.tensors.decorators import LoggingTensor
 from syft.frameworks.torch.tensors.interpreters import AdditiveSharingTensor
 from syft.frameworks.torch.tensors.interpreters import MultiPointerTensor
-from syft.frameworks.torch.tensors.interpreters import PointerTensor
 from syft.frameworks.torch.tensors.interpreters import AutogradTensor
+from syft.frameworks.torch.pointers import ObjectPointer, CallablePointer, PointerTensor
```
robert-wagner: Please put these on separate lines

syft/frameworks/torch/pointers/callable_pointer.py
```diff
@@ -0,0 +1,80 @@
...
+from typing import Union
+from typing import TYPE_CHECKING
+
+# this if statement avoids circular imports
+if TYPE_CHECKING:
+    from syft.workers import AbstractWorker
+
+
+class CallablePointer(object_pointer.ObjectPointer):
+    """ A class of pointers that are callable
```
robert-wagner: Could you add to this docstring how they would be used

syft/frameworks/torch/pointers/callable_pointer.py
```diff
@@ -0,0 +1,80 @@
...
+    location: "AbstractWorker",
+    id: (str or int),
+    id_at_location: (str or int),
+    owner: "AbstractWorker",
+    tags,
+    description,
+    garbage_collect_data: bool = True,
+    register_pointer: bool = True,
+) -> object_pointer.ObjectPointer:
+    """Creates a callable pointer to the object, identified by the pair (location, id_at_location).
```
robert-wagner: comma is not needed between object and identified

syft/frameworks/torch/pointers/callable_pointer.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Note, that there is no check whether an object with this id exists at the location.
```
robert-wagner: Should this check be added?

syft/frameworks/torch/pointers/callable_pointer.py
```diff
@@ -0,0 +1,80 @@
...
+# this if statement avoids circular imports
+if TYPE_CHECKING:
+    from syft.workers import AbstractWorker
+
+
+class CallablePointer(object_pointer.ObjectPointer):
+    """ A class of pointers that are callable
+    """
+
+    def __init__(
```
robert-wagner: Add a docstring

syft/frameworks/torch/pointers/callable_pointer.py
```diff
@@ -0,0 +1,80 @@
...
+    owner: "AbstractWorker",
+    tags,
+    description,
+    garbage_collect_data: bool = True,
+    register_pointer: bool = True,
+) -> object_pointer.ObjectPointer:
+    """Creates a callable pointer to the object, identified by the pair (location, id_at_location).
+
+    Note, that there is no check whether an object with this id exists at the location.
+    """
```
robert-wagner: Please add descriptions for each of the parameters to the docstring as well as what it returns

syft/frameworks/torch/pointers/object_wrapper.py
```diff
@@ -0,0 +1,52 @@
...
+from typing import List
+from typing import Union
+from typing import TYPE_CHECKING
+
+# this if statement avoids circular imports between base.py and pointer.py
+if TYPE_CHECKING:
+    from syft.workers import AbstractWorker
+
+
+class ObjectWrapper:
```
robert-wagner: Please add docstrings for class and all methods (Why is this class used?)

syft/frameworks/torch/pointers/pointer_tensor.py
```diff
@@ -0,0 +1,181 @@
...
+     >>> bob = sy.VirtualWorker(id="bob")
+     >>> x = sy.Tensor([1,2,3,4,5])
+     >>> y = sy.Tensor([1,1,1,1,1])
+     >>> x_ptr = x.send(bob) # returns a PointerTensor, sends tensor to Bob
+     >>> y_ptr = y.send(bob) # returns a PointerTensor, sends tensor to Bob
+     >>> # executes command on Bob's machine
+     >>> z_ptr = x_ptr + y_ptr
+    """
+
+    def __init__(
```
robert-wagner: Docstring?

syft/frameworks/torch/pointers/pointer_tensor.py
```diff
@@ -0,0 +1,181 @@
...
+        if self._shape is None:
+            self._shape = self.get_shape()
+
+        return self._shape
+
+    @shape.setter
+    def shape(self, new_shape):
+        self._shape = new_shape
+
+    def get(self, deregister_ptr: bool = True):
```
robert-wagner: how does this differ in logic from the one in ObjectPointer

syft/frameworks/torch/tensors/interpreters/autograd.py
```diff
@@ -3,7 +3,7 @@
 
 import syft
 from syft.frameworks.torch.tensors.interpreters.abstract import AbstractTensor
-from syft.frameworks.torch.tensors.interpreters import PointerTensor
+from syft.frameworks.torch.pointers.pointer_tensor import PointerTensor
```
robert-wagner: We should be able to import directly from pointers due to the init

syft/frameworks/torch/pointers/callable_pointer.py
midokura-silvia: It can make sense in some cases. It could be an argument check_existence.

midokura-silvia: We can add this check in a separate pull request. It needs either the adding of a message or the modifications of the train-config branch to call an arbitrary function of the worker.

syft/frameworks/torch/pointers/pointer_tensor.py
midokura-silvia: I refactored the code so that the difference is clearly visible.

syft/frameworks/torch/pointers/callable_pointer.py
```diff
@@ -0,0 +1,117 @@
...
+                registered if register is set to True. Note that this is
+                different from the location parameter that specifies where the
+                pointer points to.
+            id: An optional string or integer id of the PointerTensor.
+            garbage_collect_data: If true (default), delete the remote object when the
+                pointer is deleted.
+            point_to_attr: string which can tell a pointer to not point directly to\
+                an object, but to point to an attribute of that object such as .child or
+                .grad. Note the string can be a chain (i.e., .child.child.child or
+                .grad.child.child). Defaults yo None, which means don't point to any attr,
```
robert-wagner: defaults yo None -> defaults to None

PR #2098, opened 2019-04-25
Plans should not be workers

mari-linhares:
As mentioned by @midokura-silvia at #2068. Plans should not be workers. This change addresses this and also:

* Create a new abstraction called: ObjectStorage, which consist of a wrapper class to _objects. Plans now inherit from ObjectStorage.

* **BaseWorkers inherit from ObjectStorage**.

* plan.plan_blueprint -> plan.blueprint.

* Fixes docstring style of some methods + typing.

**Important**

> *3 test cases were removed, why???*

Since plans are not workers anymore they are not being hooked, and so the blueprint attribute does not exist and a plan cannot be re-built locally after being built remotely (which was the case for the last tests)


LaRiffle:
Conflict fixed here, feel free to integrate changes here or not
https://github.com/OpenMined/PySyft/pull/2099

iamtrask:
Continuing with this PR in https://github.com/OpenMined/PySyft/pull/2099

PR #2099, opened 2019-04-25
Plans should not be workers #2098 (fix conflict)

LaRiffle:
Fixing conflict in PR "Plans should not be workers #2098" as a non branch owner.

As mentioned by @midokura-silvia at #2068. Plans should not be workers. This change addresses this and also:

* Create a new abstraction called: ObjectStorage, which consist of a wrapper class to _objects. Plans now inherit from ObjectStorage.

* **BaseWorkers inherit from ObjectStorage**.

* plan.plan_blueprint -> plan.blueprint.

* Fixes docstring style of some methods + typing.

* Creates a new directory called syft/generics to store generic objects used across syft.

* Creates a new directory called syft/federated where plans live.

**Important**

> *3 test cases were removed, why???*

Since plans are not workers anymore they are not being hooked, and so the blueprint attribute does not exist and a plan cannot be re-built locally after being built remotely (which was the case for the last tests)


test/workers/test_plan.py
```diff
@@ -103,78 +103,78 @@ def forward(self, x):
     pointer_to_result = net(pointer_to_data)
     pointer_to_result.get()
 
+# def test_multiple_workers(workers):
```
iamtrask: Is there a reason these tests were removed? Should we replace them with other tests?

LaRiffle: Yes, these tests require non-plan-creators to build plans
here is our discussion with Marianne:

Marianne Monteiro 
another quick question (take your time to answer these btw): in the current implementation would it be possible to send a plan to a remote worker? apparently in the current implementation to be able to actually run a plan, you need to have direct/indirect access to the blueprint. Indirect access would be some other virtual worker has access to it and ask you to register the objects you need, but this operation, as far as I know, can not be done remotely right now, right?


Marianne Monteiro 
I was able to make plans work (not being a worker), for the 3 first test cases (https://github.com/OpenMined/PySyft/blob/dev/test/workers/test_plan.py) the 3 last ones are not working, because since plans are not workers anymore they are not being hooked, and so the blueprint attribute does not exist and a plan can not be built remotely.

Théo Ryffel 
I'm not sure Iv'e fully understood the question 
To run a plan, you need to have access to a readable_plan, not necessarily the blueprint. But to build it it's right, you need to hold the blueprint.
So it would seem reasonable that the creator of the plan (which has the blueprint) should be the one to build it


Théo Ryffel 
There I agree that the 3 last tests should not be possible now!

mari-linhares: @iamtrask  @LaRiffle just fixed these tests on PR #2108

PR #2102, opened 2019-04-25
Fix minor bug in class FederatedDataset

tushar-semwal:
Takes care of workers with integer ids.


tushar-semwal:
I am honoured. :)

PR #2108, opened 2019-04-28
Uncomment plans tests + a websocket test for plans

mari-linhares:


syft/federated/plan.py
```diff
@@ -90,9 +99,9 @@ def __init__(
 
         # Info about the plan stored
         self.plan = list()
-        self.readable_plan = list()
-        self.arg_ids = list()
-        self.result_ids = list()
+        self.readable_plan = readable_plan if readable_plan else []
```
robert-wagner: Could you add an is None here. Technically everything with a false truth value would be converted to [] here

syft/federated/plan.py
```diff
@@ -185,18 +194,43 @@ def find_location(self, args):
...
+        to_ids: List[Union[str, int]],
+        from_worker: Union[str, int] = None,
+        to_worker: Union[str, int] = None,
+    ):
         """Replaces pairs of tensor ids in the plan stored.
 
         Args:
-            from_ids: the left part of the pair: ids to change.
-            to_ids: the right part of the pair: ids to replace with.
+            from_ids: The left part of the pair: ids to change.
```
robert-wagner: The right and left part of the pair are unclear. Can you clarify this morning/just remove that part

syft/federated/plan.py
```diff
@@ -185,18 +194,43 @@ def find_location(self, args):
...
+            to_ids: The right part of the pair: ids to replace with.
+            from_worker: The previous worker that built the plan.
+            to_worker: The new worker that is running the plan.
         """
+        if not from_worker:
```
robert-wagner: What if the previous worker had id 0

syft/federated/plan.py
```diff
@@ -185,18 +194,43 @@ def find_location(self, args):
...
+            from_worker = self.id
+        if not to_worker:
```
robert-wagner: Ditto

robert-wagner:
Reviewed

test/federated/test_plan.py
```diff
@@ -189,3 +198,33 @@ def my_plan(data):
 
     x = th.tensor([-1, 2, 3])
     assert (deserialized_plan(x) == th.tensor([-42, 24, 46])).all()
+
+
+def test_plan_execute_remotelly(hook, start_proc):
```
midokura-silvia: typo in word remotely

syft/federated/plan.py
midokura-silvia: I propose to remove the reference to pairs. There are two lists and no pairs. Maybe check that the lists have the same length.

PR #2113, opened 2019-04-29
FederatedClient draft

### LINKED ISSUE #2062: plan.send() implementation is not intuitive

mari-linhares:
Work in progress.

- [x] Question: Do we need to share the actual model to another worker or just the parameters? If we need to, how do we do this?

> Answer: we only want to share the parameters... parameters + plan == everything needed.

- [x] How to build the forward method locally in the blueprint owner? Currently, I get the error: `TypeError: __bool__ should return bool, returned Tensor`

> Answer: right now you need to explicitly send the model to the local worker. This is still work in progress since is not clear how to do this correctly.

- [x] Support remote method execution: implemented by @midokura-silvia :100:.

- [ ] Support remote method execution that returns a tensor.

- [ ] Build plans automatically: placeholders / input_data_args. This fixes #2062.

- [ ] Why do we need to call model.send(me) and model.get(), we should not need to do this.

- [ ] Test remote execution.

- [ ] Add experimental notebook with federated training loop working.

- [ ] Implement TrainConfig datasets.

@midokura-silvia


iamtrask:
We only want to share the parameters... parameters + plan == everything needed :). If by "sharing the entire model" you mean actually sending a python object we're specifically trying to avoid this for security reasons.

mari-linhares:
Thanks, Trask!

LaRiffle:
From what I see, `torch.nn.functional.linear` is correctly hooked and the command is sent to a worker.
But then this worker stills call this command on pointers (it should call it on tensors), and this is where it fails because .dim() is hooked so `if input.dim() == 2` tries to assess if a pointer tensor is true or false.

LaRiffle:
So when you build your plan you have this odd thing: you as a plan order to build the plan on args which from your point of view are pointers owned by "me" to yourself.
At that point and for some reason, the call to F.linear is made on these pointers but is not hooked to be sent to the pointer location (you as a plan), therefore you end up doing `if input.dim() == 2` on pointers

LaRiffle:
Oh I got it! The reason it believes F.linear must be computed locally is because the weights of the nets are local! If you check https://github.com/OpenMined/PySyft/blob/dev/examples/tutorials/Part%209%20-%20Introduction%20to%20Plan.ipynb cell 13 you observe that we also send the weights of the model to the plan.
So this is what is missing here

midokura-silvia:
Where shall we put the function that implements the training loop? In my opinion the worker could be a possible location. 
Is it okay to add a way of calling a method of the remote worker? I am just thinking of modifying the execute_command() function to allow calling methods of self additionally to calling methods of pointers.
This way we don't need to send the implementation of the training loop. Instead we have a standard implementation that is configured by the TrainConfig. 
It could also be used to obtain information from the remote worker as returning the number of objects or their string representation.

mari-linhares:
Thanks for having a look @LaRiffle, the funny thing is that I see this error when I try to build the plan locally. When I send the TrainConfig it actually sends the model weights and the plan.

mari-linhares:
@midokura-silvia 

> Where shall we put the function that implements the training loop? In my opinion the worker could be a possible location.

Yes. This federated client abstraction is just so we can have all the training loop related logic living in a single place. If you look at base in this PR it inherits from FederatedClient.

> Is it okay to add a way of calling a method of the remote worker? I am just thinking of modifying the execute_command() function to allow calling methods of self additionally to calling methods of pointers.

Yes, that's the plan!

LaRiffle:
@mari-linhares Yes because the model is added to the args in `plan__call__` (through self.self), and those args are sent to the correct worker in the non-local setting, while they are not in the local case where you do local building because of the `if isinstance(arg, torch.Tensor):` clause in build_plan

mari-linhares:
I see, thanks @LaRiffle, everything makes a lot more sense now!

midokura-silvia:
Could we avoid deriving the BaseWorker from FederatedClient? 

The WebsocketClientWorker derives from it and doesn't need this functionality. 
VirtualWorker and WebsocketServerWorker (and WebsocketIOServerWorker) could derive directly from FederatedClient.

Any plans on how to set the dataset variable in the federated client?

mari-linhares:
> Could we avoid deriving the BaseWorker from FederatedClient?

Yes, nice catch.

> Any plans on how to set the dataset variable in the federated client?

Is not yet implemented I imagine FederatedClient could have a method to set the dataset. The data should be local. You would probably initialize when you start a WebsocketServerWorker.

robert-wagner:
> From what I see, `torch.nn.functional.linear` is correctly hooked and the command is sent to a worker.
> But then this worker stills call this command on pointers (it should call it on tensors), and this is where it fails because .dim() is hooked so `if input.dim() == 2` tries to assess if a pointer tensor is true or false.

Hey @LaRiffle @mari-linhares input.dim() called on a pointer tensor returns an int

PR #2124, opened 2019-05-06
Rename Websocket workers.

### LINKED ISSUE #2120: Rename WebsocketClientWorker and WebsocketServerWorker
I would propose to rename the two classes as follows:

WebsocketClientWorker => WebsocketRemoteProxy or WebsocketRemoteInterface or WSRemote...  
WebsocketServerWorker => WebsocketRemoteWorker

"Remote" could also be replaced by the word client, e.g. WebsocketClientProxy and WebsocketClientWorker

The idea is to avoid the words client and server. Currently they are used as given by their roles in the websocket connection. However, when talking about federated learning, we often also talk about client/server but with another meaning. In the federated learning setting, the two classes WebsocketClientWorker and WebsocketServerWorker form together a client that can participate in a training.

midokura-silvia:
ClientWorker => ClientProxy
ClientServer => ClientWorker
Note that the meaning of client worker has changed.

fixes #2120


examples/tutorials/advanced/websockets-example-MNIST/run_websocket_client_worker.py
```diff
@@ -1,7 +1,7 @@
 from multiprocessing import Process
 
 import syft as sy
-from syft.workers import WebsocketServerWorker
+from syft.workers import WebsocketClientWorker
```
robert-wagner: I do not like this name change. Traditionally the client is the one who calls commands. I would prefer if we moved away from the terms client and server regardless as they seem to add confusion. That being said I would recommend that we spend more time brainstorming names before we change them

midokura-silvia: @robert-wagner Thanks for your feedback. 
What names do you propose? And whose opinion should be heard?

mari-linhares:
I'll close this PR for now (we can re-open when we have a decision about the names).

We can keep the discussion at #2020.

Issue #1989, opened 2019-03-12
Increase test coverage to 100%


robert-wagner:
Currently we have 95% coverage. Ideal that coverage would be 100%

Ankk98:
I would like to start my contribution to OpenMined with this issue.

Ankk98:
This issue is assigned or I can work on it?

robert-wagner:
@Ankk98 Feel free to work on it! I will be iterating on it over time but would love any help.

Ankit-Dhankhar:
@robert-wagner can we integrate https://codecov.io/ for coverage test during the pull request integrated in travis.ci? 
Most of organisation use this for check coverage of pull request.

PR #2125, opened 2019-05-07
added testing of plan functions, small bugfix

midokura-silvia:
Small bugfix in function Plan.replace_worker_ids to support integer ids
Added tests

### DIFF
--- a/syft/federated/plan.py
+++ b/syft/federated/plan.py
@@ -254,12 +254,21 @@ class Plan(ObjectStorage):
             from_worker_id: Id of the the replaced worker.
             to_worker_id: Id of the new worker.
         """
-        for from_id, to_id in [
-            (from_worker_id, to_worker_id),
-            (from_worker_id.encode(), to_worker_id.encode()),
-        ]:
+
+        id_pairs = [(from_worker_id, to_worker_id)]
+        if type(from_worker_id) == str:
+            to_worker_id_encoded = (
+                to_worker_id.encode() if type(to_worker_id) == str else to_worker_id
+            )
+            id_pairs.append((from_worker_id.encode(), to_worker_id_encoded))
+
+        for id_pair in id_pairs:
             self.readable_plan = Plan._replace_message_ids(
-                obj=self.readable_plan, change_id=-1, to_id=-1, from_worker=from_id, to_worker=to_id
+                obj=self.readable_plan,
+                change_id=-1,
+                to_id=-1,
+                from_worker=id_pair[0],
+                to_worker=id_pair[1],
             )
 
     @staticmethod
--- a/test/federated/test_plan.py
+++ b/test/federated/test_plan.py
@@ -1,5 +1,3 @@
-import random
-
 import syft as sy
 
 import torch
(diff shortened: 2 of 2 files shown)


syft/federated/plan.py
```diff
@@ -254,12 +254,21 @@ def replace_worker_ids(self, from_worker_id: Union[str, int], to_worker_id: Unio
...
         """
-        for from_id, to_id in [
-            (from_worker_id, to_worker_id),
-            (from_worker_id.encode(), to_worker_id.encode()),
-        ]:
+
+        id_pairs = [(from_worker_id, to_worker_id)]
+        if type(from_worker_id) == str:
+            to_worker_id_encoded = (
+                to_worker_id.encode() if type(to_worker_id) == str else to_worker_id
```
mari-linhares: Can you cover all 4 possibilities when dealing with strings?

from_worker, to_worker
from_worker, to_worker.encode()
from_worker.encode(), to_worker
from_worker.encode(), to_worker.encode()

test/federated/test_plan.py
```diff
@@ -227,4 +224,97 @@ def my_plan(data):
 
     assert (plan_res == th.tensor([-42, 24, 46])).all()
 
+    # delete remote object before websocket connection termination
+    del x_ptr
+
     server.terminate()
+
+
+def test_replace_worker_ids(hook):
```
mari-linhares: There's a lot going on here. Break this into multiple tests and add some comments about what each test is covering.

test/federated/test_plan.py
```diff
@@ -227,4 +224,97 @@ def my_plan(data):
...
+        obj=messages, change_id=10, to_id=100, from_worker="me", to_worker="another"
+    )
+
+    # note that tuples are converted to lists
+    expected = [100, ["worker", "another"], "you", 20, 100, b"you", [30, ["you", "another", "bla"]]]
+
+    assert replaced == expected
+
+
+def test___call__(hook):
```
mari-linhares: Break this into two tests: test__call__fn, test__call__method

mari-linhares:
Thanks for this! The tests are super helpful!!!

I've fixed this issue at the train_config branch as well, if you want to have a look.

mari-linhares:
Related to #1989

syft/federated/plan.py
midokura-silvia: We don't need all four possibilities. The outdated code in dev contained the four possibilities. Then the current dev code changed it to (from_worker, to_worker) and (from_worker.encode(), to_worker.encode()) but without taking care of the integers that don't have a encode() function.

As we are replacing, it doesn't make sense to replace from_worker twice. Therefore only two outcomes, not four.

mari-linhares: Hmm, I would say we need the 4 possibilities (as in train-config) to cover all possible cases, don't we?

My question then is: can we guarantee that with only these two cases all the other cases will work?

midokura-silvia: Let's think of an example doing string replacements:
"The quick brown b'brown' fox jumps over the lazy b'brown' dog."
If we want to replace brown by black, we would call the replace method with from_worker=brown and to_worker=black.
First call would replace all brown instances with black:
"The quick black b'brown' fox jumps over the lazy b'brown' dog."
The second call would replace all b'brown' instances with b'black':
"The quick black b'black' fox jumps over the lazy b'black' dog."
And we are done.

Issue #2122, opened 2019-05-05
Backpropagation not producing gradients


DanyEle:
I'm currently working on the implementation of the federated training of a Recurrent Neural Network, and I stumbled upon the following issue: when operating on the remote pointer of the loss function initialized with Negative Log Likelihood, the grad_fn function is not set and it will still say "requires_grad=True" despite the loss function being set in the code:

![requires_grad_True](https://user-images.githubusercontent.com/4907418/57189502-f826dd80-6f0f-11e9-9787-f2f8b39be1ae.png)

Instead, in the sequential version or when the input arguments of the loss function are ".get()", the grad_fn is actually set in the loss function:

![requires_grad_fn](https://user-images.githubusercontent.com/4907418/57189535-40460000-6f10-11e9-8b8d-38f8442a703f.png)

Because of this issue, the backpropagation does not successfully generate all the model's gradient parameters (i.e.: only two parameters are generated in the remote version vs four parameters in the local version), hence preventing the model from converging in the remote version. I looked up if other people are having the same issue, and these pages seem to suggest that the graph is broken (?):  https://discuss.pytorch.org/t/list-model-parameters-0-grad-doubt-in-weight-updates/8422

To reproduce:

```
import torch.nn as nn
import torch
import syft as sy
import numpy as np

hook = sy.TorchHook(torch)  
dani = sy.VirtualWorker(hook, id="dani")  
ele = sy.VirtualWorker(hook, id="ele") 
workers_virtual = [dani, ele]

criterion = nn.NLLLoss()
output = torch.from_numpy(np.ones((1, 18)))
output.requires_grad_()
output_sent = output.send(dani)

category_single = torch.from_numpy(np.array([2]))
category_sent = category_single.send(dani)
#Non-working version, with requires_grad=True
loss = criterion(output_sent, category_sent) 
print(loss.copy().get()) #Requires_grad=True

#Working version, with grad_fn properly set.
loss_working = criterion(output_sent.copy().get(), category_sent.copy().get()) 
print(loss_working) #grad_fn = NllLossBackward

loss.backward()
```

#model just has 2 parameters with grad != None following the backpropagation. Consequently, during the training process, the model weights are not updated. 

for param in model_ptr.parameters():

Returns an error, as parameters at position [0] is not present.

mari-linhares:
I've tried to run the snippet above, and I got the following error:

```python
File "original_syft_bug.py", line 12, in <module>
    output = np.ones(1, 18)
[...]

PR #2127, opened 2019-05-07
Fixed backpropagation for models using parameters.grad

DanyEle:
Fix for issue #2122 reported while trying to implement a Recurrent Neural Network in a federated way. 

Fixed the loss.backward() function not generating gradient updates accumulated in the model parameters.

If trying to access the .grad.data attribute of a model's parameters, not all the gradient parameters were actually set, leading to the model weights not being updated. The de-commented lines actually allow the gradients to be stored in the model parameters and the weights to be updated, hence the loss function to converge.


robert-wagner:
Hey, This is a super high value pull request. It looks like flake8 is not happy with indentation which is causing travis to fail. That being said could you add a test to make sure this is not broken again in the future

DanyEle:
Sorry for the many commits, yesterday night I messed up badly with Git. Going to try and see if I can get a test up for this fix.

test/torch/test_hook.py
```diff
@@ -220,3 +220,69 @@ def test_hook_args_and_cmd_signature_malleability():
 
     r3 = a + b
     assert (r3 == syft.LoggingTensor().on(torch.tensor([2.0, 4]))).all()
+
+
+def test_RNN_grad_set_backpropagation():
+    """Perform backpropagation at a remote worker and check if the gradient updates
+    and properly computed within the model"""
+
+    import syft as sy
```
robert-wagner: You don't need this import

test/torch/test_hook.py
```diff
@@ -220,3 +220,69 @@ def test_hook_args_and_cmd_signature_malleability():
...
+
+    hook = sy.TorchHook(torch)
+    alice = sy.VirtualWorker(hook, id="alice")
```
robert-wagner: You can get Alice by including workers as a parameter and writing Alice = workers["Alice"]. The tests for most tensors have good examples for how to do this

test/torch/test_hook.py
```diff
@@ -220,3 +220,69 @@ def test_hook_args_and_cmd_signature_malleability():
...
+
+    # now let's get the model and check if its parameters are indeed there
+    model_got = alice_model.get()
+
+    learning_rate = 0.005
+
+    # If the gradients are there, then the backpropagation did indeed complete successfully
+    for param in model_got.parameters():
+        assert param.grad.data is not None
+        param.data.add_(-learning_rate, param.grad.data)
```
robert-wagner: What does this line do?

DanyEle: This is the most important line, in which the` param.grad.data` was previously yielding a `None`. It basically updates the model's parameters (e.g.: layers' weights) based on the gradients computed following the backpropagation phase.

test/torch/test_hook.py
DanyEle: Done!

test/torch/test_hook.py
DanyEle: Removed import

test/torch/test_hook.py
robert-wagner: Why is this called after the assertion though

DanyEle: Basically with this "assert2, we would like to make sure that every single gradient among the model's parameters is indeed there.

 By executing `param.data.add_(-learning_rate, param.grad.data)` without an an assert before-hand, we may already have an exception in case a grad.data is None, meaning that the code is broken again and we had no way to check the effective presence/absence of the gradient beforehand

PR #2131, opened 2019-05-08
WIP: Fix #2076

### LINKED ISSUE #2076: Search remote data will cause the data being deleted
I've experimented on getting PySyft to work with remote worker with its own dataset.
During these, I found out the `search` from local will cause the data deleted on the remote.

Relevant files in [this gist](https://gist.github.com/feigaoxyz/99dffd2f4336417f83f0cf4e1c80d8fd).

To trigger the error, 
1. First run `python remote.py` in one terminal window
2. Then run `python local.py` from another terminal window

You will see the results:

```sh
$ python remote.py
Objects: {94139650912: tensor([-1.,  2.])
	Tags: x
	Shape: torch.Size([2])}
worker <WebsocketServerWorker id:0 #tensors:1> received SEARCH (b'x',)
worker <WebsocketServerWorker id:0 #tensors:1> received OBJ_DEL 94139650912
Objects: {}
```

```sh
$ python local.py
Local objects before search {}
Search result: [(Wrapper)>[PointerTensor | me:68744693343 -> 0:13527568418]]
Local objects after search {}
```

As the first one showed: the remote worker actually received two actions `SEARCH` and `OBJ_DEL` in correspondence with the search command on Line 20 of local.py.

---

This error can also be reproduced by slightly modifying the `test/test_websocket_woker.py` by searching same term twice, see [L57 of updated test](https://gist.github.com/feigaoxyz/99dffd2f4336417f83f0cf4e1c80d8fd#file-test_websocket_woker-py).

mari-linhares:
Fixes #2076.

Main changes are:

- tensor.garbage_collected_data is now serialized/deserialized.

- Created a force remove object operation (FORCE_RM_OBJ), the only difference from regular remove object operation is that if the tensor being removed has a child (pointer tensor) than it forces the child to be garbage collected.

- FORCE_RM_OBJ is needed to force all the chain to be removed if the "end" point tensor deleted and has garbage_collected_data attribute equals to True.

- Some test improvements :smile: !

### DIFF
--- a/syft/codes.py
+++ b/syft/codes.py
@@ -7,6 +7,7 @@ class MSGTYPE(object):
     IS_NONE = 6
     GET_SHAPE = 7
     SEARCH = 8
+    FORCE_OBJ_DEL = 9
 
 
 # Build automatically the reverse map from codes to msg types
--- a/syft/generic/object_storage.py
+++ b/syft/generic/object_storage.py
@@ -112,7 +112,7 @@ class ObjectStorage:
         """Forces object removal.
 
         Besides removing the object from the permanent object registry if it exists.
-        Explictly forces removall of the object modifying the `garbage_collect_data` attribute.
+        Explicitly forces removal of the object modifying the `garbage_collect_data` attribute.
 
         Args:
             remote_key: A string or integer representing id of the object to be
--- a/syft/generic/object_storage.py
+++ b/syft/generic/object_storage.py
@@ -107,3 +107,19 @@ class ObjectStorage:
         """
         if remote_key in self._objects:
             del self._objects[remote_key]
+
+    def force_rm_obj(self, remote_key: Union[str, int]):
+        """Forces object removal.
+
+        Besides removing the object from the permanent object registry if it exists.
+        Explictly forces removall of the object modifying the `garbage_collect_data` attribute.
+
+        Args:
+            remote_key: A string or integer representing id of the object to be
+                removed.
+        """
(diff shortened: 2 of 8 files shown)


syft/generic/object_storage.py
```diff
@@ -107,3 +107,19 @@ def rm_obj(self, remote_key: Union[str, int]):
         """
         if remote_key in self._objects:
             del self._objects[remote_key]
+
+    def force_rm_obj(self, remote_key: Union[str, int]):
+        """Forces object removal.
+
+        Besides removing the object from the permanent object registry if it exists.
+        Explictly forces removall of the object modifying the `garbage_collect_data` attribute.
```
midokura-silvia: typo

syft/codes.py
```diff
@@ -7,6 +7,7 @@ class MSGTYPE(object):
     IS_NONE = 6
     GET_SHAPE = 7
     SEARCH = 8
+    FORCE_OBJ_DEL = 9
```
midokura-silvia: I know it's off-topic for this pull request. What about deriving MSGTYPE from Enum (from enum import Enum). Any reasons for not doing so?

mari-linhares: I don't think there's a reason not to use an enum... thanks for spotting it.

mari-linhares: @midokura-silvia I've just tried to change it to an enum and I get serialization/deserialization issues.

PR #2134, opened 2019-05-09
Add tests for garbage collection on a remote worker

mari-linhares:


test/torch/tensors/test_gc.py
```diff
@@ -177,3 +179,19 @@ def test_implicit_garbage_collect_logging_on_pointer(workers):
     x = "open-source"
 
     assert x_id not in workers["bob"]._objects
+
+
+def test_websocket_garbage_collection(hook, start_proc):
+    # Args for initializing the websocket server and client
+    base_kwargs = {"id": "ws_gc", "host": "localhost", "port": 8777, "hook": hook}
+    server_kwargs = base_kwargs
```
robert-wagner: What is the point of this assignment?

mari-linhares: not needed, indeed.

PR #2141, opened 2019-05-15
Provide unique ids by checking against already generated ids

midokura-silvia:
IdProvider was not guaranteed to provide an id only once. Added tests.

### DIFF
--- a/syft/generic/id_provider.py
+++ b/syft/generic/id_provider.py
@@ -1,5 +1,6 @@
 import random
 from typing import List
+from syft import exceptions
 
 
 def create_random_id():
@@ -13,7 +14,7 @@ class IdProvider:
     Can take a pre set list in input and will complete
     when it's empty.
 
-    An instance of IdProvider is accessbile via sy.ID_PROVIDER.
+    An instance of IdProvider is accessible via sy.ID_PROVIDER.
     """
 
     def __init__(self, given_ids=None):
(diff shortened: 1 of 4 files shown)


syft/generic/id_provider.py
```diff
@@ -13,7 +17,9 @@ class IdProvider:
 
     def __init__(self, given_ids=list()):
         self.given_ids = given_ids
-        self.generated = []
+        self.generated = (
+            []
+        )  # TODO: use sorted container to quickly check whether a value is contained
```
mari-linhares: Can we use a set before merging this to dev?

iamtrask: +1

midokura-silvia: I tried, but it is not as easy as it seems. There is a mechanism to retrieve the list of generated ids via a ResponseSignatureError. And the order of the elements in the list is used to assign ids to return values. So simply changing self.generated to a set() results in errors in the unittests with remote method calls that have more than one return value.

midokura-silvia: The generated ids are now stored in a set.

syft/generic/id_provider.py
```diff
@@ -27,6 +34,44 @@ def pop(self, *args) -> int:
...
+    def set_next_ids(self, given_ids: List):
+        """Sets the next ids returned by the id provider
+
+        Note that the ids are returned in reverse order of the list, as a pop() operation is applied
+
+        Args:
+            given_ids: List, next ids returned by the id provider
+
+        """
+        # TODO: should we check whether these ids conflict with the content of self.generated?
```
mari-linhares: Yes. I think an exception should be thrown.

PR #2144, opened 2019-05-15
Add support for Pytorch 1.1

### LINKED ISSUE #2126: Torch 1.1 Integration
So PyTorch just released torch 1.1 and it causes a lot of failing Unit Tests because of some breaking API updates. We need to address them.

LaRiffle:
Fix #2126 related to the new version 1.1 of Pytorch

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 lz4
 msgpack
 sklearn
-torch==1.1
+torch>=1.0.1
 torchvision
 websocket_client
 websockets>=7.0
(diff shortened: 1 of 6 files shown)


robert-wagner:
Does it work for versions greater than 1.0.1 or only 1.1

LaRiffle:
It was passing before change the requirements.txt so I assume it works :)

robert-wagner:
In that case could we change the requirements to be >=1.0.1

LaRiffle:
You made me change it! haha

LaRiffle:
I would suggest upgrade everyone to 1.1 so that we avoid cases where we have different behaviours depending of the version

PR #2147, opened 2019-05-17
Large Precision Tensor

mccorby:
This PR contains the first steps to have very large integers in the system.
It works by transforming the input into an a torch tensor that contains the big number split in components.

This PR also includes a first operation `add` that allows addition of two LPTs


syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -634,3 +634,11 @@ def combine(self, *pointers):
         ps.append(self)
 
         return sy.combine_pointers(*ps)
+
+    def large_prec(self, *args, **kwargs):
```
mccorby: This might not be needed

syft/frameworks/torch/tensors/interpreters/large_precision_tensor.py
```diff
@@ -0,0 +1,157 @@
+import numpy as np
+import torch
+
+from syft.frameworks.torch.overload_torch import overloaded
+from syft.frameworks.torch.tensors.interpreters import AbstractTensor
+
+
+class LargePrecisionTensor(AbstractTensor):
```
robert-wagner: rename file to large_precision.py

syft/frameworks/torch/tensors/interpreters/large_precision_tensor.py
```diff
@@ -0,0 +1,157 @@
...
+    """LargePrecisionTensor allows handling of numbers bigger than LongTensor
+
+    Some systems using Syft require larger types than those supported natively. This tensor type supports arbitrarily
+    large values by packing them in smaller ones.
+    If the original tensor is a PyTorch tensor it is not processed
+    """
+
+    def __init__(
+        self, owner=None, id=None, tags=None, description=None, precision=16, virtual_prec=0
```
robert-wagner: why were these chosen as defaults

robert-wagner: also add typing

syft/frameworks/torch/tensors/interpreters/large_precision_tensor.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        return {"precision": self.precision, "virtual_prec": self.virtual_prec}
+
+    def __eq__(self, other):
+        return self.child == other.child
+
+    @overloaded.method
+    def add(self, self_, *args, **kwargs):
+        other = args[0]
+        if self_.shape[-1] != other.shape[-1]:
```
robert-wagner: format this as an if elif else

syft/frameworks/torch/tensors/interpreters/large_precision_tensor.py
```diff
@@ -0,0 +1,157 @@
...
+            filler = torch.zeros(size=new_shape, dtype=to_adjust.dtype)
+        else:
+            filler = torch.ones(size=new_shape, dtype=to_adjust.dtype)
+        return torch.cat([filler, to_adjust], len(shape) - 1)
+
+    @staticmethod
+    def _split_number(number, bits):
+        """Splits a number in numbers of a smaller power
+
+        :param number: the number to split
```
robert-wagner: Change to google style

syft/frameworks/torch/tensors/interpreters/large_precision_tensor.py
```diff
@@ -0,0 +1,157 @@
...
+        return n
+
+    def fix_large_precision(self):
+        self.child = self._create_internal_tensor(self.child, self.precision, self.virtual_prec)
+        return self
+
+    def restore_precision(self):
+        """
+        Restore the tensor expressed now as a matrix for each original item.
+        :return: the original tensor
```
robert-wagner: google style

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -647,3 +647,21 @@ def combine(self, *pointers):
         ps.append(self)
 
         return syft.combine_pointers(*ps)
+
+    def large_prec(self, *args, **kwargs):
```
robert-wagner: do we need a method to explicitly convert this or should this happen automatically when calling something like fix prec

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -647,3 +647,21 @@ def combine(self, *pointers):
...
+        return (
+            syft.frameworks.torch.tensors.interpreters.LargePrecisionTensor(*args, **kwargs)
+                .on(self)
+                .enc_fix_large_prec()
+                .wrap()
+        )
+
+    large_precision = large_prec
+
+    def enc_fix_large_prec(self):
```
robert-wagner: why do we need this as a separate function on native tensor

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -647,3 +647,21 @@ def combine(self, *pointers):
...
+        return self.child.fix_large_precision()
+
+    def restore_prec(self):
```
robert-wagner: same as for enc_fix_large_prec

syft/frameworks/torch/tensors/interpreters/large_precision_tensor.py
```diff
@@ -0,0 +1,157 @@
...
+    def __init__(
+        self, owner=None, id=None, tags=None, description=None, precision=16, virtual_prec=0
+    ):
+        """Initializes a LargePrecisionTensor.
+
+        Args:
+            owner: An optional BaseWorker object to specify the worker on which
+                the tensor is located.
+            id: An optional string or integer id of the LargePrecisionTensor.
+            precision: The precision this tensor will be transformed to internally. In bits.
```
robert-wagner: what exactly is the difference between this and virtual prec

syft/frameworks/torch/tensors/interpreters/large_precision_tensor.py
```diff
@@ -0,0 +1,157 @@
...
+        # refs_ok is necessary to enable iterations of reference types.
+        # These big numbers are stored as objects in np
+        result = []
+        for x in np.nditer(tensor, flags=["refs_ok"]):
+            n = int(x.item() * 2 ** virtual_prec)
+            print(f"\nAdding number {n} for item {x.item()}\n")
+            result.append(LargePrecisionTensor._split_number(n, precision))
+        new_shape = tensor.shape + (len(max(result, key=len)),)
+        result = np.array(result).reshape(new_shape)
+        return torch.IntTensor(result)
```
robert-wagner: should this not be a long tensor for more storage

jacorbachosky:
> Initial review. Will comment more later if I see other stuff (its the end of the day here)

Thanks for the late night review. It was the end of the day here too. We should be doing something else! :)

syft/frameworks/torch/tensors/interpreters/large_precision_tensor.py
mccorby: I used `16` bits as the precision that would be needed when moving this tensor to other systems (I had JVM in mind).
It would be good to have this value set in a global configuration of the setup.
I am also thinking that instead of using a number we could use a `dtype`. I will give it a try and see how it looks like

syft/frameworks/torch/tensors/interpreters/native.py
mccorby: Not needed

mccorby: I wanted this to be separated until we had a full working solution.
Eventually `fix_prec()` will check the precision and use `FixedPrecisionTensor` or `LargePrecisionTensor`.

syft/frameworks/torch/tensors/interpreters/large_precision_tensor.py
mccorby: I would change the naming as follows:
`precision` -> `internal_type`
`virtual_prec` -> `precision`

syft/frameworks/torch/tensors/interpreters/large_precision_tensor.py
mccorby: I'll see if by using `dtype` as a param this can be done differently.

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,175 @@
...
+    @overloaded.method
+    def add(self, self_, *args, **kwargs):
+        other = args[0]
+        if self_.shape[-1] != other.shape[-1]:
+            result = self._add_different_dims(other, self_)
+        else:
+            result = self_ + other
+        return LargePrecisionTensor(precision=self.precision, virtual_prec=self.virtual_prec).on(
+            result
+        )
```
LaRiffle: The @overloaded.method decorator will automatically put LargePrecisionTensor on top of your result, so you can remove it, otherwise you have it twice (as cell 17 of the NB you shared shows)
Also, you can simplify the function signature `def add(self, self_, *args, **kwargs)` -> `def add(self, self_, other)` directly

mccorby:
> It only works for 1D tensors right?

I broke something when I passed to `np` 😱 . I'll see if I can fix it

syft/frameworks/torch/tensors/interpreters/large_precision_tensor.py
mccorby: Using `dtype` now

syft/frameworks/torch/tensors/interpreters/large_precision_tensor.py
mccorby: Using `dtype` now to create the tensor.

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,181 @@
...
+            description (str): a description of this tensor.
+            internal_type (int): The precision this tensor will be transformed to internally. In bits.
+            virtual_prec (int): The virtual precision required by the caller. In bits.
+        """
+        super().__init__(id=id, owner=owner, tags=tags, description=description)
+        self.internal_type = internal_type
+        self.virtual_prec = virtual_prec
+
+    def _create_internal_tensor(self):
+        # TODO refs_ok might not be needed as we are passing now torch.tensors here
```
LaRiffle: Maybe add a short dosctring to explain what's happening

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,181 @@
...
+        return torch.tensor(result, dtype=self.internal_type)
+
+    def get_class_attributes(self):
+        """
+        Specify all the attributes need to build a wrapper correctly when returning a response.
+        """
+        return {"internal_type": self.internal_type, "virtual_prec": self.virtual_prec}
+
+    def __eq__(self, other):
+        return self.child == other.child
```
LaRiffle: Do you need to specify `def __eq__`?

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,181 @@
...
+        ndarray = self.child.numpy()
+        result = LargePrecisionTensor._restore_tensor_into_numbers(
+            ndarray, internal_precision[self.internal_type]
+        ) / (2 ** self.virtual_prec)
+        return torch.from_numpy(result.reshape(ndarray.shape[:-1]).astype(np.float32))
+        # At this point the value is an object type. Force cast to float before creating torch.tensor
+
+    @staticmethod
+    def _restore_tensor_into_numbers(number_array, precision):
+        number_array = number_array.reshape(-1, number_array.shape[-1])
```
LaRiffle: Add a short docstring

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,181 @@
...
+        result = []
+        for elem in number_array:
+            result.append(LargePrecisionTensor._restore_number(elem, precision))
+        return np.array(result)
+
+    @staticmethod
+    def _adjust_to_shape(to_adjust, shape, fill_value=0) -> torch.Tensor:
+        # We assume only the last dimension needs to be adjusted
```
LaRiffle: Same, instead of comments

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,181 @@
...
+        """Initializes a LargePrecisionTensor.
+
+        Args:
+            owner (BaseWorker): An optional BaseWorker object to specify the worker on which
+                the tensor is located.
+            id (str or int): An optional string or integer id of the LargePrecisionTensor.
+            tags (list): list of tags for searching.
+            description (str): a description of this tensor.
+            internal_type (int): The precision this tensor will be transformed to internally. In bits.
+            virtual_prec (int): The virtual precision required by the caller. In bits.
```
LaRiffle: Maybe even further clairfy `internal_type` : you could say that the storage of the large tensor is using small tensors of size `internal_type` bits, or smthg equivalent

test/torch/tensors/test_large_precision.py
```diff
@@ -0,0 +1,63 @@
...
+def test_add_different_dims():
+    internal_type = torch.int16
+    virtual_prec = 256
+    x1 = torch.tensor([100000.0])
+    x2 = torch.tensor([20.0])
+    expected = torch.tensor([100020.0])
+    lpt1 = x1.large_prec(internal_type=internal_type, virtual_prec=virtual_prec)
+    lpt2 = x2.large_prec(internal_type=internal_type, virtual_prec=virtual_prec)
+    result = lpt1 + lpt2
+    assert torch.all(torch.eq(expected, result.restore_precision()))
```
LaRiffle: Can you also add negative values in the test and decimal numbers?

syft/frameworks/torch/tensors/interpreters/large_precision.py
LaRiffle: Also, internal_type in now a type, you should modify the type info

syft/frameworks/torch/tensors/interpreters/native.py
robert-wagner: I would make the case then that this should be a class method on large_prec tensor which we can either call manually or add to fix_prec later.

mccorby: The code now decided whether to use LPT or FPT depending on the requested precision.

syft/frameworks/torch/tensors/interpreters/large_precision.py
mccorby: Leftovers. Removed

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -13,6 +13,23 @@
 from syft.exceptions import PureTorchTensorFoundError
 
 
+def _get_maximum_precision():
```
mccorby: I'm not 100% sure about this solution though it is a way of allowing different setups. If you can think on a better alternative, happy to discuss it

LaRiffle: Yes maybe this can be a bit more flexible. What I experienced is that dealing with integers > 2**62-1 is not fun with precision tensors, so 62 could be the total limit which sum the space allocated to the fractional_precision *and* the space allocated to the integer part. I think we should find a way to take this into account.
Another point to consider: we shouldn't mix precision in bits and decimal precision: from the user interface, people usually keep the `base` at 10 so the precision they ask for (3 by default) means 3 digits after the `.` which roughly corresponds to a 10 bits precision

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,194 @@
...
+    def __init__(
+        self,
+        owner=None,
+        id=None,
+        tags=None,
+        description=None,
+        internal_type=torch.int32,
+        precision_fractional=0,
+        verbose=False,
+    ):
```
LaRiffle: With respect to my comment below maybe we should consider add a `base` attribute just like in the precision tensor

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -13,6 +13,25 @@
...
+
+    Returns:
+        The maximum value for precision allowed in this setup
+    """
+    return default_pytorch_maximum_precision()
+
+
+def default_pytorch_maximum_precision():
+    """Dealing with integers > 2**62-1 is not fun with precision tensors.
+    """
```
LaRiffle: You need to be more precise:
I would suggest doing a check as follows:
if log2(int(value)) + log2(base**precision_fractional) > 62
then go_to_large_precision

But because value can be < 1 or <0
if log2(int(abs(value))+1) + log2(base**precision_fractional) > 62
then go_to_large_precision

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,198 @@
...
+            "internal_type": self.internal_type,
+            "precision_fractional": self.precision_fractional,
+        }
+
+    @overloaded.method
+    def add(self, self_, other):
+        if self_.shape[-1] != other.shape[-1]:
+            return self._add_different_dims(other, self_)
+        else:
+            return self_ + other
```
Jasopaum: What happens when overflow occurs in one chunk of the data?

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,198 @@
...
+        We assume only the last dimension needs to be adjusted.
+        Original input tensors should have the same dimensions.
+        """
+        diff_dim = shape[-1] - to_adjust.shape[-1]
+        new_shape = to_adjust.shape[:-1] + (diff_dim,)
+
+        if not fill_value:
+            filler = torch.zeros(size=new_shape, dtype=to_adjust.dtype)
+        else:
+            filler = torch.ones(size=new_shape, dtype=to_adjust.dtype)
```
Jasopaum: Did you mean `filler = fill_value * torch.ones(...)`?

LaRiffle: What is exactly the use case for fill_value != 0?

mccorby: @Jasopaum Good catch
@LaRiffle I can only think of filling the tensor with 0s (for additions) or 1s (for multiplications)

syft/frameworks/torch/tensors/interpreters/large_precision.py
mccorby: We're dead here :) I was expecting to address the overflow when implementing the multiplication but it's true this can already be a problem with the addition.

In fact, this is probably the *main issue* with this representation of large numbers.

1) One possible solution could be to halve the storage precision. This way there will always be enough room to avoid an overflow.
This approach, however, will force to recheck the result in case the new number needs to be split again as soon a we concatenate more additions or multiplications.

2) Another option is to deal with these numbers the same way that a processor deals with numbers greater than the word they work with. This approach is slower but we know it works and we know there exists an implementation.

3) The third option would be to work directly with NumPy objects when operating. This would require splitting and restoring back the tensor that represents the large number with every operation. Probably this option is the most reasonable in terms of coding effort.

What do you think?

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,198 @@
...
+        base = 2 ** bits
+        n = 0
+        # Using tolist() to allow int type. Terrible performance :(
+        for x in number_parts.tolist():
+            n = n * base + x
+        return n / (self.base ** self.precision_fractional)
+
+
+# The internal precision used to decompose the large numbers is half of the size of the type.
+# This is because we are not considering negative numbers when decomposing.
```
Jasopaum: Why is it half and not just -1?

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,198 @@
...
+            sign = 1
+        else:
+            sign = -1
+            number = number * (-1)
+
+        base = 2 ** bits
+        number_parts = []
+        while number:
+            number, part = divmod(number, base)
+            number_parts.append(part * sign)
```
Jasopaum: Maybe we need a check to be sure that we do not try to put a negative number in a LPT made of uint

syft/frameworks/torch/tensors/interpreters/large_precision.py
LaRiffle: Agreed on the 3rd, we shouldn't resent to use numpy when appropriate.
How does this PR relates with #2257 ? Is it just a fork of this one?

mccorby: #2257 was for showing how it could be with np. I will cancel it and keep working on this PR.

LaRiffle: Ok thanks for the update!

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,205 @@
...
+            self_, internal_precision[self.internal_type]
+        )
+        b = LargePrecisionTensor._internal_representation_to_large_ints(
+            other, internal_precision[self.internal_type]
+        )
+        # We need to divide the result of the multiplication by the precision
+        return self._create_tensor_from_numpy((a * b) / (self.base ** self.precision_fractional))
+
+    __mul__ = mul
+
```
Jasopaum: Maybe you can add sub and div?

test/torch/tensors/test_large_precision.py
```diff
@@ -0,0 +1,151 @@
...
+
+def test_subtract():
+    internal_type = torch.int16
+    precision_fractional = 256
+    x1 = torch.tensor([10.0])
+    x2 = torch.tensor([-90000000000000010.0])
+    expected = torch.tensor([-90000000000000000.0])
+    lpt1 = x1.fix_prec(internal_type=internal_type, precision_fractional=precision_fractional)
+    lpt2 = x2.fix_prec(internal_type=internal_type, precision_fractional=precision_fractional)
+    result = lpt1 + lpt2
```
Jasopaum: As __add__ is the method called, I would still put this test in test_add and say that you test it when values are negative

syft/frameworks/torch/tensors/interpreters/large_precision.py
mccorby: No longer applies

syft/frameworks/torch/tensors/interpreters/large_precision.py
mccorby: Added `sub`

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -166,33 +182,42 @@ def _split_number(self, number, bits):
...
         for x in number_parts.tolist():
             n = n * base + x
-        return n / (self.base ** self.precision_fractional)
+        return n
 
 
-# The internal precision used to decompose the large numbers is half of the size of the type.
-# This is because we are not considering negative numbers when decomposing.
+# The internal precision used to decompose the large numbers is the size of the type - 1.
 internal_precision = {
```
LaRiffle: make a @property in the class to allow `self.internal_precision` instead of `internal_precision[self.internal_type]`, it's more pythonic (and will be useful)

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -61,30 +60,38 @@ def _create_internal_tensor(self):
         result = np.array(result).reshape(new_shape)
         return torch.tensor(result, dtype=self.internal_type)
 
+    def _create_tensor_from_numpy(self, ndarray):
```
LaRiffle: You can make this method a @staticmethod like _internal_representation_to_large_ints, and change the signature `_create_tensor_from_numpy(self, ndarray):` -> `_create_tensor_from_numpy(ndarray, internal_type):`. I believe you can get `self.child.shape[:-1] ` for the ndarray directly (am i right?)

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -95,51 +102,57 @@ def __iadd__(self, other):
...
-        new_shape = to_adjust.shape[:-1] + (diff_dim,)
-
-        if not fill_value:
-            filler = torch.zeros(size=new_shape, dtype=to_adjust.dtype)
-        else:
-            filler = torch.ones(size=new_shape, dtype=to_adjust.dtype)
-        return torch.cat([filler, to_adjust], len(shape) - 1)
-
-    def _split_number(self, number, bits):
+    def _split_number(self, number, bits) -> list:
```
LaRiffle: Also make this one static and ask that internal type be provided in the signature if needed

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,226 @@
...
+            bits (int): the bits used in the split.
+
+        Returns:
+            Number: the large number represented by this tensor
+        """
+        base = 2 ** bits
+        n = 0
+        for x in number_parts.tolist():
+            n = n * base + x
+        return n
```
LaRiffle: Here is a solution to have this one vectorized easily:
```
n_parts = len(number_parts)
base = 2 ** bits
powers = base ** np.arange(n_parts)
return np.sum(number_parts * powers)
```
smthg like this should work I hope

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+type_precision = {
```
LaRiffle: Maybe add a comment again to explain why you needed to do -1

syft/frameworks/torch/hook/hook_args.py
```diff
@@ -32,13 +35,15 @@
     list: lambda _args: [build_rule(a) for a in _args],
     tuple: lambda _args: tuple([build_rule(a) for a in _args]),
     dict: one,  # FIXME This is for additiveShareTensor.child, it can be confusing and AST.child
+    np.ndarray: one,
```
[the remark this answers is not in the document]
mccorby: It's needed in the hooks

syft/frameworks/torch/tensors/interpreters/large_precision.py
mccorby: Difficult with vectorisation as every element uses the result of the previous (maybe it can be done but I didn't find a solution).
I created a recursive function instead as it is cleaner and a better fit than the loop I had before

syft/frameworks/torch/tensors/interpreters/large_precision.py
mccorby: Changed the type sizes to be what the platform provides. The -1 is applied when computing the internal size (comments added to the source to explain it)

test/torch/tensors/test_large_precision.py
```diff
@@ -0,0 +1,171 @@
...
+
+def test_uint8_representation(workers):
+    x = torch.tensor([[1.5, 2.0, 3.0], [4.5, 5.0, 6.0]])
+    enlarged = x.fix_prec(internal_type=torch.uint8, precision_fractional=256)
+    restored = enlarged.float_precision()
+    # And now x and restored must be the same
+    assert torch.all(torch.eq(x, restored))
+
+
+def test_subtract():
```
robert-wagner: Can you add a test for mod?

mccorby: Hi @robert-wagner,
`mod` is not yet implemented or do you mean something else?

robert-wagner: Hey @mccorby That was a big part of why I asked for a test for it. By far the biggest thing I see my self using lpt tensor for is for converting to something like a Chinese remainder theorem tensor (since it is significantly more space efficient). However, that conversion requires mod. Theoretically we could add that as a separate pr though

mccorby:
There is an issue when plugging LPT in the normal flow of send/share that would break `fix_prec`

Fixing.

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,234 @@
...
+
+    Some systems using Syft require larger types than those supported natively. This tensor type supports arbitrarily
+    large values by packing them in smaller ones.
+    Typically a user will require to enlarge a float number by fixing its precision
+        tensor.fix_prec()
+
+    The large value is defined by `precision_fractional`.
+
+    The smaller values are of type `internal_type`. The split of the large number into the smaller values is done using
+    half the size of the selected internal type. This is to avoid overflow as types are in the range ±2**(size - 1).
```
Jasopaum: I don't think this is still true now that you use NumPy

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -0,0 +1,226 @@
...
+        # itertools.zip_longest fills on the right but we need to fill with zeros on the left
+        n_result = []
+        for a_number in result:
+            len_a_number = len(a_number)
+            if len_a_number < max_length:
+                n_result.append([0] * (max_length - len_a_number) + a_number)
+            else:
+                n_result.append(a_number)
+        result = np.array(n_result).reshape(new_shape)
+        return torch.tensor(result, dtype=self.internal_type)
```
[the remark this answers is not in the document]
Jasopaum: Avoiding the loop over all the elements is not too hard, you can just change _split_number so that it takes an array instead of a single value (I did it but I'm not sure I can push to this PR, I can send you the small changes though)

syft/frameworks/torch/tensors/interpreters/large_precision.py
mccorby: I think it still applies since the internal type is still a PyTorch type. NumPy is only used to perform the operations

Jasopaum: Ok but we don't have to worry about overflows anymore, do we?

mccorby: The overflow can happen when doing `divmod`. A way of seeing it happening is to remove the -1 in `def internal_precision(self)`

Jasopaum: I think the code is ok, I was just saying that for me, it's not half the size that is used but all of it, and that the `-1` is not here to avoid overflow but because we need to keep a bit for the sign in each chunk

mccorby: Haha... I see it now. You're right. Fixing!

test/torch/tensors/test_large_precision.py
mccorby: Added `mod`

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -110,6 +106,8 @@ def __iadd__(self, other):
 
         return self
 
+    add_ = __iadd__
```
LaRiffle: Did you declare `add_` just for the tests?
Actually I think that you can write directly `x. __iadd__(y)` or `x += y`

LaRiffle: NB actually add_ is a valid torch function

PR #2148, opened 2019-05-17
 Add Tanh and Sigmoid to autograd

mari-linhares:


syft/frameworks/torch/tensors/interpreters/derivatives.yaml
```diff
@@ -12,18 +12,25 @@
...
   self: grad * other
   other: grad * self if type(self) == type(other) else None
 
-- name: sqrt(self)
-  self: grad / (2 * result)
-
-- name: asin(self)
-  self: grad * (-self * self + 1).rsqrt()
+- name: sigmoid(self)
+  self: self.sigmoid() * (1 - self.sigmoid())
```
LaRiffle: I'm not 100% of my maths but I suppose you should make `grad` happen as in other functions right ? Otherwise you give the derivative with respect to a self whose gradient is 1.

syft/frameworks/torch/tensors/interpreters/derivatives.yaml
```diff
@@ -12,18 +12,25 @@
...
-  self: grad * self.cosh()
\ No newline at end of file
+  self: grad * self.cosh()
+
+- name: sqrt(self)
+  self: grad / (2 * result)
+
+- name: tanh(self)
+  self: 1 - self.tanh() ** 2
+
```
LaRiffle: same interrogation here

syft/frameworks/torch/tensors/interpreters/derivatives.yaml
robert-wagner: I think autograd tensor does that automatically but I'm not sure

robert-wagner: I think autograd tensor does that automatically but I'm not sure

LaRiffle: Like I would expect `grad * self.sigmoid() * (1 - self.sigmoid())`
If this is added automatically however it's fine, but I'm not sure, considering other examples

robert-wagner: Now that you mention it, I agree

mari-linhares: my bad, thanks @LaRiffle!

PR #2151, opened 2019-05-18
Update polynomial tensor 

LaRiffle:
I'm not sure I've understood all the code in this tensor, however I cleaned it a bit


kamathhrishi:
Oops , thank you @LaRiffle

syft/frameworks/torch/tensors/interpreters/polynomial.py
```diff
@@ -250,17 +235,15 @@ def sigmoid(self, x):
...
-            Parameters:
+        Method provides exponential function approximation interms of Taylor Series
 
+        Args:
             x: Torch tensor
 
-            return:
-
-            approximation of the sigmoid function as a torch tensor"""
+        return:
```
mari-linhares: "return" and "returns" should be "Returns"

PR #2155, opened 2019-05-21
Plans jit experimental

mari-linhares:
Testing serializing/deserializing capabilities of pytorch jit tracing.


examples/experimental/Plans Jit Experimental/plans-jit-virtual-workers-experimental.ipynb
```diff
@@ -0,0 +1,266 @@
+{
```
LaRiffle: <p>Oh actually I was expecting that we would consider the model as a plan (or any other forward function as a plan), and that the send would be done after serialisation is made (= plan is built): it comes for free with @torchi.jit.script and you have to build it with tracing, </p><p>Does it make sense?</p>

 _Reply via <a href='https://app.reviewnb.com/OpenMined/PySyft/pull/2155/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

syft/workers/base.py
```diff
@@ -118,6 +119,47 @@ def __init__(
                         if self.id not in worker._known_workers:
                             worker.add_worker(self)
 
+    def run_script_module(self, serialized_buffer: bin):
```
midokura-silvia: I propose to first focus on sending the object to the worker and loading it from the buffer. I don't see the need of a run_script_module() function as this would be covered by the TrainConfig developments. (Ok for testing, though :-) )

examples/experimental/Plans Jit Experimental/plans-jit-virtual-workers-experimental.ipynb
mari-linhares: <p>Yes, I thought the same thing. This was just for testing to see what error we would get.</p>

syft/workers/base.py
mari-linhares: This was for testing only :), I was checking if we were able to train a model remotely using the serialized model.

examples/experimental/Plans Jit Experimental/plans-jit-virtual-workers-experimental.ipynb
mari-linhares: @LaRiffle I think the only thing is not totally clear is if we're going to use IoBuffer as I used in this PR, are we? I considered we were so this is more of a proof of concept that serializing a torch script module we can train a model remotely :100:

LaRiffle: Ok cool! Yes I'm not completely sure but so far I don't see another easy way to do so, except perhaps maybe sending together this buffer + the torchscript.graph for non-torch models which can't use the buffer, but this is not very important at the moment I believe.

mari-linhares: Yeah, I agree it makes sense to restrain plans to torch operations for now, also using @jit.script instead of tracing can deal with if statements and for loops.

mari-linhares: I agree with @LaRiffle last suggestion of adding complementary information when using syft tensors/operations along with jit.

For us to have plans changed completely to use jit I think we need a proof of concept of jit using at least one syft's specific operation.

Issue #2110, opened 2019-04-29
Move simplifier and detailer from serde to specific objects


mari-linhares:
How serde [simplifiers](https://github.com/OpenMined/PySyft/blob/dev/syft/serde.py#L1290) and [detailers](https://github.com/OpenMined/PySyft/blob/dev/syft/serde.py#L1340) are related is not very intuitive.

Every time a new simplifier is created the index needs to be updated and the index created needs to match the index of the detail function at the detailers list. Especially because there is `simplifiers` and [forced_full_simplifiers](https://github.com/OpenMined/PySyft/blob/dev/syft/serde.py#L1312) which makes things even less obvious.

 I think that if `detailers` were a dictionary the relation would be more obvious.

robert-wagner:
I think it may be a better idea to move each simplifier and detailer to the file which defines the object. In that case serde would only need to keep the higher level logic and the simplifiers for built in python types making everything easier

mari-linhares:
Agree, I think a detailers dict/list would still be needed right?

racinger:
Hello @mari-linhares,

I am working on this issue.
Thanks,

midokura-silvia:
What was the conclusion of the discussion to move the serde functionality to the classes? It will not work for all cases, but for many it will work and make the maintenance easier.

mari-linhares:
Fixed #2230

PR #2157, opened 2019-05-21
[WIP] Move simplifier and detailer from serde to specific objects

racinger:
Hi everyone,
I moved some simplify_<object_name> functions and detail_<object_name> functions from serde.py to their object class. #2110 

After suggestion from @mari-linhares , I made those methods static. 
I am having issue running the pytest since I guess I am trying to import the class within itself.
Any idea? I need to pass as parameter an object of the class itself (maybe using self but not It can't be static anymore).

```
test/conftest.py:5: in <module>
    import syft
syft/__init__.py:3: in <module>
    from syft import frameworks
syft/frameworks/__init__.py:1: in <module>
    from . import torch
syft/frameworks/torch/__init__.py:1: in <module>
    from . import hook_args
syft/frameworks/torch/hook_args.py:6: in <module>
    from syft.frameworks.torch.tensors.interpreters import AbstractTensor
syft/frameworks/torch/tensors/__init__.py:2: in <module>
    from . import interpreters
syft/frameworks/torch/tensors/interpreters/__init__.py:1: in <module>
    from .pointer import PointerTensor  # noqa: F401
syft/frameworks/torch/tensors/interpreters/pointer.py:16: in <module>
    class PointerTensor(AbstractTensor):
syft/frameworks/torch/tensors/interpreters/pointer.py:330: in PointerTensor
    def _simplify_pointer_tensor(ptr: syft.PointerTensor) -> tuple:
E   AttributeError: module 'syft' has no attribute 'PointerTensor'
```
Thanks


syft/federated/plan.py
```diff
@@ -509,3 +512,56 @@ def __str__(self):
         out += ">"
 
         return out
+
+
+    @staticmethod
+    def _simplify_plan(plan: sy.Plan) -> tuple:
```
mari-linhares: you can call it just simply and detail in all objects now.

robert-wagner: Furthermore, you can just use Plan rather than sy.plan since it is inside the class itself

racinger: Sorry the imports I made are not correct. It looks like I can't use the object Plan as a parameter of the method within the Plan class.

Issue #2154, opened 2019-05-20
Should plans be implemented using torchscript?


mari-linhares:
Discussion started at [slack](https://openmined.slack.com/archives/C6EEFN3A8/p1558209789003100).

**Main pros:**

1) Faster development
2) Way simpler code base

**Main setbacks:**

1) Need to interface with non-pytorch environments (syft.js for federated learning across browsers, and Syft for Android for federated learning across phones)

But is not clear if torchscript works on javascript or java.

2) Torch JIT may or may not be a secure form of serialization

3) How to execute syft's operations within torchscript?

**POCS:**

- [x] Models can be serialized/deserialized and **trained** remotely (#2155)
- [ ] Plans using JIT (#2158)
- [ ] We can still serialize/deserialize syft's specific operations using jit (#2162)

**References:**
* http://blog.ezyang.com/2019/05/pytorch-internals/, inspiring to think about the features we want to build in PySyft.
* https://pytorch.org/docs/stable/jit.html also is great place for those thinking about our Plan implementation

mari-linhares:
@midokura-silvia

mari-linhares:
Question:can we re-use torch jit's implementation of [serialization](https://pytorch.org/docs/stable/jit.html#torch.jit.save) and still work PySyft specific tensors?

I believe ideally the operations performed by scriptModules would automatically adapt based on the input tensor (a.k.a if an encrypted tensor is given the output should also be an encrypted tensor), not sure if this would work out of the box.

mari-linhares:
#2155

mari-linhares:
The decision for now is: 

* plans will not be implemented using jit because we need to support syft tensors and operations inside of plans (for our best knowledge at this point we would need to change c++ code in order to do so).

* jit modules are super nice for regular FL (non-encrypted FL) so we'll support it (#2169).

PR #2158, opened 2019-05-21
Draft: Using jit script modules for serialization

midokura-silvia:
Working example to send a torch.jit.ScriptModule to a worker and call it with data.
See #2154
The work relies on generic pointers, see #2096


midokura-silvia:
Travis fails as there is a problem with torch 1.1.0, see https://github.com/pytorch/pytorch/issues/20017
All unit tests pass with torch 1.0.1.

PR #2160, opened 2019-05-21
Added support for LSTM and GRU Cells in PySyft and merged PR #1995

DanyEle:
- Merged changes of Pull Request #1995  with the latest 'dev' branch
- Added support for GRUs, GRU cells and LSTM cells in PySyft
- Added a workaround to fix the forward pass in  LSTMs with LSTM cells by returning a single tensor instead of a tuple of tensors (which seems to lead to errors in case of websockets). 
- Added a workaround to fix the torch.sort method when being executed over websockets by returning a tensor instead of a tuple of tensors.

I found out that returning multiple tensors in a tuple on remote websockets currently leads to a "ResponseSignatureError":

https://files.slack.com/files-pri/T6963A864-FJRTUCPSQ/image.png


DanyEle:
I think I have mistakenly deleted the Jupyter Notebook I created in my last Pull Request. My bad! I will also need to fix the formatting.

DanyEle:
Added back the test cases in https://github.com/OpenMined/PySyft/pull/1995/files

Note that once the tensor.dim() method is implemented for remote tensors in PySyft, also the forward pass into standard LSTMs is going to work.

syft/frameworks/torch/hook/hook.py
```diff
@@ -841,7 +861,6 @@ def _which_methods_should_we_auto_overload(self, tensor_type: type):
         """Creates a list of Torch methods to auto overload.
 
         By default, it looks for the intersection between the methods of
-        tensor_type and torch_type minus those in the exception list
         (syft.torch.exclude).
```
LaRiffle: A line has vanished

syft/frameworks/torch/hook/hook.py
```diff
@@ -828,6 +828,26 @@ def is_wrapper(self, it_is_a_wrapper):
...
+            """Hook the size to return the shape with a callable syntax"""
+            if dim is None:
+                return self.shape
+            return self.shape[dim]
+
+        torch.Tensor.native_size = torch.Tensor.size
+        torch.Tensor.size = size
+
+        def flip_hook_native_size(self):
+            """
```
LaRiffle: Thanks for re-using my hack, but actually I'm very unhappy with this hack, I don't think we should use it. I would either try to modify the PyTorch implement to have them use shape instead of size, or try another technique

syft/workers/base.py
```diff
@@ -375,6 +375,15 @@ def execute_command(self, message: tuple) -> PointerTensor:
...
 
+            # Temporary fix for websockets when returning a tuple of tensors from an LSTM cell
+            if command_name == "torch.lstm_cell":
+                response = torch.stack(response)
+
+            # Temporary fix for websockets when returning a tuple of tensors from torch.sort()
+            if command_name == "torch.sort":
+                Alpha_Tensor_Fixed = (response[0].float(), response[1].float())
+                response = torch.stack(Alpha_Tensor_Fixed)
+
```
LaRiffle: Same thing, this is a bit hacky and should not be merged in the code base. I understand that there is a pb, but let's open an Issue about this case when we use function with > 1 output and websockets

test/torch/test_federated_learning.py
```diff
@@ -88,3 +88,24 @@ def test_toy_federated_learning(self):
...
+    lstm = nn.LSTM(3, 3)
+    lstm.send(bob)
+    inputs = torch.randn(5, 1, 3).send(bob)
+    hidden = (
+        torch.randn(1, 1, 3).send(bob),
+        torch.randn(1, 1, 3).send(bob),
+    )  # clean out hidden state
+    # out, hidden = lstm(inputs, hidden)
+    # Test will fully pass once the tensor.dim() method is implemented for remote tensors
+    # assert out.shape == torch.Size([5, 1, 3])
```
LaRiffle: Would like to try implementing .dim() for remote tensors ? Could be  a cool contribution
You could rely on the .shape attribute which works for remote tensors

syft/frameworks/torch/hook/hook.py
DanyEle: Which line is no longer there exactly?

syft/frameworks/torch/hook/hook.py
DanyEle: So should we discard this fix for .size() completely?

syft/workers/base.py
DanyEle: Okay. I guess this would better fit an issue too.

test/torch/test_federated_learning.py
DanyEle: Would it be sufficient to just hook the .dim() method from the original PyTorch, or what exactly what would need to be done to implement it for remote tensors?

syft/frameworks/torch/hook/hook.py
mari-linhares: 844: tensor_type and torch_type minus those in the exception list

syft/frameworks/torch/hook/hook.py
LaRiffle: Yes, because we can't modify the hook at runtime it will be far too slow and bad practice.

test/torch/test_federated_learning.py
LaRiffle: If you hook dim() to use .shape, then it will also work for remote tensor: namely instead of returning the shape of the pointer itself it sends the call remotely to get the remote shape

DanyEle: Implemented dim() for remote tensors

syft/frameworks/torch/hook/hook.py
DanyEle: @mari-linhares Thank you for putting it back there!

syft/frameworks/torch/hook/hook.py
DanyEle: Removed the flip_hook_native_size

DanyEle:
Uhm... It looks like all tests related to remote websockets are now failing. What did I break?

DanyEle:
I looked into the remaining two tests not passing, and it seems to be related to AdditiveSharedTensors not having the dim() method. I tried to implement as:
```
def dim(self):
        sum_dim = 0
        for share in self.child.values():
            sum_dim += len(share.shape)
        return sum_dim
```

But this does not seem to be working. Any suggestions on how to implement the dim() method for AdditiveSharedTensors?

Jasopaum:
> I looked into the remaining two tests not passing, and they seem to be related to AdditiveSharedTensors not having the dim() method. I tried to implement as:
> 
> ```
> def dim(self):
>         sum_dim = 0
>         for share in self.child.values():
>             sum_dim += len(share.shape)
>         return sum_dim
> ```
> 
> But this does not seem to be working. Any suggestions on how to implement the dim() method for AdditiveSharedTensors?

Why do you sum for all the shares? I think this should work:
```
def dim(self):
    for share in self.child.values():
        return len(share.shape)
```

DanyEle:
> def dim(self):
>     for share in self.child.values():
>         return len(share.shape)

The following tests are not passing:

 test/torch/tensors/test_additive_shared.py::test_nn_linear FAILED        [ 55%]

test/torch/tensors/test_precision.py::test_torch_nn_functional_linear FAILED [ 92%]

I get the following error when running it this test:  test/torch/tensors/test_additive_shared.py::test_nn_linear:

DanyEle:
Fixed the remaining issues with the dim() method and the test cases not passing. Should be good to be merged now!

test/torch/test_federated_learning.py
LaRiffle: 👍

andrelmfarias:
Hi guys,

I am a new PySyft user and started to work with the library this weekend. 
I tried to train a toy GRU model remotely and got some errors related to the Tensor.size() method on the forward pass of the GRU layer.

I understand from #2201 that you have not implemented this method for remote tensors yet. 

What I understand from the error is that without the .size() method the GRU layer won't work.

Do you have an idea if this method is going to be implemented and when?

Thank you!

DanyEle:
> Hi guys,
> 
> I am a new PySyft user and started to work with the library this weekend.
> I tried to train a toy GRU model remotely and got some errors related to the Tensor.size() method on the forward pass of the GRU layer.
> 
> I understand from #2201 that you have not implemented this method for remote tensors yet.
> 
> What I understand from the error is that without the .size() method the GRU layer won't work.
> 
> Do you have an idea if this method is going to be implemented and when?
> 
> Thank you!

I guess you can just keep track of the conversation going on at #2201

Issue #2117, opened 2019-05-02
Conv2D not working in MPC


avitalsh:
Hi,
I tried running the code in [tutorial 10](https://github.com/OpenMined/PySyft/blob/master/examples/tutorials/Part%2010%20-%20Secure%20Deep%20Learning%20Classification.ipynb) with only changing the model to be a conv2D layer, followed by a fc layer:

```
class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.conv = nn.Conv2d(1, 3, (3,3), bias=False)
        self.fc = nn.Linear(3*26*26, 10, bias=False)

    def forward(self, x):
        x = self.conv(x)
        x = x.view(-1, 3*26*26)
        x = self.fc(x)
        return x

```
I got the following error - 
RuntimeError: _thnn_conv2d_forward is not implemented for type torch.LongTensor

I assume it has to do with the fact that conv2D supports float tesnsors, where SMPC works on integers (so we use .fix_prec()).

Is there a way to solve this issue and to perform secure inference with convolutional neural networks? 

Thanks

mari-linhares:
Hey @avitalsh,

> I assume it has to do with the fact that conv2D supports float tensors, where SMPC works on integers (so we use .fix_prec()).

I've just tried and I think this is the case as well..

> Is there a way to solve this issue and to perform secure inference with convolutional neural networks?

Yes, feel free to create a PR!

Jasopaum:
I am dealing with this issue

mari-linhares:
Fixed by #2165

PR #2165, opened 2019-05-23
Solving #2117 Conv2d not working in MPC

Jasopaum:
first commit: implementation without bias, padding, dilation


syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -476,6 +512,130 @@ def argmax(tensor, **kwargs):
...
+                else:
+                    res = res + bias
+
+            # ... And reshape it back to an image
+            res = (
+                res.permute(0, 2, 1)
+                .view(batch_size, nb_channels_out, nb_rows_out, nb_cols_out)
+                .contiguous()
+            )
+            return res
```
LaRiffle: You can do a bit simpler with the .child stuff here:
```
                res = torch.cat(res, dim=2)
            else:
                res = im_reshaped.matmul(weight_reshaped)

            # Add a bias if needed
            if bias is not None:
                res += bias

            # ... And reshape it back to an image
            res = (
                res.permute(0, 2, 1)
                .view(batch_size, nb_channels_out, nb_rows_out, nb_cols_out)
                .contiguous()
            )
            return res.child
```

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -505,6 +665,16 @@ def relu(tensor_shares):
...
+                def pad(input, pad, mode="constant", value=0):
+                    padded_shares = {}
+                    shares_dict = input.child.child if input.is_wrapper else input.child
+                    for location, shares in shares_dict.items():
+                        padded_shares[location] = torch.nn.functional.pad(shares, pad, mode, value)
+
+                    return AdditiveSharingTensor(padded_shares)
+
+                module.pad = pad
+
```
LaRiffle: You should use the decorator `@overloaded.function` which helps you to have a simpler code:
```
                @overloaded.function
                def pad(input_shares, pad, mode="constant", value=0):
                    padded_shares = {}
                    for location, shares in input_shares.items():
                        padded_shares[location] = torch.nn.functional.pad(shares, pad, mode, value)

                    return padded_shares

                module.pad = pad
```

LaRiffle:
Ah I said it is a really good PR!
From the tests I observed something is not clear - you can (should) never share additively floats (well in theory because you succeeded here ahah). Float should be converted to fixedprecision tensors
This way, you actually need to define conv2d only in fixed_precision.py not in additive_shared.py (I just tried it it works). This avoids duplicate code.
Also, the code in conv2d is really tricky to understand, you might want to add a few comments to explain how you transform your data to do the convolution in a single matrix multiplication, or link any useful reference.

Jasopaum:
Thanks for the review! I will modify the code and create a new PR.

For the additively shared floats, they actually were not floats but int tensors that's why it worked when I shared them. Do you still think we should remove conv2d from AdditivelySharingTensor? Maybe the users would like to share int tensors without using the fix_prec intermediary (I don't know if this makes sense).

PR #2173, opened 2019-05-28
Added code from udacity to unit tests

iamtrask:


test/test_udacity.py
```diff
@@ -0,0 +1,836 @@
...
+    y
+
+    x = th.tensor([0.1, 0.2, 0.3])
+
+    x = x.fix_prec().share(bob, alice, secure_worker)
+
+    y = x + x
+
+    y.get().float_prec()
+
```
robert-wagner: Delete these blank lines

test/test_udacity.py
```diff
@@ -0,0 +1,836 @@
...
+    x
+
+    bob._objects
+
+    alice._objects
+
+    assert True
+
+
+def test_section_3_securing_fl(hook):
```
robert-wagner: Should you have assertions in this code

test/test_udacity.py
```diff
@@ -0,0 +1,836 @@
...
+
+    x.move(bob)
+
+    x
+
+    bob._objects
+
+    alice._objects
+
+    assert True
```
robert-wagner: Never assert true. It's a pointless line

test/test_udacity.py
```diff
@@ -0,0 +1,836 @@
...
+
+    alice._objects
+
+    x.move(bob)
+
+    x
+
+    bob._objects
+
+    alice._objects
```
robert-wagner: should there be assertions about the contents of these objects

iamtrask: no - i'm just copy pasting the whole notebook - i'd rather not make edits

test/test_udacity.py
iamtrask: k

test/test_udacity.py
iamtrask: no - it's just checking that the code runs

test/test_udacity.py
iamtrask: they're separate cells in the notebook - at some point i'm going to have to come back and find bits of code to change,. the lines help somee

mari-linhares:
Why are polynomial tensor files commented?

iamtrask:
> Why are polynomial tensor files commented?

The test is failing in very unexpected ways (some kind of segfault which I couldn't track down).

Also, the rest of the codebase isn't yet using it, and it likely needs significant refactoring before the reset of the codebase can use it. In particular, it's not yet actually a tensor type, it's a class with static functions. We can revisit it after the course is released.

Issue #2176, opened 2019-05-29
Raise Error when trying to share additively float values


LaRiffle:
It makes to sense to do this:
```
x = torch.Tensor([0.0, 5.1])
x_shared = x.share(bob, alice, crypto_provider=james)
x_shared.get()
```
And it gives wrong results
We should check that the sharing is made on intgeres / Long / etc and not float

LaRiffle:
Hey, someone pinged me in slack about this Issue but Slack isn't very good to save un-responded message threads (https://superuser.com/questions/1199066/why-is-my-slack-dm-list-incomplete) so I lost _you_!
I would be glad to give me info but basically: moving to fixed precision is a constraint of the crypto protocols we use for additive sharing. This means that **everything** which needs to be secret shared should be in fixed precision (not only the neural network weights). Otherwise this violate the guarantees of correctness of the secret sharing protocols.
Note that you can still broadcast non private values with `.send(*workers)` which creates a MultiPointerTensor instead of an AdditiveSharedTensor. In that case it can be a float, as long as you don't want to operate it with AdditiveSharedTensors.

vghorakavi:
@LaRiffle Hey LaRiffle, I was the person who pinged you on Slack.  I see Jasopaum is working on the issue.  I will ping you for more information if I need it.  Thanks!

Note: My username on Slack used to be maxwells_daemon, but I changed it to vghorakavi to help keep things clear.

LaRiffle:
Thanks @vghorakavi !

pierrepocreau:
Hey, may I work on this issue ?

veenaneshwari:
@pierrepocreau I have not done any work on this issue. Please feel free to work on it.

pierrepocreau:
There are two problems around the maxpool.
Here, to calculate the derivative of maxpool, we use share on a torch.zeros tensor but it is of type FloatTensor.
https://github.com/OpenMined/PySyft/blob/f692ac551d41ebb4546274001ad192434b85e522/syft/frameworks/torch/mpc/securenn.py#L575-L577

Maybe we can use something else then torch.zeros ? or convert it to a LongTensor since it's only zeros.

Then to test the maxpool 2d, we call share on a FloatTensor
https://github.com/OpenMined/PySyft/blob/f6e8fd2d736422695c6c20d3d0e86e4c05600752/test/torch/crypto/test_snn.py#L171-L198

[...]

PR #2188, opened 2019-05-29
Modification to have operations between FPT>AST and FPT>torch

Jasopaum:
This seems to be a bit of a hack to me
But it can at least be the start of a discussion about how to enable scalar-FixedPrecisionTensor multiplications


mari-linhares:
Hey @Jasopaum,

Can you add some unit tests?

Jasopaum:
Sure, I will!
But I also wanted to know if someone has an idea for a better implementation

LaRiffle:
What is the use case for this?
By default, as (partially) explained in #2176, I'm opposed to mixing FixedPrecisionTensors with floats scalars, because FixedPrecisionTensors represent objects in a finite field where it is impossible to consider the notion of floats.

Jasopaum:
It was for training a model encrypted via additive sharing MPC. Specifically, we would need a float learning rate in gradient descent. But maybe we can find another way?

LaRiffle:
I see your point, but you should probably also convert the learning rate into fixed precision if you want to be coherent int terms of spaces you're operating in, even if in this specific case it is not elegant because we're used to lr being a float. So what I would suggest is having the lr being "only" FixedPrecision but operating with AdditiveSharedTensor.

But maybe I'm being too picky ahah - it's just that when we're dealing with the crypto protocols, if we change the usage and hypothesis we also change the guarantees in terms of security.

Jasopaum:
No, I don’t think you’re too picky and  changing the learning rate to Fixed Precision should be enough!
I can close this PR, then

Jasopaum:
So instead of what I did, we should enable multiplication between FixedPrecisionTensor > Tensor and FixedPrecisionTensor > AdditiveSharingTensor, which is not currently possible. Am I right?

LaRiffle:
Yes I believe you’re right !
I need to check on my laptop because we have a multiplication between an additivesharing tensor and normal tensor I think and it is called public_mul

Jasopaum:
Or we could also just multiply the learning rate by the scaling factor (same as the one used to convert to FixedPoint) and do something similar to what I did but only with integer values
Do you think it is better?

LaRiffle:
> Or we could also just multiply the learning rate by the scaling factor (same as the one used to convert to FixedPoint) and do something similar to what I did but only with integer values
> Do you think it is better?

This is exactly what FixedPrecision does for you, let's keep this abstraction in this case.


So I checked, `public_mul` exists but currently does not allow to do this. It allows you to multiply an additiveSharedTensor with a MultiPointerTensor (`other`). So cool extension would be when other is an int or long, convert it in MultiPointerTensor in the public_mul function (l.332 in additive_shared.py)
This will able you to run:
``` 
t = torch.tensor([1., 2, 3, 4]).fix_precision().share(alice, bob, crypto_provider=james)
lr = torch.tensor([0.5]).fix_precision()
t.child.child.child * lr.child.child
```
Then to have this `t * lr` working you need to modify also `mul` in precision.py

Jasopaum:
> > Or we could also just multiply the learning rate by the scaling factor (same as the one used to convert to FixedPoint) and do something similar to what I did but only with integer values
> > Do you think it is better?
> 
> This is exactly what FixedPrecision does for you, let's keep this abstraction in this case.
> 
> So I checked, `public_mul` exists but currently does not allow to do this. It allows you to multiply an additiveSharedTensor with a MultiPointerTensor (`other`). So cool extension would be when other is an int or long, convert it in MultiPointerTensor in the public_mul function (l.332 in additive_shared.py)
> This will able you to run:
> 
> ```
> t = torch.tensor([1., 2, 3, 4]).fix_precision().share(alice, bob, crypto_provider=james)
> lr = torch.tensor([0.5]).fix_precision()
> t.child.child.child * lr.child.child
> ```
> 
> Then to have this `t * lr` working you need to modify also `mul` in precision.py

I thought about something like this but I didn't know how to handle the different number of layers for each operand (here, you did it by calling child several times). I wanted to have something where the user only has to call `t * lr`

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -256,8 +256,11 @@ def add(self, shares: dict, other_shares):
                 to the tensor being added to self.
         """
 
-        # if someone passes in a constant... (i.e., x + 3)
-        if not isinstance(other_shares, dict):
+        if isinstance(other_shares, torch.LongTensor) or isinstance(other_shares, torch.IntTensor):
+            # if someone passes a torch tensor
+            other_shares = other_shares.share(*self.child.keys()).child.child
+        elif not isinstance(other_shares, dict):
```
LaRiffle: Actually there was an error in .share(): we should add the crypto_provider which can be found in self.crypto_provider

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -286,8 +289,11 @@ def sub(self, shares: dict, other_shares):
                 to the tensor being subtracted from self.
         """
 
-        # if someone passes in a constant... (i.e., x - 3), make it a shared tensor and keep the dict
-        if not isinstance(other_shares, dict):
+        if isinstance(other_shares, torch.LongTensor) or isinstance(other_shares, torch.IntTensor):
+            # if someone passes a torch tensor
+            other_shares = other_shares.share(*self.child.keys()).child.child
```
LaRiffle: Same here

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -94,8 +94,18 @@ def truncate(self, precision_fractional):
 
     @overloaded.method
     def add(self, _self, *args, **kwargs):
-        """Add two fixed precision tensors together.
-        """
+        other = args[0]
```
LaRiffle: you can change the signature to be just add(self, _self,  other)

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -109,6 +119,32 @@ def __iadd__(self, other):
 
         return self
 
+    @overloaded.method
+    def sub(self, _self, *args, **kwargs):
+        other = args[0]
```
LaRiffle: same

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -122,10 +158,32 @@ def mul(self, *args, **kwargs):
         in the fixed precision setting
         """
 
-        # Replace all syft tensor with their child attribute
-        new_self, new_args, new_kwargs = syft.frameworks.torch.hook_args.hook_method_args(
-            "mul", self, args, kwargs
-        )
+        other = args[0]
```
LaRiffle: same

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -202,11 +270,26 @@ def eq(self, _self, other):
...
+        def add(self, other):
+            return self.__add__(other)
+
+        module.add = add
+
+        def sub(self, other):
+            return self.__sub__(other)
+
+        module.sub = sub
+
```
LaRiffle: Have you added tests for those "redirections"?

Jasopaum: I did

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -109,38 +120,77 @@ def __iadd__(self, other):
...
+        if _self.is_wrapper and not other.is_wrapper:
+            # If we try to add a FPT>(wrap)>AST and a FPT>torch.tensor),
+            # we want to perform AST + torch.tensor
+            _self = _self.child
+            other = other.wrap()
+        elif other.is_wrapper and not _self.is_wrapper:
+            # If we try to subtract a FPT>torch.tensor and a FPT>(wrap)>AST,
+            # we swap operators so that we do the same operation as above
+            other = _self.wrap()
+            _self = other.child
```
LaRiffle: Some thing here, you can't swap like this because it's a sub op ;)

examples/tutorials/Part 12 - Train an Encrypted Neural Network on Encrypted Data.ipynb
```diff
@@ -96,12 +94,8 @@
...
     "model_enc = model.fix_precision().share(alice, bob, crypto_provider=james)\n",
     "\n",
-    "# For now, we also need to encode the learning rate but this will change in the near future\n",
-    "learning_rate_fp = (\n",
-    "    torch.Tensor([learning_rate])\n",
-    "    .fix_precision()\n",
-    "    .share(alice, bob, crypto_provider=james)\n",
-    ")"
+    "# The learning rate is also changed to fixed precision\n",
+    "learning_rate_fp = torch.tensor(learning_rate).fix_precision()"
```
mari-linhares: Just wondering, why  `.share(alice, bob, crypto_provider=james)` was removed?

LaRiffle: Because if some settings we don't want to hide the config used, to the learning rate should be public, ie just a normal fixed precision tensor

PR #2197, opened 2019-05-30
corrected import error for PointerTensor

amit-rastogi:
modified the import statement for PointerTensor to syft.frameworks.torch.pointers from syft.frameworks.torch.tensors.interpreters


mari-linhares:
Thanks @amit-rastogi !

PR #2204, opened 2019-05-31
Fix move() method by removing duplicate storage

LaRiffle:


syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -461,7 +461,10 @@ def mid_get(self):
 
         child_id = self.child.id
         tensor = self.child.get()
-        self.owner._objects[child_id] = tensor
+
+        self.owner.de_register_obj(tensor)
```
midokura-silvia: By first de_registering the tensor you handle the case, that the id changed, right? Otherwise you just register the object again that you just unregistered.

LaRiffle: For move(), the id should always change and be the remote one 
But by default it is not the case when doing .get() with pointers, so I remove the storage of the tensor with old id, change the id, and store it with this new id

mari-linhares: I think this line is not needed after the changes in #2199, can you check if that's true?

LaRiffle: You're right! I'll make a change

LaRiffle: #2206

Issue #2214, opened 2019-05-31
Exception forwarding across workers


LaRiffle:
**Is your feature request related to a problem? Please describe.**
When a remote workers fails, the client doesn't has access directly to the error (and traceback). In particular, the error needs to be caught for response of operations with several pointers where a ResponseSignatureError is raised by a worker and caught by another

**Describe the solution you'd like**
I'd like remote socket workers to catch all (or some high-level) errors and serialize them so that the client can get and catch them, or error with a full message

**Describe alternatives you've considered**
we can use `pip install tblib`
example: 
```
import tblib.pickling_support
tblib.pickling_support.install()
import pickle, sys 

def inner_0():
    raise Exception('fail')

def inner_1():
    inner_0()

def inner_2():
    inner_1()

try:
    inner_2()
except:
    s1 = pickle.dumps(sys.exc_info())

# deserialize and raise again error
from six import reraise
reraise(*pickle.loads(s1))
```

robert-wagner:
@LaRiffle Is this done?

LaRiffle:
Yes!

PR #2219, opened 2019-06-01
Add handling of ResponseSignatureError across socket workers

LaRiffle:
Provides an implementation to solve #2214

Version 1:
Keep all stack trace just like if the error was happening locally, but uses pickle c742c9b

Version 2:
Keep a serialized version of the stack trace which contains less information, but doesn't use pickle 8466e4f

### DIFF
--- a/syft/serde.py
+++ b/syft/serde.py
@@ -43,13 +43,12 @@ from lz4 import (  # noqa: F401
 )  # needed as otherwise we will get: module 'lz4' has no attribute 'frame'
 import io
 import numpy
+from tblib import Traceback
+import traceback
+from six import reraise
+import sys
 import warnings
 import zstd
-import tblib.pickling_support
-
-tblib.pickling_support.install()
-import pickle, sys
-from six import reraise
 
 import syft
 import syft as sy
@@ -1199,6 +1198,43 @@ def _detail_GetNotPermittedError(
     raise GetNotPermittedError(error_tuple[0])
 
 
+def _simplify_exception(e):
+    """
+    Serialize information about an Exception which was raised to forward it
+    """
+    # Get information about the exception: type of error, error obj, traceback
+    tp, value, tb = sys.exc_info()
+    # Serialize the traceback
+    traceback_str = "Traceback (most recent call last):\n" + "".join(traceback.format_tb(tb))
+    # Include special attributes if relevant
+    if hasattr(e, "get_attributes"):
+        attributes = e.get_attributes()
+    else:
+        attributes = {}
+    return tp.__name__, traceback_str, _simplify(attributes)
+
+
+def _detail_exception(worker: AbstractWorker, error_tuple: Tuple[str, str, dict]):
+    """
+    Re-raise an Exception forwarded by another worker
+    """
+    error_name, traceback_str, attributes = error_tuple
+    error_name, traceback_str = error_name.decode("utf-8"), traceback_str.decode("utf-8")
+    attributes = _detail(worker, attributes)
+    # De-serialize the traceback
+    tb = Traceback.from_string(traceback_str)
+    # Check that the error belongs to a valid set of Exceptions
+    if error_name in dir(sy.exceptions):
+        error_type = getattr(sy.exceptions, error_name)
+        error = error_type()
+        # Include special attributes if any
+        for attr_name, attr in attributes.items():
+            setattr(error, attr_name, attr)
+        reraise(error_type, error, tb.as_traceback())
+    else:
+        raise ValueError(f"Invalid Exception returned:\n{traceback_str}")
+
+
 def _force_full_simplify_worker(worker: AbstractWorker) -> tuple:
     """
 
@@ -1242,16 +1278,6 @@ def _detail_object_wrapper(
         id=obj_wrapper_tuple[0], obj=_detail(worker, obj_wrapper_tuple[1])
     )
     return obj_wrapper
-
-
-def _simplify_exception(e):
-    error_message = sys.exc_info()
-    return pickle.dumps(error_message)
-
-
-def _detail_exception(worker: AbstractWorker, error_tuple: tuple):
-    error = pickle.loads(error_tuple)
-    reraise(*error)
 
 
 def _simplify_script_module(obj: torch.jit.ScriptModule) -> str:
--- a/syft/serde.py
+++ b/syft/serde.py
@@ -1354,7 +1354,7 @@ simplifiers = {
     MultiPointerTensor: [14, _simplify_multi_pointer_tensor],
     Plan: [15, _simplify_plan],
     VirtualWorker: [16, _simplify_worker],
-    str: [17, _simplify_str],
+    str: [18, _simplify_str],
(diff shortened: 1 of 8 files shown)


LaRiffle:
There is an error with `test_serde_object_wrapper_traced_module` which was integrated recently, I'd love some help on this

syft/serde.py
```diff
@@ -1238,6 +1244,16 @@ def _detail_object_wrapper(
     return obj_wrapper
 
 
+def _simplify_exception(e):
+    error_message = sys.exc_info()
+    return pickle.dumps(error_message)
```
robert-wagner: I'm somewhat opposed to using pickle for this as pickle is not supported on other systems (Ie js) further we are doing nothing with the error passed to this function. Is there a better way to do this? For other errors that we have serialized in the past, we built a manual serializer

syft/serde.py
```diff
@@ -1238,6 +1244,16 @@ def _detail_object_wrapper(
...
+
+
+def _detail_exception(worker: AbstractWorker, error_tuple: tuple):
+    error = pickle.loads(error_tuple)
+    reraise(*error)
```
robert-wagner: Can we not just use raise for this?

syft/serde.py
LaRiffle: You're right, the good thing with this is that you get the same stacktrace with code from remote workers included, which is a bit amazing. But in this very case the exception ResponseSignatureError is caught so you don't get to see how cool the stacktrace is.

I can try an alternative solution

syft/serde.py
LaRiffle: You need reraise to include the stacktrace, which is needed in the case you want to debug or if the error was unexpected and wasn't caught by the client

syft/serde.py
```diff
@@ -1173,24 +1178,41 @@ def _detail_worker(worker: AbstractWorker, worker_tuple: tuple) -> pointers.Poin
...
-    Raises:
-        GetNotPermittedError: the error thrown when get is not permitted
+def _simplify_exception(e):
     """
-
-    raise GetNotPermittedError(error_tuple[0])
+    Serialize information about an Exception which was raised to forward it
+    """
+    # Get information about the exception: type of error, error obj, traceback
+    tp, value, tb = sys.exc_info()
```
robert-wagner: should we not get this info from e?

syft/serde.py
```diff
@@ -1173,24 +1178,41 @@ def _detail_worker(worker: AbstractWorker, worker_tuple: tuple) -> pointers.Poin
...
+    # De-serialize the traceback
+    tb = Traceback.from_string(traceback_str)
+    # Check that the error belongs to a valid set of Exceptions
+    if error_name in dir(sy.exceptions):
+        error_type = getattr(sy.exceptions, error_name)
+        error = error_type()
+        # Include special attributes if any
+        for attr_name, attr in attributes.items():
+            setattr(error, attr_name, attr)
+        reraise(error_type, error, tb.as_traceback())
```
robert-wagner: can we use raise instead of reraise?

syft/serde.py
```diff
@@ -1332,17 +1354,18 @@ def _force_full_simplify(obj: object) -> object:
     MultiPointerTensor: [14, _simplify_multi_pointer_tensor],
     Plan: [15, _simplify_plan],
     VirtualWorker: [16, _simplify_worker],
-    GetNotPermittedError: [17, _simplify_GetNotPermittedError],
-    str: [18, _simplify_str],
-    pointers.ObjectWrapper: [20, _simplify_object_wrapper],
+    str: [17, _simplify_str],
```
robert-wagner: why is str no longer 18

syft/workers/base.py
```diff
@@ -468,7 +468,7 @@ def respond_to_obj_req(self, obj_id: Union[str, int]):
 
         obj = self.get_obj(obj_id)
         if hasattr(obj, "allowed_to_get") and not obj.allowed_to_get():
-            return GetNotPermittedError()
+            raise GetNotPermittedError()
```
robert-wagner: raising this does not pass the value to the other machine in the case of websocket workers

syft/workers/websocket_server.py
```diff
@@ -103,6 +109,12 @@ def __init__(
             # send the response
             await websocket.send(response)
 
+    def _recv_msg(self, message: bin) -> bin:
+        try:
+            return self.recv_msg(message)
+        except ResponseSignatureError as e:
```
robert-wagner: Should we do this for all exceptions?

syft/serde.py
LaRiffle: It's equivalent actually

syft/serde.py
LaRiffle: No you can't do this as far as I know, if you want to keep the stack trace which we want to. You are forced to use reraise

syft/serde.py
LaRiffle: Because I removed 17:GetNotPermittedError as my exception ser/de supports also kinds of errors, so all numbers have been decremented after 17

syft/workers/base.py
LaRiffle: Ok I had missed one line, but now it does: it's all the point of my generic exception serializer

syft/workers/websocket_server.py
LaRiffle: Not sure, for debug purpose maybe, but it depends on the usecases:
*Definitely*
- Exception which are normally in use like ResponseSignatureError and are normally caught
*Surely*
- Exception related to FL stuff like GetPermittedError, WorkerNotFoundException, TensorsNotCollocatedException etc
*Maybe not*
- Exception unrelated to FL which cause a worker to die (MemoryError, etc), in which case we just want to launch a new worker

syft/serde.py
```diff
@@ -1332,17 +1354,18 @@ def _force_full_simplify(obj: object) -> object:
...
+    pointers.ObjectWrapper: [19, _simplify_object_wrapper],
+    GetNotPermittedError: [20, _simplify_exception],
```
robert-wagner: This is very nit pick but can we make exceptions 17 instead? This will allow some compatibility across versions

syft/serde.py
```diff
@@ -1388,9 +1411,9 @@ def _detail(worker: AbstractWorker, obj: object) -> object:
     _detail_multi_pointer_tensor,
     _detail_plan,
     _detail_worker,
-    _detail_GetNotPermittedError,
-    _detail_str,
     _force_full_detail_worker,
+    _detail_str,
     _detail_object_wrapper,
+    _detail_exception,
```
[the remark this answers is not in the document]
LaRiffle: I don't think it really matters, this is a crap way of doing things, we'll change it anyway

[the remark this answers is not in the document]
robert-wagner: I just mean from a correctness stand point. This list should be in the order of the simplifiers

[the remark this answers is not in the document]
LaRiffle: I think the order between simplifier and detailers is preserved

Issue #2211, opened 2019-05-31
Test share for keras does not work on macos in a virtual env


robert-wagner:
**Describe the bug**
Due to firewall issues, it is not possible to run test share on macos in a venv. This can be remedied by replacing network calls with mock.
**To Reproduce**
run make_test on a fresh macos install

**Additional context**
contact @robert-wagner to replicate as it is still broken on his system

mortendahl:
Mind taking a look at the linked PR @robert-wagner?

PR #2229, opened 2019-06-03
Run tests using tfe.LocalConfig

mortendahl:
Tests where previously running using a `tfe.RemoteConfig` but that is fragile due to process starting and concurrency. This patch changes the tests to use `tfe.LocalConfig` instead.

Closes https://github.com/OpenMined/PySyft/issues/2211.

### DIFF
--- a/syft/frameworks/keras/model/sequential.py
+++ b/syft/frameworks/keras/model/sequential.py
@@ -15,7 +15,7 @@ _args_not_supported_by_tfe = [
 ]
 
 
-def share(model, *workers, target_graph=None):  # pragma: no cover
+def share(model, *workers, target_graph=None):
     """
     Secret share the model between `workers`.
 
--- a/syft/frameworks/keras/model/sequential.py
+++ b/syft/frameworks/keras/model/sequential.py
@@ -58,7 +58,6 @@ def share(model, *workers, target_graph=None):  # pragma: no cover
 
     # Push and initialize shared model on servers
     sess = tfe.Session(graph=target_graph)
-    tf.Session.reset(sess.target)
     sess.run(initializer)
 
     model._tfe_session = sess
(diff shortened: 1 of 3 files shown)


syft/frameworks/keras/model/sequential.py
```diff
@@ -58,7 +58,6 @@ def share(model, *workers, target_graph=None):  # pragma: no cover
 
     # Push and initialize shared model on servers
     sess = tfe.Session(graph=target_graph)
-    tf.Session.reset(sess.target)
```
robert-wagner: Should we add coverage back to this function?

test/keras/test_sequential.py
```diff
@@ -43,7 +43,6 @@ def test_instantiate_tfe_layer():
     np.testing.assert_allclose(actual, expected, rtol=0.001)
 
 
-@pytest.mark.skip(reason="Currently breaks on macos")
 def test_share():  # pragma: no cover
```
robert-wagner: We should remove the no cover pragma if we are running this test again

robert-wagner:
You need to run `black syft` to get this pr to pass travis. Alternatively by running `make install_hooks` black will run automatically before every commit

syft/frameworks/keras/model/sequential.py
mortendahl: What do you mean here?

robert-wagner: at the top of the function there is a comment of "#pragma: no cover" which prevents coverage from being run on this function. We should remove that if this function is now being tested

mortendahl: Gotcha, and done!

robert-wagner: Did you push your changes?

mortendahl: jup

PR #2230, opened 2019-06-04
Create a serde folder with serde.py, native_serde.py and torch_serde.py

racinger:
Hello,
I split  _simplify_<object> and _detail_<object>  methods into 3 files:

- one for native python object
- one for torch object
- one for high level serializer

Please let me know what do you think. Thanks
It is a follow up of #2157 and #2110


racinger:
@mari-linhares kindly this is ready to be reviewed. Thanks a lot :)

syft/serde/serde.py
```diff
@@ -0,0 +1,476 @@
...
+        right = right(obj)
+
+        result = (left, right)
+    else:
+        result = _simplify(obj)
+
+    return result
+
+
+simplifiers = {
```
mari-linhares: I'm not sure if @robert-wagner @LaRiffle agree, but I would love to get rid of this dict, or maybe store simplifiers and detailers in the same dict.

mari-linhares: Do we need to keep this for efficiency reasons?

syft/serde/serde.py
```diff
@@ -0,0 +1,476 @@
...
+    _detail_collection_list,
+    _detail_collection_set,
+    _detail_dictionary,
+    _detail_ndarray,
+    _detail_torch_device,
+    _detail_pointer_tensor,
+    _detail_log_tensor,
+    _detail_additive_shared_tensor,
+    _detail_multi_pointer_tensor,
+    _detail_plan,
```
mari-linhares: _simplify_<syft_object> or _detail_<syft_object> should be replaced by sy.<syft_object>.simplify or sy.<syft_object>.detail.

So we would use sy.Plan.simplify instead of _simplify_plan, sy.PointerTensor.simplify instead of _simplify_pointer_tensor

syft/serde/torch_serde.py
```diff
@@ -1,53 +1,19 @@
...
-are the types and values are the simplification logic. For example,
-simplifiers[tuple] will return the function which knows how to simplify the
-tuple type. The same is true for all other simplifier/detailer functions.
-
-By default, the simplification/detail operations expect Torch tensors. If the setup requires other
-serialization process, it can override the functions _serialize_tensor and _deserialize_tensor
-
-By default, we serialize using msgpack and compress using lz4.
-If different compressions are required, the worker can override the function _apply_compress_scheme
+This file exists to provide one common place for all serialisation and simplify_ and _detail
```
mari-linhares: this looks good.

syft/serde/serde.py
mari-linhares: given that we really need it, my vote is: make a list of pairs containing (simplifier, detailer) pairs, a dict can later be built automatically using this list during init

syft/federated/plan.py
```diff
@@ -132,7 +131,7 @@ def _recv_msg(self, bin_message: bin):
         Returns:
             The None message serialized to specify the command was received.
         """
-        (some_type, (msg_type, contents)) = sy.serde.deserialize(bin_message, detail=False)
+        (some_type, (msg_type, contents)) = sy.serde.serde.deserialize(bin_message, detail=False)
```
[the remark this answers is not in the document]
racinger: I made that change. Thanks for the tips

syft/workers/base.py
```diff
@@ -245,14 +245,14 @@ def recv_msg(self, bin_message: bin) -> bin:
...
-        (msg_type, contents) = sy.serde.deserialize(bin_message, worker=self)
+        (msg_type, contents) = sy.serde.serde.deserialize(bin_message, worker=self)
         if self.verbose:
             print(f"worker {self} received {sy.codes.code2MSGTYPE[msg_type]} {contents}")
         # Step 1: route message to appropriate function
         response = self._message_router[msg_type](contents)
 
         # Step 2: Serialize the message to simple python objects
-        bin_response = sy.serde.serialize(response)
+        bin_response = sy.serde.serde.serialize(response)
```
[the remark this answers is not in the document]
racinger: same updated it.

syft/federated/plan.py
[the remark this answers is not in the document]
racinger: I changed all serde.serde to just serde.
I am having trouble for the next step you are suggesting. 
I try the following without success
I added to __init__ of syft
```
# Import serialization tools
from syft.serde import serde
from syft.serde import torch_serde
from syft.serde import native_serde
```
Then I added the following into serde/init.py
```
from syft.serde.serde import serialize
```
Finally in plan.py added the following, it seems that it creates some circular import which create an error with pytest
```
from syft.serde import serde
```
I am working on it, if you have a work around, please let me know. Thanks :)

syft/serde/serde.py
racinger: One tiny question :
For me It is not possible to => create a method simplify within Plan class for example where the method simplify take as an object a Plan object type.

```
class Plan
      @staticmethod
      def simplify(plan: Plan) -> tuple:
           ....
```
Should i just use a basic object but we will lost the strong Type of Plan
```
class Plan
      @staticmethod
      def simplify(plan: object) -> tuple:
           ....
```
After that it might still have some circular import issue because of recursion call within methods simplify and detail

syft/serde/serde.py
robert-wagner: I hate this dict but it does have some advantages for performance. That being said, if we could get rid of it I would be very happy

test/workers/test_virtual.py
```diff
@@ -5,7 +5,7 @@
 from syft.exceptions import GetNotPermittedError
 from syft.workers.virtual import VirtualWorker
 from syft.codes import MSGTYPE
-from syft import serde
+from syft.serde import serde
```
robert-wagner: change to `from syft import serde`

test/test_serde.py
```diff
@@ -556,17 +556,17 @@ def test_pointer_tensor_detail(id):
 
 
 def test_numpy_tensor_serde():
-    syft.serde._serialize_tensor = syft.serde.numpy_tensor_serializer
-    syft.serde._deserialize_tensor = syft.serde.numpy_tensor_deserializer
+    syft.torch_serde._serialize_tensor = syft.torch_serde.numpy_tensor_serializer
```
robert-wagner: can we change this back. I don't want to know that three files exist for these unless I have to. In syft.serde.__init__.py if you add `from torch_serde import *` it should fix this

test/workers/test_virtual.py
racinger: done thanks

syft/serde/serde.py
LaRiffle: I would suggest doing something like in hook_args: the dict is built dynamically during execution: so each time you add an entry you have a tiny overhead and then you have the dict look-up speed. This would use a list of pairs as suggested by @mari-linhares to build on this

syft/federated/plan.py
```diff
@@ -458,7 +457,7 @@ def _send(self, location: "sy.workers.BaseWorker"):
         _ = self.owner.send(self, workers=location)
 
         # Deep copy the plan without using deep copy
-        pointer = sy.serde._detail_plan(self.owner, sy.serde._simplify_plan(self))
+        pointer = sy.torch_serde._detail_plan(self.owner, sy.torch_serde._simplify_plan(self))
```
LaRiffle: Maybe this is a detail, but is there a way to make it transparent? I mean: I'd like to still call sy.serde.detail_<mytype> no matter if detail_<mytype>  is in serde/torch_serde or serde/syft_serde etc
It would require to change the way the imports are done. If this is not too complicated I think it would make it easier to use in non-standard cases like this one.

syft/serde/serde.py
LaRiffle: What if it is not static?

test/test_serde.py
racinger: done it by adding to serde/init.py
```from syft.serde.serde import *
from syft.serde.torch_serde import *
```

syft/serde/serde.py
racinger: It looks like it is possible to do it with this notation:

```
class Plan
      @staticmethod
      def simplify(plan: "Plan") -> tuple:
           ....
```
I ll try this way

syft/federated/plan.py
racinger: I can do it if change the detail method from protected to public. (from _detail_plan to detail_plan)
Is that ok?

LaRiffle: Are you forced to do this? I believe it works also with the "_" but maybe I'm wrong

syft/serde/serde.py
LaRiffle: Good!

racinger: @mari-linhares 
> _simplify_<syft_object> or _detail_<syft_object> should be replaced by sy.<syft_object>.simplify or sy.<syft_object>.detail.
> 
> So we would use sy.Plan.simplify instead of _simplify_plan, sy.PointerTensor.simplify instead of _simplify_pointer_tensor

I believe I can achieve your requirement.
For that i need to put all functions from protected to public (mainly for _detail and simplify to detail and simplify).
Is it an issue?

mari-linhares: I don't think so :)

syft/serde/serde.py
mari-linhares: sgtm

syft/federated/plan.py
racinger: You are right. At least after moving _detail_plan in Plan class, it does work

LaRiffle: Yes in practice python doesn't really care about "protected", it's more like a convention for users and developers

racinger: To conclude on that, I believe the following happened
serde/init.py :
```
from syft.serde.serde import *
from syft.serde.torch_serde import *
```
In this case it seems that it does not import methods with "_" , 
so I couldn't use sy.serde._simplify() in plan.py for example

So I added in the serde.init.py
```
from syft.serde.serde import _simplify
```
Then it does work

LaRiffle: Ah ok, good to know! Thanks

syft/serde/serde.py
racinger: Should I move the following _detail and _simplify : [torch_device, exception, worker, full_detail_worker, script_module] to the AbstractWorker class ?

racinger: Should I move more  _detail and _simplify methods? 
Exceptions, workers, torch.device, torch.jit ?

syft/serde/serde.py
racinger: Hi @mari-linhares and @LaRiffle, 

I am trying to implement the hook_args @LaRiffle suggested. A bit lost to where to start.
I checked the file frameworks/torch/hook/hook_args.py.
Is it similar?
Still confused, would love to get some guidance on it? Thanks
Is the args would be same at each execution?

LaRiffle: So the mechanism would be the following:
Instead of 
```
simplifiers = {
    torch.Tensor: [0, _simplify_torch_tensor],
    torch.nn.Parameter: [1, _simplify_torch_parameter],
    tuple: [2, _simplify_collection],
    list: [3, _simplify_collection],
```
Have
```
simplifiers = {}
map_simplifiers = {
    torch.Tensor: (_simplify_torch_tensor, _detail_torch_tensor),
    torch.nn.Parameter: (_simplify_torch_parameter, _detail_torch_parameter),
    etc
detailers = []
```
Then in def _simplify, check if simplifiers[type] succeeds, in case of key Error,  create
simplifiers[type]= [index, func] where func = map_simplifiers[key][0] and append map_simplifiers[key][1] in a detailer list ; also increment index

This is the first step, does it make sense?

mari-linhares: I think we can do this change in a new PR :relaxed:

syft/serde/serde.py
mari-linhares: My opinion: exceptions and workers should live in their own files (like you did for plans); torch.device, torch.jit should live in torch_serde.py

racinger: I do agree :)

syft/serde/serde.py
racinger: Ok I ll open another one for that right after closing this one I guess.
Thanks @LaRiffle , I do get it and working on it.

syft/federated/plan.py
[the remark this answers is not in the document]
midokura-silvia: there are different __init__.py files.
In the  `__init__.py`  in the _serde folder_ you can add
`from syft.serde import serde`
 
And in plan.py you add
`from syft import serde`

Then you can use serde.serialize throughout the file plan.py

racinger:
I believe I am ready we are ready for review @mari-linhares @robert-wagner @midokura-silvia . Thanks a lot

syft/federated/plan.py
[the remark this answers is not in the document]
racinger: By doing so,
It creates many import errors. (circular imports?)
Basically I can't import TrainConfig, Plan, AdditiveSharingTensor and MultiPointerTensor in serde.py and torch_serde.py anymore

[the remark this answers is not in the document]
racinger: @midokura-silvia , I have been working on it, unfortunately I believe we have too many circular imports. I don't see any clean solution for now. Let me know. Thanks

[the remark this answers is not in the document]
midokura-silvia: All right, I can't look into it myself at the moment. If the others are okay with it, let's go with it as it is for the moment.

PR #2237, opened 2019-06-05
Raising error if PointerTensor got called with .item

abmazhr:
- Raising error if PointerTensor got called with .item with an error message to call .get instead

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -24,9 +24,9 @@ copyright = "2019, OpenMinedContributors"
 author = "Andrew Trask"
 
 # The short X.Y version
-version = "0.1.16a1"
+version = "0.1.17"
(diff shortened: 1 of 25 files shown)


abmazhr:
@iamtrask kindly that can be reviewed so it can be merged if it's fine, thank you :smile:

abmazhr:
I guess because I transferred this repository from my old account to this new account, it had old changes, so the above changes are to get the current master merged in my fork then the latest 4 commits is related to the actual issue in this PR (since the previous changes are not even my changes right? it's from the original repository). Sure please let me know if I misunderstood something or if I should check my branches again :smile:

abmazhr:
@robert-wagner closing this for a new PR, fixed the noise of unrelated previous things #2239

PR #2238, opened 2019-06-05
TrainConfig basic implementation

mari-linhares:


test/federated/test_federated_client.py
```diff
@@ -61,14 +52,17 @@ def test_set_obj_other():
...
 
 
-def test_fit():
+@pytest.mark.skip(reason="bug in pytorch version 1.1.0, jit.trace returns raw C function")
+def test_fit(hook):  # pragma: no cover
+    fed_client = sy.VirtualWorker(hook=hook, id="test_fit_fc")
+
     data = torch.tensor([[-1, 2.0], [0, 1.1], [-1, 2.1], [0, 1.2]], requires_grad=True)
-    target = torch.tensor([[1], [0], [1], [0]])
+    target = torch.tensor([[1], [0], [1.0], [0]], requires_grad=True)
```
midokura-silvia: why do you want the targets to be float?

test/federated/test_federated_client.py
```diff
@@ -61,14 +52,17 @@ def test_set_obj_other():
     assert fed_client._objects[dummy_data.id] == dummy_data
 
 
-def test_fit():
```
midokura-silvia: Can we leave this test as it was, please?
The goal of this test was to have a unit test, that does not need to be skipped (does not depend on jit). 
Also, no workers are used and it was a UNIT test for the FederatedClient only.
I added these tests to get the coverage above 95%.

syft/frameworks/torch/pointers/object_wrapper.py
```diff
@@ -84,5 +84,6 @@ def create_pointer(
             id_at_location=id_at_location if id_at_location is not None else self.id,
             tags=self.tags,
             description=self.description,
+            garbage_collect_data=False,
         )
```
LaRiffle: Do you really want to never GC collect an object wrapper when calling create_pointer?
If not you can add the parameter `garbage_collect_data` to the function `create_pointer` as for tensors

syft/serde.py
```diff
@@ -1304,6 +1359,8 @@ def _simplify(obj: object) -> object:
         # for this type. If there is, run return
         # the simplified object
         current_type = type(obj)
+        if isinstance(obj, torch.jit.ScriptModule):
+            current_type = torch.jit.ScriptModule
```
LaRiffle: We should avoid avoid if/else condition in the `_simplify` which is really used a lot. I would suggest keep the type as an entry of the dict: `torch.jit.TopLevelTracedModule` or other: it is not a pb to have several keys with the same value in the simplify dict, but adding a if statement slower all the ser/deser process

syft/federated/plan.py
```diff
@@ -351,11 +352,20 @@ def _get_plan_output(self, result_ids, return_ptr=False):
...
         return responses
 
+    def _keep_only_tensor_args(self, args):
+        """Keeps only Tensors in args. This step is needed for method plans."""
+        tensor_args = []
+        for arg in args:
+            if isinstance(arg, torch.Tensor):
+                tensor_args.append(arg)
+        return tensor_args
+
```
LaRiffle: I'm not sure to understand why you need only to keep the tensor args, would you like to detail a bit more the dosctring?

syft/serde.py
midokura-silvia: That is weird, this code was already removed. I will take it out again.

syft/federated/plan.py
midokura-silvia: This is due to a weird behavior of plans that add self as first argument in the method case. Maybe we should check whether this code (plan.py, line 341)
```
        # Support for method hooked in plans
        if self.self is not None:
            args = [self.self] + list(args)
        return self.execute_plan(args, result_ids)
```
 is really necessary.

syft/serde.py
mari-linhares: I've just pulled the most recent version of this branch and is still here. I'll try to remove it now.

mari-linhares: done

syft/federated/plan.py
mari-linhares: I've removed this function and changed the behaviour to make more clear why do we need it

syft/federated/train_config.py
```diff
@@ -16,49 +16,62 @@ class TrainConfig:
...
         """
+        # syft related attributes
         self.owner = owner if owner else sy.hook.local_worker
         self.id = id if id is not None else sy.ID_PROVIDER.pop()
+        self.location = None
 
-        self.model_id = model_id
-        self.loss_fn_id = loss_fn_id
+        # training related attributes
+        self.model = model
```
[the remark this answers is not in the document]
mari-linhares: We need to store this in order to send it.

We can undo these changes and just pass the model along with the loss for sending train_config the same way it was being done before.

I like this wrapper idea better, makes more clear that train_config depends on the model and loss function and it's not just a wrapper for training hyper-parameters.

syft/serde.py
```diff
@@ -1131,11 +1131,11 @@ def _detail_train_config(worker: AbstractWorker, train_config_tuple: tuple) -> t
     )
```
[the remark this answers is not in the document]
mari-linhares: Yes, nice catch.

test/federated/test_federated_client.py
mari-linhares: when it was an int I got an error.

midokura-silvia: What error do you get? requires_grad=True requires a float tensor, but we don't need the gradient with respect to the target.

syft/federated/plan.py
LaRiffle: Ok thanks I'll look at the changes. That's right we needed to include the .self as we call methods , but maybe there is a work around to make it more transparent - will check your latest changes

syft/federated/federated_client.py
```diff
@@ -56,33 +51,21 @@ def _build_optimizer(self, optimizer_name: str, model, lr: float) -> th.optim.Op
...
         optimizer_name = optimizer_name.lower()
         if optimizer_name == "sgd":
-            logger.debug("sgd parameters = %s, lr = %s", model.parameters(), lr)
             self.optimizer = th.optim.SGD(model.parameters(), lr=lr)
         else:
             raise ValueError("Unknown optimizer: {}".format(optimizer_name))
         return self.optimizer
 
-    def fit(self, **kwargs):
+    def fit(self, dataset_key, **kwargs):
```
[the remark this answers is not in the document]
midokura-silvia: The signature is not the same on both side. Once, dataset_key is allowed to be provided as positional argument, and once it is not. 
Thus code that provides dataset_key as positional argument will work for VirtualWorker but not for WebsocketClientWorker.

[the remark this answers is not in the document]
mari-linhares: I see I'll fix it now. Although the notebook demo seems to be working correctly :thinking: .

iamtrask:
This is a VERY impressive PR! You guys have built so much in so little time! Excellent work!

My only question is about the line alice.fit("xor"). Can you help me understand semantically why I'm calling .fit() on alice instead of on train_config (which I assumed would choose to train on alice?)

midokura-silvia:
> My only question is about the line alice.fit("xor"). Can you help me understand semantically why I'm calling .fit() on alice instead of on train_config (which I assumed would choose to train on alice?)

Because you tell the worker to train, assuming he already obtained the training parameters. I think this just a question of how you interpret TrainConfig. I see it as a parameter set which you send to the worker and then tell the worker to train. 

If you interpreted it as a TrainingManager, then we probably would have implement it to be in charge of talking to the workers.

robert-wagner:
> If you interpreted it as a TrainingManager, then we probably would have implement it to be in charge of talking to the workers.

I am very much of the viewpoint that we should move away from having the workers themselves be a part of the api to call functions (other than send to worker). That being said, I don't think that api choice is a good reason to not merge the functionality. We can update it later

syft/federated/train_config.py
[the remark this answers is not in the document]
midokura-silvia: All right, let's leave it like this for the moment. 
Maybe later on, we should add some logic to clearly separate between TrainConfig with model and loss_fn and TrainConfig with model_id and loss_fn_id (=> this is the case either when model and loss_fn have already been sent, or if it is a TrainConfig instance after serialization).

PR #2239, opened 2019-06-05
Raising error if PointerTensor got called with .item

abmazhr:
An error will be raised if any instance of "PointerTensor" got accessed its method "item" with an error message telling to consider accessing "get" method instead so you can get the item you want safely.


abmazhr:
@LaRiffle sure, added a simple raises assertion test there :smile:

abmazhr:
@iamtrask @robert-wagner kindly this is ready to be reviewed :smile:

PR #2244, opened 2019-06-06
Plans should be built by the blueprint owner

### LINKED ISSUE #2003: Add status for Plan
Add a status attribute which gives a bit of information about the plan task: is the plan built / sent / run / failed / etc.

Inspiration can be found on the celery project syntax -> http://docs.celeryproject.org/en/latest/getting-started/first-steps-with-celery.html

### LINKED ISSUE #2241: Plans should be built before sending them and when possible automatically built

Right now plans are built only when called, the main reason for this is that in order to build a plan we need some data to be given to the blueprint. We would like to build the plan automatically at creation when possible.

* It makes sense conceptually that the worker that owns the plan blueprint should be one to build it;

* Not forcing the plan to be automatically built has some ugly implications like sending a plan being a lazy operation and checks need to be maid to make sure a plan can be sent (non-built plans should not be sent to remote workers since they will not be able to build it).

##### Optionally automatically build it

For @sy.func2plan we can add an argument called "args_shape" that will be a list of tuples describing the shape of the arguments of the plan.

This will be used to create mocked tensors (th.zeros or th.ones, doesn't really matter) that will be used to build the data.

Before:

```python
    @sy.func2plan
    def plan_abs(data):
        return data.abs()

    x = th.tensor([-1, 2, 3])
    # force build
    _ = plan_abs(x)

```

After:

```python
    @sy.func2plan(args_shape=[(1,)])
    def plan_abs(data):
        return data.abs()

    # build will occur automatically
    x = th.tensor([-1, 2, 3])
```

Unfortunately, this can not be done for @sy.method2plan since we need the "self" argument which I think is not possible to get without actually executing the method.

##### Raise an exception when trying to send an unbuilt plan

1. Create a boolean flag: self.built
2. If trying to send a plan that was not built yet raise an exception.

**Related Issues**: #2062

### LINKED ISSUE #2264: Ability to build a model plan (@sy.method2plan) locally and execute it remotely

Currently a @sy.method2plan is only well supported when [built remotely](https://github.com/OpenMined/PySyft/blob/35c01d61bdaa69601d39c99ccfc1072b7c2ad7f1/test/federated/test_plan.py#L101).

If we modify the test to build the method locally we get an error:

```python
def test_plan_built_on_method(hook):
    """
    Test @sy.meth2plan and plan send / get / send
    """
    x11 = th.tensor([-1, 2.0]).tag("input_data")
    x12 = th.tensor([1, -2.0]).tag("input_data2")
    x21 = th.tensor([-1, 2.0]).tag("input_data")
    x22 = th.tensor([1, -2.0]).tag("input_data2")

    device_1 = sy.VirtualWorker(hook, id="device_1", data=(x11, x12))
    device_2 = sy.VirtualWorker(hook, id="device_2", data=(x21, x22))

    class Net(nn.Module):
        def __init__(self):
            super(Net, self).__init__()
            self.fc1 = nn.Linear(2, 3)
            self.fc2 = nn.Linear(3, 2)

        @sy.method2plan
        def forward(self, x):
            x = F.relu(self.fc1(x))
            x = self.fc2(x)
            return F.log_softmax(x, dim=0)

    net = Net()

    # Force build
    net(th.tensor([1., 2., 3.]))

    net.send(device_1)
    pointer_to_data = device_1.search("input_data")[0]
    pointer_to_result = net(pointer_to_data)
    pointer_to_result.get()

    net.get()
    net.send(device_2)

    pointer_to_data = device_2.search("input_data")[0]
    pointer_to_result = net(pointer_to_data)
    pointer_to_result.get()
```

Error:

```
../anaconda3/envs/syft/lib/python3.6/site-packages/torch/nn/modules/module.py:489: in __call__
    result = self.forward(*input, **kwargs)
syft/federated/plan.py:321: in __call__
    return self.execute_plan(args, result_ids)
syft/federated/plan.py:384: in execute_plan
    self.build_plan(args)
syft/federated/plan.py:177: in build_plan
    res_ptr = self.blueprint(*local_args)
test/federated/test_plan.py:121: in forward
    x = F.relu(self.fc1(x))
../anaconda3/envs/syft/lib/python3.6/site-packages/torch/nn/modules/module.py:489: in __call__
    result = self.forward(*input, **kwargs)
../anaconda3/envs/syft/lib/python3.6/site-packages/torch/nn/modules/linear.py:67: in forward
    return F.linear(input, self.weight, self.bias)
syft/frameworks/torch/hook/hook.py:708: in overloaded_func
    response = TorchTensor.handle_func_command(command)
syft/frameworks/torch/tensors/interpreters/native.py:226: in handle_func_command
    response = eval(cmd)(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

input = (Wrapper)>[PointerTensor | me:10664432826 -> 35246282357:20761509375]
weight = Parameter containing:
tensor([[ 0.6533, -0.3028],
        [ 0.1399,  0.3493],
        [ 0.6281, -0.5958]], requires_grad=True)
bias = Parameter containing:
tensor([ 0.1744, -0.2408, -0.5271], requires_grad=True)

    @torch._jit_internal.weak_script
    def linear(input, weight, bias=None):
        # type: (Tensor, Tensor, Optional[Tensor]) -> Tensor
        r"""
        Applies a linear transformation to the incoming data: :math:`y = xA^T + b`.
    
        Shape:
    
            - Input: :math:`(N, *, in\_features)` where `*` means any number of
              additional dimensions
            - Weight: :math:`(out\_features, in\_features)`
            - Bias: :math:`(out\_features)`
            - Output: :math:`(N, *, out\_features)`
        """
>       if input.dim() == 2 and bias is not None:
E       TypeError: __bool__ should return bool, returned Tensor

../anaconda3/envs/syft/lib/python3.6/site-packages/torch/nn/functional.py:1350: TypeError
```

The code snippet above should work as intended (no error).

mari-linhares:
Fixes #2241, Fixes #2264, Fixes #2003


syft/federated/plan.py
```diff
@@ -20,41 +21,53 @@ def make_plan(plan_blueprint):
...
+    This class should be used only as decorator.
     """
-    plan = Plan(owner=sy.local_worker, id=sy.ID_PROVIDER.pop(), name=plan_blueprint.__name__)
-    plan.blueprint = plan_blueprint
-    return plan
+
+    def __init__(self, args_shape=None):
+        self.args_shape = args_shape
+
+    def __call__(self, plan_blueprint):
```
robert-wagner: Rather than building plans at call time, could we build them at send time?

mari-linhares: That's how is currently done, but it will not work for remote workers (aka WebSocket workers)

robert-wagner: What in particular makes them break at send time for websocket workers?

mari-linhares: The main thing is that virtual workers are able to build a plan remotely because they can run the model blueprint (original function / method) this is not true for remote workers.

syft/federated/plan.py
```diff
@@ -314,12 +351,29 @@ def __call__(self, *args, **kwargs):
...
-        assert len(kwargs) == 0, "kwargs not supported for plan"
+        if len(kwargs):
+            raise ValueError("Kwargs are not supported for plan.")
+
+        # TODO: for now only one value is returned from a plan
         result_ids = [sy.ID_PROVIDER.pop()]
+
+        # TODO: try to find a better way to deal with local execution
+        # of methods.
+        if self.is_method and not self.locations and self.owner == sy.hook.local_worker:
```
robert-wagner: What conditions result in the changing of the values of these checks

syft/federated/plan.py
```diff
@@ -437,19 +490,30 @@ def request_execute_plan(
...
 
-    def send(self, *locations):
-        """Mock send function that only specify that the Plan will have to be sent to location.
+    def send(self, *locations, force=False):
+        """Send plan to locations.
 
-        When one calls .send(), this doesn't trigger a call to remote workers, but
-        just stores "a promise" that it will be sent (with _send()) later when the plan in
-        called (and built)
+        If the plan was not built locally it will raise an exception.
```
robert-wagner: What is the api for building a plan? Does anything change about how it is built when it is sent to a location?

One problem I am currently running into is if the parameters for a plan are multipointers it sends the entire multipointer to a worker

syft/frameworks/torch/pointers/pointer_tensor.py
```diff
@@ -165,6 +165,9 @@ def attr(self, attr_name):
         self.__setattr__(attr_name, attr_ptr)
         return attr_ptr
 
+    def dim(self) -> int:
```
robert-wagner: I just had to fix this for multipointer tensor lol

test/federated/test_plan.py
```diff
@@ -15,30 +15,87 @@
...
 
 
-def test_plan_built_locally(hook):
+def test_plan_built_automatically(hook):
     # To run a plan locally the local worker can't be a client worker,
     # since it needs to register objects
     hook.local_worker.is_client_worker = False
 
-    @sy.func2plan
+    @sy.func2plan(args_shape=[(1,)])
```
robert-wagner: Could we add support for variably shaped args? ie a tensor with one set dimension and the other unset (for things like batching)

test/federated/test_plan.py
```diff
@@ -255,7 +297,50 @@ def my_plan(data):
...
+            self.fc2 = nn.Linear(3, 2)
+
+        @sy.method2plan
+        def forward(self, x):
+            x = F.relu(self.fc1(x))
+            x = self.fc2(x)
+            return F.log_softmax(x, dim=0)
+
+    net = Net()
+    # net.send(hook.local_worker)
```
robert-wagner: why are these commented out

test/federated/test_plan.py
```diff
@@ -255,7 +297,50 @@ def my_plan(data):
...
+    x = th.tensor([-1, 2.0])
+
+    local_res = net(x)
+
+    # net.get()
```
robert-wagner: ditto

robert-wagner:
As a heads up, travis is complaining about lines in test not running

syft/federated/plan.py
mari-linhares: > What is the api for building a plan? 

The way I see it is that building should be dealt with internally as a consequence of a call.

This means that the user should not call `plan.build`, but the first time they execute a plan (aka `plan(<some data>`) the plan is built.

> Does anything change about how it is built when it is sent to a location?

The plan should be built locally by the worker that knowns the blueprint (the actual function/method that the plan should execute).

So the build operation shouldn't be related to sending the plan to a location.

> One problem I am currently running into is if the parameters for a plan are multipointers it sends the entire multipointer to a worker

Yes, I think the way is currently implemented that's the expected behaviour, maybe @LaRiffle knows more about it.

syft/frameworks/torch/pointers/pointer_tensor.py
mari-linhares: I saw it and fixed here as well, thanks btw! :1st_place_medal:

test/federated/test_plan.py
mari-linhares: Yes, I've added support for -1, which usually means "unknown" or "any shape" tensors. Since for building the function we need to define the shape of the mock data, I map -1 to 1 in the mock data.

test/federated/test_plan.py
mari-linhares: nice catch, these should actually be removed since they aren't needed anymore. Before these fixes this line was needed. thanks!

syft/federated/plan.py
mari-linhares: This hack is needed to make a method (aka a module) be executed locally in a transparent way. Mainly this will not be true if the plan was sent to some location.

robert-wagner: So for clarification. Self.locations is set when send is called. Self.owner is set when?

syft/federated/plan.py
midokura-silvia: The proxy of the websocket worker is not doing anything at the moment. Just transmitting the message to the remote counterpart. And the remote part does not have access to the blueprint.
So if the proxy got intelligent and built the plan before sending it, it could work.

syft/federated/plan.py
```diff
@@ -117,8 +132,25 @@ def __init__(
         self.blueprint = blueprint
 
         # For methods
-        self.self = None
         self.is_method = is_method
+        self._self = None
+
+    def _auto_build(self, args_shape: List[Tuple[int]] = None):
+        if self.is_method:
+            self.build([self._self] + self._create_placeholders(args_shape))
```
midokura-silvia: I never understood why self is added as first argument. It will not be used as self argument anyhow. When is it really used?

syft/federated/plan.py
LaRiffle: Yes I think that when you call a Plan it only checks if the args provided are tensors are pointers, multipointers are seen as tensors and therefore the args are sent to the remote worker selected. `find_location` in plan.py is partially responsible for this.
However maybe this is slightly beyond the scope of this PR.

syft/federated/plan.py
mari-linhares: @robert-wagner @midokura-silvia Actually, I've just noticed what I've said previously was not true.

> That's how is currently done, but it will not work for remote workers (aka WebSocket workers)

[An unbuilt plan can be sent successfully to a remote worker](https://gist.github.com/mari-linhares/d1feeb45903dc15c76dde1c5874b69c8) (if you add this test to `test_plan.py` in dev branch it works).

[But the current version can't be executed locally successfully](https://gist.github.com/mari-linhares/a30ea03066a9dc5c45c613624595447b). This PR solves this.

mari-linhares: @LaRiffle do you think is good that we force the model to be built before being sent or since this is not particularly a problem we shouldn't?

syft/federated/plan.py
mari-linhares: This is needed for building only, not used for executing the plan. It can be changed here and explicitly added in the building method if you think it would be better (I do!)

syft/federated/plan.py
```diff
@@ -117,8 +132,25 @@ def __init__(
...
+        else:
+            self.build(self._create_placeholders(args_shape))
+
+    def _create_placeholders(self, args_shape):
+        # In order to support -1 value in shape to indicate any dimension
+        # we map -1 to 1 for shape dimensions.
+        mapped_shapes = []
```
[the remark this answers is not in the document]
mari-linhares: Will do

syft/federated/plan.py
LaRiffle: I think so yes, otherwise we will use the similar tricks as before with lazy .send() which don't really send the plan until a call is made. This, which was previously used, is complicated to understand, use and debug. Does it make sense?

syft/federated/plan.py
robert-wagner: Is the self.owner part of this check necessary then?

syft/federated/plan.py
robert-wagner: Sounds good. That was more a question in line with my current work that I wanted to see if this addressed

syft/federated/plan.py
LaRiffle: It’s a useful hack but it only works for virtual workers 
We need a better way to know if the worker is the local worker (aka client)

robert-wagner: Would it work if it was sent? does it keep the same reference to the local worker after it is sent over the wire?

syft/federated/plan.py
mari-linhares: Yes, I agree.

syft/federated/plan.py
mari-linhares: > Would it work if it was sent? does it keep the same reference to the local worker after it is sent over the wire?

Considering that it was already built. Yes, it will work.

No, it doesn't keep the same reference.

But as @LaRiffle said we consider that the blueprint owner is a virtual worker which is usually true since when a model is instantiated it belongs to hook.local_worker which is a virtual worker. So I don't think we need to worry about this condition right now.

PR #2248, opened 2019-06-07
Support torch.sum and torch.mean for AST

Jasopaum:


syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -432,6 +432,33 @@ def matmul(self, other):
...
+            """Overload torch.mean(x)"""
+            # We cannot directly use mean on Long tensors
+            # so we do it by hand with a sum and a division
+            sum = self.sum(*args, **kwargs)
+
+            # We need to know how many input values are used for each
+            # output value to divide
+            dims_to_reduce = args[0] if args else range(self.dim())
+            if isinstance(dims_to_reduce, int):
+                dims_to_reduce = (dims_to_reduce,)
```
LaRiffle: Small thing but I think to can merge these 3 lines in a single:
`dims_to_reduce = args if args else range(self.dim())`
Should be valide given the mean signature: `torch.mean(input, dim, keepdim=False, dtype=None) → Tensor`

Jasopaum: It doesn't work if we pass several dims along which to compute mean.
With your solution and dim=(0,1), dims_to_reduce would be ((0,1),)

LaRiffle: You're right, let's keep it like this then

PR #2249, opened 2019-06-07
More efficient max AST

Jasopaum:


syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -642,7 +642,7 @@ def max(self, dim=None, return_idx=False):
             a = values[i]
             beta = a >= max_value
             max_index = i * beta - max_index * (beta - 1)
-            max_value = a * beta - max_value * (beta - 1)
+            max_value = max_value + beta * (a - max_value)
```
LaRiffle: Can we do the same of the index?

Jasopaum: Oh, yes, of course

PR #2254, opened 2019-06-07
Fix TFEWorker Windows bug

### LINKED ISSUE #2250: Syft Keras bug on Windows
Relevant slack discussion: https://openmined.slack.com/archives/C6DEWA4FR/p1559899875021800

Bug:
image

It looks like the problem here is that the `tfe.config` is being saved in a location that is not a valid filepath in Windows.  As a result, there is likely a file with the name `/tmp/tfe.config` being saved in some folder on the machine, as opposed to a file with the name `tfe.config` being saved in the root subdirectory called `tmp`.

The fix for this should use `os.path` to figure out which filepath the tfe.config should be saved to, and then the logging messages should print the OS-specific CLI command for launching each `TFEWorker` process.

jvmancuso:
Fixes #2250.

### DIFF
--- a/syft/workers/tfe.py
+++ b/syft/workers/tfe.py
@@ -1,12 +1,15 @@
 """To be extended in the near future."""
 from collections import OrderedDict
 import logging
+import os
 import subprocess
+import tempfile
 
 import tf_encrypted as tfe
 
 
 logger = logging.getLogger("tf_encrypted")
+_TMP_DIR = tempfile.gettempdir()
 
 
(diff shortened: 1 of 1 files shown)


jvmancuso:
Would really love someone to test this on a Windows machine since I don't have one available to me!

robert-wagner:
Hey @jvmancuso have you gotten a chance to test this on a windows machine?

jvmancuso:
I haven't, but I also don't have access to one!

PR #2257, opened 2019-06-09
Large Precision Tensor with NumPy operations

mccorby:
This draft (DO NOT MERGE) shows how we could delegate in NumPy to perform the operations.
This would leave the decomposition of the big number solely for "transportation".


mccorby:
Canceling. It helped us to decide if going for the NumPy solution. Thanks, dear PR

PR #2263, opened 2019-06-10
[WIP] Fixed Precision in Zq

### LINKED ISSUE #2261: Additive Sharing and Fixed Precision in Zq
This is part of #2235 

We currently allow additive sharing in the Long field, which in particular allows negative values. However the guarantees of our protocols come from the fact the values are in specific Zq fields (ie [0, 1, ..., q-1]. Therefore we should allow Fixed precision to live in Zq fields.

**Roadmap**
The steps to take to achieve this are:
- [x] add a field argument to Fixed precision
- [x] guarantee that negative values are correctly encoded, ex: -1 --> q-1
- [x] check that no overflow happens with usual operations
- [x] fix the SecureNN protocol to add the transformation of field. See https://eprint.iacr.org/2018/442.pdf Section 4.3 in particular

Jasopaum:
fixed precision tensor only represented by positive int
fixes #2261


syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -62,15 +62,9 @@ def fix_precision(self):
         rational = self.child
 
         upscaled = (rational * self.base ** self.precision_fractional).long()
-        field_element = upscaled % self.field
+        field_element = upscaled % self.field  # Should we raise a warning if upscaled > field?
```
robert-wagner: I think it makes a lot of sense to raise an warning here. I would argue that we should raise an warning if |upscaled| > field/2 since every data point that meets that condition will be embedded incorrectly. In the past we have considered raising an error if base**precision_fractional > field but this is a much more reasonable place to check

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -168,7 +175,10 @@ def matmul(self, *args, **kwargs):
             self.precision_fractional == other.precision_fractional
         ), "In matmul, all args should have the same precision_fractional"
 
-        return response.truncate(other.precision_fractional)
+        response = response.truncate(other.precision_fractional)
+        response %= self.field  # Wrap around the field
```
robert-wagner: Theoretically we want to do this wrapping automatically. I am curious if it makes sense to add this automatically to every method (this is good for now)

robert-wagner:
> Maybe it's a bit early for a review, but don't forget to add extensive tests including negative values in the original float space, overflow cases like c = a + b when a + b > q, etc

I did not mean to approve this (meant to request changes) whoops. Thanks for the catch

syft/frameworks/torch/tensors/interpreters/precision.py
Jasopaum: Do you think I should put this somewhere in the method hooking process?

LaRiffle: I'm not sure it would be easy with the current functionality we have :/

robert-wagner: I will open an issue for this and we can address it later

LaRiffle: Yes looks like a good idea, basically it would be an extension of overload_torch.py I believe to execute a class specific transformation at the `def hook_args(`

syft/frameworks/torch/crypto/securenn.py
```diff
@@ -284,30 +293,24 @@ def share_convert(a_sh):
...
+    x = a_tilde_sh.get() % L
 
     # 5)
     x_bit = decompose(x)
     x_bit_sh = x_bit.share(*workers, field=p, crypto_provider=crypto_provider).child
     delta_sh = delta.share(*workers, field=L - 1, crypto_provider=crypto_provider).child
 
     # 6)
-    eta_p = private_compare(x_bit_sh, r, eta_pp)
+    eta_p = private_compare(x_bit_sh, r - 1, eta_pp)
```
LaRiffle: You should keep `r` (ref: Algo 4 line 6)

Jasopaum: Are you sure? I see `r-1` written...

LaRiffle: In have `r` in https://www.microsoft.com/en-us/research/uploads/prod/2018/09/securenneprint.pdf
But https://eprint.iacr.org/2018/442.pdf says `r-1` haha that's right

So... let's see. I believe you're right `r-1` should be used. The eprint has been updated in 08-Mar-2019 and `r-1` has been put instead of `r`. And it seems to make sense from the math point of view

Jasopaum:
I close this PR as it' now included in #2300 (arithmetic in Zq was needed for SecureNN protocols)

PR #2274, opened 2019-06-12
Fix 2273 - operating Fix Prec and Additive Shared tensors with constants

### LINKED ISSUE #2273: AttributeError: 'int' object has no attribute 'is_wrapper'
**Describe the bug**
Execute Part 11 - Secure Deep Learning Classification.ipynb ecounter "AttributeError: 'int' object has no attribute 'is_wrapper'"

**To Reproduce**
1. Go to 'examples/tutorials/Part 11 - Secure Deep Learning Classification.ipynb'
2. Click on 'run all'
3. Scroll down to 'cell 17 test(args, model, private_test_loader)'

**Expected behavior**
No error here:).
**Screenshots**

**Desktop (please complete the following information):**
 On Google colaboratory

LaRiffle:
Fix #2273

### DIFF
--- a/test/torch/tensors/test_additive_shared.py
+++ b/test/torch/tensors/test_additive_shared.py
@@ -180,22 +180,22 @@ def test_mul(workers):
     assert (z == (t * t)).all()
 
 
-def test_operate_with_constants(workers):
+def test_operate_with_integer_constants(workers):
     bob, alice, james = (workers["bob"], workers["alice"], workers["james"])
-    x = th.tensor([1.0])
+    x = th.tensor([2.0])
     x_sh = x.fix_precision().share(alice, bob, crypto_provider=james)
 
-    r_sh = x_sh + 0
-    assert r_sh.get().float_prec() == x
+    r_sh = x_sh + 10
+    assert r_sh.get().float_prec() == x + 10
 
-    r_sh = x_sh - 0
-    assert r_sh.get().float_prec() == x
+    r_sh = x_sh - 7
+    assert r_sh.get().float_prec() == x - 7
 
-    r_sh = x_sh * 1
-    assert r_sh.get().float_prec() == x
+    r_sh = x_sh * 2
+    assert r_sh.get().float_prec() == x * 2
 
-    r_sh = x_sh / 1
-    assert r_sh.get().float_prec() == x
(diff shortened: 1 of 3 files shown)


Jasopaum:
Good fix, I had just the same issue this morning but didn't have time to fix it :+1:

test/torch/tensors/test_additive_shared.py
```diff
@@ -180,6 +180,24 @@ def test_mul(workers):
     assert (z == (t * t)).all()
 
 
+def test_operate_with_constants(workers):
+    bob, alice, james = (workers["bob"], workers["alice"], workers["james"])
+    x = th.tensor([1.0])
+    x_sh = x.fix_precision().share(alice, bob, crypto_provider=james)
+
+    r_sh = x_sh + 0
+    assert r_sh.get().float_prec() == x
```
midokura-silvia: Could you modify the test so that the result is not equal to x? Otherwise the test won't distinguish between correct summation/multiplication and not doing anything.

mari-linhares: I've tried to, and I'm getting the following assertion error:

```python
def test_operate_with_constants(workers):
    bob, alice, james = (workers["bob"], workers["alice"], workers["james"])
    x = th.tensor([1.0])
    x_sh = x.fix_precision().share(alice, bob, crypto_provider=james)

    r_sh = x_sh + 10
    assert r_sh.get().float_prec() == x + 10

    r_sh = x_sh - 7
    assert r_sh.get().float_prec() == x - 7

    r_sh = x_sh * 2
    assert r_sh.get().float_prec() == x * 2

    r_sh = x_sh / 0.5
    assert r_sh.get().float_prec() == x / 0.5
```

Error:

```
 assert r_sh.get().float_prec() == x + 10
E       assert tensor([1.0100]) == tensor([11.])
```

LaRiffle: Thanks for testing this! I will limit myself to integers because as mentioned in another thread operating fixed_precision with float values should not be possible: float should be converted to fixed_precision explicetly (at least for the moment until we are confident enough with what we're doing)

mari-linhares: Nice, thanks @LaRiffle !

PR #2280, opened 2019-06-12
Lock requirements

### LINKED ISSUE #2171: Lock all requirements
We want to be sure that all our requirements play nicely with each other. As such we want to set minimal versions for which they work (like we have for torch now)

mari-linhares:
Fix #2171 

These versions are based on the current versions I have installed. If you know of some version that could be downgraded let me know.


robert-wagner:
@mari-linhares travis cannot find your version of sklearn

PR #2217, opened 2019-05-31
POC: asynchronous federated training on MNIST using websockets and jit

midokura-silvia:
This pull request contains code to train asynchronously on different workers. 
A concrete example is given with MNIST and the data being separated by digits on three different workers. 
The distribution of the model uses jit.trace() and the communication with the workers is a websocket connection. 
The example is located in the folder: examples/tutorials/advanced/websockets-example-MNIST-parallel

PR #2285, opened 2019-06-14
Create tutorial folder for using PySyft with remote data

racinger:
Following request to add example to access remote data with Pysyft:
https://openmined.slack.com/archives/C6AN4BLS2/p1558260827293100?thread_ts=1557573614.257800&cid=C6AN4BLS2

### DIFF
--- a/examples/tutorials/remotemachine/PredictPytorchBreastCancer.ipynb
+++ b/examples/tutorials/remotemachine/PredictPytorchBreastCancer.ipynb
@@ -1,0 +1,532 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 19,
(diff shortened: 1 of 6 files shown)


robert-wagner:
How does this compare to the tutorial made in #2217 (other than being different data?) @midokura-silvia @racinger

racinger:
> How does this compare to the tutorial made in #2217 (other than being different data?) @midokura-silvia @racinger

@robert-wagner 
I believe that this example show how to handle PySyft when the data is already in another device (worker). Most of PySyft tutorials show how to distribute the data within the client code. But it seems to not be a real use case application.

@mari-linhares thanks for the feedbacks. I ll do as you suggested.

PR #2286, opened 2019-06-14
Add set_num_threads as a function to exclude

niladell:


syft/frameworks/torch/torch_attributes.py
```diff
@@ -96,6 +96,7 @@ def __init__(self, torch: ModuleType, hook: ModuleType) -> None:
             "range",
             "save",
             "set_",
+            "set_num_threads",
```
robert-wagner: Is there an issue attached to this/what breaks?

niladell: There is no particular issue attached to this but using "set_num_threads" gets hooked and breaks whenever called.

PR #2291, opened 2019-06-16
Dinamically build simplifiers and detailers

### LINKED ISSUE #2290: Dinamically generate simplifiers and detailers list
**Describe the bug**

Serde file is not manageable in the current format.

It' very easy to unintentionally breaking `serde.py` by updating simplifiers and detailers list incorrectly. 

A fix was discussed at #2230 which consists of dynamically generating the lists during init.

**Pros**: easier to maintain code.

**Cons**: overhead during initialization.

mari-linhares:
Fixes #2290

### DIFF
--- a/syft/serde/serde.py
+++ b/syft/serde/serde.py
@@ -114,7 +114,7 @@ MAP_TO_SIMPLIFIERS_AND_DETAILERS = OrderedDict(
     }
 )
 
-MAP_TO_FORCE_FULL_SIMPLIFY = set([VirtualWorker])
+MAP_TO_FORCE_FULL_SIMPLIFY = [VirtualWorker]
 
 # If a object implements its own simplifer and detailer function it should be stored in this list.
 OBJ_SIMPLIFIER_AND_DETAILERS = [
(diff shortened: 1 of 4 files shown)


syft/serde/serde.py
```diff
@@ -92,6 +93,82 @@
...
+        torch.jit.ScriptModule: (_simplify_script_module, _detail_script_module),
+        torch.jit.TopLevelTracedModule: (_simplify_script_module, _detail_script_module),
+        torch.nn.Parameter: (_simplify_torch_parameter, _detail_torch_parameter),
+        torch.Tensor: (_simplify_torch_tensor, _detail_torch_tensor),
+        tuple: (_simplify_collection, _detail_collection_tuple),
+        type(Ellipsis): (_simplify_ellipsis, _detail_ellipsis),
+    }
+)
+
+MAP_TO_FORCE_FULL_SIMPLIFY = set([VirtualWorker])
```
LaRiffle: Why do you put this one in a set and not the  OBJ_SIMPLIFIER_AND_DETAILERS and following objects ?

mari-linhares: Oops, my bad. This should be a list as well.

mari-linhares:
Before merging this I'll try to move these structure to their own files. This will make more obvious that these structures need to be changed when these files are modified and cleaner code.

Issue #2283, opened 2019-06-14
AutogradTensor support for MPC


LaRiffle:
We have very interesting blocks with the current AutogradTensor so it's time to build on it to have it working of ML usecases.

**Suggested roadmap:**
- [x] Is .parent needed ?
- [x] Improve the handling of .grad for syft tensors #1895
- [x] Check or fix composition of functions
- [x] Have a working example with fixed precision tensor
- [x] Have a working example with additive shared tensor

LaRiffle:
Solved by #2293

PR #2293, opened 2019-06-17
Encrypted Training demo using AutogradTensor

LaRiffle:
This PR responds to #2283 and aims at providing concrete example of autograd on FixedPrecision and AdditiveShared Tensors.

As is it provides:
- [x] Cleaning of AutogradTensor
- [x] Improvement fo the handling of .data and .grad for syft tensors 
- [x] Support of basic ops (+, *, @) on additive shared tensors and Fixp tensors
- [x] Basic torch ops like .t(), .sum(), .pow()
- [x] More complex operations which need decompositions like addmm, nn.Linear
- [x] Autograd support for basic inplace ops
- [x] Autograd support for inputs of different shapes
- [x] A basic example of encrypted training


syft/frameworks/torch/tensors/interpreters/gradients_core.py
```diff
@@ -60,10 +49,117 @@ def __init__(self, tensor):
...
+    align on the right, unless there is a 1 instead of the other value: shapes (5,4,3,2) and (4,1,2) are
+    compatible, (5,4,3,2) and (5,4,2,2) are not.
+    """
+    # Initialize the dimension if needed
+    if current_dim is None:
+        current_dim = max(len(x_shape), len(y_shape)) - 1
+        # Inverse x and y if needed to satisfy len(x_shape) <= len(y_shape)
+        if len(x_shape) > len(y_shape):
+            return get_mismatch_dims(
+                y_shape, x_shape, x_squash_dims, x_squash_keep_dims, y_squash_keep_dims, current_dim
```
Jasopaum: Doesn't it reduce twice the current_dim to call the function again instead of just swapping x_shape and y_shape?

syft/frameworks/torch/tensors/interpreters/autograd.py
```diff
@@ -45,29 +43,33 @@ def __init__(
...
-
-        if self.parent is not None:
-            # self.parent is a weakref
-            self.parent().grad = self._grad
+        self._grad = value
 
     def attr(self, attr_name):
-        attr_val = self.child.attr(attr_name)
+        if attr_name == "grad":
+            return self.grad
```
Jasopaum: I am not sure I understand everything so I'll just ask this question here:
Why is grad accessible from the other types of tensors (FPT, AST, native, etc.)? Shouldn't having it in the AutogradTensor wrapping everything be sufficient?

syft/frameworks/torch/tensors/interpreters/gradients_core.py
```diff
@@ -60,10 +49,117 @@ def __init__(self, tensor):
...
+            x_squash_keep_dims,
+            y_squash_keep_dims,
+            current_dim - 1,
+        )
+
+
+def apply_dim_transformations(grad_self, grad_other, self_shape, other_shape):
+    """
+    Given computed gradients and initial shapes, reshape the gradients to match the
+    initial shapes by retro-engineering the expansion operations made by PyTorch
```
robert-wagner: the term in english is reverse engineering

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -157,9 +199,6 @@ def _is_parameter(self):
         """
         return isinstance(self, syft.hook.torch.nn.Parameter)
 
-    def copy(self):
```
robert-wagner: will we still have a copy method for native tensor?

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -209,6 +234,18 @@ def mul(self, other):
 
     __mul__ = mul
 
+    def pow(self, power):
+        """
+        Compute integer power of a number by recursion using mul
+        """
+        assert power > 0
```
robert-wagner: we should also assert that power is an int

examples/tutorials/Part 12 - Train an Encrypted Neural Network on Encrypted Data.ipynb
```diff
@@ -89,22 +100,19 @@
...
-    "data_enc = data.fix_precision().share(alice, bob, crypto_provider=james)\n",
-    "target_enc = target.fix_precision().share(alice, bob, crypto_provider=james)\n",
-    "model_enc = model.fix_precision().share(alice, bob, crypto_provider=james)\n",
-    "\n",
-    "# The learning rate is also changed to fixed precision\n",
-    "learning_rate_fp = torch.tensor(learning_rate).fix_precision()"
+    "data = sy.AutogradTensor().on(data)\n",
+    "target = sy.AutogradTensor().on(target)\n",
+    "model.weight = sy.AutogradTensor().on(model.weight)\n",
+    "model.bias = sy.AutogradTensor().on(model.bias)"
```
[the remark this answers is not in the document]
LaRiffle: Yes I will, I'm thinking about the most appropriate syntax, autograd should not come by default I think, or I least should be de-activable

syft/frameworks/torch/tensors/interpreters/gradients.py
```diff
@@ -1,5 +1,6 @@
 # This file is generated from build_gradients.py
```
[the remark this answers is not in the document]
LaRiffle: You're right!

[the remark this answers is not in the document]
LaRiffle: I had forgotten that we were doing this this way, it's a bit of a shame that we can't use it, I'll continue to think if there is a smart way to use the yaml again

syft/frameworks/torch/tensors/interpreters/gradients_core.py
LaRiffle: No I don't think so because I call it with a return

syft/frameworks/torch/tensors/interpreters/autograd.py
LaRiffle: So it depends, for native tensor or any syft tensor which could be above an autograd tensor, we need to keep .grad to propagte the information,
For FPT and AST they can't be above the autograd node in the syft chain, so I specified
```
@property
    def grad(self):
        return None
```
It makes it clear and I think is useful for the serialization. Maybe there is a better way to handle .grad however, I haven't tried to achieve the most compact form

syft/frameworks/torch/tensors/interpreters/gradients_core.py
Jasopaum: Oh yes, my bad, I thought you were doing something like `current_dim -= 1` just above. It's all good!

syft/frameworks/torch/hook/hook_args.py
```diff
@@ -76,7 +76,7 @@
 
 # Methods or functions whose signature changes a lot and that we don't want to "cache", because
 # they have an arbitrary number of tensors in args which can trigger unexpected behaviour
-ambiguous_methods = {"__getitem__", "_getitem_public", "view", "permute"}
+ambiguous_methods = {"__getitem__", "_getitem_public", "view", "permute", "add_"}
```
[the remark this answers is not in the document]
LaRiffle: You're right I should add sub_ (but not mul_) these two methods actually can be used as you expect but also like this:
```
t = torch.ones(1)
t2 = torch.ones(1)
t.sub_(0.1, t2)
>>> 0.900
```
So there are two signatures possible, hence it's ambiguous

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -93,6 +93,10 @@ def get_class_attributes(self):
         """
         return {"crypto_provider": self.crypto_provider, "field": self.field, "n_bits": self.n_bits}
 
+    @property
+    def grad(self):
+        return None
```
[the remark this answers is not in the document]
LaRiffle: Yep I will

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: Yes in abstract.py

examples/tutorials/Part 12 - Train an Encrypted Neural Network on Encrypted Data.ipynb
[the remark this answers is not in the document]
robert-wagner: `with torch.no_grad` ideally

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -393,6 +397,18 @@ def __mul__(self, other, **kwargs):
         """
         return self.mul(other, **kwargs)
 
+    def pow(self, power):
```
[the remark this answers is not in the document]
mari-linhares: FYI: There's a faster algorithm for calculating `pow` (instead of being O(power) is O(log(power)).

https://www.rookieslab.com/posts/fast-power-algorithm-exponentiation-by-squaring-cpp-python-implementation

examples/tutorials/Part 12 - Train an Encrypted Neural Network on Encrypted Data.ipynb
```diff
@@ -52,10 +55,23 @@
...
    "source": [
-    "# Create our dataset\n",
+    "# A Toy Dataset\n",
+    "data = torch.tensor([[0,0],[0,1],[1,0],[1,1.]])\n",
+    "target = torch.tensor([[0],[0],[1],[1.]])\n",
     "\n",
-    "data = torch.Tensor([[0., 0.], [0., 1.], [1., 0.], [1., 1.]])\n",
-    "target = torch.Tensor([[0.], [0.], [1.], [1.]])"
+    "# A Toy Model\n",
+    "model = nn.Linear(2,1)"
```
mari-linhares: Does it work for a model with multiple layers?

syft/frameworks/torch/tensors/interpreters/additive_shared.py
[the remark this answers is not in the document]
LaRiffle: Thanks for the info!

examples/tutorials/Part 12 - Train an Encrypted Neural Network on Encrypted Data.ipynb
LaRiffle: I didn't had the time to try it for now :)

robert-wagner: Would recommend trying with multiple layers (even if they're multiple linear layers) eg something like
```
class Net(nn.Module):

    def __init__(self):
        self.l1 = nn.linear(2,2)
        self.l2 = nn.linear(2,1)
    def forward(self, input):
        hidden = self.l1(input)
        return self.l2(hidden)
```

examples/tutorials/Part 12 - Train an Encrypted Neural Network on Encrypted Data.ipynb
```diff
@@ -116,42 +124,40 @@
...
-    "    # How much is the model wrong\n",
-    "    err = pred - target_enc\n",
-    "    # Compute gradient descent step\n",
-    "    update = data_enc.t().matmul(err)\n",
-    "    # Apply the gradient descent step\n",
-    "    model_enc = model_enc - update * learning_rate_fp\n",
-    "    \n",
-    "    loss = err.get().abs().sum().float_precision()\n",
-    "    print(loss) "
+    "opt = optim.SGD(params=model.parameters(),lr=0.1).fix_precision()\n",
```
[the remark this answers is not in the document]
LaRiffle: Agreed - for now it is best to have all our fancy operations be explicit.
But then in the middle term, I'd like high level instructions where I don't need to specify all transformation manually

examples/tutorials/Part 12 - Train an Encrypted Neural Network on Encrypted Data.ipynb
LaRiffle: Ok I can include this in this PR

examples/tutorials/Part 12 - Train an Encrypted Neural Network on Encrypted Data.ipynb
```diff
@@ -183,7 +198,7 @@
     }
    ],
    "source": [
-    "model_enc.get().float_precision()"
+    "model.weight.child.child.get().float_precision()"
```
robert-wagner: Is there anything we can do to get the user to not call .child?

LaRiffle: Sure!

LaRiffle: Done

examples/tutorials/Part 12 - Train an Encrypted Neural Network on Encrypted Data.ipynb
LaRiffle: I've added an example with this (and some activation), it works but I think there's an issue in the code because the loss doesn't decrease fast enough

syft/frameworks/torch/tensors/interpreters/gradients.py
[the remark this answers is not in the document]
robert-wagner: Bump

syft/frameworks/torch/tensors/interpreters/gradients.py
```diff
@@ -1,6 +1,7 @@
...
 
-from .gradients_core import *
+import torch
+from .gradients_core import GradFunc, apply_dim_transformations
```
robert-wagner: nit, split these into multiple lines

syft/frameworks/torch/tensors/interpreters/gradients.py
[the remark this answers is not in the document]
LaRiffle: I don't a solution for the moment

examples/tutorials/Part 12 - Train an Encrypted Neural Network on Encrypted Data.ipynb
LaRiffle: Actually it's all good

syft/frameworks/torch/tensors/interpreters/gradients.py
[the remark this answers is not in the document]
robert-wagner: I think in this case jason was just saying to delete the comment

PR #2300, opened 2019-06-18
Addition of missing SecureNN protocols

Jasopaum:
Solves #2261


syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -423,11 +444,15 @@ def __itruediv__(self, *args, **kwargs):
...
     def __truediv__(self, shares: dict, divisor):
+        # TODO: truediv in Zq
         assert isinstance(divisor, int)
 
         divided_shares = {}
-        for location, pointer in shares.items():
-            divided_shares[location] = pointer / divisor
+        for i_worker, (location, pointer) in enumerate(shares.items()):
+            if i_worker % 3 == 0:
+                divided_shares[location] = self.field - (self.field - pointer) / divisor
```
LaRiffle: Have checked that type(int / int) == int ?

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -566,6 +591,24 @@ def chunk(tensor_shares, *args, **kwargs):
 
         module.chunk = chunk
 
+        @overloaded.function
+        def roll(tensor_shares, shifts, **kwargs):
```
LaRiffle: Where is this one used?

Jasopaum: It's used in the maxpool_deriv SecureNN protocol

syft/frameworks/torch/tensors/interpreters/additive_shared.py
Jasopaum: It's not the case with native Python types but `/`does not change the type of torch tensors

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -423,11 +443,19 @@ def __itruediv__(self, *args, **kwargs):
...
 
         divided_shares = {}
-        for location, pointer in shares.items():
-            divided_shares[location] = pointer / divisor
+        for i_worker, (location, pointer) in enumerate(shares.items()):
+            # Still no solution to perform a real division on a additive shared tensor
+            # without a heavy crypto protocol.
+            # For now, the solution works in most cases (we compute Q - (Q - pointer) / divisor
+            # for as many worker as the number of times the sum of shares "crosses" Q/2, in average)
+            if i_worker % 3 == 0:
```
LaRiffle: Why 3 here ? The theorem 1 of https://eprint.iacr.org/2017/396.pdf is for 2 parties, have you tried with more?

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -566,6 +594,24 @@ def chunk(tensor_shares, *args, **kwargs):
 
         module.chunk = chunk
 
+        @overloaded.function
+        def roll(tensor_shares, shifts, **kwargs):
```
LaRiffle: Can you add a docstring to explain the behaviour of this function for people who don't know the original torch.roll?

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -566,6 +594,24 @@ def chunk(tensor_shares, *args, **kwargs):
...
+            results = {}
+
+            for worker, share in tensor_shares.items():
+                if isinstance(shifts, dict):
+                    results[worker] = torch.roll(share, shifts[worker].get().item(), **kwargs)
```
LaRiffle: the local client (so P2 in the paper implem) learns here shifts[worker].get().item() which is `i - g mod n` right? Can P2 learn `g` then, and then `r` and then deduce `ind_n`?

syft/frameworks/torch/crypto/securenn.py
```diff
@@ -389,6 +414,153 @@ def relu(a_sh):
...
+        # 4)
+        beta_sh = relu_deriv(w_sh)
+
+        # 5)
+        max_sh = select_share(beta_sh, max_sh, x_sh[i])
+
+        # 6)
+        k = (
+            torch.tensor([i]).share(alice, bob, crypto_provider=crypto_provider).child
+        )  # I did not manage to create an AST with 0 and i as shares
```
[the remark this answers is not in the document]
Jasopaum: Can I leave it as it is?

syft/frameworks/torch/tensors/interpreters/additive_shared.py
Jasopaum: I didn't follow you. `shifts[worker].get().item()` is `-r % n` (see 4) in the max pool_deriv function).
Thinking again about it, I think it is not secure to have P2 learning the shift because it could retrieve the output of the function...

syft/frameworks/torch/tensors/interpreters/additive_shared.py
Jasopaum: It's a bit hard to explain and I'm not 100% sure it's true but I think that for more shares, one third of them should be transformed with `Q - (Q - val) / div`. It's because it's needed as many times as the sum of shares crosses `Q / 2` when wrapped around the field. As values are concentrated around 0 for a big field and shares are generated randomly, I think that the number of times `Q / 2` is crossed is indeed one third of the number of shares.
Tell me if this is not clear, or even not true.

syft/frameworks/torch/tensors/interpreters/additive_shared.py
Jasopaum: Oh, I remember now!
If we consider the max pool_deriv protocol, `share` and `shifts[worker]` are PointerTensors here, so there should be a way not to get them. But as `shifts[worker]` is a tensor and the roll function takes an int here, I need to use the .item() method which raises an error when called instead of get() on a PointerTensor.
It works if I remove this error raising and if I use shifts[worker].item() but as I don't know why the error was there in the first place, I don't want to break anything

syft/frameworks/torch/crypto/securenn.py
```diff
@@ -121,96 +121,92 @@ def select_share(alpha_sh, x_sh, y_sh):
...
 
     # 15)
-    # transform remotely the AdditiveSharingTensor back to a torch tensor
-    d_ptr = remote_mask.remote_get()
-    beta_prime = (d_ptr == 0).sum(1)
-    # Get result back
-    res = beta_prime.get()
-    return res
+    d = permuted_mask.get()
+    beta_prime = ((d == 0).sum(-1) > 0).long()
```
LaRiffle: The sum should be able to exceed 1 right? It should be only 0 or 1

syft/frameworks/torch/crypto/securenn.py
```diff
@@ -121,96 +121,92 @@ def select_share(alpha_sh, x_sh, y_sh):
...
     # 14)
     # Hide c values
-    s = torch.randint(1, p, c.shape).send(alice, bob).child
     mask = s * c
+
     # Permute the mask
-    perm = torch.randperm(c.shape[1]).send(alice, bob).child
-    permuted_mask = mask[:, perm]
-    # Send it to another worker
-    remote_mask = permuted_mask.wrap().send(crypto_provider)
```
LaRiffle: The reason why I had sent the mask to another worker is the following: in our construction, the local worker knows the randomness shared by P0 and P1 (because we didn't managed to hide it from him, but maybe we should). In that case, we can't allow the local worker to get and see permuted_mask because otherwise it can inverse the permutation and remove s to get c. So opening the permuted_mask should be made by a worker which don't have access to the randomness, that's why I chose the crypto_provider, but I agree this is not very elegant.

syft/frameworks/torch/crypto/securenn.py
Jasopaum: I think you're right!

syft/frameworks/torch/crypto/securenn.py
Jasopaum: It's a good reason, I'll change this back

LaRiffle:
The major point of failure is that given any AST x, `x * 0 == 0` fails.

robert-wagner:
Hey @Jasopaum @LaRiffle what is the status of this pr?

Jasopaum:
I think we have what we wanted to have there, maybe Theo can confirm before we merge

LaRiffle:
I'm gonna run a final test tomorrow and then we'll finally merge this PR :)

PR #2305, opened 2019-06-20
Implement simplify and detail for FixedPrecisionTensor

### LINKED ISSUE #2304: Serialize / Deserialize FixedPrecisionTensors

I want to be able to serialize/deserialize FixedPrecisionTensors.

Tests should be implemented.

mari-linhares:
Fixes #2304

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/precision.py
+++ b/syft/frameworks/torch/tensors/interpreters/precision.py
@@ -1,5 +1,7 @@
-import syft
 import torch
+
+import syft
+from syft.workers import AbstractWorker
 from syft.frameworks.torch.tensors.interpreters.abstract import AbstractTensor
 from syft.frameworks.torch.overload_torch import overloaded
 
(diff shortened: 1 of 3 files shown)


syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -515,3 +517,60 @@ def get(self):
...
+
+        Returns:
+            tuple: a tuple holding the unique attributes of the fixed precision tensor.
+        """
+        chain = None
+        if hasattr(tensor, "child"):
+            chain = syft.serde._simplify(tensor.child)
+
+        return (
+            syft.serde._simplify(tensor.id),
```
robert-wagner: do we need to serialize these? they are strings/ints which msgpack handles fine

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -515,3 +517,60 @@ def get(self):
...
+            syft.serde._simplify(tensor.field),
+            syft.serde._simplify(tensor.base),
+            syft.serde._simplify(tensor.precision_fractional),
+            syft.serde._simplify(tensor.kappa),
+            syft.serde._simplify(tensor.tags),
```
robert-wagner: This is the one exception to the above comment

syft/frameworks/torch/tensors/interpreters/precision.py
mari-linhares: Oops, indeed.

For ints I'm sure we don't need to simplify. But for strings we're using encoding something:

```
def _simplify_str(obj: str) -> tuple:
    return (obj.encode("utf-8"),)

```

mari-linhares: Given that said, I'll keep simplify for attributes that can be strings (potentially) if that's okay.

robert-wagner: sounds good to me

PR #2307, opened 2019-06-21
Allow syft to be imported without tf_encrypted being installed.

midokura-silvia:
Many applications of syft don't depend both on pytorch and tensorflow. This code makes it possible to run `import syft` without having tf_encrypted (and thereby tensorflow) installed. 

The goal is to reduce the resources needed for the library, especially for embedded devices.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,6 +1,7 @@
 import pytest
 import torch
 from multiprocessing import Process
+import builtins
 
 import syft
 from syft import TorchHook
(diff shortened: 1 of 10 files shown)


test/conftest.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 import torch
 from multiprocessing import Process
+import builtins
```
robert-wagner: Could you add a skip if to the tfe tests if keras is not available?

midokura-silvia: and what would be the condition of the skipif?

robert-wagner: if not depency_check 🙂

PR #2309, opened 2019-06-25
CRT tensor

Jasopaum:


syft/frameworks/torch/hook/hook_args.py
```diff
@@ -71,6 +74,7 @@
     AutogradTensor: lambda i: AutogradTensor(data=i).on(i, wrap=False),
     AdditiveSharingTensor: lambda i, **kwargs: AdditiveSharingTensor(**kwargs).on(i, wrap=False),
     MultiPointerTensor: lambda i, **kwargs: MultiPointerTensor(**kwargs).on(i, wrap=False),
+    CRTTensor: lambda i, **kwargs: CRTTensor(i, **kwargs),  # I am not sure how the .on() works here
```
Jasopaum: Can someone tell me what I did wrong to be unable to use .on() as with other types?

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -0,0 +1,155 @@
...
+    __mul__ = mul
+
+    def solve_system(self):
+        """ Build the tensor in Zq with q = prod(self.child.keys())
+        satisfying the modular system represented by the tensor.
+        The algorithm consists in:
+        1) Compute N = prod(self.child.keys())
+        2) yi = N / ni where the ni's are modulos in the system
+        3) zi = yi^(-1) mod ni (we know zi exists because all the original modulos are pairwise coprime
+        4) The result is x = sum(ai * yi * zi)
```
robert-wagner: what is ai in this case?

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -0,0 +1,155 @@
...
+            ai = ai.float_prec() % mod
+            yi = N // mod
+            zi = modular_inverse(yi, mod)
+            res += ai * yi * zi
+
+        return FixedPrecisionTensor(field=N, precision_fractional=0).on(res % N)
+
+    @staticmethod
+    @overloaded.module
+    def torch(module):
```
robert-wagner: what does this do?

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
Jasopaum: It's the residue modulo ni, I'll update the doc

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
Jasopaum: I wanted to allow the usage of torch.operation(a, b) with CRT tensors

syft/frameworks/torch/hook/hook_args.py
LaRiffle: .on() allows you to create a empty node before and then to insert it in the syft chain of a tensor, between the head wrapper and the first node. Here you already provide the syft chain as an argument of init so it works in a different way.

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
LaRiffle: This is standard for overloading torch functions for special syft tensors

syft/frameworks/torch/hook/hook_args.py
Jasopaum: But (if that's what you meant) `CRTTensor(**kwargs).on(i, wrap=False)` was not working, that's why I chose this. I'll try to see what's wrong with that

iamtrask:
(I'm very excited about this PR)

test/torch/tensors/test_chinese_remainder.py
```diff
@@ -0,0 +1,97 @@
...
+    res_7c = torch.tensor([4, 4]).fix_prec(field=7, precision_fractional=0)
+    residues_c = {3: res_3c, 7: res_7c}
+    crt_c = syft.CRTTensor(residues_c).wrap()
+
+    assert crt_a == crt_b
+    assert not (crt_a == crt_c)
+
+
+"""
+def test_ne():
```
robert-wagner: Why are these commented out?

Jasopaum: It's commented out because the __ne__ method is not hooked for any other syft tensor, and also because I think some upstream work (in native.py or something like that) needs to be done to enable the use of `!=`.
Maybe hooking `__ne__` for all syft tensors can be another PR?

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -55,26 +55,23 @@ def __repr__(self):
...
-    def eq(self, other):
-        self_ = self.child
-        other_ = other.child
+    @overloaded.method
+    def eq(self, self_, other_):
         assert (
             self_.keys() == other_.keys()
         ), "Cannot compare 2 CRT that don't have the same modulos"
 
+        result = torch.ones(self.shape).fix_prec(precision_fractional=0)
```
robert-wagner: The result should be in the same precision fractional as the original tensor

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -0,0 +1,174 @@
...
+        result = {}
+
+        for mod in self_.keys():
+            result[mod] = self_[mod] * other[mod]
+
+        return result
+
+    __mul__ = mul
+
+    def solve_system(self):
```
mortendahl: not sure where this is being called, so may already be the case, but most of this can be precomputed and reused. reconstruction in that case then boils down to a weighted sum.

Jasopaum: I see what you mean. Should I have the coefficients of this reconstruction as attributes computed when the tensor is instantiated and used when needed?

mortendahl: do you have an example of where/how the CRT tensors are created? the ideal point to compute and cache these is where the moduli are provided (eg along these lines: https://github.com/tf-encrypted/tf-encrypted/blob/master/tf_encrypted/tensor/int100.py#L757-L760)

mortendahl: btw, i've looked quite a lot at CRT tensors and happy to discuss!

Jasopaum: We still don't use CRT tensors so I don't have any real life example to show you... In test_chinese_remainder.py (in this PR), you can see how I create these tensors for the moment (I create the tensors of residues and build a CRTTensor instance with them).
Actually, it's still a bit blurry to me how and when this kind of representation is used, how it is created in the first place, etc. So I might come to ask you some questions if you don't mind :)

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -0,0 +1,312 @@
...
+                assert (
+                    math.gcd(pair[0], pair[1]) == 1
+                ), f"{pair[0]} and {pair[1]} are not coprime, you cannot build a CRTTensor with these as modulos"
+
+        # Check that all the residues have the same precision
+        self.base = base
+        self.precision_fractional = precision_fractional
+        if residues is not None:
+            b = next(iter(residues.values())).child.base
+            prec_frac = next(iter(residues.values())).child.precision_fractional
```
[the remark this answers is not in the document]
Jasopaum: I also think keeping the precision makes sense if the user wants to use float values.
Actually, CRT tensors are just another representation of FixedPrecision tensors.

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -0,0 +1,312 @@
...
+            ai = (ai.float_prec() % mod).long()
+            res += ai * self.solve_system_coeffs[i]
+            N *= mod
+
+        return FixedPrecisionTensor(
+            field=N, base=self.base, precision_fractional=self.precision_fractional
+        ).on(res % N)
+
+    def compute_solve_system_coeffs(self):
+        def modular_inverse(a, b):
```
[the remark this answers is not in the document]
Jasopaum: Good idea, the name `solve_system` was indeed an artefact from my previous implementation

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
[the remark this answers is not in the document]
LaRiffle: Yes, so there si a distinction to make between the precision of the CRTTensor which is good to use float values and the precision of the residues which should always be 0 right?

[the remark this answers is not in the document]
Jasopaum: I'm not sure I see what you mean.
In what I did for the moment, the precision of the CRT is just the same as the precision of all the residues tensors.
I think this precision (which the precision of the residues) can be different from 0. It should be equal to the precision of the FPT we want to represent. I was thinking it was not necessary to use it in any operation but just when we want reconstruct the FPT but I just thought about multiplication where we might want to use it...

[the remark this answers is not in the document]
Jasopaum: Oh no, multiplication has to be taken care of in FPT so it's fine

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -242,17 +252,22 @@ def modular_inverse(a, b):
...
         def mul(self, other):
-            return self.__mul__(other)
+            return self.mul(other)
+
+        module.mul = mul
+
+        def div(self, other):
+            return self.div(other)
 
         module.mul = mul
```
LaRiffle: typo: this should be module.div = div

robert-wagner:
Hey @LaRiffle @Jasopaum what is the current status of this pr?

Jasopaum:
It’s waiting for reviews

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -0,0 +1,394 @@
...
+        owner=None,
+        id=None,
+        tags: set = None,
+        description: str = None,
+    ):
+        super().__init__(owner=owner, id=id, tags=tags, description=description)
+
+        # Check that all the moduli are pairwise coprime
+        if residues is not None:
+            for pair in itertools.combinations(residues.keys(), r=2):
```
robert-wagner: Can we add a flag to skip this test, if the list of keys gets long, this test takes a long time and if we could verify it offline, that would a ton of time (also after the initial spin up) (or if we could cache the results of the function)

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -0,0 +1,394 @@
...
+        self.field = field
+
+        # Check that all the residues have the same precision and
+        # check that all the shapes are the same
+        # while filling tensor
+        self.base = base
+        self.precision_fractional = precision_fractional
+        if residues is not None:
+            r = next(iter(residues.values()))  # Take one arbitrary residue
+            assert isinstance(r.child, FixedPrecisionTensor)
```
robert-wagner: Can you add a human readable message for this assertion so that people know what needs to be passsed

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -0,0 +1,394 @@
...
+            b = r.child.base
+            prec_frac = r.child.precision_fractional
+            res_shape = r.shape
+            self.child = {}
+            for f, r in residues.items():
+                assert isinstance(r.child, FixedPrecisionTensor)
```
robert-wagner: same as above for the assertions

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -0,0 +1,394 @@
...
+                assert r.child.precision_fractional == prec_frac
+                assert f == r.child.field
+                assert (
+                    r.shape == res_shape
+                ), "All residue tensors of CRTTensor must have the same shape"
+
+                self.child[f] = r
+
+            if base is not None:
+                assert b == base
```
robert-wagner: Ditto

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -0,0 +1,394 @@
...
+            if precision_fractional is not None:
+                assert prec_frac == precision_fractional
```
robert-wagner: ditto

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
Jasopaum: This test will occur only when the user gives residues, not when the "usual" way of converting to CRT tensors is used (by calling .fix_prec(storage="crt"). So this should not be that often...
But I can still only warn the user that the check is not done if the list is too long. What would you consider a long list in this case?

Jasopaum:
I added the possibility to choose big fields for CRT tensors but I'm not sure it works yet, I cannot test that before next week

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -0,0 +1,406 @@
...
+        by operations on self because 1) the operations can only be done between tensors with the same moduli
+        and 2) the output tensor will have the same moduli.
+        This is why we store them after having computed them once.
+        """
+        res = 0
+        N = 1
+
+        # We need to sort the dict to be sure that the coeffs are retrieved in the same order they were stored
+        for i, (mod, ai) in enumerate(sorted(self.child.items())):
+            # We need LargePrecisionTensors to reconstruct the original tensor because large values are used in the process
```
mortendahl: fyi there are alternative reconstruction algorithms to avoid this issue (given that the final number is small)

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
```diff
@@ -0,0 +1,406 @@
...
+            "precision_fractional": self.precision_fractional,
+            # Solving coefficients can be reused in results of operations (see docstring of the reconstruct method)
+            "reconstruction_coeffs": self.reconstruction_coeffs,
+        }
+
+
+_moduli_for_fields = {
+    21: [3, 7],  # Still here for some small tests
+    "int64": [257, 263, 269, 271, 277, 281, 283, 293],
+    "int100": [1201, 1433, 1217, 1237, 1321, 1103, 1129, 1367, 1093, 1039],
```
mortendahl: what is the size of the underlying types holding the values? int32? int64?

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
Jasopaum: Could you point me to the method, please?

syft/frameworks/torch/tensors/interpreters/chinese_remainder_theorem.py
Jasopaum: FixedPrecisionTensors are used to hold the residue and they always are int64, I think.
I indeed think that it would make more sense to use int32 tensors behind CRTTensors, which means choosing the type of tensor when creating a FPT

test/torch/tensors/test_chinese_remainder.py
LaRiffle: Definitely

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -223,47 +225,100 @@ def mul(self, other):
...
+
+            # Give back its sign to response
+            pos_res = response * sgn
+            neg_res = (self.field - response) * (1 - sgn)
+            response = neg_res + pos_res
 
         return response
 
     __mul__ = mul
+    __imul__ = mul
```
LaRiffle: inplace operation can't be equivalent to out of place ops, maybe remove this line or make sure you don't create inconsistencies

syft/frameworks/torch/tensors/interpreters/__init__.py
```diff
@@ -3,5 +3,6 @@
 from .precision import FixedPrecisionTensor  # noqa: F401
 from .additive_shared import AdditiveSharingTensor  # noqa: F401
 from .multi_pointer import MultiPointerTensor  # noqa: F401
+from .chinese_remainder_theorem import CRTTensor  # noqa: F401
```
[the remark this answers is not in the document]
Jasopaum: I do agree but I didn't find any good names... Thanks for the name propositions :)

syft/frameworks/torch/tensors/interpreters/precision.py
Jasopaum: Is it more correct to overload __imul__ as I did it now?

PR #2311, opened 2019-06-25
Deleting this tutorial as of now since backpropagation is not working…

amit-rastogi:
… properly

Need to re-work on this tutorial since backpropagation is not working properly.


amit-rastogi:
Yes that sound good. I'll move it to experimental.

PR #2312, opened 2019-06-26
LPT: optimize a bit _split_number and functions using it

Jasopaum:


syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -55,20 +55,10 @@ def __init__(
...
-            result.append(
-                LargePrecisionTensor._split_number(n, self.internal_precision, self.internal_type)
-            )
-        max_length = len(max(result, key=len))
-        new_shape = self.child.shape + (max_length,)
-        # List with different lengths will result in np.array building an array of lists instead of the matrices
-        # itertools.zip_longest fills on the right but we need to fill with zeros on the left
-        result = [self._expand_item(a_number, max_length) for a_number in result]
-        result = np.array(result).reshape(new_shape)
+        self_scaled = self.child.numpy() * self.base ** self.precision_fractional
```
mccorby: A comment:
This is transforming the values into `float` type objects. The tests are passing so it looks fine

robert-wagner: Turning them into floats is potentially bad. Can we avoid that?

Jasopaum: I think multiplying a numpy array of int by an int still gives a numpy array of int at the end. If self_scaled is an array of float, I guess it's because self.child was a tensor of float.
Also, self_scaled is not use as is, it's just an intermediary step that goes to _split_number which returns an array of int.
But I still can cast self_scaled to an array of int easily if needed

Jasopaum: @robert-wagner, should I still change something here?

robert-wagner: I think this is good for now since if it doesn't break, the casting will just slow things down. However could you add a comment about this potential bug in case it does cause a problem in the future

PR #2321, opened 2019-06-29
run static type checker in CI, fix annotations in syft/grid.py

redshiftzero:
Towards #1456

While there are many static type hints in the codebase, it looks like the checker isn't currently ran anywhere, so there are many typing issues on the `dev` branch, which you can see via:

```
$ mypy --python-version 3.6 --ignore-missing-imports syft
syft/grid.py:16: error: Need type annotation for 'tag_counter'
syft/grid.py:22: error: Need type annotation for 'worker_tag_ctr'
syft/grid.py:56: error: Incompatible return value type (got "Dict[Any, Any]", expected "Tuple[Dict[Any, Any], Counter[Any]]")
syft/frameworks/torch/torch_attributes.py:36: error: Module has no attribute "functional"
syft/frameworks/torch/torch_attributes.py:37: error: Module has no attribute "nn"
syft/frameworks/torch/torch_attributes.py:111: error: Module has no attribute "Tensor"
syft/frameworks/torch/torch_attributes.py:123: error: Module has no attribute "FloatTensor"
syft/frameworks/torch/torch_attributes.py:124: error: Module has no attribute "DoubleTensor"
[...snipped due to very long output...]
```

One way to incrementally address this would be to add a mypy call in CI, and gradually fix the typing issues, at which point you can cause type checking errors to fail the Travis build (like is done for other linters). This PR does that and then fixes the first few typing issues seen above in `syft/grid.py`. Let me know if another approach is preferred here.


redshiftzero:
thanks for review @mari-linhares - I added a note to the contributing docs and a ticket (#2323) to track the remaining work. I don't think it would be a ton of work to address the remaining issues, though it's worth noting a few of the issues mypy flags may be legitimate type errors.

syft/grid.py
```diff
@@ -9,17 +12,17 @@ def __init__(self, *workers):
 
     def search(
         self, *query, verbose: bool = True, return_counter: bool = True
-    ) -> Tuple[Dict, Counter]:
+    ) -> Union[Tuple[Dict[Any, Any], CounterType], Dict[Any, Any]]:
         """Searches over a collection of workers, returning pointers to the results
         grouped by worker."""
 
-        tag_counter = Counter()
```
mari-linhares: Oh, I've just noticed you've used typing in the code itself, for pysyft codebase we use typing just for docstrings. Could you remove this before we merge it?

syft/grid.py
```diff
@@ -9,17 +12,17 @@ def __init__(self, *workers):
...
+        tag_counter: CounterType[int] = Counter()
         result_counter = 0
 
         results = {}
         for worker in self.workers:
 
-            worker_tag_ctr = Counter()
+            worker_tag_ctr: CounterType[int] = Counter()
```
mari-linhares: This should also be removed.

syft/grid.py
redshiftzero: So there will be times where a static type checker will need additional information about the types of variables beyond the annotations in function/method argument and return types (e.g. one of the issues in `syft/grid.py` was `"error: Need type annotation for 'tag_counter'"`). If we want to be able to run mypy without errors on the codebase, there are two paths: 

* Add an annotation in variable definitions where needed such that mypy can infer the types of variables it doesn't yet understand (the approach I took in this diff), or

* Add the same type information in a comment e.g. `tag_counter = Counter()  # type: CounterType[int]`

Either is a fine solution, though the latter might be a bit more approachable for a contributor who is familiar with Python but not yet type annotations. I didn't take this path as there is [a linting check](https://github.com/OpenMined/PySyft/blob/dev/.pre-commit-config.yaml#L21) disallowing type information in comments. We _could_ disable that if it's preferable - what do you think?

robert-wagner: Hey @redshiftzero
What changes need to be made on a repository wide scale to get mypy happy with the typing in the code itself (ie do we need to add these comments/annotations all over the place or will it just be isolated.)

I personally am biased towards not having them as comments as I think it looks a bit cleaner and is more in line with how we do it for the method definitions. That being said I would love to hear @iamtrask @LaRiffle @mari-linhares and @midokura-silvia 's opinions on this as well

mari-linhares: I agree with @robert-wagner. 

If these are isolated changes I'm pro supporting mypy by the cost of having some files with comments/static notations.

To be honest I don't have a strong preference for comments/static typing.

Issue #2327, opened 2019-07-02
Fixed precision on LongTensor


LaRiffle:
**Describe the bug**
`torch.tensor([1]).fix_precision()` fails

**To Reproduce**
Run `torch.tensor([1]).fix_precision()`
See:
```python
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-21-837e1ffc8c5a> in <module>
----> 1 torch.tensor([1]).fix_precision()

~/code/PySyft/syft/frameworks/torch/tensors/interpreters/native.py in fix_prec(self, *args, **kwargs)
    643         prec_fractional = kwargs.get("precision_fractional", 3)
    644         max_precision = _get_maximum_precision()
--> 645         if self._requires_large_precision(max_precision, base, prec_fractional):
    646             return (
    647                 syft.LargePrecisionTensor(*args, **kwargs)

~/code/PySyft/syft/frameworks/torch/tensors/interpreters/native.py in _requires_large_precision(self, max_precision, base, precision_fractional)
    666         """
    667         base_fractional = math.log2(base ** precision_fractional)
--> 668         return torch.any((self.abs() + 1).log2() + base_fractional > max_precision)
    669 
    670     def share(

~/code/PySyft/syft/frameworks/torch/hook/hook.py in overloaded_native_method(self, *args, **kwargs)
    661                 except BaseException as e:
    662                     # we can make some errors more descriptive with this method
--> 663                     raise route_method_exception(e, self, args, kwargs)
    664 
    665             else:  # means that there is a wrapper to remove

~/code/PySyft/syft/frameworks/torch/hook/hook.py in overloaded_native_method(self, *args, **kwargs)
    655                 try:
    656                     if isinstance(args, tuple):
--> 657                         response = method(*args, **kwargs)
    658                     else:
    659                         response = method(args, **kwargs)

RuntimeError: log2_vml_cpu not implemented for 'Long'
```

**Expected behavior**
Should convert transparently into fixed precision

mccorby:
I'll pick this one (unless someone wants to do it as a first in the project)

mccorby:
I found out that when restoring a `LongTensor`, `float_precision()` is not returning the original values. 

This test would fail:
```
def test_long():
    x = torch.tensor([1])
    enlarged = x.fix_prec()
    restored = enlarged.float_precision()
    # And now x and restored must be the same
    assert torch.all(torch.eq(x, restored))
```
Are we happy with this?

PR #2329, opened 2019-07-03
Use NumPy to compute log2 as LongTensor does not support it

mccorby:
Fixed issue when calculating whether to use LPT or FPT for LongTensor tensors.

Issue: https://github.com/OpenMined/PySyft/issues/2327

### DIFF
--- a/test/torch/tensors/test_large_precision.py
+++ b/test/torch/tensors/test_large_precision.py
@@ -1,5 +1,6 @@
 import pytest
 import torch
+
 from syft.frameworks.torch.tensors.interpreters import LargePrecisionTensor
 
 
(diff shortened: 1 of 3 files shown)


test/torch/tensors/test_large_precision.py
```diff
@@ -218,3 +218,12 @@ def test_mod():
     lpt2 = x2.fix_prec(internal_type=internal_type, precision_fractional=precision_fractional)
     result = lpt1 % lpt2
     assert torch.all(torch.eq(expected, result.float_precision()))
+
+
+def test_long():
```
mari-linhares: can you parametrize this test and use multiple more input + expected values? Like an array with more than one value, float numbers with more decimal places (example 1.234123), negative numbers...

test/torch/tensors/test_precision.py
```diff
@@ -18,6 +18,15 @@ def test_wrap(workers):
     assert isinstance(x.child.child, torch.Tensor)
 
 
+def test_long():
```
mari-linhares: same

test/torch/tensors/test_large_precision.py
```diff
@@ -218,3 +218,19 @@ def test_mod():
     lpt2 = x2.fix_prec(internal_type=internal_type, precision_fractional=precision_fractional)
     result = lpt1 % lpt2
     assert torch.all(torch.eq(expected, result.float_precision()))
+
+
+test_data = [
```
mari-linhares: I don't think is needed to create a variable just to use it in the decorator. Also, can you add tests with more decimal places (I assume this would work shouldn't it?) and negative numbers?

mccorby: Thanks for the reviews!

"I don't think is needed to create a variable just to use it in the decorator"
What would you suggest instead? Should I loop inside the test itself?

mari-linhares: You can move this array to the decorator itself. You'll have something like:

```python
 @pytest.mark.parametrize("x, expected", [(torch.tensor([1]), torch.tensor([1.0])),
    (torch.tensor([1.0]), torch.tensor([1.0])),
    (torch.tensor([2000.0, 1.0]), torch.tensor([2000.0, 1.0])),
    (torch.tensor([2000.0, 1]), torch.tensor([2000.0, 1.0])),
])
```

mccorby: Done

test/torch/tensors/test_large_precision.py
```diff
@@ -218,3 +219,22 @@ def test_mod():
...
+@pytest.mark.parametrize(
+    "x, expected",
+    [
+        (torch.tensor([1]), torch.tensor([1.0])),
+        (torch.tensor([1.0]), torch.tensor([1.0])),
+        (torch.tensor([2000.0, 1.0]), torch.tensor([2000.0, 1.0])),
+        (torch.tensor([2000.0, 1]), torch.tensor([2000.0, 1.0])),
+        (torch.tensor([-2000.0]), torch.tensor([-2000.0])),
+        (torch.tensor([-2000.0, -50]), torch.tensor([-2000.0, -50.0])),
+        (torch.tensor([-2000.0123458910]), torch.tensor([-2000.0123458910])),
```
robert-wagner: Could you add a test case where there are both positive and negative values in a tensor and one with a positive tensor with values after the decimal place

robert-wagner:
@mccorby please run black on the code

test/torch/tensors/test_large_precision.py
mccorby: Done

Issue #2334, opened 2019-07-04
Add sy.__version__


LaRiffle:
We should add `sy.__version__` attribute for people to easily retrieve which version they are using

amit-rastogi:
picking this up

DanyEle:
Would indeed be very useful!

PR #2335, opened 2019-07-04
Add sy.__version__ #2334

amit-rastogi:
import syft as sy
print(sy.__version__)

Output: 0.1.19a1


syft/version.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.1.19a1"
```
robert-wagner: Could you add this location to setup.cfg as well for bump version support

amit-rastogi: Done.

Issue #2333, opened 2019-07-04
Multiple workers with same id issues


LaRiffle:
**Describe the bug**
When multiple workers with the same id are declared, we observe some weird behaviour.
Could be just raise an Error when this happen? Alternatively, handle this in a consistent way.

**To Reproduce**
```python
import torch
import syft as sy
hook = sy.TorchHook(torch)

worker = sy.VirtualWorker(hook, id="worker")
worker.clear_objects()
ptr = torch.tensor([1,2,3]).send(worker)
print(worker._objects) # >>> Has 1 object!

worker = sy.VirtualWorker(hook, id="worker")
worker.clear_objects()
ptr = torch.tensor([1,2,3]).send(worker)
print(worker._objects) # >>> Empty!
```

amit-rastogi:
@LaRiffle I would like to take this up.

LaRiffle:
Sure!

PR #2339, opened 2019-07-05
Multiple workers with same id issues #2333

amit-rastogi:
The cause of this behavior is due to the way "_objects" dictionary is reset inside clear_objects() method using the statement "self._objects = {}"

When we create a worker with the same id as a pre-existing worker then its "__dict__" object is shared with the pre-existing worker using "self.__dict__.update(known_workers[self.id].__dict__)"

Due to this when we do "self._objects = {}" inside clear_object(), "__objects" now points to a new dictionary object and hence "known_workers" dictionary and "__object" dictionary are not in sync anymore.

To fix this we would need to call the clear() method rather than assigning {} inside clear_objects() method
self._objects.clear()

After the above change the sample code output is shown below-

```
import torch
import syft as sy
hook = sy.TorchHook(torch)

worker = sy.VirtualWorker(hook, id="worker")
worker.clear_objects()
ptr = torch.tensor([1,2,3]).send(worker)
print(worker._objects) # >>> Has 1 object!

worker = sy.VirtualWorker(hook, id="worker")
worker.clear_objects()
ptr = torch.tensor([1,2,3]).send(worker)
print(worker._objects) # >>> Empty!

```
Output-
{48603286308: tensor([1, 2, 3])}
{13171397894: tensor([1, 2, 3])}

### DIFF
--- a/syft/workers/base.py
+++ b/syft/workers/base.py
@@ -678,7 +678,7 @@ class BaseWorker(AbstractWorker, ObjectStorage):
     def clear_objects(self):
         """Removes all objects from the worker."""
 
-        self._objects = {}
+        self._objects.clear()
         return self
 
     @staticmethod
--- a/test/workers/test_base.py
+++ b/test/workers/test_base.py
@@ -33,7 +33,6 @@ def test_clear_object_for_worker_created_with_pre_existing_id(hook):
     worker.clear_objects()
 
     ptr = th.tensor([1, 2, 3]).send(worker)
-    print(worker._objects)
 
     assert len(worker._known_workers[worker.id]._objects) == len(worker._objects)
     assert len(worker._objects) == 1
@@ -46,7 +45,6 @@ def test_clear_object_for_worker_created_with_pre_existing_id(hook):
     assert len(worker._objects) == 0
 
     ptr = th.tensor([1, 2, 3]).send(worker)
-    print(worker._objects)
 
     assert len(worker._known_workers[worker.id]._objects) == len(worker._objects)
     assert len(worker._objects) == 1
--- a/test/workers/test_base.py
+++ b/test/workers/test_base.py
@@ -27,6 +27,31 @@ def test_create_already_existing_worker(hook):
     _ = x + y * z
 
 
+def test_clear_object_for_worker_created_with_pre_existing_id(hook):
+
+    worker = sy.VirtualWorker(hook, id="worker")
+    worker.clear_objects()
+
+    ptr = th.tensor([1, 2, 3]).send(worker)
+    print(worker._objects)
+
+    assert len(worker._known_workers[worker.id]._objects) == len(worker._objects)
+    assert len(worker._objects) == 1
+
+    # create worker with pre-existing id
+    worker = sy.VirtualWorker(hook, id="worker")
+    worker.clear_objects()
+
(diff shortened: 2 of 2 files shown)


syft/workers/base.py
```diff
@@ -678,7 +678,7 @@ def __getitem__(self, idx):
     def clear_objects(self):
         """Removes all objects from the worker."""
 
-        self._objects = {}
+        self._objects.clear()
```
robert-wagner: Could you add a test for this?

amit-rastogi: Done. 
However, Travis fails due to 1 test case failing test/federated/test_federated_client.py::test_fit[gaussian_mixture-1]

the failed assertion is -
assert loss_after < loss_before
assert tensor(0.6931, grad_fn=<NllLossBackward>) < tensor(0.6931, grad_fn=<NllLossBackward>)

The above test case passes on my local development setup though. 
Should we trigger Travis build again? 
Or, I think the assert should be changed to "assert loss_after <= loss_before" we shouldn't assert if the loss remains the same.

robert-wagner: https://gitlab.com/pycqa/flake8-docstrings/merge_requests/15

test/workers/test_base.py
```diff
@@ -27,6 +27,31 @@ def test_create_already_existing_worker(hook):
...
+
+    # create worker with pre-existing id
+    worker = sy.VirtualWorker(hook, id="worker")
+    worker.clear_objects()
+
+    assert len(worker._known_workers[worker.id]._objects) == len(worker._objects)
+    assert len(worker._objects) == 0
+
+    ptr = th.tensor([1, 2, 3]).send(worker)
+    print(worker._objects)
```
mari-linhares: remove print statement

test/workers/test_base.py
```diff
@@ -27,6 +27,31 @@ def test_create_already_existing_worker(hook):
     _ = x + y * z
 
 
+def test_clear_object_for_worker_created_with_pre_existing_id(hook):
+
+    worker = sy.VirtualWorker(hook, id="worker")
+    worker.clear_objects()
+
+    ptr = th.tensor([1, 2, 3]).send(worker)
+    print(worker._objects)
```
mari-linhares: remove print statement

amit-rastogi: Done.

mari-linhares:
Fixes #2333

amin-nejad:
Hi, I'm not super familiar with the library yet but I've noticed that this is still a problem. It seems like the fix was merged into `dev` rather than `master`. Is there a reason for this? Thanks

karlhigley:
Could you create a new issue that contains steps to reproduce and the behavior/errors you're seeing?

PySyft used to use a separate `dev` branch that was merged into `master` periodically, but has since moved to merging directly into `master`. The fix in this PR is in the `master` branch, which you can verify by running:
```
$ git branch --contains 21c976a494a43a2aa614e4a28003663e08f3cb49
``` 
which outputs something like:
```
  [... many branches ...]
  karl/trace-from-placeholders
* master
  nolski/tfjs-translation
  overload
  [... many more branches ...]
```

karlhigley:
^ @amin-nejad

amin-nejad:
Thanks @karlhigley, I've created an issue now: #3455

PR #2340, opened 2019-07-06
Add architecture diagram

mccorby:
A high-level diagram of PySyft architecture


robert-wagner:
hey @mccorby I really like the spirit of this pr but I think there are some things that we could do to make it better.
Firstly, I think we should write out the words for federated learning and differential privacy. Secondly I think we should add edges to the diagram to show how the different pieces interact. eg a worker has tensors which are modified to perform different algorithms

Issue #2341, opened 2019-07-08
SecureNN with n_workers >= 3


LaRiffle:
**Describe the bug**
When used with n>=3 workers with SecureNN

```
wks = [ sy.VirtualWorker(hook, id="w#%d" % i) for i in range(3) ]
crypto_prov = sy.VirtualWorker(hook, id="crypto_prov")
t = torch.zeros(3,3)
t = t.fix_prec().share(*wks, crypto_provider=crypto_prov)
t = t * 2
t = t.get()
t = t.float_prec()
print(t)
```

This doesn't work correctly:
```
tensor([[-1.0000e-03,  0.0000e+00,  1.0000e-03],
        [ 0.0000e+00,  0.0000e+00,  0.0000e+00],
        [-1.0000e-03,  1.0000e-03,  1.8447e+13]])
```

Jasopaum:
In this particular case where we multiply by a scalar, I think we do something useless:
we first multiply the multiplier by `base ** precision_fractional` and then truncate. When I remove this, the results are ok.
But I agree that there are still some problems when multiplying 2 ASTs shared between more than 2 workers

LaRiffle:
Trask: Use a crypto provider and that should be sufficient

vvmnnnkv:
Hmm @LaRiffle, this doesn't seem sufficient to fix the problem of >2 workers:

```
wks = [ sy.VirtualWorker(hook, id="w#%d" % i) for i in range(3) ]
crypto_prov = sy.VirtualWorker(hook, id="crypto_prov")
t = torch.zeros(3,3)
t = t.fix_prec().share(*wks, crypto_provider=crypto_prov)
t = t * 2
t = t.get()
t = t.float_prec()
print(t)

tensor([[-1.8447e+13,  1.8447e+13,  0.0000e+00],
        [ 0.0000e+00,  0.0000e+00,  0.0000e+00],
        [-1.0000e-03, -1.8447e+13,  1.8447e+13]])
```

joseilberto:
@LaRiffle, I recommend as a quick fix using the in-place operation:
```
wks = [ sy.VirtualWorker(hook, id="w#%d" % i) for i in range(3) ]
t = torch.zeros(3,3)
t = t.fix_prec().share(*wks)
t *= 2
t = t.get()
t.float_prec()
```

By the end of today, I will submit a pull request fixing this specific bug.

vvmnnnkv:
[...]

PR #2342, opened 2019-07-09
Fixing __mul__ error for int multiplication issue 2341

joseilberto:
When multiplying by integers, the expansion on other * self.base ** self.precision_fractional is not necessary. The values in self.child are already on fixed precision and the integer is also fixed precision, asl the previous implementation had some serious instabilities with float precision resulting in the errors found here https://github.com/OpenMined/PySyft/issues/2341.


Jasopaum:
Hey, thanks for the PR but we were already working on same function you modified in another PR (#2300) and I already added the changes you've made there...

joseilberto:
@Jasopaum, oh thank you.

Issue #2201, opened 2019-05-31
Add size() method for remote tensors


DanyEle:
The last remaining method that we would need to get the forward pass into remote LSTMs is the implementation of the size() method for remote tensors, as described in Pull Request #2160.

LaRiffle:
This will be complicated as is, because .size() is used in many places to get informations about the current object and not to what it's pointing at. (like torch.save())

Namely we would like to have `t.size() == torch.Size([0])` for `t = torch.tensor([1, 2, 3, 4]).send(bob)` while .shape is used to look at the target object so `t.shape == torch.Size([4])`

LaRiffle:
@robert-wagner I'd love to have your view on this point (namely we have a method which could apply for the pointer object AND the point target)

robert-wagner:
> @robert-wagner I'd love to have your view on this point (namely we have a method which could apply for the pointer object AND the point target)

Sorry for the delay in getting back to you on this. I need to adjust my notification preferences for GitHub (I currently get way to many messages to deal with)

My question is why does the size of the pointer matter (what utility do we get out of it? I can definitely see the case for getting the size of the object that it is pointing to but I am not sure what the point of getting the size of the pointer is (other than knowing it is a pointer but in many cases there are better ways to do that)

LaRiffle:
It is used internally when you're printing Tensors objects for example, so hooking .size() might (and does) break many processes

LaRiffle:
I'm closing this since I don't see obvious way of fixing this.

Issue #2221, opened 2019-06-02
Could not find a version that satisfies the requirement torch>=1.0.1 (from syft) (from versions: 0.1.2, 0.1.2.p


hr21:
I am using a windows machine, with anaconda navigator I installed as per the instructions given in the readme.md 
1. created an enviornment named pysyft and activated it
2. initialized pip install syft
3. The error mentioned above is displayed

I am  Udacity Student for Secure and Private AI

![image](https://user-images.githubusercontent.com/39687652/58757245-8e0d5280-8526-11e9-962f-7d231cec825e.png)

kamathhrishi:
Looks you haven't installed the appropriate version of torch or don't have torch installed in your conda environment. So activate your conda environment and install pytorch. Conda environments are a good way to keep separate dependencies. So if you have had a different version of torch in your system which you usually use. You can use a specific version of torch separate using virtual environments such as conda.

amit-rastogi:
Did you execute the following command in your pysyft environment? If not, then please run this command and check.

conda install pytorch torchvision -c pytorch

iamtrask:
You should update conda and pip. If it doesn't have torch 1.x it's likely very old.

hr21:
I have already tried all the mentioned commands I even tried using it within the enviornment in which I had pytorch Installed and I got back the same error my current conda version is 4.5x and the torch version is 1.x @iamtrask

colin-byrneireland:
I've the exact same issue seems to me that torch 1.2 (latest) isn't compatible with syft ?.
Any suggestions

PR #2343, opened 2019-07-09
Added size() method for remote tensors

DanyEle:
Added a basic version of the size() method for remote tensors.

A lot of tests are still not passing though. Suggestions on what could be fixed so as to make all tests pass. If the size() method is successfully implemented, we would also "unlock" the usage of LSTMs in PySyft.

Implementation for issue #2201


LaRiffle:
Can you add a test to this change to seee what is now possible?

DanyEle:
Now the forward pass into LSTMs is working. I added a couple of tests for the size of remote tensors and the usage of the forward phase with LSTMs. 


```
def test_lstm(workers):
    bob = workers["bob"]
    lstm = nn.LSTM(3, 3)
    inputs = torch.randn(5, 1, 3)
    hidden = (torch.randn(1, 1, 3), torch.randn(1, 1, 3))  # clean out hidden state
    out, hidden = lstm(inputs, hidden)
    assert out.shape == torch.Size([5, 1, 3])
    lstm = nn.LSTM(3, 3)
    lstm.send(bob)
    inputs = torch.randn(5, 1, 3).send(bob)
    hidden = (
        torch.randn(1, 1, 3).send(bob),
        torch.randn(1, 1, 3).send(bob),
    )  # clean out hidden state
    out, hidden = lstm(inputs, hidden)
    #This test will pass once the .size() method is implemented for
    #remote tensors
    assert out.shape == torch.Size([5, 1, 3])
```

LaRiffle:
Good, add also basic .size() tests for all tensors where you added ( in the test_<tensor_type>.py files), or don't forget to push the work if you already added them :)

DanyEle:
> Good, add also basic .size() tests for all tensors where you added ( in the test_<tensor_type>.py files), or don't forget to push the work if you already added them :)

Yep, I re-named a test for the size of local tensors to test_local_size and create a new test case for testing the size() method using remote tensors.

LaRiffle:
The numerous tests failing in this PR confirm the hypothesis made in #2221 
https://github.com/OpenMined/PySyft/issues/2201#issuecomment-497712040
Hooking size() is not possible because it is used internally

DanyEle:
Once #2349 is merged, we may as well close this PR.

DanyEle:
Actually, it looks like https://github.com/OpenMined/PySyft/pull/2349 is not overriding the forward pass of LSTMs / GRUs, so this method would still be needed for them.

PR #2344, opened 2019-07-09
commented out flake8 docstrings from travis

robert-wagner:


robert-wagner:
I'm merging this without a review to make it so that we can merge pull requests until https://gitlab.com/pycqa/flake8-docstrings/merge_requests/15 is merged. This change will be reverted at that point

PR #2345, opened 2019-07-09
Modularize creation of a websocket worker for tests

mari-linhares:
Clean up tests.

Replace `start_proc` to `start_remote_worker` when possible to avoid redundant code.


robert-wagner:
hey @mari-linhares Some tests are breaking due to connection errors. I am not sure why that is

mari-linhares:
Hey @robert-wagner,

I believe it's fixed now, had some minor problems updating the tests but they should pass now.

I've also changed the ports of a few WebSocket tests for precaution (avoiding use the same ports, just in case a connection is not closed).

test/federated/test_train_config.py
```diff
@@ -450,10 +441,10 @@ def forward(self, x):
 
     # Create and send train config
     train_config = sy.TrainConfig(model=model, loss_fn=loss_fn, batch_size=2, epochs=1)
-    train_config.send(local_worker)
+    train_config.send(remote_worker)
```
midokura-silvia: Could we use remote_proxy instead of remote_worker? Because the object previously called local_worker is actually local and not remote. Yet it represents a proxy of the remote worker.

midokura-silvia: Please take a look at the changes in #2386 and #2388. They address another aspect of the websocket connection setup. I wasn't aware (forgot) about your changes to these files.

test/conftest.py
```diff
@@ -1,29 +1,72 @@
...
-        p = Process(target=target)
-        p.start()
-        return p
+@pytest.fixture()
+def start_remote_worker():  # pragma: no cover
+    """Helper function for starting a websocket worker."""
 
-    return _start_proc
+    def _start_remote_worker(
+        id, hook, dataset: str = None, host="localhost", port=8768, max_tries=5, sleep_time=0.01
```
midokura-silvia: Do we want a default value for port? Or should it always be a required argument?

test/conftest.py
```diff
@@ -1,29 +1,72 @@
...
+    ):
+        kwargs = {"id": id, "host": host, "port": port, "hook": hook}
+        server = _start_proc(WebsocketServerWorker, dataset=dataset, **kwargs)
+        remote_worker = instantiate_websocket_client_worker(
```
midokura-silvia: I would prefer another name for remote_worker (as remote_proxy). As this remote_worker instance is the local proxy of the remote websocket server.

test/federated/test_train_config.py
```diff
@@ -299,7 +299,7 @@ def test_send_model_and_loss_fn(workers):
     # TODO check reason for error (RuntimeError: This event loop is already running) when starting websocket server from pytest-asyncio environment
     # dataset = sy.BaseDataset(data, target)
 
-    # process_remote_worker = start_proc(WebsocketServerWorker, dataset=(dataset, dataset_key), **kwargs)
+    # process_remote_worker = start_remote_worker(id="async_fit", hook=hook, dataset=(dataset, dataset_key))
```
midokura-silvia: We can leave the comment, but the syntax is no longer correct. start_remote_worker returns a tuple of values.

test/federated/test_train_config.py
```diff
@@ -435,9 +428,10 @@ def forward(self, x):
...
         print("Loss after training: {}".format(loss_after))
 
-    local_worker.ws.shutdown()
-    del local_worker
+    remote_proxy.close()
+    del remote_proxy
 
-    process_remote_worker.terminate()
+    time.sleep(0.1)
+    server.terminate()
```
midokura-silvia: could you check whether the two lines:
del remote_proxy
and 
time.sleep(0.1)
are necessary?
The main issue with the websocket connections in the tests are connection requests after closing the connections.
These are mostly due to:
 - pointer objects that refer to remote objects (and are in charge of garbage collection) that are deleted at the end of the test

test/message/test_plan.py
```diff
@@ -287,26 +287,23 @@ def my_plan(data):
     x = th.tensor([-1, 2, 3])
     local_res = my_plan(x)
 
-    kwargs = {"id": "test_plan_worker", "host": "localhost", "port": 8799, "hook": hook}
-    server = start_proc(WebsocketServerWorker, **kwargs)
+    server, remote_proxy = start_remote_worker(id="test_plan_worker", hook=hook)
```
midokura-silvia: port not specified. Is this wanted?
Is there a way to make sure that the process is terminated even when the test fails? I don't see an easy way to implement that right now. It would be easier if the pytest fixture returned a server instance and not the function. (Then you could put it in the teardown code after yield).

If I remember correctly this is one of the reasons for having different port numbers for each test.

test/workers/test_websocket_worker.py
```diff
@@ -81,166 +64,142 @@ def create_self_signed_cert(cert_path, key_path):
...
     assert nr_objects == 1
 
-    # get remote object before terminating the websocket connection
-    y.get()
-    local_worker.close()
-    local_worker.remove_worker_from_local_worker_registry()
-    process_remote_worker.terminate()
-
-
-def test_clear_objects_remote(hook, start_proc):
```
midokura-silvia: why did you remove this test?

test/conftest.py
mari-linhares: I think we can keep a default value port.

test/federated/test_train_config.py
mari-linhares: indeed, thank you!

test/federated/test_train_config.py
mari-linhares: will do

test/message/test_plan.py
mari-linhares: > Is there a way to make sure that the process is terminated even when the test fails?
I think this is out of the scope of this change. But I agree this would be great to have.

> If I remember correctly this is one of the reasons for having different port numbers for each test.

Yes, indeed. As I said before I'm okay with keeping a default port for tests for now, but in a different PR we can deal with automatic different port allocation and handling fails. What do you think?

test/workers/test_websocket_worker.py
mari-linhares: good catch! that was me failing to merge the two files, will add it back.

test/federated/test_train_config.py
mari-linhares: I don't think we need them, getting rid of them.

test/message/test_plan.py
midokura-silvia: I agree that it is out of scope to catch port issues. 
However there were no port conflicts possible before this pull request, as we used a different port for each test. So I would propose to stick with separate port numbers.

PR #2349, opened 2019-07-10
Implementation of RNNs in python with linear layers for MPC

andrelmfarias:
Implemented all RNN modules available in PyTorch in Python with linear layers for MPC.

Tests of these modules were fine in pure Pytorch. However, I still get errors when I hook torch and run tests under syft.


DanyEle:
Very interested in this pull request! Are tests passing now? From the Travis CI trace, it looks like the only error reported is the following:

> 1 file would be reformatted, 131 files would be left unchanged.

Running `black syft .` would be enough to fix it.

andrelmfarias:
I still have to implement some tests with RNNs modules to check if they behave as expected (i.e. similar results in comparison to pure PyTorch RNNs)

Also I still need to work on some improvement to the implementations:

https://github.com/OpenMined/PySyft/pull/2349/commits/c0a7921a6a3c0729f7f66ad228aa675152a2d42f

https://github.com/OpenMined/PySyft/pull/2349/commits/e6db59d648570cc02824bc004e4c0af44936e9c5

andrelmfarias:
Last bugs with LSTMs were fixed.

Still need to implement grads for `tanh` and `sigmoid` for MPC training

syft/frameworks/torch/nn/rnn.py
```diff
@@ -0,0 +1,399 @@
...
+    ):
+
+        super(LSTM, self).__init__(
+            input_size, hidden_size, num_layers, bias, batch_first, dropout, bidirectional, LSTMCell
+        )
+
+
+def init_hidden(input, hidden_size, n_layers=None):
+    if n_layers is None:
+        h = torch.zeros(input.shape[0], hidden_size, dtype=input.dtype, device=input.device)
```
mari-linhares: Just checking... shouldn't this be input.shape[1] instead of input.shape[0]?

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -563,6 +563,25 @@ def mod(self, shares: dict, modulus: int):
     def __mod__(self, *args, **kwargs):
         return self.mod(*args, **kwargs)
 
+    @overloaded.method
+    def chunk(self, shares, *args, **kwargs):
```
mari-linhares: Add docstring.

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -431,6 +431,41 @@ def addmm(bias, input_tensor, weight):
...
+            """
+            Overloads torch.sigmoid to be able to use MPC
+            Approximation with polynomial interpolation of degree 10 over [-10,10]
+            Ref: https://mortendahl.github.io/2017/04/17/private-deep-learning-with-mpc/#approximating-sigmoid
+            """
+
+            weights = [0.5, 0.216578258, -0.0083312848, 0.0001876528, -1.9669e-06, 7.5e-09]
+            degrees = [0, 1, 3, 5, 7, 9]
+
+            max_degree = 9
```
mari-linhares: `max_degree = degrees[-1]`

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -431,6 +431,41 @@ def addmm(bias, input_tensor, weight):
...
+            max_idx = degrees.index(max_degree)
+            selected_weights = weights[1:max_idx]
```
mari-linhares: I don't think these two variables (selected_weights, selected_degrees) are needed. You can just use weights[1:max_idx] and degrees[1:max_idx] in the for loop.

test/torch/nn/test_rnn.py
```diff
@@ -0,0 +1,182 @@
...
+    out, h = model(x, h0)
+    assert out.get().float_precision().shape == th.Size(
+        [batch_size, seq_len, directions * hidden_size]
+    )
+    assert h.get().float_precision().shape == th.Size(
+        [batch_size, directions * num_layers, hidden_size]
+    )
+
+
+def test_gru(hook, workers):
```
mari-linhares: Is this test exactly like `test_simple_rnn` but for GRU? If so, it would be better to add the cell type as a parameter at @pytest.mark.parametrize

test/torch/nn/test_rnn.py
```diff
@@ -0,0 +1,182 @@
...
+    out, h = model(x, h0)
+    assert out.get().float_precision().shape == th.Size(
+        [batch_size, seq_len, directions * hidden_size]
+    )
+    assert h.get().float_precision().shape == th.Size(
+        [batch_size, directions * num_layers, hidden_size]
+    )
+
+
+def test_lstm(hook, workers):
```
mari-linhares: same comment.

test/torch/nn/test_rnn.py
```diff
@@ -0,0 +1,182 @@
...
+    out, (h, c) = model(x, (h0, c0))
+    assert out.get().float_precision().shape == th.Size(
+        [batch_size, seq_len, directions * hidden_size]
+    )
+    assert h.get().float_precision().shape == th.Size(
+        [batch_size, directions * num_layers, hidden_size]
+    )
+    assert c.get().float_precision().shape == th.Size(
+        [batch_size, directions * num_layers, hidden_size]
+    )
```
mari-linhares: I think a test showing that a RNN trains properly (aka loss decreasing) is a good idea.

syft/frameworks/torch/nn/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .rnn import RNN
```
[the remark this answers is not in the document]
andrelmfarias: Will do

syft/frameworks/torch/nn/rnn.py
```diff
@@ -0,0 +1,399 @@
...
+        self.input_size = input_size
+        self.hidden_size = hidden_size
+        self.bias = bias
+        self.num_chunks = num_chunks
+        self.nonlinearity = nonlinearity
+        self.fc_xh = nn.Linear(input_size, self.num_chunks * hidden_size, bias=bias)
+        self.fc_hh = nn.Linear(hidden_size, self.num_chunks * hidden_size, bias=bias)
+        self.reset_parameters()
+
+    def reset_parameters(self):
```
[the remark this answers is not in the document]
andrelmfarias: Will do

syft/frameworks/torch/nn/rnn.py
```diff
@@ -0,0 +1,399 @@
...
+
+    def __init__(self, input_size, hidden_size, bias=True, nonlinearity="tanh"):
+        super(RNNCell, self).__init__(input_size, hidden_size, bias, num_chunks=1)
+
+        if nonlinearity == "tanh":
+            self.nonlinearity = torch.tanh
+        elif nonlinearity == "relu":
+            self.nonlinearity = torch.relu
+        else:
+            raise RuntimeError("Unknown nonlinearity: {}".format(nonlinearity))
```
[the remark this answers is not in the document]
andrelmfarias: Will do

syft/frameworks/torch/nn/rnn.py
```diff
@@ -0,0 +1,399 @@
...
+        super(GRUCell, self).__init__(input_size, hidden_size, bias, num_chunks=3)
+
+    def forward(self, x, h=None):
+
+        if h is None:
+            h = init_hidden(x, self.hidden_size)
+
+        gate_x = self.fc_xh(x)
+        gate_h = self.fc_hh(h)
+        x_r, x_z, x_n = gate_x.chunk(3, 1)
```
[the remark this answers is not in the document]
andrelmfarias: Indeed it is more coherent

syft/frameworks/torch/nn/rnn.py
```diff
@@ -0,0 +1,399 @@
...
+        self.bias = bias
+        self.batch_first = batch_first
+        self.dropout = float(dropout)
+        self.bidirectional = bidirectional
+        self.num_directions = 2 if bidirectional else 1
+        self.is_lstm = base_cell is LSTMCell
+        self.nonlinearity = nonlinearity
+
+        # Dropout layers
+        # TO DO: implement a nn.Dropout class for PySyft
```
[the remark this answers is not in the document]
andrelmfarias: Ok

syft/frameworks/torch/nn/rnn.py
```diff
@@ -0,0 +1,399 @@
...
+            if h is not None:
+                if self.is_lstm:
+                    h, c = h
+                    h = torch.transpose(h, 0, 1)
+                    c = torch.transpose(c, 0, 1)
+                    h = (h, c)
+                else:
+                    h = torch.transpose(h, 0, 1)
+
+        batch_size = x.shape[1]
```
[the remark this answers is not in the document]
andrelmfarias: This doesn't need to change, as discussed on Slack

syft/frameworks/torch/nn/rnn.py
```diff
@@ -0,0 +1,399 @@
...
+                c_forward = c
+
+        # Forward direction
+        output_forward = x.new(seq_len, batch_size, self.hidden_size).zero_()
+        for t in range(seq_len):
+            h_next = h_forward.new(h_forward.shape).zero_()
+            if self.is_lstm:
+                c_next = c_forward.new(c_forward.shape).zero_()
+
+            for layer in range(self.num_layers):
```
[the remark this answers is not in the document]
andrelmfarias: I agree. I will do it

syft/frameworks/torch/nn/rnn.py
```diff
@@ -0,0 +1,399 @@
...
+        dropout=0,
+        bidirectional=False,
+    ):
+
+        super(LSTM, self).__init__(
+            input_size, hidden_size, num_layers, bias, batch_first, dropout, bidirectional, LSTMCell
+        )
+
+
+def init_hidden(input, hidden_size, n_layers=None):
```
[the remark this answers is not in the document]
andrelmfarias: Will do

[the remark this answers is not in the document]
andrelmfarias: OK, will do

syft/frameworks/torch/nn/rnn.py
andrelmfarias: here input.shape[0] is the batch_size and input.shape[1] is the embedding size (exactly as in the PyTorch original RNNCell module: https://pytorch.org/docs/stable/nn.html#torch.nn.RNNCell)

We need to initialize h with shape (batch_size, hidden_size)

syft/frameworks/torch/tensors/interpreters/additive_shared.py
andrelmfarias: Will do

syft/frameworks/torch/tensors/interpreters/precision.py
andrelmfarias: Agree

syft/frameworks/torch/tensors/interpreters/precision.py
andrelmfarias: OK

test/torch/nn/test_rnn.py
andrelmfarias: Not exactly equal, as GRU and LSTM don't have the arg `nonlinearity` that RNN has. I prefer to keep them separate or we will have to iterate twice for GRU and LSTM (by using two values for `nonlinearity`) with no need to do so.

test/torch/nn/test_rnn.py
andrelmfarias: I can do only one test for GRU and LSTM and add the pytest decorator, but it might be a little bit messy as I will have some `if`s to handle it, since we have two states (hidden and cell) with LSTM and only one with GRU

test/torch/nn/test_rnn.py
andrelmfarias: Agree

test/torch/nn/test_rnn.py
mari-linhares: hmmm, I think you could just use one argument (something like 'model-type') and you could set it to: gru, lstm, rnn-tanh or rnn-relu. Internally you could deal with the specificity of each model (aka setting the nonlinearity when needed) What do you think?

I think this is better than having 3 different tests that are not actually that different.

andrelmfarias: I see. Ok, I will do like that

test/torch/nn/test_rnn.py
andrelmfarias: I just found out the RNNs are not working with AutogradTensors. It will take a bit of time to debug and implement a test for RNN training.

andrelmfarias: @mari-linhares I implemented a test with federated learning instead. I will work on AutogradTensors in another PR

mari-linhares: great!

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -489,6 +489,39 @@ def addmm(bias, input_tensor, weight):
...
+            """
+            Overloads torch.sigmoid to be able to use MPC
+            Approximation with polynomial interpolation of degree 10 over [-10,10]
+            Ref: https://mortendahl.github.io/2017/04/17/private-deep-learning-with-mpc/#approximating-sigmoid
+            """
+
+            weights = [0.5, 0.216578258, -0.0083312848, 0.0001876528, -1.9669e-06, 7.5e-09]
+            degrees = [0, 1, 3, 5, 7, 9]
+
+            max_degree = degrees[2]
```
mari-linhares: shouldn't it be -1 instead of 2?

andrelmfarias: It's temporally.
After merges from last week, these activation functions got too slow (probably due to several multiplications with degrees higher than 3), so I needed to reduce the max_degree to keep it still doable. The precision we lose by doing this is not that significant and I plan to get back to max_degree=9 after I investigate further the problem with this increase in time execution.

PR #2350, opened 2019-07-11
Plaintext speed regression

akontaxis:
Adds a notebook containing:
- torch implementations of a few linear algebra techniques:
    - forward- and back-solving 
    - LDLt decomposition
    - QR decomposition via Householder reflections

- initial implementations of secure linear regression and Jonathan Bloom's [DASH](https://github.com/jbloom22/DASH/) that leverage PySyft for secure computation.

These implementations of linear regression and DASH are not strictly secure (yet), in that a few final steps are performed on the local worker for now. That's because our implementations of LDLt decomposition, QR decomposition, etc. don't quite work for the PySyft `AdditiveSharingTensor` just yet. They definitely do in principle (because they're compositions of operations the underlying protocol supports), but there are still a few details to hammer out.


iamtrask:
Heads up @andrelmfarias

Issue #2318, opened 2019-06-28
Backward function not called for torch.autograd.function


DanyEle:
**Describe the bug**
The `backward` function of a custom class in PySyft inherited from the `torch.autograd.function` is not invoked when running the `backward()` method on remote tensors. This feature is essential for my current research work based on PySyft. 

**To Reproduce**
Just run this code:

```
import torch
import torch.autograd as autograd
import torch.nn as nn
import syft as sy

class MyFun(torch.autograd.Function):
    def forward(self, inp):
        return inp

    def backward(self, grad_out):
        grad_input = grad_out.clone()
        print('Custom backward called!')
        return grad_input

class MyMod(nn.Module):
    def forward(self, x):
        return MyFun()(x)

#local version - the backward method is indeed invoked. 
mod = MyMod()
y = autograd.Variable(torch.randn(1), requires_grad=True)
z = mod(y)
z.backward()

#remote version - the backward method is not invoked.
hook = sy.TorchHook(torch)
worker = sy.VirtualWorker(hook, id="alice")
y_remote = y.send(worker)
z_remote = z.send(worker)
z_remote.backward()
```

**Expected behavior**
I would expect the backward() method from the custom class to be actually called. 

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04 LTS

**Additional context**
I believe the problem lies in the fact that the `torch.autograd.function` is not implemented in PySyft, and hence the `backward` method of a custom class is not invoked.

LaRiffle:
As discussed with @DanyEle one reason for this can be:
Regarding nn.Modules operated on remote data, the forward pass method is run by the client (ie the local instance) and then all ops inside (like nn.relu or nn.Linear) are hooked and a message is sent for each of them to the workers which holds the data. Therefore, the remote worker doesn't have the big picture of your nn.module model.
However, when you call backward, the cmd `backward()` is sent directly to the remote worker. In this case, as it doesn't have the big picture of your nn.module model, it doesn't see the custom backward of yours and just performs the default backward.

DanyEle:
Thank you for the response @LaRiffle . Any idea on how to possibly fix this issue and introduce support for torch.autograd.function in PySyft over remote tensors?

robert-wagner:
hey @DanyEle On another note, how does this behave if you call y.send() before you call the mod on it?

DanyEle:
> hey @DanyEle On another note, how does this behave if you call y.send() before you call the mod on it?

[...]

PR #2358, opened 2019-07-14
Added serialization and deserialization methods for AutogradTensor

DanyEle:
In response to issues raised in #2318 , adding methods to serialize and deserialize an AutogradTensor.


test/torch/tensors/test_native.py
```diff
@@ -117,15 +117,15 @@ def test_copy():
     assert tensor is not coppied_tensor
 
 
-def test_size():
+def test_local_size():
```
mari-linhares: why were these tests renamed?

syft/frameworks/torch/tensors/interpreters/autograd.py
```diff
@@ -265,3 +265,60 @@ def float_precision(self):
         if isinstance(self.child, torch.Tensor) and not self.child.is_wrapper:
             return self.child
         return self
+
```
mari-linhares: Can you add tests for this?

It would be great to have simple tests for serialize / deserialize and also a test where you:

- train a model (with some toy data)
- run inference in the model (with another batch of toy data)
- serialize and deserialize the model
- check that the evaluation has the same result as the original model
- continue training the model and check that the weights are changing

DanyEle: Thanks for the comments. I tried to perform training of an RNN using Autogradtensors just yesterday, and I bumped into some issues related to backpropagation yielding errors, because tensors were wrapped into AutogradTensors. I fixed this issue by extracting the tensors via an "hotfix" into the Pytorch's library code at `torch/autograd/__init__.py` in the `def backward` method.

At line 87, we just had:

```python grad_tensors = list(grad_tensors)```

I replaced that line  with:

```python
 try:
            grad_tensors = list(grad_tensors)
        except:
            print("Gradients extracted!")
            #extract the tensor from the autogradTensor in every tensor
            grad_tensors = extract_grads_from_autograd_tensors(grad_tensors)

def extract_grads_from_autograd_tensors(auto_grad_tensors):
    grad_tensors = []
    for i in range(len(auto_grad_tensors)):
        #if the child was not yet extracted, then do extract it
        
        tensor_grad = auto_grad_tensors[i].child
        grad_tensors.append(tensor_grad)
        
    return(grad_tensors)
```










Any suggestions on how this hotfix could be introduced into PySyft for autograd_training with Autogradtensors?

test/torch/tensors/test_native.py
DanyEle: This is just some leftover from #2343

mari-linhares: Would you mind undoing the changes in this file if possible? so we can keep the PR as simple and narrow as possible

syft/frameworks/torch/tensors/interpreters/autograd.py
mari-linhares: What's the exception you get when you try to execute `list(grad_tensors)`?

mari-linhares: Probably adding these changes to hook should do the job, and in my opinion, is not that hacky.

test/torch/tensors/test_native.py
DanyEle: Yeah, I'm going to do that tomorrow morning for sure

syft/frameworks/torch/tensors/interpreters/autograd.py
DanyEle: Yeah, hooking these changes should probably be sufficient. I just didn't have much of a clue as to where they should be placed.

DanyEle:
Also, Issue #2318 has finally been resolved and closed, with no need for AutogradTensors. Nevertheless, it may still be useful to keep this PR for serializing and deserializing AutogradTensors.

DanyEle:
I added two test cases for AutogradTensor. 

1) Sending and receiving an AutogradTensor remotely.

2) Training a simple NN model with AutogradTensors. However, there seem to be some issues in the forward propagation phase. Maybe @LaRiffle  may take a look at the test case `test_train_remote_autograd_tensor`

here is the error I get

image

syft/frameworks/torch/tensors/interpreters/autograd.py
DanyEle: The error I was getting is the following one, BTW:

image

DanyEle:
Hi @mari-linhares , I managed to fix the test case not passing. Silly me, I was extracing the model's weights locally and then trying to perform a forward pass then no weights were even present in the model!

The training procedure I used is simply the same one as in the test case `test_encrypted_training_with_linear_model` in the file `test_autograd.py`

.spyproject/codestyle.ini
```diff
@@ -0,0 +1,6 @@
+[codestyle]
```
midokura-silvia: the files in the .spyproject folder should not be part of this pull request.

test/torch/tensors/test_autograd.py
```diff
@@ -511,3 +511,63 @@ def test_get_float_prec_on_autograd_tensor(workers):
...
+    data_remote = data_local.send(worker, local_autograd=True)
+    target_remote = target_local.send(worker, local_autograd=True)
+    model_remote = model_local.send(worker, local_autograd=True)
+
+    loss_remote, model_remote_trained = train(model_remote, data_remote, target_remote)
+
+    # let's check if the local version and the remote version have the same weigth
+    equal_weights = torch.all(
+        torch.eq(
+            model_local_trained.weight.copy().get().data, model_remote.weight.copy().get().data
```
midokura-silvia: why do you need to copy the weights before calling get()?

DanyEle: There is no explicit reason here, but usually calling .copy().get() allows the weights to remain in the remote model for further operations, whereas .get() would just get them locally and remove them from the remote model.

test/torch/tensors/test_autograd.py
```diff
@@ -511,3 +511,63 @@ def test_get_float_prec_on_autograd_tensor(workers):
...
+        )
+    )
+
+    assert equal_weights == 1
+
+    assert loss_remote.copy().get().item() < 500 and loss_local.item() < 500
```
midokura-silvia: Let's see how we can improve this second assert statement.
A loss < 500 seems very lax to me. Does the fitting work on this dataset work with 10 epochs? I guess it does, given that the classifier just needs to look at the first coordinate (linearly separable problem, much easier than xor). 
If the training is stable, we could ask for both remote and local loss to decrease through training. 
I just saw that you send the local model after training to the remote location to train again. That makes it a bit more tricky to check the loss. Any propositions?

test/torch/tensors/test_autograd.py
```diff
@@ -511,3 +511,63 @@ def test_get_float_prec_on_autograd_tensor(workers):
...
+            predictions = model_input(data_input)
+            # 3) calculate how much we missed
+            loss = ((predictions - target_input) ** 2).sum()
+            # 4) Figure out which weights caused us to miss
+            loss.backward()
+            # 5) change those weights
+            opt.step()
+        return (loss, model_input)
+
+    # Training procedure to train an input model, be it remote or local
```
midokura-silvia: The comment ( # Training procedure) should be attached to the nested train() function

test/torch/tensors/test_autograd.py
```diff
@@ -511,3 +511,63 @@ def test_get_float_prec_on_autograd_tensor(workers):
...
+
+    # Some Toy Data
+    data_local = torch.tensor([[0, 0], [0, 1], [1, 0], [1, 1.0]])
+    target_local = torch.tensor([[0], [0], [1], [1.0]])
+
+    # Toy local model
+    model_local = nn.Linear(2, 1)
+
+    # Local training
+    loss_local, model_local_trained = train(model_local, data_local, target_local)
```
midokura-silvia: model_local and model_local_trained point to the same model, right?

test/torch/tensors/test_autograd.py
DanyEle: Yeah, it is indeed a bit more tricky because it looks like the local model is being affected too, following the remote training. I tried to detach all the parameters of the local model, but that didn't make a difference.

As you suggested, we could probably check for a monotonic decrease of the loss during the training procedure both for the local and the remote training procedure.

test/torch/tensors/test_autograd.py
DanyEle: Yeah, I'll fix it.

test/torch/tensors/test_autograd.py
DanyEle: Yep, they actually point to the same model, so we could possibly get rid of the model_local_trained variable.

midokura-silvia: If you calculate the loss before training you could see a decrease. But probably not for the second training round (the remote one), as there are only 4 data points and you already trained the 3 parameters of the model for 10 steps.

DanyEle:
Should be good to go now! I added the checks for loss decrease both for the remote and the local version.

syft/frameworks/torch/tensors/interpreters/multi_pointer.py
```diff
@@ -103,7 +103,6 @@ def shape(self) -> torch.Size:
     def dim(self) -> int:
         """This method fixes the error that the result of dim was a list of ints
         stored inside a multipointer tensor"""
-
```
mari-linhares: Can you undo this change? Just so we don't track in unedeed file.

test/torch/tensors/test_autograd.py
```diff
@@ -511,3 +511,79 @@ def test_get_float_prec_on_autograd_tensor(workers):
...
+
+
+def test_serialize_deserialize_autograd_tensor(workers):
+    # let's try to send an autogradTensor to a remote location and get it back
+    alice = workers["alice"]
+    random_tensor = torch.randn(5, 3)
+    remote_tensor = random_tensor.send(alice, local_autograd=True)
+    local_tensor = remote_tensor.get()
+    # check if the tensor sent is equal to the tensor got from the remote version
+    equal_tensors = torch.all(torch.eq(local_tensor, random_tensor))
```
mari-linhares: You can just have a boolean assertion here: `assert torch.all(...)` instead of asserting  that is equal to 1.

test/torch/tensors/test_autograd.py
```diff
@@ -511,3 +511,79 @@ def test_get_float_prec_on_autograd_tensor(workers):
...
+
+    # Remote training, setting autograd for the data and the targets
+    data_remote = data_local.send(worker, local_autograd=True)
+    target_remote = target_local.send(worker, local_autograd=True)
+    model_remote = model_local.send(worker, local_autograd=True)
+
+    loss_remote, model_remote_trained = train(model_remote, data_remote, target_remote, remote=True)
+
+    # let's check if the local version and the remote version have the same weigth
+    equal_weights = torch.all(
```
mari-linhares: Same comment about boolean assertion

test/torch/tensors/test_autograd.py
mari-linhares: Agree.

syft/frameworks/torch/tensors/interpreters/autograd.py
```diff
@@ -265,3 +265,60 @@ def float_precision(self):
...
+        Args:
+            tensor: an AutogradTensor.
+
+        Returns:
+            tuple: a tuple holding the unique attributes of the AutogradTensor.
+        """
+        chain = None
+        # send the child over
+        if hasattr(tensor, "child"):
+            chain = syft.serde._simplify(tensor.child)
```
midokura-silvia: Could be contracted into a ternary operator:
chain = syft.serde._simplify(tensor.child) if hasattr(tensor, "child") else None

DanyEle: Done!

test/torch/tensors/test_autograd.py
DanyEle: Done!

syft/frameworks/torch/tensors/interpreters/multi_pointer.py
DanyEle: Should be done now

DanyEle:
Should be ready to be merged?

test/torch/tensors/test_autograd.py
```diff
@@ -541,3 +541,73 @@ def test_get_float_prec_on_autograd_tensor(workers):
...
     x2 = x2.share(bob, alice, crypto_provider=james, requires_grad=True)
     assert (x2.get().float_precision() == x).all()
+
+
+def test_serialize_deserialize_autograd_tensor(workers):
+    # let's try to send an autogradTensor to a remote location and get it back
+    alice = workers["alice"]
+    random_tensor = torch.randn(5, 3)
+    remote_tensor = random_tensor.send(alice, local_autograd=True)
+    local_tensor = remote_tensor.get()
```
LaRiffle: This tests doesn't work: there are no autogradtensor created here because:
- you need to set requires_grad to true on tensors
- send(..., local_autograd=True) creates an autogradtensor locally, you would need to send this object to another worker to have the expected behaviour

I would suggest check types of your local and remote objects in the tests to avoid this and to improve readability :)

test/torch/tensors/test_autograd.py
```diff
@@ -541,3 +541,73 @@ def test_get_float_prec_on_autograd_tensor(workers):
...
+            loss.backward()
+            # 5) change those weights
+            opt.step()
+        return (loss, model_input)
+
+    worker = workers["alice"]
+
+    # Some Toy Data
+    data_local = torch.tensor([[0, 0], [0, 1], [1, 0], [1, 1.0]])
+    target_local = torch.tensor([[0], [0], [1], [1.0]])
```
LaRiffle: Same here!

test/torch/tensors/test_autograd.py
```diff
@@ -541,3 +541,73 @@ def test_get_float_prec_on_autograd_tensor(workers):
...
+    # Toy local model
+    model_local = nn.Linear(2, 1)
+
+    # Local training
+    loss_local, model_local_trained = train(model_local, data_local, target_local, remote=False)
+
+    # Remote training, setting autograd for the data and the targets
+    data_remote = data_local.send(worker, local_autograd=True)
+    target_remote = target_local.send(worker, local_autograd=True)
+    model_remote = model_local.send(worker, local_autograd=True)
```
LaRiffle: So there is a bug here: Autograd should be created because model parameters have require_grad to True but it's not currently working :/

test/torch/tensors/test_autograd.py
DanyEle: So, I need to add a requires_grad=True on these tensors?

test/torch/tensors/test_autograd.py
DanyEle: Alright, so would it be fine  to have:

`random_tensor = torch.randn(5, 3)` ?

iamtrask:
Am I correct in reading that when you move a set of AutogradTensors to another machine, that they actually create the graph on that machine?

iamtrask:
Also, it doesn't seem like the gradient gets serialized if it exists. Is this true?

test/torch/tensors/test_autograd.py
LaRiffle: Better to not use randn I think to be more elegant, try `tensor = torch.tensor([[3.0, 2], [-1, 2]], requires_grad=True)` for example

test/torch/tensors/test_autograd.py
LaRiffle: try `data_local = torch.tensor([[0, 0], [0, 1], [1, 0], [1, 1.0]], requires_grad=True) ` instead

DanyEle:
> Also, it doesn't seem like the gradient gets serialized if it exists. Is this true?

I confirm this. This doesn't just happen for Autograd tensors though, but also for remote tensors being transmitted with PySyft.

DanyEle:
I implemented the fixes suggested by @LaRiffle. However, when trying to send a tensor with a `requires_grad=True`, is being produced. I thought this error was due to the fact that I was not serializing/deserializing the requires_grad attribute, so I went ahead and added serialization and de-serialization for that attribute. Nevertheless, the error is still there in the forward propagation phase into the model.

image
image

PR #2360, opened 2019-07-16
Monitor network usage with pyshark

### LINKED ISSUE #2256: Monitoring network usage
This is part of #2235 

**Description**
We need a tool to monitor how much information we share over the network, as SMPC is often said to consume a lot of bandwidth. This is very important to benchmark our implementation compared to similar existing works.

kakirastern:
Fixes #2256. 

Implementing the monitoring of network usage with pyshark, by adding two static methods `get_packets` and `read_packet` to `metrics.py` in `PySyft.syft.generic`. The `get_packets` function takes in the network `interface` as the basic argument and returns the LiveCapture object with the packet info being sniffed on, plus the number of packets captured in the event. On the other hand, the `read_packet` is a convenience function for showing payload details regarding a single packet in a LiveCapture object.


iamtrask:
Wow! Very cool PR! Would you be willing to also include an example (prefarably a notebook or two) showing how to use it? Perhaps in this folder (https://github.com/OpenMined/PySyft/tree/dev/examples/tutorials/advanced)

kakirastern:
I understand many things needed to be changed in this PR for the current function to capture data transmitted through a particular TCP port used by a `worker`. But I would like to get feedback in this PR in order to carry it to fruition.

kakirastern:
> Wow! Very cool PR! Would you be willing to also include an example (prefarably a notebook or two) showing how to use it? Perhaps in this folder (https://github.com/OpenMined/PySyft/tree/dev/examples/tutorials/advanced)

Sure, I will do so once it has been polished further and approved.

kakirastern:
Noted there is some `test_fit[gaussian_mixture-10]` failure which resulted in the following `Assertion Error`: 
```
E           assert tensor(0.6931, grad_fn=<NllLossBackward>) < tensor(0.6931, grad_fn=<NllLossBackward>)
test/federated/test_federated_client.py:164: AssertionError
```
This is unrelated to the code committed herein, but is causing the Travis CI checks of this PR to fail. 

Also noted from the Travis tests output the four warnings below:
```
/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/_pytest/mark/structures.py:324
  /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/_pytest/mark/structures.py:324: PytestUnknownMarkWarning: Unknown pytest.mark.asyncio - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
    PytestUnknownMarkWarning,

test/federated/test_plan.py::test_plan_method_execute_locally
  /home/travis/build/OpenMined/PySyft/syft/frameworks/torch/tensors/interpreters/native.py:308: UserWarning: Implicit dimension choice for log_softmax has been deprecated. Change the call to include dim=X as an argument.
    response = eval(cmd)(*args, **kwargs)

test/torch/differential_privacy/test_pate.py::test_base_dataset_torch

test/torch/differential_privacy/test_pate.py::test_torch_ref_match
  /home/travis/build/OpenMined/PySyft/syft/frameworks/torch/hook/hook.py:764: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
    current_tensor = hook_self.torch.native_tensor(*args, **kwargs)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

kakirastern:
Rebased and dropped a commit accidentally added on during `git poll origin dev` merge that is not coming from me. Now everything is passing. Will stick to using `git rebase -i ...` for this repo from now on to be safe. Seems like that commit was problematic.

kakirastern:
Other arguments I could add (besides `interface`) to the method are as follows: 

```
* bpf_filter: BPF filter to use on packets.
* display_filter: Display (wireshark) filter to use.
* only_summaries: Only produce packet summaries, much faster but includes very little information
* disable_protocol: Disable detection of a protocol (tshark > version 2)
* decryption_key: Key used to encrypt and decrypt captured traffic.
* encryption_type: Standard of encryption used in captured traffic (must be either 'WEP', 'WPA-PWD', or 'WPA-PWK'. Defaults to WPA-PWK).
* tshark_path: Path of the tshark binary
* output_file: Additionally save captured packets to this file.
```

kakirastern:
Original method was too verbose despite being functional, so I have split the `get_packet_info` functions into two, one called `get_packets` to get pyshark capture object and the number (or `length`) of packets transmitted with a timeout set to 50 seconds, and the other called `read_packet` to read one packet from the `capture` output of `get_packets` based on the index argument which currently is set to be between `0` and `length - 1`. 

I think this is a better solution than the previous more naive way suggested by former users in some forums.

kakirastern:
Added the arguments `timeout`, `bpf_filter`, `display_filter`, `tshark_path`, and `output_file` to the `get_packets` method in WebsocketClientWorker of syft.workers.websocket_client to facilitate better control of sniffing and to enable specification of a particular capture filter and/or display filter, say `tcp port 80` for the former and `tcp.port == 80` for the latter. Have tested the code locally. 

So effectively now `read_packet` is simply a convenience function.

kakirastern:
Will add a notebook to https://github.com/OpenMined/PySyft/tree/dev/examples/tutorials/advanced to give examples on how to use the two new static methods next. If something needs to be changed in the PR please advise. Otherwise I will probably leave things as is.

syft/workers/websocket_client.py
```diff
@@ -8,6 +8,7 @@
 import logging
 import ssl
 import time
+import pyshark
```
mccorby: This is a great work. It looks very good and it will be very helpful.
My only concern is whether this utility could be somewhere else and not within this class.

kakirastern: Yeah, I could move them to somewhere more appropriate... Any suggestion(s)?

kakirastern: I could move the `pyshark` utility elsewhere, say further up like at [syft/workers/base.py](https://github.com/OpenMined/PySyft/blob/1d216ae362bc35a2238cbaf12a9c02f047debefd/syft/workers/base.py)? Would that be more appropriate? I checked `torch` is imported there hence my suggestion.

mccorby: I would move it to its own file/module and let the users that want to have those metrics import it in their projects.
What about `metrics.py`?
Second opinions are welcomed!

LaRiffle: It depends a bit of how it is used I would say, if you have an example to share it would be super cool!
I would say, either this is very related to workers is that case we can have it as part of the abstract.py worker file, or if it is more agnostic of the concept of worker, then a `metrics.py` file maybe in syft/generic or syft/federated would be more appropriate!

kakirastern: Thanks for the suggestions! I have moved the utility to a spot called `metrics.py` in the `PySyft/syft/generic` directory. I think this would be a lot more fitting than its former location. Also, I will add examples and explanations for the newly added methods soon, say within the next few days.

kakirastern:
Just added tutorial example to `examples/tutorials/advanced/Monitor_Network_Traffic/metrics.ipynb` which can be accessed [here](https://github.com/OpenMined/PySyft/blob/091de2cba7ffd9cbe22a78e34f5626f6abb428ab/examples/tutorials/advanced/Monitor_Network_Traffic/metrics.ipynb).

kakirastern:
Just added a few details to the example notebook to make things more complete. Think the PR is ready for a review. Any feedback would be much appreciated. Thanks!

examples/tutorials/advanced/Monitor_Network_Traffic/metrics.ipynb
```diff
@@ -0,0 +1,179 @@
...
+   "metadata": {},
+   "source": [
+    "Author: Kris Stern"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**Intro**: The `metrics` utility is located at the `syft/generic` directory. In it is defined the `NetworkMonitor` class, with `get_packets` and `read_packet` as two separate statis methods. The `get_packets` function is the go-to tool for sniffing on the network with a Python wrapper for TShark/Wireshark, itself a free and open-source packet analyzer. More info regarding Wireshark can be found at the official Wiki at https://wiki.wireshark.org/FrontPage. So `get_packets` yields the `capture` and `length` objects, where `capture` is a LiveCapture generated by `pyshark`, and wheere `length` is the number of packets in `capture`. Furthermore, the `read_packet` is a convenience function designed to read a single packet given the LiveCapture output from `get_packets` and an index which is between `0` and `length - 1` (conforming to Python indexing convention which starts from `0`)."
```
robert-wagner: wheere->where

kakirastern: Thanks, corrected along with other typos.

kakirastern:
Ready to be merged...

DanyEle:
Looks good to me too. Eagerly looking forward to trying this one one as soon as it gets merged :)

PR #2364, opened 2019-07-17
Added fix_prec for Linear Object

shashigharti:
I encountered the issue
"AttributeError: 'Linear' object has no attribute 'fix_prec'"

while I replaced fix_precision with fix_prec as shown in this tutorial https://github.com/OpenMined/PySyft/blob/dev/examples/tutorials/Part%2012%20-%20Train%20an%20Encrypted%20Neural%20Network%20on%20Encrypted%20Data.ipynb

### DIFF
--- a/syft/frameworks/torch/hook/hook.py
+++ b/syft/frameworks/torch/hook/hook.py
@@ -1037,6 +1037,7 @@ class TorchHook:
             return nn_self
 
         self.torch.nn.Module.fix_precision = module_fix_precision_
+        self.torch.nn.Module.fix_prec = module_fix_precision_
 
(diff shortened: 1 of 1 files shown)


Ankit-Dhankhar:
@robert-wagner Why are there two variable for same function ? Can we replace all fix_prec with fix_precision or vice-versa if they are redundant?

robert-wagner:
@Ankit-Dhankhar At their core they are the same function (in the source of native tensor for example, one just references the other. The main idea for this is so that people have a shorthand they can write if they want to). You could use whichever one you want and it does the same thing

PR #2372, opened 2019-07-19
Re-factor Pointers, AdditiveSharedTensor, MultiPointer to remove wrappers

LaRiffle:
This PR is a first step toward an encrypted training demo.

I have tried here to remove all un-necessary parts to improve computation speed. The result of this work is not a great increase of the speed (at most 5%) but a small re-factor of the wrappers: now wrappers are increasingly seen as real wrappers:
1. They do less and forward more instructions downward, including (partially) send, share, mid_get, remote_get etc.
2. Their goal is to make sure users play with torch tensor typed object and to provide the structure for parameters (keeping the .data binding).
3. They are removed in the shares of AST and children of MPT as users don't see these objects, they are also remove between FPT and AST.

This helps removing extra layers, all the more as in crypto protocol we were constantly doing `.send(...).child` or `.share(...).child`: we now do `.send(..., no_wrap=True)` and `.share(..., no_wrap=True)`

TODO:
- [x] Confirm moving all non-necessary func out of native.py
- [x] Rm Fix Precision wrapper when using AST  (the ...>FPT>(Wrapper)AST ) if possible


syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -224,11 +232,20 @@ def zero(self):
...
             .long()
-            .share(*self.locations, field=self.field, crypto_provider=self.crypto_provider)
-            .child
+            .share(
+                *self.locations, field=self.field, crypto_provider=self.crypto_provider, **no_wrap
+            )
         )
         return zero
 
+    def refresh(self):
```
robert-wagner: same as above

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -206,7 +206,7 @@ def id(self):
         else:
             try:
                 return self._id
-            except:
+            except AttributeError:
```
robert-wagner: again should this be `if hasattr`?

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -448,89 +457,34 @@ def create_pointer(
     ) -> PointerTensor:
         """Creates a pointer to the "self" torch.Tensor object.
 
-        This method is called on a torch.Tensor object, returning a pointer
```
robert-wagner: should the above definition be here with the docs removed?

syft/serde/serde.py
```diff
@@ -329,10 +334,13 @@ def _compress(decompressed_input_bin: bin) -> bin:
 
     """
     compress_stream, compress_scheme = _apply_compress_scheme(decompressed_input_bin)
-    if len(compress_stream) < len(decompressed_input_bin):
-        return compress_scheme.to_bytes(1, byteorder="big") + compress_stream
-    else:
-        return NO_COMPRESSION.to_bytes(1, byteorder="big") + decompressed_input_bin
+    try:
```
robert-wagner: Same question as above

syft/serde/serde.py
```diff
@@ -329,10 +334,13 @@ def _compress(decompressed_input_bin: bin) -> bin:
...
+        z = scheme_to_bytes[compress_scheme] + compress_stream
+        return z
+    except KeyError:
+        raise CompressionNotFoundException(
+            "compression scheme not found for compression code:" + str(compress_scheme)
```
robert-wagner: capitalize compression

syft/serde/serde.py
```diff
@@ -361,7 +369,7 @@ def _decompress(binary: bin) -> bin:
         return binary
     else:
         raise CompressionNotFoundException(
-            "compression scheme not found for" " compression code:" + str(compress_scheme)
+            "compression scheme not found for compression code:" + str(compress_scheme)
```
robert-wagner: capitalize compression

test/test_serde.py
```diff
@@ -333,21 +333,6 @@ def test_compressed_serde(compress_scheme):
     assert numpy.array_equal(arr, arr_serialized_deserialized)
 
 
-@pytest.mark.parametrize("compress_scheme", [1, 2, 3, 100])
-def test_invalid_decompression_scheme(compress_scheme):
```
robert-wagner: why is this test removed

syft/frameworks/torch/hook/hook.py
```diff
@@ -707,14 +707,29 @@ def get_hooked_func(hook_self, attr):
...
 
+        cmd_name = f"{attr.__module__}.{attr.__name__}"
+
         @wraps(attr)
         def overloaded_func(*args, **kwargs):
             """
             Operate the hooking
             """
-            cmd_name = f"{attr.__module__}.{attr.__name__}"
+            try:
```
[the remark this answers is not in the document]
LaRiffle: They are faster yes because in this case the try succeeds in 99% of the cases, so it would be useless to add a if there for only 1% of the case. 
I believe this part of code is critical (it is run even when people use plain torch operations after they hooked) so speed should prevail on readability, but I must say I haven't benchmarked try / if here. I could if you think it's important

syft/frameworks/torch/hook/hook.py
```diff
@@ -707,14 +707,29 @@ def get_hooked_func(hook_self, attr):
...
+                tensor_type = (
+                    type(args[0]) if not isinstance(args[0], (tuple, list)) else type(args[0][0])
+                )
+            except IndexError:
+                tensor_type = TorchTensor
+
             command = (cmd_name, None, args, kwargs)
-            response = TorchTensor.handle_func_command(command)
+
+            try:
```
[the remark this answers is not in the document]
LaRiffle: Same here, the try should succeed in 95% of the case

syft/frameworks/torch/hook/hook.py
```diff
@@ -960,17 +975,18 @@ def create_grad_objects(model):
                 o.backward()
                 p.grad -= p.grad
 
-        def module_send_(nn_self, dest, force_send=False, **kwargs):
+        def module_send_(nn_self, *dest, force_send=False, **kwargs):
             """Overloads torch.nn instances so that they could be sent to other workers"""
+            print(nn_self, dest, force_send, **kwargs)
```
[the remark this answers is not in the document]
LaRiffle: no indeed!

syft/frameworks/torch/pointers/pointer_tensor.py
```diff
@@ -144,6 +145,115 @@ def data(self, new_data):
     def is_none(self):
         return self.owner.request_is_remote_tensor_none(self)
 
+    @staticmethod
+    def create_pointer(
+        tensor,
+        location: BaseWorker = None,
```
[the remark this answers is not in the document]
LaRiffle: Indeed, but I'm not sure what AbstractWorker is there for, it is an empty class

syft/frameworks/torch/pointers/pointer_tensor.py
```diff
@@ -239,6 +349,9 @@ def item(self) -> None:
             "so you can be safely getting the item you need."
         )
 
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
LaRiffle: I don't know exactly, but I think comparison operator are a bit tricky to handle and moving from __ eq__ to .eq really helped
That's what we do in native.py and because we now operate on pointers (when wrappers are removed), we need them to have the same trick

syft/frameworks/torch/tensors/interpreters/abstract.py
```diff
@@ -245,6 +245,16 @@ def wrap(self) -> torch.Tensor:
     def copy(self):
         return self + 0
 
+    def refresh(self):
```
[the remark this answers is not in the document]
LaRiffle: It's in abstract ;)

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -122,6 +124,14 @@ def get(self):
 
         return result
 
+    def mid_get(self):
```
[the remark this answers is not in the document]
LaRiffle: Not exactly sure, because it require .get() to be implemented in abstract.py which is not the case.
But maybe .get() should also have a default implementation in abstract.py I agree.
I'll look at it and try to do this!

syft/frameworks/torch/tensors/interpreters/additive_shared.py
LaRiffle: I believe it is answered: it overwrites the default behaviour of refresh() for AST

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: Same thing, I believe we get a good speedup with this but I might be wrong

syft/serde/serde.py
LaRiffle: Same answer ;) This should work in 99.99% of the casee

test/test_serde.py
LaRiffle: The test was not valid actually: you can't overwrite `serde._apply_compress_scheme = some_other_compression_scheme` like it is done here, it doesn't work

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: What do you mean exactly?

syft/serde/serde.py
LaRiffle: 👍

syft/serde/serde.py
LaRiffle: 👍

syft/frameworks/torch/hook/hook.py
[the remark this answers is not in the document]
robert-wagner: sounds good!

syft/frameworks/torch/tensors/interpreters/native.py
robert-wagner: The change here deletes much of the documentation but leaves the method

syft/frameworks/torch/tensors/interpreters/abstract.py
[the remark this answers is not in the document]
robert-wagner: That was the point. It seems most other methods are in native, why the difference here?

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: Yes, because basically this method doesn't do anything anymore, so I didn't wanted to duplicate the docstring, as we now how hard it is to already keep one version up to date ;)

syft/frameworks/torch/tensors/interpreters/abstract.py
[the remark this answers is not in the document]
LaRiffle: Methods in abtract.py are available for all syft tensors, while those in native.py are only added to torch Tensor
Here everyone should knnow about refresh in order to forward the message to AST

PR #2381, opened 2019-07-22
Numpy transforms for dataset

Ankit-Dhankhar:
I have a doubt about https://github.com/OpenMined/PySyft/blob/9965dad87b0a38d221fe3923d55f2284bf00381f/syft/frameworks/torch/federated/dataset.py#L53 
and 
https://github.com/OpenMined/PySyft/blob/9965dad87b0a38d221fe3923d55f2284bf00381f/syft/frameworks/torch/federated/dataset.py#L49
why there are two ways of applying transform while second one prevalent?


Ankit-Dhankhar:
Test cases are passing on local machine but failing at travis. What can be the problem and how should I resolve this?

syft/frameworks/torch/federated/dataset.py
```diff
@@ -60,14 +61,9 @@ def transform(self, transform):
         """
 
         # Transforms cannot be applied to Pointer  , Fixed Precision or Float Precision tensors.
-        if type(self.data) == torch.Tensor:
```
kamathhrishi: I would prefer to keep it this way with an additional elif statement for checking numpy type and applying the transform. Since using datatypes we haven't explicitly tested could lead to untraceble errors. 

So
`if type(self.data) == torch.Tensor:
        self.data = transform(self.data)
  else type(self.data)==<the numpy format>:
        self.data=transform(self.data)
   else:
         raise TypeError("Transforms can be applied only on torch tensors")
 `

syft/frameworks/torch/federated/dataset.py
```diff
@@ -60,14 +61,9 @@ def transform(self, transform):
...
-
+        if transform is not None:
             self.data = transform(self.data)
 
-        else:
```
kamathhrishi: In accordance to above comment keep this.

kamathhrishi:
Also write a test case to ensure Numpy transform works well.

midokura-silvia:
Concerning the failing unit test: The one failing test contains some randomness that can make it fail from time to time. If it does not fail repeatedly, I wouldn't worry.

kamathhrishi:
@midokura-silvia I think it's a good idea to make the tests deterministic as much as possible ? Travis needs to fail only when the intended changes made in a pr have a effect on the codebase. Randomized test make it harder to debug.

robert-wagner:
> Concerning the failing unit test: The one failing test contains some randomness that can make it fail from time to time. If it does not fail repeatedly, I wouldn't worry.

With that in mind we should make the test deterministic

midokura-silvia:
Feel free to fix the seeds for the random number generation in pytorch. But this would be a separate pull request. Another option would be to make sure that the dataset is so easy and the hyperparameters are well-chosen such that the training can't fail/get stuck.

syft/frameworks/torch/federated/dataset.py
Ankit-Dhankhar: Then how would be differentiate between transforms, one way is to apply transforms independent of datatype if is either torch.tensor or numpy array. e.g
```
if type(self.data) == torch.Tensor or type(self.data) == numpy.ndarray:
    self.data = transform(self.data)
else: raise TypeError("Transforms can be applied only on torch tensors or numpy array")
```

Ankit-Dhankhar:
I was implementing test cases, but I am facing implementation conflict as self data is required to be tensor (as per the requirement of fix_prec function). I don't know how to proceed further, please guide.

jvmancuso:
@robert-wagner @midokura-silvia, has anyone addressed the issue holding up @Ankit-Dhankhar here?

syft/frameworks/torch/federated/dataset.py
```diff
@@ -60,13 +61,13 @@ def transform(self, transform):
...
-
+        if transform is not None:
+            if type(self.data) == torch.Tensor or type(self.data) == numpy.ndarry:
+                self.data = transform(self.data)
+            else:
+                raise TypeError("Transforms can be applied only on torch tensors or numpy array")
         else:
-
-            raise TypeError("Transforms can be applied only on torch tensors")
+            raise TypeError("Transform cann't be None")
```
murarugeorgec: TYPO: ```cann't``` --> ```can't```

iamtrask:
> I was implementing test cases, but I am facing implementation conflict as self data is required to be tensor (as per the requirement of fix_prec function). I don't know how to proceed further, please guide.

Looks like this PR is blocked until we hook numpy and potentially even integrate it with FixedPrecisionTensor?

iamtrask:
Closing due to inactivity.

PR #2382, opened 2019-07-23
Update README.md

OliverMaerz:
Add instructions on how to run docker image on a Mac


OliverMaerz:
There is no error. It starts ok. But the IP of the container is not accessible from macOS and opening the url with the container IP times out (see https://docs.docker.com/docker-for-mac/networking/#per-container-ip-addressing-is-not-possible)

PR #2386, opened 2019-07-23
Helper function to try websocket connection

Jasopaum:
Sometimes, a ConnectionRefusedError was raised when trying to instantiate a WebsocketClientWorker in the tests for websocket workers.
The introduced helper function waits if the websocket client could not be instantiated before trying once more (5 tries max)

### DIFF
--- a/test/workers/test_websocket_worker.py
+++ b/test/workers/test_websocket_worker.py
@@ -10,6 +10,26 @@ from syft.workers import WebsocketClientWorker
 from syft.workers import WebsocketServerWorker
 
 
+def instantiate_websocket_client_worker(**kwargs):  # pragma: no cover
+    """ Helper function to instantiate the websocket client.
+    If connection is refused, we wait a bit and try again.
+    After 5 failed tries, a ConnectionRefusedError is raised.
+    """
+    retry_counter = 0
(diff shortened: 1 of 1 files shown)


mccorby:
Hi Jason,
I think we keep falling into the same issue with regard to the sockets or any other network connection. This is a problem we have not yet solved with in our tests.
I think the solution should be based on mocking the `websocket` instead of trying to use an actual one.

By mocking the responses of the methods of the socket, we can stop relying on it being connected, not falling, etc... We don't want to test the `websocket` class, we want to test our classes

This will require more work though as we have to figure out the responses (part of the issue to describe the serde process).

```

midokura-silvia:
@mccorby 
I discussed this helper function with Jason, and IMHO it is the easiest thing to get around the websocket connection problem that he experienced in #2373.
You are right, we should split unit tests from integration tests (websocket tests are definitely integration tests). But this will be a separate bigger task.

midokura-silvia:
Just noted that I was a bit too quick to merge the pull request. I will open a new one that will remove all the redundant time.sleep() commands.

Jasopaum:
Oops, I thought I had removed them! Thanks

PR #2395, opened 2019-07-25
reduced the timeout interval on windows due to C timeval overflowerror

amit-rastogi:
WebsocketClientWorker call to connect() method fails on Windows due to "OverflowError: timeout doesn't fit into C timeval". Reduced the TIMEOUT_INTERVAL in syft/workers/websocket_client.py. After this change WebsocketClientWorker connect() call is successful on Windows.


DanyEle:
Well, such a high timeout was introduced to avoid having timeouts on Raspberry PIs when transferring data. However, reducing it 999999 seconds (277.7 hours) is still a reasonably high amount.

amit-rastogi:
Yes that makes sense. I can think of 2 alternatives here-
1. Reduce the timeout to 999999 seconds for every platform (including Raspberry PIs). 999999 seconds should be a higher value for Raspberry PIs as well.
2. If there is a way to detect that we are executing on a Raspberry PI platform than keep the higher value 9999999 seconds for the Raspberry PI platform and reduce the timeout to a small interval maybe 300 seconds for non-Raspberry PIs.

amit-rastogi:
I'm going ahead with reducing the timeout to 999999 seconds (277.7 hours) for every platform (including Raspberry PIs). It should still be a higher value for Raspberry PIs and connect() would also not break on Windows due to C timeval overflow error. Also, the code would not get cluttered with platform checks. Please let me know in case there's any feedback on this.

PR #2398, opened 2019-07-25
Add field attribute to LPT

Jasopaum:
The LargePrecisionTensor class was missing the field that was in the original FixedPrecisionTensor


test/torch/tensors/test_large_precision.py
```diff
@@ -17,23 +17,23 @@ def test_wrap(workers):
 
 def test_fix_prec(workers):
     x = torch.tensor([1.5, 2.0, 3.0])
-    enlarged = x.fix_prec(internal_type=torch.int16, precision_fractional=256)
+    enlarged = x.fix_prec(internal_type=torch.int16, precision_fractional=128)
```
LaRiffle: Can you give an intuition for moving 256->128?

Jasopaum: For a fixed_precision of 256 and this tensor, we would need to have a field of size more than 10**512 and I found it to be a bit high. But I admit that it's kind of a lightweight justification, I can change it back if needed!

test/torch/tensors/test_large_precision.py
```diff
@@ -162,23 +162,17 @@ def test_concat_ops():
     assert torch.all(torch.eq(expected, result.float_precision()))
 
 
-def test_uint8_representation_not_allowed_with_negative_values(workers):
-    x = torch.tensor([[-1.5, 2.0, 3.0], [4.5, 5.0, 6.0]])
-    with pytest.raises(AssertionError):
-        x.fix_prec(internal_type=torch.uint8, precision_fractional=256)
-
-
```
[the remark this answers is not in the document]
Jasopaum: Because as we work in a field, we don't need to think about negative numbers anymore

PR #2402, opened 2019-07-28
[WIP] Refactor Polynomial Tensor 

kamathhrishi:
Polynomial tensor was previously commented out due to some test errors due to it.  This PR aims to resolve it. Also this pr aims to make polynomial tensor part of syft chain with end goal being passing additive and fixed precision tensors passing tests.


test/torch/tensors/test_polynomial.py
```diff
@@ -1,152 +1,159 @@
...
+
+    x = torch.tensor(np.linspace(-3, 3, 10), dtype=torch.double)
+
+    expected = torch.tensor(
+        [0.0498, 0.0970, 0.1889, 0.3679, 0.7165, 1.1176, 2.9503, 5.1088, 10.2501, 20.2955],
+        dtype=torch.double,
+    )
+
+    # allclose function to compare the expected values and approximations with fixed precision
+    result = poly_tensor.get_val("exp", x)
```
iamtrask: This is incorrect. PolynomialTensor should be a part of the class hierarchy of "x", not a function on which x is called.

syft/frameworks/torch/tensors/interpreters/polynomial.py
```diff
@@ -1,257 +1,346 @@
...
+        # reduce precision of interpolated coefficients
+        precision = self.precision
+        coefs = [int(x * 10 ** precision) / 10 ** precision for x in coefs]
+
+        # approximation function
+        f_interpolated = np.poly1d(coefs)
+
+        return f_interpolated
+
+    def sigmoid(self):
```
iamtrask: this method is running sigmoid inline. This is incorrect given that sigmoid() does not have an underscore (also this means it will not work with autograd). It should be returning a new tensor with the results. self.child should be unchanged. 

If the other methods are this way as well they will also need to be modified.

syft/frameworks/torch/tensors/interpreters/polynomial.py
```diff
@@ -1,257 +1,355 @@
...
+                    (len(self.sigmoid_coeffs) - 1) - i
+                ].child
+
+        else:
+
+            for i in range(0, len(self.sigmoid_coeffs)):
+
+                val += (x ** i) * self.encrypt_fn["sigmoid"][(len(self.sigmoid_coeffs) - 1) - i]
+
+        return val
```
iamtrask: This doesn't look like it's properly wrapped. 

On line 185: 
x = self.child

On line 190:
val += (x ** i) + .....

which means val is the same type as x

Thus, val is the same type as self.child which means it's the wrong type. 

Relatedly, shouldn't there be a hooking decorator on this method?

syft/frameworks/torch/tensors/interpreters/polynomial.py
```diff
@@ -1,257 +1,355 @@
...
+            self.exp_coeffs = [1, (1 / 2), (1 / 6), (1 / 24), (1 / 120), (1 / 840), (1 / 6720)]
+            self.tanh_coeffs = [(1), 0, (-1 / 3), 0, (2 / 15), 0, (-17 / 315)]
+
+        else:
+
+            self.sigmoid_coeffs = torch.tensor(self.func_approx["sigmoid"].coef)
+            self.exp_coeffs = torch.tensor(self.func_approx["exp"].coef)
+            self.tanh_coeffs = torch.tensor(self.func_approx["tanh"].coef)
+
+    def get_encrypt_function(self):
```
Jasopaum: Are you filling a dict with the coefficients for all the functions here?
If so couldn't you wait to use a function to create the appropriate tensor?

And also, why are `self.exp_coeffs`, etc. lists and not directly tensors

kamathhrishi: Thanks for the suggestion, Jason. This is a refractor of the old PR. I realized its inefficient considering how only the required coefficients could be used when required.

syft/frameworks/torch/tensors/interpreters/polynomial.py
kamathhrishi: It works fine now :)

iamtrask:
> Interesting progress going on!
> I have a question:
> For FixedPrecision I have .fix_precision() and float_precision() to revert
> For AdditiveShared I have .share() and .get()
> For Polynomial we have .poly() _but what's the one to revert_?

once you go .poly() you never go back!

kamathhrishi:
I will be working on this issue https://github.com/OpenMined/PySyft/issues/2622 and then taking up this. We need a way generalized way to handle unequal chains.

Issue #2392, opened 2019-07-25
One-worker Bug


wentaiwu92:
**Describe the bug**
When we set **only one client/work** in federated learning task, the corresponding _FederatedDataLoader_ does not work:

"....\syft\frameworks\torch\federated\dataloader.py", line 245, in __next__
    iterator = self.iterators[0]
IndexError: list index out of range

I checked the code of **dataloader.py** and found the probable problem here:
```       
        if iter_per_worker:
            self.num_iterators = len(self.workers)
        else:
            # You can't have more iterators than n - 1 workers, because you always
            # need a worker idle in the worker switch process made by iterators
            self.num_iterators = min(num_iterators, len(self.workers) - 1)
```
where the num_iterators is set to 1, making the last statement of the codes above yields 0 (because min(1, 1-1) = 0) when len(self.workers) = 1. Consequently, no iterators available and thus the dataloader fails to work.

I wanted to test one-work FL for some reason and this bug appeared.

ShivamSRS:
I'll look into this

Ankit-Dhankhar:
@wentaiwu92 you can use ```iter_per_worker=True``` while calling FederatedDataLoader. That will allow you to have one iterator for one worker.
I hope that will help if not let me know I would be happy to help :)

kamathhrishi:
Close this issue :) ?

wentaiwu92:
Thanks ShivamSRS and kamathhrishi, **PySyft** is really helpful for FL.

PR #2407, opened 2019-07-28
One worker bug

Ankit-Dhankhar:
In reference to issue #2392, Single iterator does need to shift worker IMHO.  Thus, for case of 1 worker we can have 1 iterator insetead of general (worker - 1) iterators.


DanyEle:
I also experienced this bug personally when running some tutorials. Would be very nice to have a fix for it when using one single worker!

robert-wagner:
@Ankit-Dhankhar Could you add a test to ensure we don't get this bug in the future

Ankit-Dhankhar:
Sure I will add a test case for same :)

Ankit-Dhankhar:
Test case added, please have a look

PR #2410, opened 2019-07-29
Clear objects for ObjectStorage with websocket connection

midokura-silvia:
In this pull request, we make clear_objects() callable on a ObjectStorage that is located within a WebsocketServerWorker. 
The following modifications are contained:
1) clear_objects() moved from BaseWorker to ObjectStorage
2) return value removed (returned self before, not sure why) => NOTE: modification in test_udacity.py!
3) added option to send() function to specify whether the resulting pointer should be in charge of garbage collection.


syft/workers/websocket_client.py
```diff
@@ -80,15 +79,17 @@ def _send_msg(self, message: bin, location) -> bin:
             "make hook.local_worker a WebsocketClientWorker?",
         )
 
-    def _receive_action(self, message: bin) -> bin:
+    def _forward_to_websocket_server_worker(self, message: bin) -> bin:
         self.ws.send(str(binascii.hexlify(message)))
-        response = binascii.unhexlify(self.ws.recv()[2:-1])
+        # response = binascii.unhexlify(self.ws.recv()[2:-1])
+        raw_response = self.ws.recv()
+        response = binascii.unhexlify(raw_response[2:-1])
```
LaRiffle: Do you want to remove the comment?

syft/frameworks/torch/pointers/object_wrapper.py
```diff
@@ -92,7 +92,7 @@ def create_pointer(
             id_at_location=id_at_location if id_at_location is not None else object.id,
             tags=object.tags,
             description=object.description,
-            garbage_collect_data=garbage_collect_data,
+            garbage_collect_data=False if garbage_collect_data is None else garbage_collect_data,
```
LaRiffle: Not sure why this is different from the previous implem where the default is also False but stated in the signature

syft/frameworks/torch/pointers/pointer_tensor.py
```diff
@@ -219,7 +219,7 @@ def create_pointer(
                 id_at_location=id_at_location,
                 owner=owner,
                 id=ptr_id,
-                garbage_collect_data=garbage_collect_data,
+                garbage_collect_data=True if garbage_collect_data is None else garbage_collect_data,
```
LaRiffle: Not sure why this is different from the previous implem where the default is also False but stated in the signature

syft/frameworks/torch/pointers/object_wrapper.py
midokura-silvia: The issue is that I have a chain of functions that will end up calling the create_pointer() function. 
And in the upper layers I would like to be able to distinguish between 1) argument not provided => None and 2) argument provided => True/False. And the two separate implementations of create_pointer have different default values for garbage_collect_data.

LaRiffle: Ok so the main argument would be not to set a default for this boolean value in the signature which differs for 2 functions?

midokura-silvia:
@iamtrask Could you take a look whether you are okay with the change in test_udacity.py, please?

midokura-silvia:
I see the point. But returning self then tries to serialize an instance of WebsocketServerWorker to send it over the network. 
We can catch that case and have two different behaviors (VirtualWorker vs WebsocketServerWorker). But it gets more complicated. Or create a separate function with another name, that does the same but does not return self.

midokura-silvia:
> Hey Silvia - if we make a change to that test then I probably have to go back and re-record a non-trivial number of videos. Can we leave the "return self" in for now. That would be quite a bit of work for me for a relatively small change.

I put the "return self" back in. And I added the possibility to deactivate the return value via an argument to the function. This way, I avoid the problem of serializing the return value.

PR #2415, opened 2019-07-30
make use of secureNN div for AdditiveSharingTensor

Jasopaum:
Some problems occurred:
- [x] In the secureNN division, we multiply the divisor by `2**Q_BITS` at some point, which overflows.
- [x] When dividing a FPT>AST by another FPT>AST, the scaling (`base ** precision_fractional`) disappears. We need to override the FPT division to do the inverse of a truncation after the division.

Also secureNN division is super slow


iamtrask:
Epic!

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -215,90 +215,135 @@ def t(self, _self, *args, **kwargs):
...
             # we swap operators so that we do the same operation as above
-            new_self, new_other, _ = syft.frameworks.torch.hook_args.unwrap_args_from_method(
-                "mul", self, other, None
-            )
+            if cmd == "mul":
+                new_self = other.child
+                new_other = self
+            elif cmd == "div":
+                # TODO how to divide by AST?
+                raise NotImplementedError("Division by AST not implemented")
```
LaRiffle: "Division of a FPT by an AST not implemented" would be more clear
Also let's not use FPT & AST abbreviations in our debug messages, people might not know them

andrelmfarias:
Commenting here just to follow this PR. Will need it for Encrypted Linear Regression.

Issue #2430, opened 2019-08-01
Local Worker unaware of itself


aristizabal95:
**Describe the bug**
When creating the torch hook, the local worker is initialized but unaware of itself. This means that running `sy.hook.local_worker._known_workers` returns an empty dictionary, when it should include the local worker. This makes every subsequent worker unaware of the local worker, since each new worker copies the local worker known workers.

**To Reproduce**
Working script demonstrating the issue:
```python
import torch as th
import syft as sy

hook = sy.TorchHook(th)
print(hook.local_worker._known_workers)
# => {}
```

**Expected behavior**
`hook.local_worker._known_workers` should return `{'me': <VirtualWorker id:me #objects:0>}`

**Desktop (please complete the following information):**
 - OS: MacOS Mojave 10.14.5
 - Version `syft==0.1.19a1`

**Additional context**
This issue is related to #2426

Issue #2426, opened 2019-08-01
Local worker *me* not recognized by other workers when explicitly declared as crypto provider


aristizabal95:
**Describe the bug**
Sharing a sent tensor between workers, and explicitly declaring the crypto provider as `sy.local_worker` causes the warning `Worker [worker_id] couldn't recognize worker me`. This warning causes problems when trying to execute get() on the sent, shared tensor, throwing an exception `AttributeError: 'str' object has no attribute 'id'`. Strange enough, this doesn't happen if `sy.local_worker` is implicitly assigned as crypto provider by not providing any crypto provider while sharing the tensor.

**To Reproduce**
Working script to demonstrate the problem.
```python
import torch as th
import syft as sy

hook = sy.TorchHook(th)

# Declare workers
alice = sy.VirtualWorker(id="alice", hook=hook)
bob = sy.VirtualWorker(id="bob", hook=hook)
jack = sy.VirtualWorker(id="jack", hook=hook)

# create a tensor
a = th.zeros(1).long()

# send the tensor
a = a.send(alice)

# share the tensor with me as an implicit crypto provider
a = a.share(bob, jack) # No warning is shown here

# get the shared tensor
a = a.get() # AdditiveSharingTensor returned, *crypto provider: me*

# Redeclare tensor
a = th.zeros(1).long()

# send the tensor
a = a.send(alice)

# share the tensor with me as an explicit crypto provider
a = a.share(bob, jack, crypto_provider=sy.local_worker) 
# W0801 11:56:01.096644 4346246592 base.py:577] Worker alice couldn't recognize worker me

# get the shared tensor
a = a.get() # AttributeError: 'str' object has no attribute 'id'
```

**Expected behavior**
There should be no difference between implicit and explicit declaration of the local worker as crypto provider.

**Desktop (please complete the following information):**
 - OS: MacOS Mojave 10.14.5
 - Version `syft==0.1.19a1`

**Additional context**
A similar script can be seen inside `syft/frameworks/torch/crypto/securenn.py`, for example in `_shares_of_zero()` declaration:
```python
def _shares_of_zero(size, field, crypto_provider, *workers):
    """
    Return n in [0, max_value-1] chosen by a worker and sent to all workers,
    in the form of a MultiPointerTensor
    """
    u = (
        torch.zeros(size)
        .long()
        .send(workers[0])
        .share(*workers, field=field, crypto_provider=crypto_provider, **no_wrap)
        .get()
        .child
    )

    return u
[...]

PR #2431, opened 2019-08-01
Make the local worker aware of itself on TorchHook creation.

aristizabal95:
This PR solves issue #2430 and issue #2426. When the local_worker is being created, `hook.local_worker` is `None`. Because of this, the local_worker is not added to the `hook.local_worker._known_workers` dictionary, making himself and any subsequent worker unaware of local_worker's existence.


robert-wagner:
Hey @aristizabal95 could you add a test to make sure this bug does not pop up again (basically just testing to see that the local worker is in the list of known workers)

aristizabal95:
Just realized there's a test file `test_local_worker.py`, might be better to put the test there?

aristizabal95:
@robert-wagner I added the test case. Hope everything looks good!

PR #2436, opened 2019-08-02
Created an actual Message type and moved Plan out of federated

iamtrask:


syft/federated/__init__.py
```diff
@@ -1,10 +1,5 @@
-from syft.federated.plan import Plan
 from syft.federated.train_config import TrainConfig
 from syft.federated import federated_client
 from syft.federated.federated_client import FederatedClient
 
-from syft.federated.plan import func2plan
-from syft.federated.plan import method2plan
-from syft.federated.plan import make_plan
-
```
robert-wagner: All needs to change as well

syft/serde/serde.py
```diff
@@ -48,7 +48,7 @@
 from syft.workers import AbstractWorker
 from syft.workers import VirtualWorker
 
-from syft.federated import Plan
+from syft.msg import Plan, Message
```
robert-wagner: Put these on separate lines

syft/workers/base.py
```diff
@@ -85,6 +85,8 @@ def __init__(
         self.verbose = verbose
         self.auto_add = auto_add
         self.msg_history = list()
+        self.promises = {}
+        self.tensor2promise_id = {}
```
robert-wagner: What is this parameter for?

syft/msg/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .message import Message
```
robert-wagner: Use absolute imports please

syft/msg/message.py
```diff
@@ -0,0 +1,38 @@
+import syft as sy
+
+from syft.workers import AbstractWorker
+
+
+class Message:
```
robert-wagner: Should plans inherit from message? Also this class seems like it might be a security risk, what prevents me from creating an arbitrary message?

syft/workers/base.py
```diff
@@ -217,8 +220,9 @@ def send_msg(self, msg_type: int, message: str, location: "BaseWorker") -> objec
         """
         if self.verbose:
             print(f"worker {self} sending {msg_type} {message} to {location}")
+
         # Step 0: combine type and message
-        message = (msg_type, message)
+        message = sy.msg.Message(msg_type, message)
```
robert-wagner: This would look much cleaner if you imported message. That being said, what is this point if this change, why not just keep it a tuple

syft/workers/base.py
```diff
@@ -251,7 +255,10 @@ def recv_msg(self, bin_message: bin) -> bin:
             self.msg_history.append(bin_message)
 
         # Step 0: deserialize message
-        (msg_type, contents) = sy.serde.deserialize(bin_message, worker=self)
+        msg = sy.serde.deserialize(bin_message, worker=self)
```
robert-wagner: Same as above for my question about tuples

syft/workers/base.py
```diff
@@ -843,6 +850,9 @@ def generate_triple(
         c_shared = c.share(*locations, field=field, crypto_provider=self).child
         return a_shared, b_shared, c_shared
 
+    def save_promise(self):
```
robert-wagner: Why does this method that does nothing exist?

syft/workers/base.py
```diff
@@ -863,4 +873,6 @@ def create_message_execute_command(
         """
         if return_ids is None:
             return_ids = []
-        return tuple([codes.MSGTYPE.CMD, [[command_name, command_owner, args, kwargs], return_ids]])
+        return sy.Message(
```
robert-wagner: Again, what advantage does this give us over a tuple

test/workers/test_virtual.py
```diff
@@ -93,7 +93,7 @@ def test_recv_msg():
     # Test 2: get tensor back from alice
 
     # Create message: Get tensor from alice
-    msg = (MSGTYPE.OBJ_REQ, obj.id)
+    msg = sy.msg.Message(MSGTYPE.OBJ_REQ, obj.id)
```
robert-wagner: Why are there no other tests for messages?

syft/federated/__init__.py
iamtrask: done

syft/serde/serde.py
iamtrask: Done

syft/workers/base.py
iamtrask: oops - bit early for that - this was originally a bigger PR but i had to stop because it was already getting quite large (better to merge along the way)

syft/msg/__init__.py
iamtrask: done

syft/msg/message.py
iamtrask: Not any more of a security risk than it was before, messages are still whitelisted based on the msg_type.  This is literally just a class where the tuple used to be which carried messages. Starting small.

syft/__init__.py
```diff
@@ -48,11 +48,14 @@
 from syft.grid import VirtualGrid
 
 # Import federate learning objects
-from syft.federated import Plan
```
midokura-silvia: Could we limit this pull request to only contain the introduction of the message type?
It is not clear to me, why Plan should be in a message folder.

syft/workers/base.py
iamtrask: It's a myriad of reasons, mostly organization. I'm working on some more advanced async/protocol/promise functionality and it's become quite annoying that messages aren't an actual type with actual to-strings which help with introspection.

syft/codes.py
```diff
@@ -8,6 +8,7 @@ class MSGTYPE(object):
     GET_SHAPE = 7
     SEARCH = 8
     FORCE_OBJ_DEL = 9
+    PROMISE = 10
```
midokura-silvia: is this within the scope of this pull request?

syft/workers/base.py
iamtrask: eventually Message objects will be passed to send_msg but I didn't want to do too much in one PR. This is an incremental PR for sure.

syft/workers/base.py
iamtrask: same answer as above

syft/workers/base.py
iamtrask: removed

syft/workers/base.py
iamtrask: i'm working toward something which will give us more advantages - but i didn't want to make too many changes for one PR - so this is an incremental one.

test/workers/test_virtual.py
iamtrask: There is no new functionality for Message yet. I literally just replaced the tuple with a class with two attributes. Zillions of other tests already use it.

syft/workers/base.py
```diff
@@ -843,6 +850,9 @@ def generate_triple(
         c_shared = c.share(*locations, field=field, crypto_provider=self).child
         return a_shared, b_shared, c_shared
 
+    def save_promise(self):
+        ""
```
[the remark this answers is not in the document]
iamtrask: removed it instead

syft/__init__.py
iamtrask: The message folder will contain all messages, groups of messages, and graphs of messages. Plan was in the federated folder which implied that it's only useful for Federated Learning.

syft/codes.py
iamtrask: removed

test/workers/test_virtual.py
midokura-silvia: There might be tests that are using it, but all of them are integration tests (meaning, that they use something that uses the class Message). It would be good to have a basic unit test as well.

iamtrask: fair - i'll make a test :)

iamtrask: i'll test serde - which is new

syft/workers/base.py
robert-wagner: Still would look cleaner importing message

iamtrask: Also we should be testing message construction - which we are not.

syft/workers/base.py
robert-wagner: Still should import message though

syft/workers/base.py
iamtrask: This will also allow us to add functionality inside the messages to make sure they're well formed. At present we only find out if they end up causing an error on the other side.

syft/__init__.py
midokura-silvia: I agree that Plan can be used for more than federated learning. But I am not sure that it should be in a message folder neither.

syft/workers/base.py
iamtrask: Also this will make the code easier to write - at present it's totally undocumented what goes into a message (aka - our protocol is undocumented). A good first step here is setting out messages types and what information goes in each message.

iamtrask: This will also allow us to do smarter simplification/compression in the future

iamtrask: So yeah - lots of good reasons I think. :)

test/workers/test_virtual.py
iamtrask: groovy - added a test for message serde

syft/workers/base.py
iamtrask: Done

syft/workers/base.py
iamtrask: also - changed it to import Message directly

syft/msg/message.py
```diff
@@ -0,0 +1,38 @@
...
+            ptr (Message): a Message
+        Returns:
+            tuple: a tuple holding the unique attributes of the message
+        Examples:
+            data = simplify(ptr)
+        """
+
+        return (ptr.msg_type, sy.serde._simplify(ptr.contents))
+
+        # a more general but slower/more verbose option
```
robert-wagner: Should this comment be here?

test/message/test_message.py
```diff
@@ -0,0 +1,27 @@
+import syft as sy
+import syft
+
+import torch as th
```
robert-wagner: Why are all these imports here?

syft/workers/base.py
midokura-silvia: My prefered style is to use:
from sy import msg
and then use msg.Message(...)
That way you always have the source referenced. 
And additionally you follow the google style guide, see http://google.github.io/styleguide/pyguide.html, section 2.2

syft/__init__.py
iamtrask: a plan is a cache of messages - seemed pretty close. Other ideas for what to name the folder?

iamtrask: maybe rename the folder "messaging"?

syft/workers/base.py
iamtrask: lol - ok - will fix

iamtrask: fixed

syft/msg/message.py
iamtrask: removed

test/message/test_message.py
iamtrask: removed

test/message/test_plan.py
```diff
@@ -420,9 +421,9 @@ def forward(self, x):
 
 def test_replace_worker_ids_two_strings(hook):
     plan = sy.Plan(id="0", owner=hook.local_worker, name="test_plan")
-    _replace_message_ids_orig = sy.federated.Plan._replace_message_ids
+    _replace_message_ids_orig = sy.msg.Plan._replace_message_ids
```
midokura-silvia: Could you unify the way the import is done? (sy.Message vs sy.msg.Message vs sy.msg.Plan)
sy.msg.Plan => msg.Plan

test/message/test_message.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def test_message_serde(hook):
+
+    x = sy.Message(0, [1,2,3])
+    x_bin = sy.serde.serialize(x)
+    y = sy.serde.deserialize(x_bin, sy.local_worker)
+
+    assert x.contents == y.contents
+    assert x.msg_type == y.msg_type
```
midokura-silvia: missing newling

syft/__init__.py
midokura-silvia: @mari-linhares what do you think about the naming of the folder and the location of Plan?

test/message/test_message.py
iamtrask: done

test/message/test_plan.py
iamtrask: done

syft/__init__.py
mari-linhares: I agree that we should rename it eventually. Not sure if "message" is the best option but I don't have a good naming in mind neither.

iamtrask: msg, cache,... something. 

It's relatively easy to change in the future with modern IDEs. But I think all are better than having it in federated.

PR #2443, opened 2019-08-06
Add support for float_prec() on pointer tensors

aristizabal95:
Currently `float_precision()` is not supported for pointer tensors. This code adds support and includes testing for both `fix_precision()` and `float_precision()` on pointer tensors.

Had to manually wrap the response to the `send_command()` as it returned an unwrapped pointer. Might cause issues in the future if later on any change makes the response return a wrapped pointer.


syft/frameworks/torch/pointers/pointer_tensor.py
```diff
@@ -307,6 +307,24 @@ def fix_prec(self, *args, **kwargs):
...
+        Returns:
+            A pointer to a Tensor
+        """
+
+        # Send the command
+        command = ("float_prec", self, args, kwargs)
+
+        response = self.owner.send_command(self.location, command)
+
+        # For some reason the response has no Wrapper. Must be manually added here
```
mari-linhares: Hmm, I don't like that you have to wrap the response for this specific case but not for the others... I think it's worth exploring why this is the case...

test/torch/pointers/test_pointer_tensor.py
```diff
@@ -309,3 +309,36 @@ def test_raising_error_when_item_func_called(workers):
...
+    Ensure .fix_precision() works as expected.
+    """
+    bob = workers["bob"]
+
+    tensor = torch.tensor([1, 2, 3, 4.0])
+    ptr = tensor.send(bob)
+
+    ptr = ptr.fix_precision()
+
+    assert type(ptr.child) is syft.frameworks.torch.pointers.pointer_tensor.PointerTensor
```
mari-linhares: can you import PointerTensor and use it here?

test/torch/pointers/test_pointer_tensor.py
```diff
@@ -309,3 +309,36 @@ def test_raising_error_when_item_func_called(workers):
...
+    Ensure .float_precision() works as expected.
+    """
+    bob = workers["bob"]
+
+    tensor = torch.tensor([1, 2, 3, 4.0]).fix_prec()
+    ptr = tensor.send(bob)
+
+    ptr = ptr.float_precision()
+
+    assert type(ptr.child) is syft.frameworks.torch.pointers.pointer_tensor.PointerTensor
```
mari-linhares: same about importing.

test/torch/pointers/test_pointer_tensor.py
```diff
@@ -309,3 +309,36 @@ def test_raising_error_when_item_func_called(workers):
...
+
+    tensor = torch.tensor([1, 2, 3, 4.0])
+    ptr = tensor.send(bob)
+
+    ptr = ptr.fix_precision()
+
+    assert type(ptr.child) is syft.frameworks.torch.pointers.pointer_tensor.PointerTensor
+    assert (
+        type(ptr.get().child)
+        is syft.frameworks.torch.tensors.interpreters.precision.FixedPrecisionTensor
```
mari-linhares: can you import FixedPrecisionTensor and use it here?

test/torch/pointers/test_pointer_tensor.py
```diff
@@ -309,3 +309,36 @@ def test_raising_error_when_item_func_called(workers):
     pointer = PointerTensor(id=1000, location=workers["alice"], owner=workers["me"])
     with pytest.raises(RuntimeError):
         pointer.item()
+
+
+def test_fix_prec_on_pointer_tensor(workers):
```
mari-linhares: Can you add the same test but for a pointer of a pointer?

test/torch/pointers/test_pointer_tensor.py
```diff
@@ -309,3 +309,36 @@ def test_raising_error_when_item_func_called(workers):
...
+    ptr = ptr.fix_precision()
+
+    assert type(ptr.child) is syft.frameworks.torch.pointers.pointer_tensor.PointerTensor
+    assert (
+        type(ptr.get().child)
+        is syft.frameworks.torch.tensors.interpreters.precision.FixedPrecisionTensor
+    )
+
+
+def test_float_prec_on_pointer_tensor(workers):
```
mari-linhares: Can you add the same test but for a pointer of a pointer?

syft/frameworks/torch/pointers/pointer_tensor.py
aristizabal95: I've found why, but I'm not sure how to fix it. the definition of `float_prec()` in `interpreters/native.py` is 
```
def float_prec(self):
    return self.child.float_precision()
```
which differs to `fix_prec()` here
```
if self.is_wrapper:
    self.child = self.child.fix_prec(*args, **kwargs)
    return self
```
What I can see is that it is assumed that a float_precision tensor is wrapped, and must return an unwrapped torch tensor. I see two ways of solving this:

1. check if `self.child` is a pointer, in which case return the wrapped tensor
```
def float_prec(self):
    if self.child is PointerTensor:
        self.child = self.child.float_prec()
        return self
    return self.child.float_precision()
```
2. Check that if a torch tensor has a wrapper, then remove the wrapper, as it is useless. 

The first option is the easiest, but I feel is somewhat dirty to have to manually check for type and specify different behavior. The second one might be out of my knowledge on how to implement it.

What do you think?

mari-linhares: I agree that 2. looks like a better option. @LaRiffle @robert-wagner do you know if 2. is possible/worth implementing?

test/torch/pointers/test_pointer_tensor.py
```diff
@@ -309,3 +310,67 @@ def test_raising_error_when_item_func_called(workers):
...
+    """
+    bob = workers["bob"]
+
+    tensor = torch.tensor([1, 2, 3, 4.0]).fix_prec()
+    ptr = tensor.send(bob)
+
+    ptr = ptr.float_precision()
+
+    assert type(ptr.child) is PointerTensor
+    assert type(ptr.get()) is torch.Tensor
```
mari-linhares: This should be a FixedPrecisionTensor, right?

test/torch/pointers/test_pointer_tensor.py
```diff
@@ -309,3 +310,67 @@ def test_raising_error_when_item_func_called(workers):
...
+    alice = workers["alice"]
+
+    tensor = torch.tensor([1, 2, 3, 4.0]).fix_prec()
+    ptr = tensor.send(bob)
+    ptr = ptr.send(alice)
+
+    ptr = ptr.float_precision()
+
+    assert type(ptr.child) is PointerTensor
+    assert type(ptr.get().get()) is torch.Tensor
```
mari-linhares: This should be a FloatPrecisionTensor, right?

aristizabal95:
I think I realized something that may be related with `wrap()` special cases, this time with `fix_precision()`. I'm getting an error with doing `model.fix_precision()` when that model has been previously sent to a worker. 

The parameters are in this form `Parameter>[PointerTensor | me:8280105820 -> S1:89778439233]`.

`fix_prec()` assumes that if it isn't wrapped, then it is a tensor:
```
if self.is_wrapper:
    self.child = self.child.fix_prec(*args, **kwargs)
    return self
```
and so in this case it tries to operate directly into the `Parameter` object, causing an `AttributeError: 'numpy.ndarray' object has no attribute 'wrap'`

This is another issue altogether, but seems like checking for `wrap` and doing special behavior is messing up things.

Note: This used to work for me on syft 0.1.19a1.

aristizabal95:
Since there was no further discussion about possible solutions for the wrapping problem, I decided to implement solution 1.

aristizabal95:
@mari-linhares Hi, this PR has been silent for a while, even though I believe the changes requested should be reviewed. I'm not sure if in that case I should mark the conversations as resolved, so that another approval is done.

Just in case, I indicate in each revision that a `FloatPrecisionTensor` is a normal torch `Tensor`, and there is no class called `FloatPrecisionTensor` (I checked). If this is the case, then the tests need no modification. 

I'd be happy if I could recieve some feedback from you. I almost forgot I did this PR and would like to see float precision on pointers working for some projects I'm working on.

Thanks!

test/torch/pointers/test_pointer_tensor.py
LaRiffle: - Call fix_precision() on the ptr before calling float_precision() (otherwise it doesn't make sense even if it doesn't fail)
- You can use bob._objects[ptr.id_at_location] to inspect a remote object (if you want to check types for example)
- Prefer maybe isinstance(..., ...) instead of type() == ..., it is more robust for sub classes

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -628,6 +628,10 @@ def enc_fix_prec(self):
         return self.child.fix_precision()
 
     def float_prec(self):
+        if type(self.child) is PointerTensor:
```
LaRiffle: Prefer isinstance(...) to type(...) ==

aristizabal95:
@LaRiffle @mari-linhares just made the changes, please check if everything looks right. Thanks!

Issue #2444, opened 2019-08-06
Add "disable_garbage_collection" and "enable_garbage_collection" flags to native.py


iamtrask:
**Is your feature request related to a problem? Please describe.**
By default, when we delete a pointer to a tensor, it sends a command to the remote worker telling it to delete the thing being pointed to automatically.

This is a GREAT default, BUT, sometimes we want to turn it off. Currently it's quite cumbersome. Let's say we have a tesnor.

x = th.tensor([1,2,3,4,5])

...and we send it to someone

x_pointer = x.send(bob)

...if we want Bob to KEEP that tensor even when we delete our pointer, we have to do this....

x_pointer.child.garbage_collect_data = False

This is cumbersome... it would be better if we could just do

x_pointer.disable_garbage_collection

In fact, it would be even BETTER if this flag returned ```self``` allowing us to do

x_pointer = x.send(bob).disable_garbage_collection

It would also be convenient to have a shorthand version

x_pointer = x.send(bob).disable_gc

AND to also have a specific parameter we can update

x_pointer = x.send(bob)
x_pointer.gc = False
x_pointer.garbage_collection = False

iamtrask:
For reference on how this is implemented, the implementation of a "description" feature was very similar, which you can see here.

https://github.com/OpenMined/PySyft/blob/dev/syft/frameworks/torch/tensors/interpreters/native.py#L93

The whole project should be doable by only changing native.py (see the link above)

Shashankjain12:
Hey @iamtrask can you assign me this issue.!?

hanare:
@iamtrask I have created a PR can you please review it ?
https://github.com/OpenMined/PySyft/pull/2445

hanare:
@iamtrask  This issue can be closed.

iamtrask:
Thank you!

PR #2445, opened 2019-08-06
Add disable_garbage_collection  flags to native.py  

hanare:


robert-wagner:
Hey @hanare Could you give this pr a more informative name?

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -210,6 +210,60 @@ def id(self):
...
+
+    @property
+    def disable_gc(self):
+        self.child.garbage_collect_data = False
+        if self.has_gc():
+            self.gc = False
+            self.garbage_collection = False
+        return self
+
+    def has_gc(self):
```
robert-wagner: This always returns true

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -210,6 +210,60 @@ def id(self):
...
+        return hasattr(self, "gc")
+
+    @property
+    def garbage_collection(self):
```
robert-wagner: what is the difference between this and gc?

test/torch/pointers/test_pointer_tensor.py
```diff
@@ -23,6 +23,42 @@ def test_send_default_garbage_collector_true(workers):
...
+    x = torch.Tensor([-1, 2])
+    x_ptr = x.send(workers["alice"])
+    x_ptr.gc = False
+    assert x_ptr.child.garbage_collect_data == False
+    assert x_ptr.gc == False, "property GC is not in sync"
+    assert x_ptr.garbage_collection == False, "property garbage_collection is not in sync"
+
+
+def test_send_gc_true(workers):
+    """Pointer tensor should be not garbage collected."""
```
robert-wagner: This string is wrong

test/torch/pointers/test_pointer_tensor.py
```diff
@@ -23,6 +23,42 @@ def test_send_default_garbage_collector_true(workers):
     assert x_ptr.child.garbage_collect_data
 
 
+def test_send_garbage_collect_data_false(workers):
+    """Pointer tensor should be not garbage collected."""
+    x = torch.Tensor([-1, 2])
+    x_ptr = x.send(workers["alice"])
+    x_ptr.garbage_collection = False
+    print(x_ptr.gc)
```
robert-wagner: remove this print statement

syft/frameworks/torch/tensors/interpreters/native.py
iamtrask: It should be the same - I thought it might be nice to have a shorthand version. Open to comments there.

iamtrask:
I agree with Bobby's comments

syft/frameworks/torch/tensors/interpreters/native.py
robert-wagner: My point for this is that repoducing the logic in 2 places is bad. If we change one of them and forget to change the other, that leads to unexpected behavior

hanare: Nothing, gc just shorthand form. It was mentioned in the description https://github.com/OpenMined/PySyft/issues/2444

syft/frameworks/torch/tensors/interpreters/native.py
hanare: removed

test/torch/pointers/test_pointer_tensor.py
hanare: Fixed

test/torch/pointers/test_pointer_tensor.py
hanare: fixed

syft/frameworks/torch/tensors/interpreters/native.py
iamtrask: go with Bobby's suggestion - he's usually right :)

robert-wagner: For clarification, I think we should keep both, just abstract them so that they call the same thing 🙂

test/torch/pointers/test_pointer_tensor.py
```diff
@@ -17,12 +17,47 @@ def test_create_pointer(workers):
 
 
 def test_send_default_garbage_collector_true(workers):
-    """Pointer tensor should be garbage collected by default."""
+    """ emote tensor should be garbage collected by default on deletion of the Pointer tensor pointing to remote tensor"""
```
robert-wagner: emote -> remote. Also could you split this up into 2 lines (its a bit wide on my screen)

PR #2449, opened 2019-08-07
Update CONTRIBUTING.md

iamtrask:


CONTRIBUTING.md
```diff
@@ -2,6 +2,10 @@
 
 ## Getting Started
 
+### Learn Git and Github
+
+If you're not too familiar with Git and Github, start by reviewing this guide. https://guides.github.com/activities/hello-world/
```
robert-wagner: One small nit on tone, This might be better phrased as "All our development is done using Git and Github. If you are not familiar with these a great guide can be found <link>"

iamtrask: done

PR #2451, opened 2019-08-07
Remove uneeded TODOs at test_gc.py

mari-linhares:


robert-wagner:
@mari-linhares travis did not like the change of the fixture name

mari-linhares:
> Overwriting variable names at sending in the test, is on purpose, to be sure nothing weird happens when people do this

I'll add a comment.

test/torch/tensors/test_gc.py
```diff
@@ -196,20 +199,18 @@ def test_implicit_garbage_collect_logging_on_pointer(workers):
...
+    sample_ptr = sample_data.send(local_worker)
 
     _ = sample_ptr.get()
-    assert sample_data not in client_worker._objects
+    assert sample_data not in local_worker._objects
 
-    client_worker.ws.shutdown()
-    client_worker.ws.close()
+    local_worker.ws.shutdown()
+    local_worker.ws.close()
```
midokura-silvia: I think local_worker.close() should be sufficient (replacing both local_worker.ws.shutdown() and local_worker.ws.close())

Issue #2478, opened 2019-08-09
Encrypted Training on MNIST tutorial test model on training data itself


Ankit-Dhankhar:
Max Zenk pointed out on slack(beginner channel) that in the tutorial: **Part 12 bis - Encrypted Training on MNIST** ```get_private_data_loaders``` functions loads the training set for training and testing.

Ankit-Dhankhar:
Issue Fixes :)

PR #2479, opened 2019-08-09
fix test_loader to load train data instead of test

Ankit-Dhankhar:
This PR fixes issue #2478 :)


mari-linhares:
@robert-wagner can you double-check the changes in requirements.txt?

PR #2480, opened 2019-08-09
add assertion at test_hook.py

mari-linhares:
Remove TODO.


test/torch/test_hook.py
```diff
@@ -147,8 +147,7 @@ def test_hook_tensor(workers):
     x = torch.tensor([1.0, -1.0, 3.0, 4.0], requires_grad=True)[0:2]
     x.send(workers["bob"])
 
-    # TODO: shouldn't there be an assertion here?
-    # assert True
+    assert True
```
midokura-silvia: Could we replace the assert True by something being set by the hooking process?
For example, we could check that some native attribute exists.

PR #2288, opened 2019-06-15
Added Advanced Tutorial Example - SMS spam prediction with handcrafted GRU

andrelmfarias:

PR #2491, opened 2019-08-10
Fix build issues by downgrading torch

mari-linhares:
For now, we had to downgrade torch (force version 1.1) and torchvision (force version 0.3)

PR #2492, opened 2019-08-10
Remote Gradient Clipping for PySyft Tensors

DanyEle:
Currently, as noticed by @andrelmfarias in his tutorial, the plain PyTorch `torch.nn.utils.clip_grad.py` method does not work for remote tensors, like the parameters of a remote RNN. This method is essential for preventing the exploding / vanishing gradient problem from occuring with RNNs and their LSTMs or GRU versions. 

I modified the in-place `torch.nn.utils.clip_grad.py` method from the PyTorch 1.1.0 version and adapted it to work in a remote manner, then hooked it to the torch.nn.utils.clip_grad method for PySyft remote tensors.

In the current implementation, I'm overriding the `torch.nn.utils.clip_grad` method, so the plain PyTorch `torch.nn.utils.clip_grad` method is no longer available when hooking PySyft. 

Should I use a different name or hooking strategy for this method, so that both the plain `torch.nn.utils.clip_grad` and the clip_grad_remote can live side-by-side? (one method would operate on plain PyTorch tensors and the other one on PySyft tensors) 

I tested it with the Federated SMS Spam Detection tutorial #2288  and it does indeed seem to be working.

image

P.S: Why are there so many commits being shown below??


andrelmfarias:
Well done on this PR @DanyEle 
Why aren't the tests passings? You have just added a new function to hook... weird...

DanyEle:
> Well done on this PR @DanyEle
> Why aren't the tests passings? You have just added a new function to hook... weird...

Thanks. It's most likely due to the PyTorch 1.2 being installed in the Travis CI build, because of the pytorch>=1.1 version in the requirements.txt file. There is already a PR open #2491 to address this issue. Other PRs seem to have the same tests not passing too.

syft/frameworks/torch/hook/hook.py
```diff
@@ -1134,3 +1135,47 @@ def optim_float_precision_(optim_self):
...
+            if isinstance(parameters, torch.Tensor):
+                parameters = [parameters]
+            parameters = list(filter(lambda p: p.grad is not None, parameters))
+            max_norm = float(max_norm)
+            norm_type = float(norm_type)
+            if norm_type == inf:
+                total_norm = max(p.grad.data.abs().max() for p in parameters)
+            else:
+                # Init tensor with value 0 and dimension 1 (plain scalar)
+                total_norm = torch.Tensor(1)
```
midokura-silvia: Could you replace torch.Tensor(1) by torch.zeros(1)?

DanyEle: Done!

syft/frameworks/torch/hook/hook.py
```diff
@@ -1134,3 +1135,47 @@ def optim_float_precision_(optim_self):
             return optim_self
 
         self.torch.optim.Optimizer.float_precision = optim_float_precision_
+
+        # Modification of torch/nn/utils/clip_grad.py
+        def clip_grad_norm_remote_(parameters, max_norm, worker=None, norm_type=2):
+            r"""Clips gradient norm of an iterable of parameters stored over a remote model
```
mari-linhares: remove `r`

syft/frameworks/torch/hook/hook.py
```diff
@@ -1134,3 +1135,47 @@ def optim_float_precision_(optim_self):
             return optim_self
 
         self.torch.optim.Optimizer.float_precision = optim_float_precision_
+
+        # Modification of torch/nn/utils/clip_grad.py
```
mari-linhares: Can you make more explicit in the comment why do we need a specific function for clipping remote tensors? maybe add comments to the lines that were changed and why/how they were changed?

syft/frameworks/torch/hook/hook.py
```diff
@@ -27,6 +27,7 @@
 
 from syft.exceptions import route_method_exception
 from syft.exceptions import TensorsNotCollocatedException
+from torch._six import inf
```
robert-wagner: Is there a better place to import inf from? six is mostly geared for 2:3 compatability

test/torch/test_hook.py
```diff
@@ -293,3 +293,15 @@ def initHidden(self):
         # so we better check it beforehand
         assert param.grad.data is not None
         param.data.add_(-learning_rate, param.grad.data)
+
+
+def test_remote_gradient_clipping(workers):
```
robert-wagner: Can you add a test to make sure that gradient clipping works locally as well

syft/frameworks/torch/hook/hook.py
DanyEle: I replaced it with 'from math import inf'. I suppose that should work too?

test/torch/test_hook.py
DanyEle: Added! And modified the method to support the modification of local tensors with gradient clipping as well.

syft/frameworks/torch/hook/hook.py
DanyEle: Made the comment at the beginning of the function more explicit about the motivation for having this extra function. Actually, the only two lines changed are the following ones:

```
total_norm = torch.zeros(1)
#Let's send the total norm over to
total_norm = total_norm.send(worker)
```

syft/frameworks/torch/hook/hook.py
DanyEle: Done!

PR #2497, opened 2019-08-12
Add error message when passing internal_type if not needed

Jasopaum:
Just a small modification to send a more informative error message when the user sets `internal_type` using `.fix_prec()` on a tensor that does not need large precision.

When doing, for instance, 
`torch.tensor([1.5, 2.0, 3.0]).fix_prec(internal_type=torch.int16, precision_fractional=3)`

The error message was:
TypeError: __init__() got an unexpected keyword argument 'internal_type'

Now, it is:
TypeError: internal_type not needed if data does not need LargePrecisionTensor to be stored


syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -682,6 +682,10 @@ def fix_prec(self, *args, storage="auto", field_type="int100", **kwargs):
             )
         else:
             assert not need_large_prec, "This tensor needs large precision to be correctly stored"
+            if "internal_type" in kwargs:
+                raise TypeError(
+                    "internal_type not needed if data does not need LargePrecisionTensor to be stored"
+                )
```
midokura-silvia: Does this need to be a TypeError? Or is a warning sufficient? 
What happens if you just ignore the internal_type argument? Would the code would work correctly?

Jasopaum: The error originally sent was a TypeError so what I did is just to catch it just before and send it with another message.
But I agree, we could remove this attribute from the kwargs and raise a warning, it would work the same. I also thought about this and wasn't sure what was the best. Do you prefer this solution?

midokura-silvia: My preference would be on removing the needless argument and emit a warning. And maybe change the message to make it clear that internal_type should not be provided (not just not needed) in this case.

Jasopaum: Done
It is now clear that the argument should not be provided but I find it weird because the user might not know that his tensor does not need large precision

midokura-silvia: Looks good to me. @robert-wagner Any comments on the usage of warnings?

iamtrask:
I'm very grateful for your attention to detail on these usability messages.

PR #2508, opened 2019-08-14
Fix MNIST parallel demo

midokura-silvia:
- Adapt code to new Message class 
- Update tests
- Remove downloading of MNIST dataset for tracing purposes


test/workers/test_websocket_worker.py
```diff
@@ -270,7 +270,7 @@ def test_evaluate(hook, start_proc):  # pragma: no cover
 
     dataset = sy.BaseDataset(data=data, targets=target)
 
-    kwargs = {"id": "evaluate_remote", "host": "localhost", "port": 8780, "hook": hook}
+    kwargs = {"id": "evaluate_remote", "host": "localhost", "port": 8790, "hook": hook}
```
robert-wagner: why the port change?

iamtrask: are you kidding? 8790 is clearly vastly superior.

midokura-silvia: I changed it in order to have a different port number than the ones used in the MNIST parallel demo. And yes, it's superior :-)

PR #2523, opened 2019-08-16
added torch.cat to ambiguous functions in hook_args.py

Jasopaum:
torch.cat wasn't working correctly when changing the number of tensors to cat


mari-linhares:
Hey @Jasopaum, can you add some tests for this?

Jasopaum:
Sure! But I don't where would be the best as this can be call by every type of tensor

mari-linhares:
Hey @Jasopaum,

Sorry for the delay! I think you can move this test to the [test_hook.py](https://github.com/OpenMined/PySyft/blob/dev/test/torch/test_hook.py) and create a new function (something like: test_torch_cat)

PR #2526, opened 2019-08-18
Modified Gradient Clipping to work with real remote/local model parameters

DanyEle:
Fixed the clip_grad_remote method to work with real-life model parameters, so as to work with the SMS prediction dataset and other real RNN/LSTM/GRU models.

### DIFF
--- a/syft/frameworks/torch/hook/hook.py
+++ b/syft/frameworks/torch/hook/hook.py
@@ -1140,6 +1140,9 @@ class TorchHook:
         # Modification of torch/nn/utils/clip_grad.py. The plain PyTorch method was not compatible with
         # PySyft remote tensors, so this method adds support for gradient clipping of remote tensors,
         # and keeps functionalities from PyTorch to clip local PyTorch tensors.
+      # Modification of torch/nn/utils/clip_grad.py. The plain PyTorch method was not compatible with
+        # PySyft remote tensors, so this method adds support for gradient clipping of remote tensors,
+        # and keeps functionalities from PyTorch to clip local PyTorch tensors.
         def clip_grad_norm_remote_(parameters, max_norm, norm_type=2):
             """Clips gradient norm of an iterable of parameters stored over a remote model
         
@@ -1160,10 +1163,6 @@ class TorchHook:
             """
             if isinstance(parameters, torch.Tensor):
                 # Remote PySyft tensor
-                if hasattr(parameters, "child") and isinstance(
-                    parameters.child, syft.frameworks.torch.pointers.pointer_tensor.PointerTensor
-                ):
-                    worker = parameters.location
                 parameters = [parameters]
             parameters = list(filter(lambda p: p.grad is not None, parameters))
             max_norm = float(max_norm)
@@ -1172,18 +1171,23 @@ class TorchHook:
                 total_norm = max(p.grad.data.abs().max() for p in parameters)
             else:
                 # Remote PySyft tensor
-                if hasattr(parameters, "child") and isinstance(
-                    parameters.child, syft.frameworks.torch.pointers.pointer_tensor.PointerTensor
+                if hasattr(parameters[0], "child") and isinstance(
+                    parameters[0].child, syft.frameworks.torch.pointers.pointer_tensor.PointerTensor
                 ):
                     total_norm = torch.zeros(1)
                     # Let's send the total norm over to the remote worker where the remote tensor is
-                    total_norm = total_norm.send(worker)
-                # Local PyTorch tensor
+                    total_norm = total_norm.send(parameters[0].location)
                 else:
                     total_norm = 0
                 for p in parameters:
                     param_norm = p.grad.data.norm(norm_type)
-                    total_norm += param_norm.item() ** norm_type
+                    #Remote PySyft tensor
+                    if hasattr(p, "child") and isinstance(p.child, syft.frameworks.torch.pointers.pointer_tensor.PointerTensor):
+                        total_norm += param_norm ** norm_type
+                    #Local PySyft tensor
+                    else:
+                        total_norm += param_norm.item() ** norm_type
+                        
                 total_norm = total_norm ** (1.0 / norm_type)
             clip_coef = max_norm / (total_norm + 1e-6)
             if clip_coef < 1:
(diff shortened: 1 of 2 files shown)


test/torch/test_hook.py
```diff
@@ -293,23 +293,64 @@ def initHidden(self):
         param.data.add_(-learning_rate, param.grad.data)
 
 
-def test_remote_gradient_clipping(workers):
+# Real test case of gradient clipping
```
mari-linhares: Move this comment to docstring

DanyEle: Done!

syft/frameworks/torch/hook/hook.py
```diff
@@ -1172,23 +1168,30 @@ def clip_grad_norm_remote_(parameters, max_norm, norm_type=2):
                 total_norm = max(p.grad.data.abs().max() for p in parameters)
             else:
                 # Remote PySyft tensor
-                if hasattr(parameters, "child") and isinstance(
-                    parameters.child, syft.frameworks.torch.pointers.pointer_tensor.PointerTensor
+                if hasattr(parameters[0], "child") and isinstance(
```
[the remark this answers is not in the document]
DanyEle: Created a method to create whether a parameter is remote and refactored the whole clip_grad_norm_remote_ method so as to check whether a list of remote parameters was being passed to it.

DanyEle:
@mari-linhares The `clip_grad_norm_remote_` should be much cleaner now, with the introduced checks for remote parameters.

syft/frameworks/torch/hook/hook.py
```diff
@@ -1169,37 +1169,46 @@ def clip_grad_norm_remote_(parameters, max_norm, norm_type=2):
...
-                    parameters.child, syft.generic.pointers.pointer_tensor.PointerTensor
-                ):
+                # all parameters are remote
+                if all([param_is_pointer_tensor(param) for param in parameters]):
                     total_norm = torch.zeros(1)
-                    # Let's send the total norm over to the remote worker where the remote tensor is
-                    total_norm = total_norm.send(worker)
-                # Local PyTorch tensor
+                    # Let's send the total norm over to the remote where the remote tensor is
+                    total_norm = total_norm.send(parameters[0].location)
```
robert-wagner: why is this only sent to one location

test/torch/test_hook.py
```diff
@@ -293,23 +293,68 @@ def initHidden(self):
...
-    remote_vanishing_tensor = vanishing_tensor_test.send(alice)
-    vanishing_remote_tensor_clipped = torch.nn.utils.clip_grad(remote_vanishing_tensor, 2)
-    # Has the remote gradient indeed increased?
-    greater_tensor_check = (vanishing_remote_tensor_clipped > remote_vanishing_tensor).copy().get()
-    one_tensors = torch.ones([1], dtype=torch.uint8)
-    assert torch.eq(greater_tensor_check, one_tensors)
-
-
-def test_local_gradient_clipping():
-    # Vanishing gradient test
```
robert-wagner: is there a reason this was removed? it isn't inherently a bad test

test/torch/test_hook.py
```diff
@@ -293,23 +293,68 @@ def initHidden(self):
...
+    local parameters of an RNN
+    """
     alice = workers["alice"]
-    vanishing_tensor_test = torch.Tensor([-9.8367e23])
-    remote_vanishing_tensor = vanishing_tensor_test.send(alice)
-    vanishing_remote_tensor_clipped = torch.nn.utils.clip_grad(remote_vanishing_tensor, 2)
-    # Has the remote gradient indeed increased?
-    greater_tensor_check = (vanishing_remote_tensor_clipped > remote_vanishing_tensor).copy().get()
-    one_tensors = torch.ones([1], dtype=torch.uint8)
-    assert torch.eq(greater_tensor_check, one_tensors)
```
robert-wagner: you could change this to greater_tensor_check.all()

syft/frameworks/torch/hook/hook.py
DanyEle: Because the location of all parameters in one model is the same. So, I could have picked the location any random parameter within the model.

test/torch/test_hook.py
DanyEle: Because in a real scenario, this method is supposed to be applied to a generator of parameters of a model and not to one simple tensor. The new tests checks pretty much the same things as the previous anyway.

test/torch/test_hook.py
DanyEle: Yeah, that's a good point.

DanyEle: Next time I create one such test, I'll use your syntax, thanks! It's much more compact than time. However, this test case was removed.

test/torch/test_hook.py
robert-wagner: There is a difference between unit and integration tests. The deleted test seems a bit closer to a unit test but it seems you replaced it with an integration test

DanyEle:
@robert-wagner Do you think the two previous "unit" tests should be kept?

robert-wagner:
@DanyEle I do

DanyEle:
> @DanyEle I do

I've put back the test cases related to gradient clipping for local and remote tensors, slightly modifying them to introduce checks over the `tensor.grad` property of tensors instead of the tensors themselves.

test/torch/test_hook.py
```diff
@@ -293,23 +293,125 @@ def initHidden(self):
...
+    # Method operates in place
+    nn.utils.clip_grad_norm_(local_tensor, 2)
+    tensor_comparison_grad = torch.Tensor([[4.5, 4.5], [4.5, 4.5]])
+    # Has the gradient decreased w.r.t. the initial value (i.e. was it clipped)?
+    smaller_tensor_check = local_tensor.grad < tensor_comparison_grad
+
     one_tensors = torch.ones([1], dtype=torch.uint8)
-    assert torch.eq(greater_tensor_check, one_tensors)
+
+    assert torch.eq(one_tensors, smaller_tensor_check.all())
```
robert-wagner: You don't need the torch.eq or ones tensor. .all returns a boolean (not a tensor)

DanyEle: Well, it looks to me like it actually returns a tensor containing a scalar being 0 for False and 1 for True.

image

So I just changed the check into an equality among integers.

DanyEle:
I guess it should be good now @robert-wagner ?

Issue #2531, opened 2019-08-19
Better integration of Message class throughout codebase


jvmancuso:
The `ObjectPointer` class has not been updated to use the Message class from #2450. This could be the case elsewhere, so we should do a once-over for files that still reference `from syft.codes import MSGTYPE` and update them.

PR #2534, opened 2019-08-19
Cleanup MSGTYPE 

lc0:
Minor clean up of MSGTYPE accordingly with #2531

I did not touch `plan.py` and there is also `FL Experiment.ipynb`

### DIFF
--- a/syft/generic/pointers/object_pointer.py
+++ b/syft/generic/pointers/object_pointer.py
@@ -5,7 +5,6 @@ from typing import TYPE_CHECKING
 import syft
 from syft import exceptions
 from syft import messaging
-from syft.codes import MSGTYPE
 from syft.generic.object import AbstractObject
 
 # this if statement avoids circular imports between base.py and pointer.py
--- a/syft/workers/websocket_client.py
+++ b/syft/workers/websocket_client.py
@@ -11,7 +11,6 @@ import time
 
 import syft as sy
 from syft import messaging
-from syft.codes import MSGTYPE
 from syft.generic.tensor import AbstractTensor
 from syft.workers import BaseWorker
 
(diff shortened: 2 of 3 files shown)


lc0:
cc @jvmancuso @robert-wagner

lc0:
plan works with message internals kind of https://github.com/OpenMined/PySyft/blob/dev/syft/messaging/plan.py#L179

So I am not sure if we intended to replace it the same way as files above

lc0:
I would push notebook change in a separate PR, in order not too make it mess here :)

jvmancuso:
I think it's okay if you push notebook changes with it, please do! We use NBReview to be able to quickly and easy review notebooks for that exact reason.

I think Plan should be a subclass of Message (feel like that would be in the spirit of the work, cc: @iamtrask), but I don't think we need to take care of that in this PR.

mari-linhares:
> I think Plan should be a subclass of Message (feel like that would be in the spirit of the work, cc: @iamtrask), but I don't think we need to take care of that in this PR.

TTBOMK, yes!

lc0:
btw, even with current mentioned notebook is failing for me, so I would rather merge this one first and after figure out other parts :)

image


And if I print y[1:]
```
b'\x04"M\x18h@T\x00\x00\x00\x00\x00\x00\x00\xa6>\x00\x00\x00\xf0\x0f\x92\x06\x92\x01\x92\x06\x94\x92\x05\x91\xa7__add__\x92\x14\x96\xcf\x00\x00\x00\x06\xa2\xc5\\\x9c\t\x00\xff\x02\x14\x0e\x97\xcd-\xa2me\xc0\x92\r\x91\x04\xc3\x92\x06\x91!\x00\tP\x04\xc3\x92\x00\x90\x00\x00\x00\x00'
```

lc0:
so what do you think? shall we merge this part and follow the plan and notebook as the next step?

PR #2540, opened 2019-08-21
Added test with send in plan

### LINKED ISSUE #2539: Test send operation inside a plan
Make tests for using send inside a plan, the test will be something like:

```python
@sy.func2plan([th.Size((2,3))])
def plan_double_abs(x):
    x = x.send(bob)
    x = x + x
    x = th.abs(x)
    return x

ptr_result = plan_double_abs(th.zeros(3))
assert type(ptr_result, sy.PointerTensor)
result = ptr_result.get()
assert torch.equal(result, th.tensor([2., 2., 2.]))
```

mccorby:
Test for plans using `send`.

Closes #2539

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@ docs/_modules
 .vscode/
 **/pip-wheel-metadata
 
+# IDE specific
+out/
+*.iml
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,3 @@ docs/_modules
 .vscode/
 **/pip-wheel-metadata
 
-# IDE specific
-out/
-*.iml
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -20,3 +20,6 @@ docs/_modules
 .vscode/
 **/pip-wheel-metadata
 
+# IDE specific
```
mari-linhares: This is not related to the PR. Can you remove these changes?

test/message/test_plan.py
```diff
@@ -8,14 +8,10 @@
 from syft.serde.serde import deserialize
 from syft.serde.serde import serialize
 from syft import messaging
-import time
```
mari-linhares: Thanks for the cleanup!

.gitignore
mccorby: Done

test/message/test_plan.py
iamtrask: +1

PR #2541, opened 2019-08-21
Generalize exceptions and Plan

jvmancuso:
Work done to continue toward supporting modular frameworks.
- `PureTorchTensorFound` is now `PureFrameworkTensorFound`
- Plans move toward being framework agnostic


syft/messaging/message.py
```diff
@@ -8,9 +8,8 @@
 """
 
 import syft as sy
-
-from syft.workers import AbstractWorker
 from syft import codes
+from syft.workers import AbstractWorker
```
iamtrask: So glad you changed the order of these two imports to be non-alphabetical. Hardcore bothered me last night.

iamtrask:
I don't really see where this did much to Plan, but it looks good to me.

syft/messaging/message.py
jvmancuso: 🙂

jvmancuso:
>I don't really see where this did much to Plan, but it looks good to me.

Generalized references to `torch.Tensor` and changed how you access the hook to make sure it's being pulled from the right place! V light changes, but needed to be in line with the rest of the work.

jvmancuso:
>Travis is complaining - looks like a flake 8 error

Also I fixed this last night but didn't push because while testing I discovered that the recently merged `test_rnn_mpc.py` has slowed testing by nearly an order of magnitude. Working on a fix for that now with @robert-wagner

PR #2545, opened 2019-08-22
Fixing RNN tests

jvmancuso:
After a recent merge, the tests for RNNs take a _very_ long time to complete, slowing down CI and local dev testing by nearly an order of magnitude. This PR patches it by moving the tests to a separate integration folder that are only run by CI on merge. I've also reduced the hyperparameters all to 2 to try to improve speed as much as possible.

As a side note, the length of time it takes to complete each test leads me to believe that RNNs are generally not suitable to be computed fully with MPC, and I'd suggest users search for alternatives as often as possible :)


.travis.yml
```diff
@@ -33,7 +33,7 @@ before_script:
   # https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds
   - flake8 . --count --exit-zero --statistics --select=E,F,W,C90
 script:
-  - coverage run -m pytest
+  - coverage run -m pytest test
```
robert-wagner: Could you add pytest integration after the coverage checking? We want to make sure that integration tests pass before merging (but as a separate stage)

Makefile
```diff
@@ -26,8 +26,8 @@ lab: venv
 test: venv
 	(. venv/bin/activate; \
 		python setup.py install; \
-		venv/bin/coverage run setup.py test;\
-		venv/bin/coverage report -m --fail-under 95;\
+		venv/bin/coverage run -m pytest test --ignore=test/integration/; \
```
robert-wagner: I was thinking having integration as a top level folder

jvmancuso: the tests rely on shared fixtures with other tests. I was thinking of pulling those fixtures from a common common module elsewhere but this seemed better

.travis.yml
jvmancuso: this includes the integration tests since the `integration/` folder is a subdirectory

Makefile
robert-wagner: That a good point. That being said could you also add a target which only runs the integration tests?

.travis.yml
robert-wagner: I was saying to run them seperately. The idea being that if unit tests fail, we don't want to bother running the integration tests since they're slow

Makefile
jvmancuso: will do

.travis.yml
jvmancuso: gotcha, will do!

andrelmfarias:
This long time-execution with RNNs is due one of the last commits that integrated multiplication and division in only one block, which slowed down cumulated multiplications in MPC. RNNs' activation functions (`tanh` and `sigmoid`) are approximated by a polynomial for MPC, so one computation of `tanh` have several multiplications.
For instance, the time-execution of `tanh` for a 3x3 tensor increased around 50x after that commit. It will be corrected in the near future so that the RNNs will be suitable for MPC.

PR #2549, opened 2019-08-23
Fix 2503 related to calling backward in multiple contexts

### LINKED ISSUE #2503: URGENT: bug in encrypted autograd
**Describe the bug**
For some reason, calling loss.backward() on a pointer using normal autograd then breaks our ability to call loss.backward() on an encrypted loss variable even on two totally separate examples.

**To Reproduce**
```
import torch
import torch as th
import torch.nn as nn
import torch.optim as optim
import syft as sy

# Set everything up
hook = sy.TorchHook(torch) 

big_hospital = sy.VirtualWorker(hook, id="big_hospital2")
small_hospital = sy.VirtualWorker(hook, id="small_hospital2")
crypto_provider = sy.VirtualWorker(hook, id="crypto_provider2")

# A Toy Model
class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.fc1 = nn.Linear(2, 2)
        self.fc2 = nn.Linear(2, 1)

    def forward(self, x):
        x = self.fc1(x)
        x = F.relu(x)
        x = self.fc2(x)
        return x
    
def federated():
    # A Toy Dataset
    data = th.tensor([[0,0],[0,1],[1,0],[1,1.]])
    target = th.tensor([[0],[0],[1],[1.]])

    model = Net()

    # Training Logic
    opt = optim.SGD(params=model.parameters(),lr=0.1)

    data = data.send(big_hospital)
    target = target.send(big_hospital)

    # NEW) send model to correct worker
    model.send(data.location)

    opt.zero_grad()

    # 2) make a prediction
    pred = model(data)

    loss.backward()
    
    print("Done!")
    
def encrypted():
    # A Toy Dataset
    data2 = th.tensor([[0,0],[0,1],[1,0],[1,1.]])
    target2 = th.tensor([[0],[0],[1],[1.]])

    model2 = Net()

    # We encode everything
    data2 = data2.fix_precision().share(big_hospital, small_hospital, crypto_provider=crypto_provider, requires_grad=True)
    target2 = target2.fix_precision().share(big_hospital, small_hospital, crypto_provider=crypto_provider, requires_grad=True)
    model2 = model2.fix_precision().share(big_hospital, small_hospital, crypto_provider=crypto_provider, requires_grad=True)

    opt2 = optim.SGD(params=model2.parameters(),lr=0.1).fix_precision()


    opt2.zero_grad()

    # 2) make a prediction
    pred2 = model2(data2)

    loss2 = ((pred2 - target2)**2).sum()

    loss2.backward()

#     # 5) change those weights
#     opt2.step()

#     # 6) print our progress
#     print(loss2.get().float_precision())
        
    print("Done")
    
run_broken = True

# make sure to re-start your jupyter notebook / environment with each test.
if(run_broken):
    # Breaks
    federated()
    encrypted() # breaks here - something about loss2.backward() causes the federated() demo to break 
else:
    # Works fine
    encrypted()
    federated()
```

Throws error:

```
AttributeError                            Traceback (most recent call last)
<ipython-input-1-db6dbbeffaa2> in <module>()
    100     # Breaks
    101     federated()
--> 102     encrypted() # breaks here - something about loss2.backward() causes the federated() demo to break
    103 else:
    104     # Works fine

<ipython-input-1-db6dbbeffaa2> in encrypted()
     84 
     85     # 4) figure out which weights caused us to miss
---> 86     loss2.backward()
     87 
     88 #     # 5) change those weights

/Users/atrask/anaconda/lib/python3.6/site-packages/syft-0.1.22a1-py3.6.egg/syft/frameworks/torch/hook/hook.py in overloaded_native_method(self, *args, **kwargs)
    683                 # Put back the wrappers where needed
    684                 response = syft.frameworks.torch.hook_args.hook_response(
--> 685                     method_name, response, wrap_type=type(self), new_self=self
    686                 )
    687 

/Users/atrask/anaconda/lib/python3.6/site-packages/syft-0.1.22a1-py3.6.egg/syft/frameworks/torch/hook/hook_args.py in hook_response(attr, response, wrap_type, wrap_args, new_self)
    243         response_hook_function = hook_method_response_functions[attr_id]
    244         # Try running it
--> 245         new_response = response_hook_function(response)
    246 
    247     except (IndexError, KeyError, AssertionError):  # Update the function in cas of an error

/Users/atrask/anaconda/lib/python3.6/site-packages/syft-0.1.22a1-py3.6.egg/syft/frameworks/torch/hook/hook_args.py in <lambda>(x)
    502         f = many_fold
    503 
--> 504     return lambda x: f(lambdas, x)
    505 
    506 

/Users/atrask/anaconda/lib/python3.6/site-packages/syft-0.1.22a1-py3.6.egg/syft/frameworks/torch/hook/hook_args.py in two_fold(lambdas, args, **kwargs)
    520 
    521 def two_fold(lambdas, args, **kwargs):
--> 522     return lambdas[0](args[0], **kwargs), lambdas[1](args[1], **kwargs)
    523 
    524 

/Users/atrask/anaconda/lib/python3.6/site-packages/syft-0.1.22a1-py3.6.egg/syft/frameworks/torch/hook/hook_args.py in <lambda>(i)
    480         if isinstance(r, (list, tuple))  # if the rule is a list or tuple.
    481         # Last if not, rule is probably == 1 so use type to return the right transformation.
--> 482         else lambda i: backward_func[wrap_type](i, **wrap_args)
    483         for a, r in zip(response, rules)  # And do this for all the responses / rules provided
    484     ]

/Users/atrask/anaconda/lib/python3.6/site-packages/syft-0.1.22a1-py3.6.egg/syft/frameworks/torch/hook/hook_args.py in <lambda>(i)
     73 backward_func = {
     74     TorchTensor: lambda i: i.wrap(),
---> 75     torch.Tensor: lambda i: i.wrap(),
     76     torch.nn.Parameter: lambda i: torch.nn.Parameter(data=i),
     77     PointerTensor: lambda i: i,

AttributeError: 'NoneType' object has no attribute 'wrap'
```

**Additional context**
latest version of PySyft from PyPI ('0.1.22a1')

LaRiffle:
Fix #2503


test/torch/hook/test_hook_args.py
```diff
@@ -1,18 +1,111 @@
...
     assert result == ([1, 1], 0)
 
 
 def test_build_rule_numpy():
     arr = np.array([2.0, 3.0, 4.0])
     result = hook_args.build_rule([arr, arr + 2, [2, 4, "string"]])
     assert result == [1, 1, [0, 0, 0]]
+
+
+def test_backward_multiple_use(workers):
```
robert-wagner: why are there no assertions in this test? how do we know backwards actually worked?

LaRiffle: I'm only interested in having backward not failing due to hook_args, so I didn't checked that it works correctly because it's checked elsewhere
But I agree this test leaves both a feeling of deja vu and incomplteness

LaRiffle: Ok @robert-wagner this is addressed!

syft/frameworks/torch/hook/hook_args.py
```diff
@@ -55,6 +55,7 @@
     "chunk",
     "torch.functional.split",
     "split",
+    "backward",
```
jvmancuso: why is this an ambiguous function but not an ambiguous method?  I also feel we don't need it to be ambiguous since this will commonly be done in a loop with the same or very similar args (however I'm not entirely sure how similar signatures need to be for this to save/waste time).

LaRiffle: > why is this an ambiguous function but not an ambiguous method? 
Yes excellent point!
Thats because of the response type, not the input type! For response wrapping we use "ambiguous function" (see hook_response and register_response)

>  this will commonly be done in a loop with the same or very similar args
Indeed, but in a same session I might want to do FL AND encrypted training, which fails, as pointed in #2503

jvmancuso: >Indeed, but in a same session I might want to do FL AND encrypted training, which fails, as pointed in #2503

Hm, I'm not sure why that wouldn't work -- I was under the impression that ambiguous functions/methods are only different because of their signature type, and moreover that they were entirely optional.  If the signature changes, I feel like that should be noticed lazily, i.e. if I call an ambiguous function 1000 times in a row with the same signature and then call it once more after with a new signature, we shouldn't have to rebuild the signature the first 1000 times.  Maybe I'm missing something though, I'm still new to this part of the code!

LaRiffle: >  Further, I thought backward always returns None -- I must be missing something?

Theoretically that's right. However with distant commands we do pre-allocation of responses: meaning that we create a pointer to the response before the command is sent and executed. If the response contained more than 1 tensor, we have a fallback mechanism and we interactively create several pointers. But if the response is empty, no such mechanism exist to limit communication: you end up with a stale pointer, and usually people don't try to play with it (like no one does ptr = x.backward() ). This stale pointer goes in register_response.

> I thought these ambiguous functions/methods were entirely optional.

They are not. hook_args has a failure mode when the signature contains more tensors than expected, **but not when are are less than expected**. This is for efficiency purpose, as making length comparison all over the calls would certainly kill the benefit from having hook_args. Therefore, functions with changing signatures have to be registered as *ambiguous*

jvmancuso: Got it, that makes sense.  Thank you for clarifying!

mari-linhares: That was very helpful thanks @LaRiffle

PR #2552, opened 2019-08-24
Module for linear algebra operations in MPC

andrelmfarias:
Created a module for standard linear algebra operations in MPC. This module will be key for the implementation of [Jonathan Bloom's paper](https://arxiv.org/pdf/1901.09531.pdf) in the context of Project 3 of [RAAIS Grant](https://blog.openmined.org/raais/).

In this first PR, I implemented the inverse function for symmetric matrices, which is useful for Section 2 of Bloom's paper as we only need to compute the inverse of the Gram matrix, which is symmetric.

I used LDLt decomposition and the algorithm proposed in section 4 of [https://arxiv.org/abs/1111.4144](https://arxiv.org/abs/1111.4144) in order to optimize the number of computations using division in MPC. LDLt decomposition is only possible with symmetric matrices and it is better suited for our case than Cholensky decomposition as it does not need to compute square roots.


syft/frameworks/torch/linalg/operations.py
```diff
@@ -0,0 +1,59 @@
...
+            if i == j:
+                t_inv[i, j] = inv_d[i] - (l_t[i, i + 1 : n] * t_inv[i + 1 : n, j]).sum()
+            else:
+                t_inv[j, i] = t_inv[i, j] = -(l_t[i, i + 1 : n] * t_inv[i + 1 : n, j]).sum()
+
+    return t_inv
+
+
+def _ldl(t):
+    """
```
LaRiffle: Use google style for docstrings

test/torch/linalg/test_linalg.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    Testing inverse of symmetric matrix with MPC
+    """
+    bob = workers["bob"]
+    alice = workers["alice"]
+    crypto_prov = sy.VirtualWorker(hook, id="crypto_prov")
+
+    x = th.Tensor([[0.4627, 0.8224], [0.8224, 2.4084]])
+
+    x = x.fix_precision(precision_fractional=6).share(bob, alice, crypto_provider=crypto_prov)
```
LaRiffle: Just a curiosity: does is works with precision_fractional=3 ?

andrelmfarias: Yes, but it's less precise (sometimes 0.1 difference from true result)

andrelmfarias: This could be a good use-case for long precision tensors...

syft/frameworks/torch/linalg/operations.py
andrelmfarias: Done

Issue #2554, opened 2019-08-26
.federate() does not work with torchvision's ImageFolder dataset


edgarinvillegas:
**Describe the bug**
When you try to federate a `torchvision.datasets.ImageFolder`, you get the custom exception AttributeError("Could not find inputs in dataset")

**To Reproduce**

1) Have the following file structure locally:
```
+-- dataset
|   +-- class1
|   |   +-- whatever1.jpg
|   |   +-- whatever2.jpg
```
Those jpgs can be any image of at least 256x256 size.

2) Run: 
```
import torch
from torchvision import datasets, transforms
import syft as sy  

hook = sy.TorchHook(torch)  
bob = sy.VirtualWorker(hook, id="bob")  
alice = sy.VirtualWorker(hook, id="alice")

imagefolder_dataset = datasets.ImageFolder(root='./dataset/', transform = transforms.Compose([
        transforms.Resize(size=256),
        transforms.CenterCrop(size=224),
        transforms.ToTensor()
    ]))


fed_dataset = imagefolder_dataset.federate((bob, alice)) # Raises exception
```

**Expected behavior**
`fed_dataset` should be created without raising any exception

**Actual behavior**
An exception "Could not find inputs in dataset" is raised:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-21c009d642db> in <module>
      7 
----> 8 fed_dataset = imagefolder_dataset.federate((bob, alice))
      9 

c:\...\syft\frameworks\torch\federated\dataset.py in dataset_federate(dataset, workers)
    148             dataset.data = dataset.test_data
    149         else:
--> 150             raise AttributeError("Could not find inputs in dataset")
    151     if not hasattr(dataset, "targets"):
    152         if hasattr(dataset, "train_labels"):

AttributeError: Could not find inputs in dataset
```

**Desktop (please complete the following information):**
 - OS: Windows 10

**Additional context**
 - torchvision: 0.3.0
 - pytorch: 1.1.0
 - pysyft: 0.1.22a1

**Workaround**
Before calling `.federate()` you can create the following dummy fields to prevent the exception:
`imagefolder_dataset.data = imagefolder_dataset.targets = True`

edgarinvillegas:
Created PR to fix this: https://github.com/OpenMined/PySyft/pull/2555

[...]

PR #2555, opened 2019-08-26
fix: Fix bug that prevents ImageFolder dataset to be federated #2554

edgarinvillegas:
This also fixes the bug for any dataset that doesn't have data/target
or equivalent fields.

The defensive code removed was checking for the data/train_data/test_data/targets/train_labels/test_labels fields in the dataset in a mandatory way. Which is not needed.
What matters is that on each iteration of the dataset, it has to return a tuple in the form (data, target), which ImageFolder complies.


syft/frameworks/torch/federated/dataset.py
```diff
@@ -140,22 +140,6 @@ def dataset_federate(dataset, workers):
...
-            dataset.data = dataset.test_data
-        else:
-            raise AttributeError("Could not find inputs in dataset")
-    if not hasattr(dataset, "targets"):
-        if hasattr(dataset, "train_labels"):
-            dataset.targets = dataset.train_labels
-        elif hasattr(dataset, "test_labels"):
-            dataset.targets = dataset.test_labels
-        else:
-            raise AttributeError("Could not find targets in dataset")
```
mari-linhares: Instead of getting rid completely we could show a warning instead of raising attribute error.

PR #2558, opened 2019-08-26
Fixed error with torch.mm / torch.Tensor.mm and implemented test

andrelmfarias:


syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -426,6 +426,7 @@ def matmul(self, *args, **kwargs):
         return response
 
     __matmul__ = matmul
+    __mm__ = mm = matmul
```
LaRiffle: Not sure `__mm__` exists...
`__matmul__` stands for `@`

Issue #2385, opened 2019-07-23
StatefulPlans


iamtrask:
**Is your feature request related to a problem? Please describe.**
Our current "Plan" data-structures are limited in several ways:

Plans do not have state (you cannot ship a plan WITH tensors which the plan uses / updates)

**Describe the solution you'd like**

I'd like to be able to package a plan with tensors which the plan can update, such that when I call .send(), the plan and its tensors go with it, and when I call .get(), the plan AND it's (potentially updated) tensors come back.

**Describe alternatives you've considered**
None

**Additional context**
This is critical functionality to the integration of PySyft wtih syft.js

iamtrask:
CC: @midokura-silvia, @LaRiffle, @robert-wagner

mari-linhares:
Hey @iamtrask,

A few questions...

1. Can you add one (or more) use cases where the "state" is critical?

2. Do you have thoughts about how should be the interface for StatefulPlans?

3. All plans are going to be stateful from this point on, or we'll have two kinds of plans? Regular Plans and StatefulPlans?

iamtrask:
1. federated learning - you send a plan with a model
2. two options. one is where you pass a list of objects in func2plan() and the other is where you instantiate a class with class objects where you want to package teh whole class (all its methods) and all its objects into a plan.
3. haven't decided yet - seems like they could all be stateful but where sometimes the state is empty :)

Jasopaum:
Do you think we could have a class (similar to the Plan class we already have) from which user's classes would inherit? For instance, we could have:
```
class Net(nn.Module, StatefulPlan):
    def __init__(self):
        super(Net, self).__init__()
        self.fc1 = nn.Linear(2, 3)
        self.fc2 = nn.Linear(3, 2)
        self.foo = torch.tensor(...)  # Tensor that will be modified

    def forward(self, x):
        x = F.relu(self.fc1(x))
        x = self.fc2(x)
        self.foo = something
        return F.log_softmax(x, dim=0)
``` 
I don't know if it's what you were thinking about but it seems to me like it's quite easy to use and not harder than something else to implement.

iamtrask:
Love this! I greatly hope StatefulPlans supports this style of API (and
possibly others as well).

amit-rastogi:
I tried the multiple inheritance approach but bumped into method resolution order (MRO) gotchas with init() method invocation.

PR #2569, opened 2019-08-28
Introduce Stateful Plans

LaRiffle:
Response implementation to #2385

I made several choices in terms of API which can be summarized by these examples, they are of course very open to discussion if smthg more elegant can be found.

**Function 2 Plan**
```python
    @sy.func2plan(args_shape=[(1,)], state={"bias": th.tensor([1.0])})
    def foo(x, state):
        bias = state.read("bias")
        x = x * 2
        return x + bias
```
**Class 2 Plan**
```python
    class Net(sy.Plan):
        def __init__(self):
            super(Net, self).__init__()
            self.fc1 = nn.Linear(2, 1)
            self.bias = th.tensor([1000.0])

            self.add_to_state("fc1", "bias")

        def forward(self, x):
            x = self.fc1(x)
            return F.log_softmax(x, dim=0) + self.bias
```

**Known weaknesses**
- sy.Plan inherits from torch.nn.Module
- State elements are sent separately from the Plan (multiple msgs only when sending, not at execution of course)
- The code remains really hard to read


syft/messaging/plan.py
```diff
@@ -33,62 +33,134 @@ class func2plan(object):
...
+def method2plan(*args, **kwargs):
+    raise SyntaxError(
+        "method2plan is not supported anymore. Consider instead subclassing your object from sy.Plan."
     )
 
-    @property
-    def method(self: object) -> Plan:
+
+class State(object):
+    """The State is a Plan attribute and is used to send tensors along functions
```
mari-linhares: Missing period.

syft/messaging/plan.py
```diff
@@ -33,62 +33,134 @@ class func2plan(object):
...
+
+    It references Plan tensor or parameters attributes using their name, and make
+    sure they are provided to remote workers who are sent the Plan.
+    """
+
+    def __init__(self, plan):
+        self.keys = []
```
mari-linhares: Could self.keys be a set?

syft/messaging/plan.py
```diff
@@ -33,62 +33,134 @@ class func2plan(object):
...
+        self.plan = plan
+
+    def __repr__(self):
+        return "State: " + ", ".join(self.keys)
+
+    def append(self, key):
```
mari-linhares: Can you add more details to this docstring? There's a lot going on here, not sure if I follow.

syft/messaging/plan.py
```diff
@@ -33,62 +33,134 @@ class func2plan(object):
...
+                raise ValueError(
+                    f"Obj of type {type(t)} is not supported in the state.\n"
+                    "Use instead tensors or parameters."
+                )
+        return self
+
+    def __iadd__(self, other_keys):
+        return self.append(other_keys)
+
+    def is_accepted_type(self, obj):
```
mari-linhares: Maybe rename this to `is_valid_type`?

syft/messaging/plan.py
```diff
@@ -33,62 +33,134 @@ class func2plan(object):
...
+
+    def read(self, key):
+        """
+        Returns one element referenced by the store
+        """
+        return getattr(self.plan, key)
+
+    def set_(self, dict_state):
+        """
+        Reset inplace the state by feeding him a dict of tensors or params
```
mari-linhares: him -> it

syft/messaging/plan.py
```diff
@@ -33,62 +33,134 @@ class func2plan(object):
...
+class State(object):
+    """The State is a Plan attribute and is used to send tensors along functions
+
+    It references Plan tensor or parameters attributes using their name, and make
+    sure they are provided to remote workers who are sent the Plan.
+    """
+
+    def __init__(self, plan):
+        self.keys = []
+        self.plan = plan
```
mari-linhares: The state is a plan attribute, but a state has a plan attribute, can you detail why this is the case?

syft/messaging/plan.py
```diff
@@ -99,30 +171,38 @@ class Plan(ObjectStorage):
...
+        self.name = self.__class__.__name__ if name == "" else name
+        self.owner = sy.local_worker if owner is None else owner
         self.verbose = verbose
 
         # Info about the plan stored
         self.plan = list()
         self.readable_plan = readable_plan if readable_plan is not None else []
+        self.state = State(self)
+        self.include_state = include_state
+        if state is not None:
```
mari-linhares: A state can store another state?

syft/messaging/plan.py
```diff
@@ -134,16 +214,20 @@ def __init__(
...
+
+    def add_to_state(self, *elements):
+        for elem in elements:
+            self.state.append(elem)
 
     def _auto_build(self, args_shape: List[Tuple[int]] = None):
         placeholders = self._create_placeholders(args_shape)
-        args = [self._self] + placeholders if self.is_method else placeholders
-        self._build(args)
+        args = placeholders
```
mari-linhares: I think you can get rid of this variable and just call placeholders args instead

syft/messaging/plan.py
```diff
@@ -190,59 +288,53 @@ def _recv_msg(self, bin_message: bin):
 
         return sy.serde.serialize(None)
 
-    def _prepare_for_running_local_method(self):
```
mari-linhares: I'm very happy to see this is gone :smiley:

syft/frameworks/torch/hook/hook.py
```diff
@@ -546,6 +546,7 @@ def create_grad_objects(model):
                 o = p.sum()
                 o.backward()
                 p.grad -= p.grad
+            pass
```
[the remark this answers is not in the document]
LaRiffle: nope I thought it was removed!

syft/frameworks/torch/hook/hook.py
```diff
@@ -646,24 +649,24 @@ def module_copy(nn_self):
...
+        # def location(nn_self):
+        #     try:
+        #         for p in nn_self.parameters():
+        #             return p.location
+        #     except AttributeError:
+        #         raise AttributeError(
+        #             "Module has no attribute location, did you already send it to some location?"
+        #         )
+        #
+        # self.torch.nn.Module.location = location
```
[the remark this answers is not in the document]
LaRiffle: Will put it back!

syft/messaging/plan.py
LaRiffle: totally

syft/messaging/plan.py
LaRiffle: The state should theoretically be in the plan class, but I moved it outside for clarity. But the state does not store the elements, juste references to elements, which are in thee plan. That's why the both reference each other.
Which is not optimal...

syft/messaging/plan.py
LaRiffle: Not at all, but it can be fed with a dict like in self.state.set_(state)

test/message/test_plan.py
```diff
@@ -127,15 +240,31 @@ def forward(self, x):
...
+            self.state += ["fc1", "bias"]
+
+        def forward(self, x):
+            x = self.fc1(x)
+            return F.log_softmax(x, dim=0) + self.bias
+
+    model = Net()
+
+    # Force build
+    assert model(th.tensor([1.0, 2])) == th.tensor([1000.0])
```
mari-linhares: there's a build method implemented, I think we should use it explicitly.

mari-linhares: Like you've done below with net.build(tensor)

test/message/test_plan.py
```diff
@@ -226,30 +381,47 @@ def plan_abs(data):
...
 
-    hook.local_worker.is_client_worker = True
+    x_ptr = th.tensor([-1, 9, 3]).send(alice)
+    p = plan_abs(x_ptr)
+    x_abs = p.get()
+    assert (x_abs == th.tensor([2, 10, 4])).all()
+
+
+# TODO: Clarify this test
+# def test_fetch_plan(hook):
```
mari-linhares: The idea of fetching a plan is to get a copy of an existing plan that I (the worker) can execute. We need this for providing secure serving for Grid (which I'm working right now), I'm happy to discuss this further on slack.

test/message/test_plan.py
```diff
@@ -473,62 +643,6 @@ def test__replace_message_ids():
     assert replaced == expected
 
 
-def test___call__function(hook):
```
mari-linhares: there's no way to reuse these tests?

syft/messaging/plan.py
```diff
@@ -190,59 +311,53 @@ def _recv_msg(self, bin_message: bin):
...
+            res_ptr = self.forward(*local_args)
         res_ptr.child.garbage_collect_data = False
 
-        worker = self.find_location(args)
+        # We put back the original state. There is no need to keep a reference
+        # To the element sent to the plan.
+        self.state.set_(copied_state)
 
+        # The readable plan is now built, we hide the fact that it was run on
+        # the plan and on the owner by replacing the workers ids
```
mari-linhares: the plan and on the owner by replacing the workers ids -> the plan and not on the owner by replacing the workers ids

test/message/test_plan.py
LaRiffle: I don't know really what to do with these tests, I can hardly them fit with the new notions of stateful plans :/

mari-linhares: I see, that's okay, I guess. :disappointed:

syft/messaging/plan.py
```diff
@@ -736,6 +736,9 @@ def __str__(self):
 
         return out
 
+    def __repr__(self):
```
mari-linhares: Nice

PR #2575, opened 2019-08-29
Fixed performance issue due to mul_and_div

andrelmfarias:
This fix solves the problem of performance of RNNs and activation functions such as `sigmoid` and `tanh` in MPC.

Time execution of a `sigmoid` on a (2, 1) tensor decreased from 15s to 0.18s


andrelmfarias:
This also fix RNN tests, what was also proposed on #2545.

RNN can now be tested with higher hyperparameters and moved back from `integration` repo

test/efficiency-tests/activations.py
```diff
@@ -0,0 +1,25 @@
...
+    alice = workers["alice"]
+    crypto_prov = sy.VirtualWorker(hook, id="crypto_prov")
+
+    if activation == "tanh":
+        func = th.tanh
+    else:
+        func = th.sigmoid
+
+    x = th.randn([10, 10]).fix_precision().share(bob, alice, crypto_provider=crypto_prov)
+    t0 = time.time()
```
mari-linhares: Would be nice to modularize time monitoring. I suggest to create a function `assert_time_is_less_than(func, time=1)`, that implements this logic. This should be stored in another file at the efficiency-tests folders named `efficiency_assertions.py` (or something similar).

kamathhrishi: Love this idea !

kamathhrishi: Great idea!

kamathhrishi: It would help benchmark different parts of our codebase and see how various optimization would help.

andrelmfarias: Yeah, good idea! What do you think about a decorator for that?

test/efficiency-tests/activations.py
```diff
@@ -1,25 +1,17 @@
...
 import pytest
-import time
 
 import syft as sy
 import torch as th
 
 
-@pytest.mark.parametrize("activation", ["tanh", "sigmoid"])
+@pytest.mark.parametrize("activation", [th.tanh, th.sigmoid])
+@assert_time(time=1)
```
mari-linhares: nice

test/efficiency-tests/activations.py
mari-linhares: Yes!! Decorators are great for this.

PR #2579, opened 2019-08-30
QR decomposition

andrelmfarias:
Implemented QR decomposition for local and pointer tensors. Decomposition made using [Householder reflections](https://en.wikipedia.org/wiki/QR_decomposition#Using_Householder_reflections)

This function is one of the main parts of the DASH algorithm, which corresponds to section 4 of [Jonathan Bloom's paper](https://arxiv.org/abs/1901.09531)


syft/frameworks/torch/linalg/operations.py
```diff
@@ -63,6 +65,192 @@ def _ldl(t):
...
-def qr(t):
-    # TODO
-    pass
+def qr(t, mode="reduced", norm_factor=None):
+    """
+    This function performs the QR decomposition of a matrix (2-dim tensor). The
+    decomposition is performed using Householder Reflection.
+
+    Args:
+        t: 2-dim tensor, shape(M, N). It should be whether a local tensor, a
```
LaRiffle: Maybe add a check that 2-dim is valid

syft/frameworks/torch/linalg/operations.py
```diff
@@ -63,6 +65,192 @@ def _ldl(t):
...
+            pointer to a remote tensor or an AdditiveSharedTensor
+
+        mode: {'reduced', 'complete', 'r'}. If K = min(M, N), then
+            - 'reduced' : returns q, r with dimensions (M, K), (K, N) (default)
+            - 'complete' : returns q, r with dimensions (M, M), (M, N)
+            - 'r' : returns r only with dimensions (K, N)
+
+        norm_factor: float. The normalization factor used to avoid overflow when
+            performing QR decomposition on an AdditiveSharedTensor. For example i
+            n the case of the DASH algorithm, this norm_factor should be of the
```
LaRiffle: typo: "in"

syft/frameworks/torch/linalg/operations.py
```diff
@@ -63,6 +65,192 @@ def _ldl(t):
...
+                "norm_factor argument.",
+            )
+        workers = t.child.child.locations
+        crypto_prov = t.child.child.crypto_provider
+        prec_frac = t.child.precision_fractional
+    elif isinstance(t, torch.Tensor):
+        t_type = "local"
+    else:
+        raise TypeError(
+            "The provided matrix should be a local torch.Tensor, a PointerTensor,",
```
LaRiffle: Add a place after 'PointerTensor,'

syft/frameworks/torch/linalg/operations.py
```diff
@@ -63,6 +65,192 @@ def _ldl(t):
...
+                down_zeros = down_zeros.send(t.child.location)
+                up_zeros = up_zeros.send(t.child.location)
+            if t_type == "ast":
+                down_zeros = down_zeros.fix_prec(precision_fractional=prec_frac).share(
+                    *workers, crypto_provider=crypto_prov
+                )
+                up_zeros = up_zeros.fix_prec(precision_fractional=prec_frac).share(
+                    *workers, crypto_provider=crypto_prov
+                )
+            left_cat = torch.cat((I[:i, :i], down_zeros), dim=0)
```
LaRiffle: Can you reuse `I_i` ?

syft/frameworks/torch/linalg/operations.py
```diff
@@ -63,6 +65,192 @@ def _ldl(t):
...
+
+        # If it is not the 1st iteration
+        # expand matrix H with Identity at diagonal and zero elsewhere
+        if i > 0:
+            down_zeros = torch.zeros([n_rows - i, i])
+            up_zeros = torch.zeros([i, n_rows - i])
+
+            # Send them to remote worker if t is pointer, secret share it if its an AST
+            if t_type == "pointer":
+                down_zeros = down_zeros.send(t.child.location)
```
LaRiffle: Can you reuse the variable `location`?

syft/frameworks/torch/linalg/operations.py
```diff
@@ -63,6 +65,192 @@ def _ldl(t):
...
+    """
+    Computation of a norm of a vector in MPC. The vector should be an AdditiveSharedTensor.
+
+    It performs the norm calculation by masking the tensor with a multiplication
+    by a big random number drawn from a uniform distribution, and computing the
+    square root of the squared norm of the masked tensor, which is computed
+    beforehand with a dot product in MPC.
+
+    In order to maintain stability and avoid overflow, this functions uses a
+    norm_factor that scales the tensor for MPC computations and rescale it at the end.
```
LaRiffle: "scales down" maybe ?

syft/frameworks/torch/linalg/operations.py
```diff
@@ -63,6 +65,192 @@ def _ldl(t):
...
+        r: lower triangular matrix as a 2-dim tensor with same type as t
+
+    """
+
+    workers = t.child.child.locations
+    crypto_prov = t.child.child.crypto_provider
+    prec_frac = t.child.precision_fractional
+    norm_factor = int(norm_factor)
+    t_normalized = t / norm_factor
+    Q = int(1e8)
```
LaRiffle: Can you make Q adaptative to the field and prec_frac ?

syft/frameworks/torch/linalg/operations.py
```diff
@@ -63,6 +65,192 @@ def _ldl(t):
...
+
+    norm_sq = (t_normalized * t_normalized).sum().squeeze()
```
LaRiffle: Does  `**2`  work here?

syft/frameworks/torch/linalg/operations.py
```diff
@@ -63,6 +65,192 @@ def _ldl(t):
...
+        .share(*workers, crypto_provider=crypto_prov)
+        .random_(0, Q)
+    )
+
+    # Compute masked norm
+    masked_norm_sq = r ** 2 * norm_sq
+
+    # Get compute square root
+    masked_norm_sq = masked_norm_sq.get().float_precision()
+    masked_norm = torch.sqrt(masked_norm_sq)
```
LaRiffle: Can you send this stuff to be computed by the crypto provider instead of disclosing this to the local worker?

test/efficiency_tests/test_linalg_time.py
```diff
@@ -17,3 +18,16 @@ def test_inv_sym(hook, workers):
...
+def test_norm_mpc(hook, workers):
+    torch.manual_seed(42)  # Truncation might not always work so we set the random seed
+    bob = workers["bob"]
+    alice = workers["alice"]
+    crypto_prov = sy.VirtualWorker(hook, id="crypto_prov")
+
+    n = 100
+    t = torch.randn([n])
+    t_sh = t.fix_precision(precision_fractional=6).share(bob, alice, crypto_provider=crypto_prov)
+    norm_sh = _norm_mpc(t_sh, norm_factor=n ** (1 / 2))
```
LaRiffle: Add an assertion to check the norm returned is not to bad (ie ||result - expected|| < e-2 par exemple))

LaRiffle: oh actually you did it, this test is a duplicate of the one in test/torch/linalg/test_operations.py
where the `@assert_time(max_time=20)` is missing however

andrelmfarias: Yes... actually one test is located in the `efficiency_tests` repo and the other in `linalg/test_operations.py`. I agree this is repeated code, but as we have created a repo only for efficiency tests I was kind of obligated to do like that... But to be honest I did not like it very much.

Do you think I should remove the repeated test from `efficiency_tests` and just include the decorator in `linalg/test_operations.py` ?

LaRiffle: Ah ok I wasn't aware of efficiency_tests 
Not sure why we created this instead of just putting the @assert_time(max_time=20)  where needed 
But let's keep it like this for the moment I suppose

andrelmfarias: It was an idea from @mari-linhares and me that was added to the tests 2 weeks ago.

Now, I think we will have a lot of repeated code (and longer tests) if we keep doing like that. In my opinion, it's better to just put the `assert_time` decorators on the tests in their respective repos when needed.

If you agree I will start doing it in this PR.

LaRiffle: Excellent let's do this!
I think this initial idea was a good one, but that we should put there only test of demos where we want to benchmark the efficiency, not specific tests. It would give us a way to monitor how our changes affect out flagship usecases

syft/frameworks/torch/linalg/operations.py
andrelmfarias: `I_i` is equal to `I[i:, i:]`, not to `I[:i, :i]`

syft/frameworks/torch/linalg/operations.py
andrelmfarias: I am not sure PySyft supports sending the shares to another remote worker, without calling `.get()` before. I tried to find a smart way to do it avoiding passing the shares to the local worker, but I didn't find a proper solution...

Do you have any ideas about how I can do that?

PR #2581, opened 2019-09-02
Share LPT

mccorby:
This PR adds `share` to Large Precision Tensors


test/torch/tensors/test_large_precision.py
```diff
@@ -233,8 +233,21 @@ def test_mod():
         (torch.tensor([2000.0123458910]), torch.tensor([2000.0123458910])),
     ],
 )
-def test_types(x, expected):
-    enlarged = x.fix_prec(internal_type=torch.int16, precision_fractional=128)
-    restored = enlarged.float_precision()
-    # And now x and restored must be the same
-    assert torch.all(torch.eq(expected, restored))
+class TestParametrized:
```
LaRiffle: Why is it needed to create a class?
Maybe it's a good practice, but we haven't done it so far, so I'm just curious :)

test/torch/tensors/test_large_precision.py
```diff
@@ -233,8 +233,21 @@ def test_mod():
...
+
+    def test_share_and_get(self, workers, x, expected):
+        alice, bob, james = (workers["alice"], workers["bob"], workers["james"])
+
+        enlarged = x.fix_prec(internal_type=torch.int16, precision_fractional=128)
+        expected = expected.fix_prec(internal_type=torch.int16, precision_fractional=128)
+
+        shared = enlarged.share(alice, bob, crypto_provider=james)
+
+        result = shared.get()
```
LaRiffle: Are operations supported on LPT shared data?
Is so, add tests for them to make sure everything goes fine :)

test/torch/tensors/test_large_precision.py
mccorby: It makes it easier to share the parameters for the tests enclosed in the class.
Happy to change it for something else

LaRiffle: All good, but I wonder if we shouldn't do that for more test actually

test/torch/tensors/test_large_precision.py
mccorby: This one is tricky. Working on it

test/torch/tensors/test_large_precision.py
mccorby: It might be easier to put the test data in a var and pass it to the parametrize annotation.
```
test_data = [
    (torch.tensor([1]), torch.tensor([1.0])),
    (torch.tensor([1.0]), torch.tensor([1.0]))
]

@pytest.mark.parametrize("x, expected", test_data)
def test_types(x, expected):
    ...
```

LaRiffle: Yes, this looks like a good idea!

test/torch/tensors/test_large_precision.py
mccorby: A first for `add` is now ready. Let's go through it, polish it and once we're happy I can apply the same to the other operations.

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -106,6 +108,13 @@ def get_class_attributes(self):
 
     @overloaded.method
     def add(self, self_, other):
+        # At this point we can have LPT>AST
+        # If that's the case, delegate into AST to do the operation.
+        # TODO I don't like the idea of this dependency here. An LPT should not know about AST
+        if hasattr(self_, "child") and isinstance(self_.child, AdditiveSharingTensor):
+            self_ = self_.child
+        if hasattr(other, "child") and isinstance(other.child, AdditiveSharingTensor):
+            other = other.child
```
LaRiffle: I don't think you need these line, if @ overloaded.method was done correctly then self_ and other are already the lpt child so they are AST

LaRiffle: Without them, this runs fine on my side
```python
lpt = torch.Tensor([-2., 4]).fix_prec(precision_fractional=100).share(alice, bob, crypto_provider=dan)
lpt + lpt
```

syft/frameworks/torch/hook/hook_args.py
```diff
@@ -77,16 +79,30 @@
...
     MultiPointerTensor: lambda i, **kwargs: MultiPointerTensor(**kwargs).on(i, wrap=False),
     CRTPrecisionTensor: lambda i, **kwargs: CRTPrecisionTensor(**kwargs).on(i, wrap=False),
     "my_syft_tensor_type": lambda i, **kwargs: "my_syft_tensor_type(**kwargs).on(i, wrap=False)",
 }
 
+
+def _lpt_forward_func(i):
+    if isinstance(i.child, AdditiveSharingTensor):
+        return i.child
+    return i._internal_representation_to_large_ints()
```
LaRiffle: to be more generic, you could put the condition on the case where `_internal_representation_to_large_ints` has to be called (myabe if the child is torch.Tensor and is not a wrapper). This way, if in the future you can new syftTensor which can be a lpt.child, it won't break

LaRiffle: In addition, you don't have to know what a AdditiveSharingTensor is:
```
if I.child is a tensor and not a wrapper:
   return i._internal_representation_to_large_ints()
else:
    return i.child
```
```

File: syft/frameworks/torch/hook/hook_args.py [hunk 5]
```diff
@@ -77,16 +79,30 @@
...
+
+
+def _lpt_backward_func(i, kwargs):
+    if isinstance(i, AdditiveSharingTensor):
+        return LargePrecisionTensor(**kwargs).on(i, wrap=False)
+
+    return LargePrecisionTensor(**kwargs).on(
+        LargePrecisionTensor.create_tensor_from_numpy(i, **kwargs), wrap=False
+    )
```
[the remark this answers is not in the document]
LaRiffle: Also, same comment: reverse the condition, and you won't have to know what an AdditiveSharingTensor is

syft/frameworks/torch/tensors/interpreters/large_precision.py
mccorby: It was previous to have the backward/forward funcs working :)
Removed

syft/frameworks/torch/tensors/interpreters/large_precision.py
```diff
@@ -158,6 +152,8 @@ def mul(self, self_, other):
             trunc_res = np.where(gate, neg_nums, pos_nums)
 
             return trunc_res
+        else:
+            return self_ + other
```
LaRiffle: This is a * not a + :)

mccorby: Eagle eye! :)
Also... quite concerned the tests are passing. I'm reviewing them

mccorby: Once `.float_precision()` is applied to the result it does not work. I'll keep working on it to fix it

mccorby: I'm postponing the implementation of `mul` when the tensors are shared. It will come in another PR as it is a bit tricky.

PR #2585, opened 2019-09-03
Support fetch_plan remotely

mari-linhares:


syft/workers/websocket_client.py
```diff
@@ -71,6 +71,14 @@ def search(self, *query):
...
         return sy.serde.deserialize(response)
 
+    def fetch_plan(self, plan_id):
+        # Prepare a message requesting the websocket server to search among its objects
+        message = messaging.FetchPlanMessage(plan_id)
+        serialized_message = sy.serde.serialize(message)
+        # Send the message and return the deserialized response.
+        response = self._recv_msg(serialized_message)
+        return sy.serde.deserialize(response)
+
```
LaRiffle: Just a question of naming (i'm not familiar with socket workers), why don't we implement _send_msg to call _recv_msg, which would make this line `response = self._recv_msg(serialized_message)` much easier to read for newcomers?

mari-linhares: In my opinion, I think this is a great idea and I don't see why this could be a problem. But this is out of the scope of this change, it would be a good first issue though!

PR #2590, opened 2019-09-05
Support fetch and sending encrypted shares with stateful plans

mari-linhares:
What this PR implements:
- [x] Support fetch plan for stateful plans
- [x] Support fetch plan for stateful plans + remote_execution using Grid
- [x] Support for sending a plan after is shared
- [x] Support fetch plan for stateful plans + websocket remote  execution
- [x] Sending an "encrypted plan"

What is still missing and should be addressed in other PRs:
- [ ] Support for changing the state of a plan `plan.update_state("key", new_value)`


syft/workers/base.py
```diff
@@ -761,7 +791,13 @@ def request_remote_tensor_shape(
         shape = self.send_msg(messaging.GetShapeMessage(pointer), location=pointer.location)
         return sy.hook.create_shape(shape)
 
-    def fetch_plan(self, plan_id: Union[str, int]) -> "Plan":  # noqa: F821
+    def fetch_plan(self, plan_id):
+        if plan_id in self._objects:
+            candidate = self._objects[plan_id]
+            if isinstance(candidate, sy.Plan):
+                return candidate.copy_to_worker(sy.hook.local_worker)
```
midokura-silvia: it would be good to explicit what happens when the if statements are False. That is, to add a return None in the end (if this is the desired behaviour)

mari-linhares: Indeed. Done!

syft/workers/base.py
```diff
@@ -761,9 +791,29 @@ def request_remote_tensor_shape(
...
+        This method is executed for local execution.
+
+        Args:
+            plan_id: A string indicating the plan id.
+
+        Returns:
+            A plan if a plan with the given `plan_id` exists. Returns None otherwise.
+        """
+        if plan_id in self._objects:
+            candidate = self._objects[plan_id]
```
LaRiffle: use self.get_obj instead

syft/workers/base.py
```diff
@@ -761,9 +791,29 @@ def request_remote_tensor_shape(
         shape = self.send_msg(messaging.GetShapeMessage(pointer), location=pointer.location)
         return sy.hook.create_shape(shape)
 
-    def fetch_plan(self, plan_id: Union[str, int]) -> "Plan":  # noqa: F821
+    def fetch_plan(self, plan_id):
```
LaRiffle: we lost the typing infos :(

syft/workers/websocket_client.py
```diff
@@ -71,13 +71,31 @@ def search(self, *query):
...
         response = self._recv_msg(serialized_message)
-        return sy.serde.deserialize(response)
+        plan = sy.serde.deserialize(response)
+
+        if plan.state_ids:
+            state_ids = []
+            for state_id in plan.state_ids:
+                ptr = self.get_ptr(state_id)
+                ptr.owner = sy.hook.local_worker
+                sy.hook.local_worker._objects[ptr.id] = ptr
```
LaRiffle: That's not OK haha

Ok so I see what you need, instead of creating def get_ptr, you should use `register_response` of `hook_args` (extract of the doctring: "When a remote worker execute a command sent by someone else, the response is inspected: all tensors are stored by this worker and a Pointer tensor is made for each of them.").
this is what is used in baseworker.py::execute_command, it's very practical and you have nothing to care about.

test/message/test_plan.py
```diff
@@ -431,10 +431,91 @@ def plan_mult_3(data):
...
+
+    alice = workers["alice"]
+    sent_plan = plan.send(alice)
+
+    # Fetch plan
+    fetched_plan = alice.fetch_plan(sent_plan.id)
+
+    # Execute it locally
+    x = th.tensor([-1.26])
+    assert (fetched_plan(x) == sent_plan(x)).all()
```
LaRiffle: I wonder if plan.send() is inplace and if sent_plan == plan
I this is the case, then you can write `assert (fetched_plan(x) == plan(x)).all()`
If not, I would expect sent_plan(x) to return a ptr

test/message/test_plan.py
```diff
@@ -449,10 +530,9 @@ def plan_mult_3(data):
     # Execute it locally
     x = th.tensor([-1, 2, 3])
     assert (get_plan(x) == th.tensor([-3, 6, 9])).all()
-    assert fetched_plan.is_built
     assert (fetched_plan(x) == th.tensor([-3, 6, 9])).all()
-
-    hook.local_worker.is_client_worker = True
```
LaRiffle: Keep this line
when we call hook.local_worker.is_client_worker = False in a test we should **always** revert it at the end. 
Otherwise, I suffer a lot :D

syft/messaging/plan.py
```diff
@@ -378,6 +378,15 @@ def find_location(self, args):
...
         return sy.framework.hook.local_worker
 
+    def _replace_reference_ids(self, old_id):
+        # Replace ids that reference self.id
+        self.replace_ids(
+            from_ids=self.arg_ids, to_ids=self.arg_ids, from_worker=old_id, to_worker=self.id
+        )
+        self.replace_ids(
+            from_ids=self.result_ids, to_ids=self.result_ids, from_worker=old_id, to_worker=self.id
+        )
```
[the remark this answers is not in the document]
mari-linhares: > As from_ids == to_ids and from_worker -> to_worker is identicial in the 2 calls, only one call to replace_ids looks necessary

I just checked `self.replace_ids`, we need both of the calls since the first replate the arg_ids and the second replace the result_ids.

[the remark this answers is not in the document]
mari-linhares: > Maybe also clarify the comment: we're updating the worker id to own the operations on the plan

Done

syft/messaging/plan.py
```diff
@@ -388,13 +397,49 @@ def copy(self):
             result_ids=self.result_ids,
             readable_plan=self.readable_plan,
             is_built=self.is_built,
+            state_ids=self.state_ids,
```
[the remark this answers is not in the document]
mari-linhares: Can we then:

1. Create a copy of the states
2. Get the new ids
3. Send the state as this new ids

Is this okay?

syft/workers/base.py
```diff
@@ -431,6 +432,23 @@ def execute_command(self, message: tuple) -> "pointers.PointerTensor":
...
+
+        Args:
+            message: A tuple specifying the command and args.
+        """
+        args, command_name = message
+        if command_name == "fetch_plan":
+            return self._fetch_plan_remote(*args)
+        elif command_name == "get_ptr":
+            return self.get_ptr(*args)
+        else:
```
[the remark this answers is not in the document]
mari-linhares: done

syft/workers/base.py
```diff
@@ -431,6 +432,23 @@ def execute_command(self, message: tuple) -> "pointers.PointerTensor":
...
+    def execute_plan_command(self, message: tuple):
+        """Executes commands related to plans.
+
+        This method is intended to execute all commands related to plans and
+        avoiding having several new message types specific to plans.
+
+        Args:
+            message: A tuple specifying the command and args.
+        """
+        args, command_name = message
```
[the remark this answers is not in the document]
mari-linhares: done

test/message/test_plan.py
mari-linhares: nice catch, thanks!

syft/workers/websocket_client.py
mari-linhares: I'm not sure how to use `register_response`, can you send a snippet with pseudo-code of how that would look like?

From the documentation + example in base I understand that `base.get_ptr` should use register_response like this:

```
    def get_ptr(self, obj_id: Union[str, int]):
        """Returns a pointer to an object from the registry.

        Args:
            obj_id: A string or integer id of an object to look up.
        """
        obj = self.get_obj(obj_id)
        if hasattr(obj, "allowed_to_get") and not obj.allowed_to_get():
            raise GetNotPermittedError()
        else:
            # response = obj.create_pointer(garbage_collect_data=False, owner=sy.local_worker).wrap()
            ptr = sy.frameworks.torch.hook_args.register_response(
                "get_ptr", obj, list(tuple([sy.ID_PROVIDER.pop()])), self
            )
            return ptr
```

And I don't need `websocket_client.get_ptr` anymore, so `websocket.fetch_plan` will look like this, is that right?

```
    def fetch_plan(self, plan_id):
        message = messaging.PlanCommandMessage((plan_id,), "fetch_plan")
        serialized_message = sy.serde.serialize(message)
        # Send the message and return the deserialized response.
        response = self._recv_msg(serialized_message)
        plan = sy.serde.deserialize(response)

        if plan.state_ids:
            state_ids = []
            for state_id in plan.state_ids:
                ptr = self.get_ptr(state_id)
                sy.hook.local_worker._objects[ptr.id] = ptr
                state_ids.append(ptr.id)
            plan.replace_ids(plan.state_ids, state_ids)
            plan.state_ids = state_ids

        plan.replace_worker_ids(self.id, sy.hook.local_worker.id)
        return plan
```

syft/messaging/plan.py
```diff
@@ -388,13 +397,49 @@ def copy(self):
...
+            readable_plan=self.readable_plan,
+            is_built=self.is_built,
         )
+
+        # Sent state to the new owner
+        if self.state_ids:
+            state_ids = []
+            for state_id in self.state_ids:
+                state_ids.append(
+                    self.owner._objects[state_id]
```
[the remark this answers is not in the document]
mari-linhares: Done, but garbage_collect_data = False is still needed.

syft/messaging/plan.py
```diff
@@ -388,13 +397,49 @@ def copy(self):
...
+                    .send(worker, garbage_collect_data=False)
+                    .id_at_location
+                )
+
+            plan.replace_ids(self.state_ids, state_ids)
+            plan.state_ids = state_ids
+
+        plan._replace_reference_ids(self.id)
+        # Replace old owner id to the new owner id
+        plan.replace_worker_ids(self.owner.id, worker.id)
```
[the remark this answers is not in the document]
mari-linhares: Added some comments that should make my intentions clear :)

syft/workers/base.py
mari-linhares: fixed :)

test/message/test_plan.py
mari-linhares: > I wonder if plan.send() is inplace and if sent_plan == plan

Yes they should be

syft/messaging/plan.py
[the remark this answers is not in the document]
LaRiffle: Ok this is not a pb :)

syft/messaging/plan.py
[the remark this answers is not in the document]
LaRiffle: Yes let's keep this for another PR!

syft/messaging/plan.py
```diff
@@ -388,13 +397,49 @@ def copy(self):
...
+            worker: Owner of the plan copy.
+        """
+        plan = Plan(
+            sy.ID_PROVIDER.pop(),
+            worker,
+            self.name,
+            arg_ids=self.arg_ids,
+            result_ids=self.result_ids,
+            readable_plan=self.readable_plan,
+            is_built=self.is_built,
```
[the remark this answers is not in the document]
mari-linhares: done

syft/workers/base.py
mari-linhares: done

syft/workers/websocket_client.py
mari-linhares: Hey Theo, can you check if this is better? https://github.com/OpenMined/Grid/pull/288

syft/workers/base.py
```diff
@@ -439,6 +443,21 @@ def execute_command(self, message: tuple) -> PointerTensor:
...
+        """Executes commands related to plans.
+
+        This method is intended to execute all commands related to plans and
+        avoiding having several new message types specific to plans.
+
+        Args:
+            message: A tuple specifying the command and args.
+        """
+        command_name, args = message
+        if command_name in self._plan_command_router:
```
robert-wagner: How often do we expect to get a command not in the command router? This would probably be faster as a try except since I expect that in general we not have incorrect commands very often (or at all)

mari-linhares: The problem with try except is that if we get a `keyError` while running the command itself we'll get the wrong error message (this happened to me a few times and it was very confusing).

syft/messaging/message.py
```diff
@@ -382,31 +382,62 @@ def detail(worker: AbstractWorker, msg_tuple: tuple) -> "SearchMessage":
...
+        # call the parent constructor - setting the type integer correctly
+        super().__init__(codes.MSGTYPE.PLAN_CMD)
+
+        self.message = message
+        self.command_name = command_name
+
+    @property
+    def contents(self):
+        """Returns a tuple with the contents of the operation (backwards compatability)."""
+        return (self.message, self.command_name)
```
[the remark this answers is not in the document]
mari-linhares: fixed

syft/messaging/message.py
```diff
@@ -382,31 +382,62 @@ def detail(worker: AbstractWorker, msg_tuple: tuple) -> "SearchMessage":
...
-    """Message used to fetch a plan from another worker."""
+class PlanCommandMessage(Message):
+    """Message used to execute a command related to plans."""
 
     # TODO: add more efficient detailer and simplifier custom for this type
     # https://github.com/OpenMined/PySyft/issues/2512
 
-    def __init__(self, contents):
-        """Initialize the message using default Message constructor.
+    def __init__(self, message: tuple, command_name: str):
```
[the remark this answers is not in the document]
mari-linhares: fixed

syft/messaging/plan.py
```diff
@@ -390,12 +426,24 @@ def copy(self):
...
             is_built=self.is_built,
         )
-        plan.replace_ids(
-            from_ids=plan.arg_ids, to_ids=plan.arg_ids, from_worker=self.id, to_worker=plan.id
-        )
-        plan.replace_ids(
-            from_ids=plan.result_ids, to_ids=plan.result_ids, from_worker=self.id, to_worker=plan.id
-        )
+
+        # TODO: we shouldn't use the same state_ids since we want to avoid
```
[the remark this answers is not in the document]
mari-linhares: done

syft/workers/base.py
robert-wagner: So if we do 
```
try:
    code here
except KeyError as e:
    raise PlanCommandUnknownError(command name)
```

mari-linhares: Yes, but if <code here> throws a KeyError by another reason (for instance we try to get_obj for an unknown obj) we would raise the wrong error. I think this could be fixed on another PR where we create a get_obj exception. What do you think?

mari-linhares: I'll use try except for now, but we should probably raise a syft exception at get_obj https://github.com/OpenMined/PySyft/blob/750bd2c3987a570ef1dc8554b933696d416724c7/syft/generic/object_storage.py#L84

robert-wagner: Agreed

mari-linhares: I've created an issue for this: https://github.com/OpenMined/PySyft/issues/2608

PR #2591, opened 2019-09-05
Generalize and improve hook_args.py implementation

jvmancuso:
Note to reviewer: code review will be easiest here if taken commit-by-commit. Care was taken to make sure that the commit messages are comprehensive and descriptive.

The main purpose of this PR is to generalize hook_args to non-torch frameworks. This was actually a very small amount of work and can be seen in the first and third commits in this PR, however I noticed several problems across the code base that I'm also addressing here:

1. Importing was a struggle after moving files around, because we were often using absolute-ish paths (non-relative paths but importing a lot of things in various __init__.py files throughout the repo).  I've standardized imports to be completely absolute across the repo.  This means that you should _never_ import things from an __init__.py other than the main one in the `syft/` package (which should mainly consist of the user-facing API). If there happen to be any leftover exceptions to this, they are only in packages with very low connectivity with their submodules and other modules, and they were left untouched by accident.
2. After debugging all the import issues related to the above, I realized that the `syft.frameworks.torch.hook_args` module had quite a few circular dependencies with modules that define their own tensor types.  The main reason for this was that hook_args needs these tensor types in its `type_rule`, `forward_func`, and `backward_func` registries, however hook_args was also being used in many of these tensor types' definitions.  To fix this, registering tensor types should now happen in the same module as their definition using [these new functions](https://github.com/OpenMined/PySyft/commit/5f305a538b810e88fea6d2ce2005abd0a647f341#diff-fb445e63f5c3a9effbdbea9651ff3496R93).


test/conftest.py
```diff
@@ -87,9 +87,9 @@ def workers(hook):
 
     # Reset the hook and the local worker
     syft.local_worker.clear_objects()
-    syft.frameworks.torch.hook.hook_args.hook_method_args_functions = {}
-    syft.frameworks.torch.hook.hook_args.hook_method_response_functions = {}
-    syft.frameworks.torch.hook.hook_args.get_tensor_type_functions = {}
+    syft.generic.frameworks.hook.hook_args.hook_method_args_functions = {}
```
LaRiffle: Can you add an import line to make those 3 lines more readable?

syft/__init__.py
```diff
@@ -3,12 +3,17 @@
...
 from syft import frameworks
-from syft import workers
 from syft import codes
-from syft import federated
-from .version import __version__
+from syft.version import __version__
+
+# This import statement is strictly here to trigger registration of syft
+# tensor types inside hook_args.py.
+import syft.frameworks.torch.hook.hook_args
```
LaRiffle: I understand why you did this, but I wonder if there is not a more pythonic way to trigger this registration?

syft/__init__.py
```diff
@@ -51,30 +54,30 @@
...
-from syft.messaging import method2plan
-from syft.messaging import make_plan
+from syft.messaging.plan import Plan
+from syft.messaging.plan import func2plan
+from syft.messaging.plan import method2plan
+from syft.messaging.plan import make_plan
 
 # Import Worker Types
-from syft.workers import VirtualWorker
+from syft.workers.virtual import VirtualWorker
```
LaRiffle: Can you add all main workers as well (socket, etc)?

syft/federated/train_config.py
```diff
@@ -4,9 +4,9 @@
 import torch
 
 import syft as sy
-from syft import workers
-from syft.generic import pointers
-from syft.workers import AbstractWorker
+from syft.generic.pointers.object_wrapper import ObjectWrapper
```
midokura-silvia: I would prefer to follow the google style guide for import statements: https://github.com/google/styleguide/blob/gh-pages/pyguide.md, saying (section 2.2):

> Use import statements for packages and modules only, not for individual classes or functions. Note that there is an explicit exemption for imports from the typing module.

The advantage is that you always know, from which module your classes are coming from.

syft/generic/frameworks/hook/hook_args.py
```diff
@@ -0,0 +1,780 @@
...
+    "_getitem_public",
+    "__setitem__",
+    "view",
+    "permute",
+    "add_",
+    "sub_",
+    "new",
+    "chunk",
+}
+ambiguous_functions = {
```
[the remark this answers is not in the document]
jvmancuso: agreed, totally forgot about these

syft/workers/base.py
```diff
@@ -22,7 +29,8 @@
 
 # this if statement avoids circular imports between base.py and pointer.py
 if TYPE_CHECKING:
-    from syft.generic import pointers
+    from syft.generic.pointer import ObjectPointer
```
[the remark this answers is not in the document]
jvmancuso: I recall it breaking several versions back when I added this code, will try again

test/conftest.py
jvmancuso: I wasn't too sure about python's scoping when you import objects from modules vs. modules -- will give it a shot

syft/__init__.py
jvmancuso: open to ideas!

syft/federated/train_config.py
jvmancuso: I'm not sure how I feel about it, but I suggest moving this discussion into an issue or the changes into a separate PR. This one is already quite large, and it should be easier to do once imports are standardized after this merges.

syft/workers/base.py
```diff
@@ -210,7 +218,7 @@ def load_data(self, data: List[Union[FrameworkTensorType, AbstractTensor]]) -> N
                 self.register_obj(tensor)
                 tensor.owner = self
 
-    def send_msg(self, message: "messaging.Message", location: "BaseWorker") -> object:
+    def send_msg(self, message: "Message", location: "BaseWorker") -> object:
```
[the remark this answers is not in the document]
jvmancuso: In certain cases I think we do, e.g. when we need to avoid circular imports or when referring to BaseWorker in its own definition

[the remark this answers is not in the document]
jvmancuso: but this line can be changed I believe!

syft/workers/base.py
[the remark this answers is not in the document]
jvmancuso: no longer need it for these pointer classes, but we do need it for FrameworkHook still, so gonna make that change

syft/federated/train_config.py
midokura-silvia: It is definitely a discussion that is out of the scope of this pull request. However you are also changing the import statements that fulfilled the google style guide. So we cannot just postpone the discussion.

jvmancuso: There were very few existing import statements that fulfilled the requirement you quoted, and the existing import statements are clearly not yet standardized, so I don't think that should be a blocker to this PR. I will make an issue and link this thread to it.

jvmancuso:
>However, it will make it harder for people to import objects in their notebook and so on, so I think we should be generous in syft/init.py to provide many shortcuts just like we can currently call sy.PointerTensor(...). This should include all tensors, workers, and objects a normal user is expected to play with (I believe many of them are already there).

In fact, I don't think many things other than the hook and workers are used in the notebooks. It seems very few tensors actually need to be exposed -- PointerTensor/MultiPointerTensor seem to be abstracted away from the user, and similar for the various precision & crypto tensors.

I'm inclined to remove things that are only exposed for internal development. The more we add to `syft/__init__.py`, the more complex our import graph gets. That quickly becomes a much bigger headache than having to add a specific import statement for the class I'm using inside of syft internals.

jvmancuso:
>I'm inclined to remove things that are only exposed for internal development.

Well, unfortunately there are a number of places that rely on tensor being exposed in the `syft` namespace.  I'm taking this to mean that we have a lot more circularity than we want throughout the code (in particular, tensors seem to be highly intertwined).  Whether this is okay or not, and whatever we end up doing about it, is almost definitely out of scope.  I will keep tensor types in the `syft` namespace.

jvmancuso:
This is ready for another review!

syft/generic/frameworks/hook/hook_args.py
```diff
@@ -101,6 +84,16 @@ def register_backward_func(new_backward_rules: Dict):
     backward_func = {**backward_func, **new_backward_rules}
 
 
+def register_ambiguous_method(*method):
+    global ambiguous_methods
```
LaRiffle: Why is it need to call global?

test/test_serde.py
```diff
@@ -10,6 +10,7 @@
 from torch import Tensor
 
 import syft
+from syft.frameworks.torch.tensors.interpreters.additive_shared import AdditiveSharingTensor
```
LaRiffle: Why is this one needed? just calling syft. AdditiveSharingTensor doesn't work?

LaRiffle: or, are you preparing the future? :)

jvmancuso: yes, this change was from when I was hoping to remove tensors from the main `syft` namespace. I can change it back if we want!

syft/generic/frameworks/hook/hook_args.py
jvmancuso: Scoping in Python functions is one-way from global --> local, so modifications made to this variable will not be retained outside of this function unless we call `global` (or at least `nonlocal`).

jvmancuso: But to the larger question of why we need to do this whole registration business, it's so that we can register classes from outside of this module in these dictionaries/sets without importing such classes.

test/test_serde.py
LaRiffle: Let's keep it for now :)

syft/generic/frameworks/hook/hook_args.py
LaRiffle: > But to the larger question of why we need to do this whole registration business, it's so that we can register classes from outside of this module in these dictionaries/sets without importing such classes.

I totally understand and agree with this



> Scoping in Python functions is one-way from global --> local, so modifications made to this variable will not be retained outside of this function unless we call `global` (or at least `nonlocal`).

Have you tested it without global? I tried a simplified version of it, and it didn't seem necessary to turn this global

jvmancuso: Yes, it should work if you call them all in the same module, but `global` is necessary when calling these registration functions from a different module, which is the intended usage.

PR #2516, opened 2019-08-15
[WIP] PromiseTensor

iamtrask:
Fixes #2515

PR #2610, opened 2019-09-10
Continue the work on Promises and PromiseTensors

Jasopaum:
Continue the work done here: https://github.com/OpenMined/PySyft/pull/2516

TODO:
- [x] Operations with scalars
- [x] Promises that can be kept several times
- [x] Bufferize output of Plans with promises
- [x] Plans with promises
- [x] Operations between promises and non-promises
- [x] Promises on sent plans
- [x] Operations on sent promises
- [x] Protocols with promises
- [x] Tests


test/torch/tensors/test_promise.py
```diff
@@ -0,0 +1,157 @@
...
+@pytest.mark.parametrize("cmd", ["__add__", "sub", "__mul__"])
+def test_operations_with_concrete(hook, cmd):
+    hook.local_worker.is_client_worker = False
+
+    a = syft.Promises.FloatTensor(shape=torch.Size((2, 2)))
+    b = torch.tensor([[-8.0, -7], [6, 5]]).wrap()  # TODO fix need to wrap
+
+    actual = getattr(a, cmd)(b)
+
+    ta = torch.tensor([[1.0, 2], [3, 4]]).wrap()
```
iamtrask: It is unfortunate that you have to wrap this. We previously haven't had good support when two tensors are interacting with each other where one has a different class hierarchy than the other. @LaRiffle any ideas?

test/torch/tensors/test_promise.py
```diff
@@ -0,0 +1,157 @@
...
+
+    x = a.send(bob)
+
+    x.keep(torch.ones((2, 2)))
+
+    assert (x.value().get() == torch.ones((2, 2))).all()
+
+
+@pytest.mark.parametrize("cmd", ["__add__", "sub", "__mul__"])
+def test_remote_operations(workers, cmd):
```
iamtrask: Can we also check that .move() works with Promises as well, meaning you could chain promises across a long series of workers and they would work?

test/torch/tensors/test_promise.py
```diff
@@ -0,0 +1,157 @@
...
+    ty = torch.tensor([[-8.0, -7], [6, 5]])
+    x.keep(tx)
+    y.keep(ty)
+
+    expected = getattr(tx, cmd)(ty)
+
+    assert (actual.value().get() == expected).all()
+
+
+def test_bufferized_results(hook):
```
iamtrask: WOAH!!! This is incredible!!! Well done!

test/torch/tensors/test_promise.py
```diff
@@ -0,0 +1,157 @@
+import pytest
+import torch
+import syft
+
+
```
iamtrask: Is there a way we can test the asynchronous execution of these promises? Perhaps we can create a "ThreadWorker" which is like a VirtualWorker but it runs in its own thread (might have to communicate with sockets... in fact this could just be a wrapper around SocketWorker)? In theory the execution time for doing promises on a series of ThreadWorkers should be much faster?

Also it would be really amazing if we could check/test that our MPC stuff can be used to create a graph of promises. Meaning we want the ability to call .share() on a promise, then do some executions, and then call .get() (which is still a promise holding the results). Then call .keep() on the input and the whole thing should run automatically.

test/torch/tensors/test_promise.py
Jasopaum: Hum, that's right, I didn't think about .move(), it does not work for now.
Can you give a small example of what you mean by chaining promises across a series of workers?

test/torch/tensors/test_promise.py
Jasopaum: I think it still needs some work before we can share Promises but I'll think about that!

And for the first part, I'm not sure I got what you mean... Do you want to compare with "classic" plans for which (I guess) the results are waited for each time the plan is called?

test/torch/tensors/test_promise.py
iamtrask: Sure thing!

The first approach is via .send() which sends the pointers around.
```
x = sy.Promises.FloatTensor([2,2])
x_bob = x.send(bob)
x_john = x_bob.send(john)
x_bill = x_john.send(bill)
```
So now we have a promise pointing to bill, which is a promise pointing to john, which is a promise pointing to bob, and if we keep the first one it'll send the data all the way down the chain.
```x_bill.keep(th.zeros([2,2])```

We can do something identical using .move()
```
x = sy.Promises.FloatTensor([2,2])
x_bob = x.send(bill)
x_john = x_bob.move(john)
x_bill = x_john.move(bill)
x_bob = x_bill.move(bob)
```
This is identical, but as you can see we can construct it in reverse. Often the order matters. I hope this is useful.

File: test/torch/tensors/test_promise.py (hunk 4 above)
Comment (iamtrask) [reply test/torch/tensors/test_promise.py]: No more compare with doing everything synchronously. Aka... if you .keep() a promise on VirtualWorkers, it should be slower than if you .keep() a promise on ThreadWorkers.

However, if you execute a command on a VirtualWorker or a ThreadWorker normally (without promises) they should run in the same amount of time (because your executing thread blocks and forces them to be synchronous).

File: syft/frameworks/torch/hook/hook.py [hunk 5]
```diff
@@ -468,6 +473,62 @@ def overloaded_attr(self, *args, **kwargs):
...
+            def method(self, *args, **kwargs):
+
+                arg_shapes = list([self._shape])
+                arg_ids = list([self.id])
+
+                # Convert scalar arguments to tensors to be able to use them with plans
+                args = list(args)
+                for ia in range(len(args)):
+                    if not isinstance(args[ia], (torch.Tensor, AbstractTensor)):
+                        args[ia] = torch.tensor(args[ia])
```
LaRiffle: I'm surprised that this is needed: you can usually use scalars with torch operation:
A cmd like `th.tensor([1.]) + 1` will (unless I'm wrong) be serialized and store the 1 as a scalar

syft/frameworks/torch/hook/hook.py
```diff
@@ -468,6 +473,62 @@ def overloaded_attr(self, *args, **kwargs):
...
         return overloaded_attr
 
+    def _hook_promise_tensor(hook_self):
+
+        methods_to_hook = hook_self.to_auto_overload[torch.Tensor]
+
+        def generate_method(method_name):
+            def method(self, *args, **kwargs):
+
+                arg_shapes = list([self._shape])
```
LaRiffle: Does it work with .shape ?

syft/frameworks/torch/hook/hook.py
LaRiffle: I understand it will make your next call to .shape fail, but you can test there if it has a shape

syft/frameworks/torch/tensors/interpreters/promise.py
```diff
@@ -0,0 +1,200 @@
...
+
+        if owner is None:
+            owner = sy.local_worker
+
+        # I did check that the two __init__ methods below only get called once, but it
+        # was exhibiting some strange behavior when I used super() for both of them.
+
+        # constructor for AbstractTensor
+        super().__init__(id=id, owner=owner, tags=tags, description=description)
+
```
LaRiffle: Call `AbstractTensor.__init__(self, ...)` for more clarity indeed

syft/frameworks/torch/tensors/interpreters/promise.py
```diff
@@ -0,0 +1,200 @@
...
+        self._shape = shape
+
+        del self.child
+
+    def torch_type(self):
+        return self.obj_type
+
+    @property
+    def shape(self):
```
LaRiffle: Why exactly is it a property?
Does it need to be "un-changeable" ?

syft/frameworks/torch/tensors/interpreters/promise.py
```diff
@@ -0,0 +1,200 @@
...
+        return self._shape
+
+    @property
+    def grad(self):
+        return None
+        # if not hasattr(self, "_grad"):
+        #     self._grad = PromiseTensor(shape=self._shape, tensor_type=self.torch_type()).wrap()
+        #
+        # return self._grad
```
LaRiffle: Agreed, we're a bit too early for this stuff^^

syft/frameworks/torch/tensors/interpreters/promise.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def on(self, tensor: "AbstractTensor", wrap: bool = True) -> "AbstractTensor":
```
LaRiffle: That's the same piece of code than for AbstractTensor? If so we can remove it :)

syft/frameworks/torch/tensors/interpreters/promise.py
```diff
@@ -0,0 +1,200 @@
...
+        Args:
+            tensor: a FixedPrecisionTensor.
+
+        Returns:
+            tuple: a tuple holding the unique attributes of the fixed precision tensor.
+        """
+
+        return (
+            sy.serde._simplify(self.id),
+            sy.serde._simplify(self._shape),
```
LaRiffle: Use .shape instead

syft/frameworks/torch/tensors/interpreters/promise.py
```diff
@@ -0,0 +1,200 @@
...
+        id = sy.serde._detail(worker, id)
+        shape = sy.serde._detail(worker, shape)
+        tensor_type = sy.serde._detail(worker, tensor_type)
+        plans = sy.serde._detail(worker, plans)
+
+        tensor = PromiseTensor(
+            owner=worker, id=id, shape=shape, tensor_type=tensor_type, plans=plans
+        )
+
+        initialize_tensor(
```
LaRiffle: Is this needed? I'm not 100% sure it does

syft/frameworks/torch/tensors/interpreters/promise.py
```diff
@@ -0,0 +1,200 @@
...
+            owner=worker,
+            id=id,
+            init_args=[],
+            kwargs={},
+        )
+
+        return tensor
+
+
+def CreatePromiseTensor(shape, tensor_type: str, *args, **kwargs):
```
LaRiffle: Is this needed? Looks a bit confusing to me when and why I should use this alias

syft/generic/object_storage.py
```diff
@@ -119,6 +121,9 @@ def force_rm_obj(self, remote_key: Union[str, int]):
         """
         if remote_key in self._objects:
             obj = self._objects[remote_key]
+            if hasattr(obj, "child") and obj.child is not None:
```
LaRiffle: This looks like a duplicate code right?

syft/generic/pointers/pointer_tensor.py
```diff
@@ -350,6 +451,28 @@ def share(self, *args, **kwargs):
...
+
+        # Send the command
+        command = ("keep", self, args, kwargs)
+
+        response = self.owner.send_command(self.location, command)
+
+        return response
+
+    def value(self, *args, **kwargs):
+        command = ("value", self, args, kwargs)
```
LaRiffle: Add a docstring to tell what it does

syft/messaging/plan.py
```diff
@@ -477,12 +487,14 @@ def replace_ids(
             # for every message of the plan
             for j, msg in enumerate(self.readable_plan):
                 # look for the old id and replace it with the new one
-                self.readable_plan[j] = Plan._replace_message_ids(
-                    obj=msg,
-                    change_id=from_ids[i],
-                    to_id=to_ids[i],
-                    from_worker=from_worker,
-                    to_worker=to_worker,
+                self.readable_plan[j] = list(
```
LaRiffle: what's the benfit of casting it into a list?

syft/messaging/plan.py
```diff
@@ -562,15 +576,30 @@ def __call__(self, *args, **kwargs):
...
                     return self.forward(*args)
             else:
                 return self._execute_readable_plan(*args)
         else:
-            if not self.is_built:
+            if any(
+                [
+                    hasattr(arg, "child")
+                    and isinstance(
+                        arg.child.location._objects[arg.child.id_at_location].child, PromiseTensor
```
LaRiffle: That's not allowed ;)
But as discussed it's out of scope for this PR, maybe juste remove it

syft/messaging/plan.py
```diff
@@ -817,9 +858,53 @@ def __str__(self):
...
         return self.__str__()
 
+    def setup_plan_with_promises(self, *args):
+        """ Slightly modifies a plan so that it can work with promises.
+        The plan will also be sent to location with this method.
+        """
+        location = args[0].child.location if isinstance(args[0].child, PointerTensor) else None
+        if location is not None:
+            # Get remote arguments
+            args = [arg.child.location[arg.child.id_at_location] for arg in args]
```
LaRiffle: This won't work - and is not allowed ;)

syft/messaging/plan.py
```diff
@@ -817,9 +858,53 @@ def __str__(self):
...
+            if hasattr(p, "child") and isinstance(p.child, PromiseTensor):
+                p.child.plans.add(self.id)
+                if prom is None:
+                    prom = p
+
+        res = sy.PromiseTensor(
+            owner=prom.owner,
+            shape=self.output_shape,
+            tensor_id=self.result_ids[0],
+            tensor_type=prom.child.obj_type,
```
LaRiffle: not sure you can infer the type of the res from the input type: maybe just let it empty for now and replaced by a default "th.floattensor"

syft/messaging/plan.py
LaRiffle: Maybe just remove this case

syft/messaging/promise.py
```diff
@@ -0,0 +1,165 @@
...
+            self.owner.register_obj(obj)
+
+        self.queue_obj_ids.append(obj.id)
+
+        # If some plans were waiting for this promise...
+        for plan_id in self.plans:
+            # If the promise has been moved, the plan waiting for it might not be on the same worker
+            # TODO is this ok?
+            if plan_id in self.owner._objects:
+                plan = self.owner._objects[plan_id]
```
LaRiffle: it's ok because the owner is support to be local, but use instead .get_obj()
`_objects` should only be used for debug purpose

syft/messaging/promise.py
```diff
@@ -0,0 +1,165 @@
...
+                continue
+
+            # ... execute them if it was the last argument they were waiting for.
+            if plan.has_args_fulfilled():
+                # Collect args
+                orig_ids = plan.arg_ids
+                args = []
+                for i, arg_id in enumerate(plan.arg_ids):
+                    if isinstance(self.owner._objects[arg_id].child, Promise):
+                        id_to_add = self.owner._objects[arg_id].child.queue_obj_ids[0]
```
LaRiffle: again, use .get_obj()
and also pop(0) to be consistent with the next lines

syft/messaging/promise.py
```diff
@@ -0,0 +1,165 @@
...
+                plan.arg_ids = orig_ids
+
+                # Remove objects from queues:
+                for arg_id in plan.arg_ids:
+                    arg = self.owner._objects[arg_id].child
+                    if isinstance(arg, Promise):
+                        to_rm = arg.queue_obj_ids.pop(0)
+                        self.owner.rm_obj(to_rm)
+
+                self.owner._objects[plan.promise_out_id].keep(result)
```
LaRiffle: .get_obj()

syft/messaging/promise.py
```diff
@@ -0,0 +1,165 @@
...
+    def value(self):
+        """ Returns the next object in the queue of results.
+        """
+        if not self.queue_obj_ids:
+            # If the promise has still not been kept
+            # or if the queue of results has been emptied
+            # TODO this doesn't work as I want with pointerTensors
+            return None
+        ret_id = self.queue_obj_ids.pop(0)
+        ret = self.owner._objects[ret_id]
```
LaRiffle: .get_obj()

test/torch/tensors/test_promise.py
LaRiffle: So, we have done specific stuff to allow AST to operate with FPT in specific cases, but nothing general
I don't really get why calling .wrap() here make the thing work :/

examples/tutorials/advanced/websockets-example-MNIST-parallel/run_websocket_client.py
```diff
@@ -111,7 +108,7 @@ def define_and_get_arguments(args=sys.argv[1:]):
         max_nr_batches=max_nr_batches,
         epochs=1,
         optimizer="SGD",
-        optimizer_args={"lr": lr},
+        optimizer_args={"lr": 0.1},
```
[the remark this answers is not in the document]
Jasopaum: I don't know, it was part of the branch I continued the work on. @iamtrask?

syft/frameworks/torch/tensors/interpreters/promise.py
Jasopaum: This line is added compared to AbstractTensor:
`self.obj_type = tensor.type()`

examples/tutorials/advanced/websockets-example-MNIST-parallel/run_websocket_client.py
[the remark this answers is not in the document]
LaRiffle: You can safely revert it :)

syft/frameworks/torch/hook/hook.py
Jasopaum: It seems to work

syft/frameworks/torch/tensors/interpreters/promise.py
Jasopaum: I guess it's because it is a property for AbstractObject, which is a parent class of Promise

syft/frameworks/torch/hook/hook.py
Jasopaum: It was to be able to use plans with scalar arguments.
I don't think it works to call a plan for which some arguments are Python integers, for instance.

syft/frameworks/torch/tensors/interpreters/promise.py
Jasopaum: I think it's more common this way, to avoid using `.wrap()` and setting an attribute to a string representing a type each time we need a promise.
But it's @iamtrask's idea, maybe he can be more precise?

syft/messaging/plan.py
Jasopaum: I don't see any, I think that the only difference is that the list could be modified but it's not done anywhere.
@iamtrask, can I remove the conversion?

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -367,10 +368,12 @@ def send(
 
             location = location[0]
 
-            if hasattr(self, "child") and isinstance(self.child, PointerTensor):
-                self.child.garbage_collect_data = False
-                if self._is_parameter():
-                    self.data.child.garbage_collect_data = False
+            if hasattr(self, "child"):
+
```
robert-wagner: why was this check split up? Alternatively, I think it might make sense to move both of these checks to a seperate function since it seems we write the combination alot

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -822,3 +825,16 @@ def combine(self, *pointers):
         ps.append(self)
 
         return syft.combine_pointers(*ps)
+
+    def keep(self, *args, **kwargs):
```
robert-wagner: What does keep do? Could you add a docstring/explain why this is needed for native tensor

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -822,3 +825,16 @@ def combine(self, *pointers):
...
+        return self.child.keep(*args, **kwargs)
+
+    def value(self):
```
robert-wagner: Same with value

syft/frameworks/torch/tensors/interpreters/promise.py
robert-wagner: This should absolutely be a property. What a property means in python is that it returns the value when you do not add parenthesis to it

syft/frameworks/torch/tensors/interpreters/promise.py
robert-wagner: Could you add that line then call super.on to avoid rewriting the same code

syft/frameworks/torch/tensors/interpreters/promise.py
```diff
@@ -0,0 +1,185 @@
...
+    return PromiseTensor(shape, *args, tensor_type=tensor_type, **kwargs).wrap()
+
+
+class Promises:
```
robert-wagner: Why are these here and should they live in their own file?

syft/generic/pointers/callable_pointer.py
```diff
@@ -9,6 +9,9 @@
 if TYPE_CHECKING:
     from syft.workers.base import BaseWorker
 
+# TODO: move this file/class to the generic folder because the functionality is not PyTorch specific
```
robert-wagner: This is in the generic folder

syft/generic/pointers/object_pointer.py
```diff
@@ -12,6 +12,9 @@
 if TYPE_CHECKING:
     from syft.workers.base import BaseWorker
 
+# TODO: move this file/class to the generic folder because the functionality is not PyTorch specific
```
robert-wagner: same

syft/generic/pointers/object_wrapper.py
```diff
@@ -11,6 +11,9 @@
 if TYPE_CHECKING:
     from syft.workers.base import BaseWorker
 
+# TODO: move this file/class to the generic folder because the functionality is not PyTorch specific
```
robert-wagner: same

syft/generic/pointers/pointer_tensor.py
```diff
@@ -274,6 +274,107 @@ def remote_get(self):
         self.owner.send_command(message=("mid_get", self, (), {}), recipient=self.location)
         return self
 
+    @staticmethod
+    def create_pointer(
```
robert-wagner: Should this live here or in a more generic pointer?

syft/frameworks/torch/tensors/interpreters/promise.py
Jasopaum: They are the interface to create promised tensors, the user can do something like `a = sy.Promises.FloatTensor(shape=torch.Size((3, 3)))`.
But I agree, maybe they should be in another file, and we may want to think again about the naming (almost the same as the class Promise). Do you propose something?
@iamtrask

syft/__init__.py
```diff
@@ -67,11 +67,15 @@
 from syft.frameworks.torch.tensors.interpreters.autograd import AutogradTensor
 from syft.frameworks.torch.tensors.interpreters.precision import FixedPrecisionTensor
 from syft.frameworks.torch.tensors.interpreters.large_precision import LargePrecisionTensor
+from syft.frameworks.torch.tensors.interpreters.promise import CreatePromiseTensor as PromiseTensor
```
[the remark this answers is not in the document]
Jasopaum: CreatePromiseTensor is a helper function to build PromiseTensor, I do agree that this is a bit weird, I'll try to make it better

syft/frameworks/torch/tensors/interpreters/promise.py
Jasopaum: Would be a good idea!
Can someone just remind me what .on() is used for? I'm not sure I understand when it would be needed for promises (actually, it's never used at this point). If it's never useful, I'll remove that

syft/frameworks/torch/tensors/interpreters/native.py
Jasopaum: It keeps the child promise with the argument given.
I need it to be able to call the method on wrapped promise but thinking about it, I'm not 100% sure it's the right way to do it.
(This is the same for value.)

syft/messaging/plan/plan.py
```diff
@@ -331,6 +354,47 @@ def run(self, args: Tuple, result_ids: List[Union[str, int]]):
...
+        """
+        prom = None
+        for p in args:
+            if hasattr(p, "child") and isinstance(p.child, PromiseTensor):
+                p.child.plans.add(self.id)
+                if prom is None:
+                    prom = p
+
+        # TODO how to infer result type? torch.result_type() in torch 1.3?
+        result_type = args[0].torch_type()  # NOTE torch specfific here
```
LaRiffle: can't we find another strategy to get the type which is torch agnostic?

syft/messaging/plan/plan.py
```diff
@@ -331,6 +354,47 @@ def run(self, args: Tuple, result_ids: List[Union[str, int]]):
...
+            shape=self.output_shape,
+            tensor_id=self.procedure.result_ids[0],
+            tensor_type=result_type,
+            plans=set(),
+        )
+
+        self.procedure.update_args(args, self.procedure.result_ids)
+        self.promise_out_id = (
+            res.id
+        )  # NOTE should I put this in self.procedure.promise_out_id instead?
```
LaRiffle: Maybe yes

syft/messaging/plan/plan.py
```diff
@@ -439,6 +503,9 @@ def simplify(plan: "Plan") -> tuple:
             sy.serde._simplify(plan.state),
             sy.serde._simplify(plan.include_state),
             sy.serde._simplify(plan.is_built),
+            sy.serde._simplify(plan.input_shapes),
+            sy.serde._simplify(plan._output_shape) if plan._output_shape is not None else None,
```
LaRiffle: you can simplify None so sy.serde._simplify(plan._output_shape) should be sufficient

syft/messaging/promise.py
```diff
@@ -0,0 +1,155 @@
...
+Plan objects which will be executed when the promise is kept (assuming
+the plan has all of the inputs it requires). """
+
+import syft as sy
+
+from abc import ABC
+from syft.workers.abstract import AbstractWorker
+
+
+class Promise(ABC):
```
LaRiffle: what is the role of ABC?

syft/messaging/promise.py
```diff
@@ -0,0 +1,155 @@
...
+                continue
+
+            # ... execute them if it was the last argument they were waiting for.
+            if plan.has_args_fulfilled():
+                # Collect args
+                orig_ids = plan.procedure.arg_ids
+                args = []
+                ids_to_rm = []
+                for i, arg_id in enumerate(plan.procedure.arg_ids):
+                    arg = self.owner._objects[arg_id]
```
LaRiffle: use get_obj

syft/messaging/promise.py
```diff
@@ -0,0 +1,155 @@
...
+        if self.id in self.owner._objects:
+            self.owner.register_obj(obj)
+
+        self.queue_obj_ids.append(obj.id)
+
+        # If some plans were waiting for this promise...
+        for plan_id in self.plans:
+            # If the promise has been moved, the plan waiting for it might not be on the same worker
+            # TODO is this ok?
+            if plan_id in self.owner._objects:
```
LaRiffle: Hacky
use get_obj and except a ObjectNotFoundError

syft/messaging/promise.py
```diff
@@ -0,0 +1,155 @@
+"""This file contains the object we use to tell another worker that they will
```
LaRiffle: What is the exact motivation of putting this in messaging and not in promiseTensor?
Do we have reason to believe this will be reused by other types?

syft/frameworks/torch/tensors/interpreters/promise.py
Jasopaum: That's what I thought but in this case, I don't think we would ever need it in the case of Promises, am I wrong?
I'll remove it

syft/messaging/promise.py
Jasopaum: I think we could indeed have promise for objects that are not tensors, maybe for instance strings like we plan to have with SyferText

syft/messaging/promise.py
Jasopaum: I just removed the if/else, I think things should just break if we move a promised object away from the plan to which it has been promised. Do you agree?

syft/frameworks/torch/hook/hook.py
Jasopaum: I think it depends on the use case: for NN stuff maybe it's not needed but for crypto protocols, it might be more often (I'm not a crypto expert though).
And the problem is that we need the arg to have an id so that we can find it back. Or we can just store the scalar value and use it instead of an id but this needs some changes

syft/messaging/plan/plan.py
```diff
@@ -160,6 +163,19 @@ def parameters(self):
         """
         return self.state.tensors()
 
+    @property
+    def output_shape(self):
+        if self._output_shape is None:
+            args = self._create_placeholders(self.input_shapes)
+            # NOTE do I really need to regiser and them remove objects to use the method?
```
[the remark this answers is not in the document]
Jasopaum: Should this be another PR?

syft/messaging/promise.py
Jasopaum: I think it was to indicate that this class is an abstract class but as no abstract method is declared, I don't think it's useful

Jasopaum: Maybe simplify and detail should be abstract, what do you think?

LaRiffle: Yeah sounds good!

syft/messaging/plan/plan.py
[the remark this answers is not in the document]
LaRiffle: Definitely I'll open an Issue  for the context "with ..." and one for "hook torch.zeros, torch.one, and every torch. which creates a tensor to have it registered"

syft/frameworks/torch/hook/hook.py
LaRiffle: I think you can "just store the scalar value" in the plan, it looks like it works from what I've seen

syft/generic/pointers/pointer_tensor.py
```diff
@@ -265,6 +265,16 @@ def move(self, location):
         ptr.garbage_collect_data = False
         return ptr
 
+    def remote_send(self, destination):
+        """ Request the worker where the tensor being pointed to belongs to send it to destination.
+        For instance, if C holds a pointer, ptr, to a tensor on A and calls ptr.remote_send(B),
+        C will hold a pointer to a pointer on A which points to the tensor on B.
+        """
+        self.owner.send_command(
+            message=("send", self, (destination,), {"inplace": True}), recipient=self.location
```
LaRiffle: break this line:
args = (destination,)
kwargs = {"inplace": True}

syft/frameworks/torch/hook/hook.py
Jasopaum: I still think it's easier to change the scalars to tensors because this way, I can store and retrieve them from the worker when needed (when the promises are fulfilled). I can't see how to do that with scalars (but maybe it's just me :))

syft/messaging/plan/plan.py
Jasopaum: I forgot this comment, @LaRiffle !
I'm not sure if this is a problem here because PromiseTensors are torch specific

syft/frameworks/torch/hook/hook.py
```diff
@@ -504,6 +510,104 @@ def overloaded_attr(self, *args, **kwargs):
 
         return overloaded_attr
 
+    def _hook_promise_tensor(hook_self):
```
midokura-silvia: Why do we need to hook the promise tensor? Why can't we just define all the methods directly in the PromiseTensor class? The class resides within PySyft, so why don't we define the methods as for example `DoubleTensor` there? It doesn't seem to rely on the DoubleTensor being hooked before defining the method.

Jasopaum: Methods like `DoubleTensor` were in the file where the class is defined before but I was asked in some comments to move them here ^^
For the other methods, I think this file was supposed to be where this kind of method generation happen but maybe not?

midokura-silvia: Ok, that sounds like a contradiction. I'd like to hear the opinion of @LaRiffle and @robert-wagner :smile:

test/torch/tensors/test_promise.py
iamtrask: status on this?

test/torch/tensors/test_promise.py
iamtrask: status on this?

test/torch/tensors/test_promise.py
```diff
@@ -0,0 +1,198 @@
...
+    y.keep(2 * torch.ones(3, 3))
+    y.keep(4 * torch.ones(3, 3))
+
+    assert (res_ptr.value().get() == 3 * torch.ones(3, 3)).all()
+    assert (res_ptr.value().get() == 7 * torch.ones(3, 3)).all()
+
+    hook.local_worker.is_client_worker = True
+
+
+def test_protocol_waiting_promise(hook, workers):
```
iamtrask: a protocol should be a list of VirtualWorkers not a list of plans

iamtrask:
Worth re-iterating that this is a mammoth task and that Jason is an absolute saint for all the work he's done on it. <3

test/torch/tensors/test_promise.py
Jasopaum: Currently, when you use `.send()` on a Promise, you don't get a promise but a pointer pointing to a promise.
Was it a typo on your side or do we have a different understanding on what should happen when sending promises?
And currently, using `.move()`moves the promise to another worker but shouldn't create a chain as with consecutive sends, am I wrong?

test/torch/tensors/test_promise.py
Jasopaum: I think ThreadWorkers could indeed be useful for benchmarking!
But maybe it should be another PR?

syft/frameworks/torch/hook/hook.py
iamtrask: Historically PromiseTensor has always been a little bit of an exception because of the way it works. I'm ok with this for now.

test/torch/tensors/test_promise.py
iamtrask: agreed

test/torch/tensors/test_promise.py
iamtrask: I implemented a basic version of getting things to interact which dont' have the exact same class path. Not sure if it would work in this case but just FYI

iamtrask: I implemented a basic version of getting things to interact which dont' have the exact same class path. Not sure if it would work in this case but just FYI

test/torch/tensors/test_promise.py
iamtrask: Ahh, I see what you mean by consecutive sends. I think that this can work. I'm happy for now but we might modify this in the future.

PR #2614, opened 2019-09-11
[WIP] Fix nested structure in Serde output

### LINKED ISSUE #2424: Serde is potentially simplifying information twice, creating a bad nested structure
# Describe the bug
I'll attempt to explain this one as best as I can.  Forgive me, there may be a few places where I'm not entirely familiar with the Python language or Serde/PySyft.

I've noticed while writing the Serde parser in Javascript a few strange places where Serde in PySyft appears to unnecessarily nest information.  I think it would be in the best interest if PySyft cleaned up the Serde simplify functions to get a cleaner, shorter, more consistent exported value.  I know that myself and @mccorby would greatly benefit from some de-duplication in the data structure being produced, and it would cause information being sent over the wire to be significantly smaller.

I want to be clear: **I'm not familiar with all of the places where Serde is duplicating information... this issue is merely my initial understanding of where duplication may be taking place.**  Grain of salt.  😉 

## To Reproduce
```python
import torch as th
import syft as sy
sy.create_sandbox(globals(), download_data=False, verbose=False)
```

```python
@sy.func2plan(verbose=False)
def plan_double_abs(x):
    x = x + x
    x = th.abs(x)
    return x

x = th.tensor([1,2,3]).send(bob)

sy.serde._simplify(x.child)
sy.local_worker.verbose=False
plan_double_abs.build(th.tensor([1., -2.]))

sy.serde._simplify(plan_double_abs)
```

The above will produce the following output:

```python
(18,
 (((6,
    (1,
     (6,
      ((6,
        ((5, (b'__add__',)),
         (19,
          (23885703668, 30300883787, 85156589176, None, (13, (2,)), False)),
         (6,
          ((19,
            (23885703668,
             30300883787,
             85156589176,
             None,
             (13, (2,)),
             False)),)),
         (0, ()))),
       (6, (53361601662,)))))),
   (6,
    (1,
     (6,
      ((6,
        ((5, (b'torch.abs',)),
         None,
         (6,
          ((19, (50671613206, 53361601662, 85156589176, None, None, True)),)),
         (0, ()))),
       (6, (68554228008,)))))),
   (6, (9, 53361601662))),
  85156589176,
  (1, (30300883787,)),
  (6, (68554228008,)),
  (5, (b'plan_double_abs',)),
  None,
  None,
  True))
```

To reproduce this exact object in Javascript I must write the following code:

```js
const firstPointerTensor = new PointerTensor(
  23885703668,
  30300883787,
  85156589176,
  null,
  new TorchSize([2]),
  false
);

const secondPointerTensor = new PointerTensor(
  50671613206,
  53361601662,
  85156589176,
  null,
  null,
  true
);

const operations = [
  Tuple(
    1,
    Tuple(
      Tuple(
        '__add__',
        firstPointerTensor,
        Tuple(firstPointerTensor),
        new Map()
      ),
      Tuple(53361601662)
    )
  ),
  Tuple(
    1,
    Tuple(
      Tuple('torch.abs', null, Tuple(secondPointerTensor), new Map()),
      Tuple(68554228008)
    )
  ),
  Tuple(9, 53361601662)
];

const plan = new Plan(
  operations,
  85156589176,
  [30300883787],
  Tuple(68554228008),
  'plan_double_abs',
  null,
  null,
  true
);
```

## Expected behavior
You'll see in the above Javascript code that the majority of the duplication comes in the `operations` variable.

1. We have multiple nested Tuples, most of which I feel could probably be eliminated.
2. We also see in the first operation two references to the same `PointerTensor`: `firstPointerTensor`, the second of which is contained within a `Tuple` for some reason. The second operation, for whatever reason, does not have the duplicate reference.
3. Each `PointerTensor` has an empty `Map` (a Dictionary in Python) as the final argument.  This may serve a purpose, but I'm not aware of one.

These are the duplications that I've been able to suss out.  Either way, this issue, combined with my other two issues on Serde referenced below, are crucial to having a common language between Python and Javascript (and Kotlin and other languages that will be ported in the future).

## Desktop
 - OS: MacOS
 - Version 10.14.6

## Additional context
I've written about the addition of unnecessary commas in another issue (https://github.com/OpenMined/PySyft/issues/2423) as well as the use of parens over square brackets in another issue (https://github.com/OpenMined/PySyft/issues/2422)

kakirastern:
Fixes #2424. 

To try to either get rid of duplicated PySyft `serde` output details expressed in the form of a nested structure, or to simplify the `serde` output by supplying some "globally verbose=False" option to turn off the printout of the unnecessary info. 

The problem is as laid out in https://github.com/OpenMined/PySyft/issues/2424#issue-475676662.


kakirastern:
And it does indeed look like it double simplification is intended, as in "simplified (bool): in some cases we want to pass in data which has already been simplified - in which case we must skip double simplification - which would be bad.... so bad... so... so bad" in [L246 of serde.py](https://github.com/OpenMined/PySyft/blob/dev/syft/serde/serde.py#L246). So the current focus is on simplifying the output and not the `simplify` process in `serde`. Will need time to study the "data structures" used in PySyft a bit more before such a goal can be achieved.

jvmancuso:
Hi @kakirastern, any update on this?

kakirastern:
Hi @jvmancuso. Sorry about the delay in updating this PR. Have been meaning to ask someone regarding the different options I have to move forward, but didn't know who to ask. I think I changed the code base a bit to attempt a fix on a related issue first, but in the process broke many tests. Do you think it is a good idea to carry on with this PR by fixing these tests? Or, is the breaking of such tests at this stage any indication of something undesirable, meaning that I should approach the original issue a little differently? Any suggestions?

jvmancuso:
it would be great to know of examples where double serialization is happening, as it seems like that's not the case everywhere -- do you have any of those? It might point to a different bug in the codebase that would be more useful to fix, and it would also be a good idea to update the PR as it's gotten stale and the issue might have been fixed elsewhere in the meantime.

iamtrask:
Closing this PR due to inactivity. Definitely interested in solving this issue though. Please re-submit at the appropriate time.

PR #2617, opened 2019-09-17
Read version number from version.py for use in setup.py

midokura-silvia:
Version upgrade now possible with change in a single location. No longer in version.py and setup.py.


robert-wagner:
This is done automatically

PR #2627, opened 2019-09-25
Refactor Plan

LaRiffle:
This PR aims to make Plans much simpler:
- When sent, now the reference is kept through a PlanPointer systematically
- The Plan class uses sub-objects classes State and Procedure to handle small operations

**New features**
- Add a clone functionality
- Improve inplace fix_prec_ and share_
- Make PlanPointer work with single and multiple locations

TODO:
- [x] remove keys and change the fun2plan state dict -> list
- [x] improve the .send() by maybe making a global clone of the plan instead of repairing the stuff after it is sent (not practical, instead I simplified the build method)


iamtrask:
Should this have WIP in the title?

syft/generic/frameworks/types.py
```diff
@@ -19,10 +20,18 @@
     framework_tensors.append(torch.nn.Parameter)
     framework_shapes.append(torch.Size)
 
+    framework_layer_module = torch.nn.Module
```
jvmancuso: Well, this looks interesting :D

syft/messaging/plan.py
```diff
@@ -3,61 +3,50 @@
 from typing import List
 from typing import Tuple
 from typing import Union
+from typing import Dict
 
 import torch
```
jvmancuso: Since when is this file torch-specific? Maybe I missed this during my refactors.  It shouldn't be outside of the frameworks/torch folder if that's the case -- but since Plans should be general in theory, we'll need to rethink this in the future

syft/messaging/plan.py
```diff
@@ -104,111 +95,256 @@ def append(self, key):
...
-            delattr(self.plan, key)
+        assert list(self.state_ids) == list(state_dict.keys())
 
-        for key in dict_state.keys():
-            setattr(self.plan, key, dict_state[key])
+        for state_id, new_tensor in state_dict.items():
+            tensor = self.owner.get_obj(state_id)
 
-        self.keys = list(dict_state.keys())
+            with torch.no_grad():
```
jvmancuso: This is torch-specific, and I don't think it should need to be here. Seems like we're in need of a better abstraction if we're going to remain framework agnostic here

syft/messaging/plan.py
```diff
@@ -220,50 +356,42 @@ class Plan(ObjectStorage, torch.nn.Module):
...
-        self.plan = list()
-        self.readable_plan = readable_plan if readable_plan is not None else []
-        self.state = State(self)
+        self.name = name or self.__class__.__name__
+        self.owner = owner
+
+        # Info about the plan stored vua the state and the procedure
+        self.procedure = procedure or Procedure(readable_plan, arg_ids, result_ids)
+        self.state = state or State(owner=owner, plan=self, state_ids=state_ids)
+        if state_dict is not None:
```
jvmancuso: this also feels pretty torch-specific

syft/workers/base.py
```diff
@@ -809,30 +817,31 @@ def fetch_plan(
...
-                    state_elem = state_ptr.copy().get()
-                else:
-                    state_elem = self.request_obj(state_id, location)
-                self.register_obj(state_elem)
-                state_ids.append(state_elem.id)
-            plan.replace_ids(plan.state_ids, state_ids)
-            plan.state_ids = state_ids
+        plan.procedure.update_worker_ids(location.id, self.id)
+
+        # Might be useless now
```
jvmancuso: should this be removed?

syft/messaging/plan.py
```diff
@@ -104,111 +95,256 @@ def append(self, key):
...
+            if tensor.grad is not None:
+                tensor.grad -= tensor.grad
+
+    def send_for_build(self, location, **kwargs):
+        """
+        Send functionality that can only be used when sending the state for
+        building the plan. Other than this, you shouldn't need to send the
+        state separately.
+        """
+        assert location == self.plan  # ensure this is a send for the build
```
Jasopaum: I didn't follow everything but it looks a bit weird to compare a plan and a location

syft/messaging/plan.py
```diff
@@ -220,50 +356,42 @@ class Plan(ObjectStorage, torch.nn.Module):
...
-        self.verbose = verbose
-
-        # Info about the plan stored
-        self.plan = list()
-        self.readable_plan = readable_plan if readable_plan is not None else []
-        self.state = State(self)
+        self.name = name or self.__class__.__name__
+        self.owner = owner
+
+        # Info about the plan stored vua the state and the procedure
```
Jasopaum: Typo: "via"

syft/messaging/plan.py
```diff
@@ -293,19 +418,17 @@ def _create_placeholders(self, args_shape):
...
 
-    # Override property of nn.Module
+    # For backward compatibility
     @property
-    def location(nn_self):
-        raise AttributeError("Plan has no attribute location")
+    def readable_plan(self):
+        return self.procedure.commands
+
+    def parameters(self):
```
Jasopaum: It might be a bit confusing to call this `parameters()` if it only returns the state tensors and not the parameters given when calling the plan

syft/messaging/plan.py
```diff
@@ -363,190 +482,82 @@ def build(self, *args):
...
         else:
-            res_ptr = self.forward(*local_args)
-        res_ptr.child.garbage_collect_data = False
+            res_ptr = self.forward(*build_args)
 
-        # We put back the original state. There is no need to keep a reference
-        # To the element sent to the plan.
-        self.state.set_(copied_state)
+        # We put back a copy of the original state
+        self.state.set_(cloned_state)
```
Jasopaum: Would a tensor method `.send_copy()` that does not remove the tensor from the original location make things cleaner here (no need to copy state and set it back)? Could it be useful somewhere else?

syft/messaging/plan.py
```diff
@@ -363,190 +482,82 @@ def build(self, *args):
...
+        # so we update the procedure
```
Jasopaum: What's before the so?

syft/messaging/plan.py
```diff
@@ -363,190 +482,82 @@ def build(self, *args):
...
+        self.procedure.update_ids(from_ids=build_state_ids, to_ids=self.state.state_ids)
 
         # The readable plan is now built, we hide the fact that it was run on
```
Jasopaum: The term readable plan has been replaced

syft/messaging/plan.py
```diff
@@ -363,190 +482,82 @@ def build(self, *args):
...
-    def _execute_readable_plan(self, *args):
-        # TODO: for now only one value is returned from a plan
-        result_ids = [sy.ID_PROVIDER.pop()]
+        object.__setattr__(self, name, value)
 
-        plan_res = self.execute_plan(args, result_ids)
-
-        return plan_res
+        if isinstance(value, FrameworkTensor):
+            self.state.state_ids.append(value.id)
```
Jasopaum: Is it not needed to add `name` in `self.state.keys`?

syft/messaging/plan.py
```diff
@@ -104,111 +95,256 @@ def append(self, key):
...
-                for param in ptr.parameters():
-                    id_at_location.append(param.id_at_location)
-            else:
-                id_at_location.append(ptr.id_at_location)
-
-        return id_at_location
+        tensors = []
+        for state_id in self.state_ids:
+            tensor = self.owner.get_obj(state_id)
+            assert tensor.id == state_id
```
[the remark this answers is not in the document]
LaRiffle: Yes usually yes, but because we use .clone(), and because of how set_() is done, we need to make sure it is still true. Actually I had a bug because this rule was violated^^
But we will remove it when the code is stable

syft/messaging/plan.py
LaRiffle: Ok I'll change it to id comparison
Actually plans and workers are kinds of cousins: when you build a plan you send tensors and commands to the plans, and he stores the commands (=they form the "readable_plan") but doesn't run them.
Look some methods like _recv_msg are held by both

syft/generic/pointers/pointer_plan.py
```diff
@@ -80,7 +119,11 @@ def request_execute_plan(
         # )
         args = [args, response_ids]
 
-        command = ("execute_plan", self.id_at_location, args, kwargs)
+        id_at_location = None
+        for loc, id_at_loc in zip(self._locations, self._ids_at_location):
+            if loc == location:
+                id_at_location = id_at_loc
```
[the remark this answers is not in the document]
LaRiffle: Sure!

syft/generic/pointers/pointer_plan.py
```diff
@@ -80,7 +119,11 @@ def request_execute_plan(
...
+        command = ("run", id_at_location, args, kwargs)
```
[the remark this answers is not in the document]
LaRiffle: Good point let's do this!

syft/messaging/plan.py
```diff
@@ -3,61 +3,50 @@
...
         self.args_shape = args_shape
-        self.verbose = verbose
-        if state is not None:
-            self.state = state
-            self.include_state = True
-        else:
-            self.state = {}
-            self.include_state = False
+        self.state_dict = state or {}
+        self.include_state = state is not None
```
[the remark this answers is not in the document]
LaRiffle: We do, include_state actually means: is my initial plan a function or a class: and if it's a function I need to provide the state in the args, so to "include state". i'll make this clearer

syft/messaging/plan.py
LaRiffle: will add a comment: I'm trying here to match the torch api of nn.Module where .parameters() return the model parameters

syft/generic/pointers/pointer_plan.py
```diff
@@ -80,7 +119,11 @@ def request_execute_plan(
         # )
```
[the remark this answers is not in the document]
LaRiffle: sure let's do it

syft/messaging/plan.py
```diff
@@ -220,50 +356,42 @@ class Plan(ObjectStorage, torch.nn.Module):
...
         result_ids: List[Union[str, int]] = None,
         readable_plan: List = None,
         blueprint=None,
-        state=None,
-        include_state: bool = False,
-        is_built: bool = False,
-        verbose: bool = False,
-        *args,
-        **kwargs,
+        state_dict=None,
```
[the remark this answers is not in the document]
LaRiffle: done

syft/messaging/plan.py
```diff
@@ -104,111 +95,256 @@ def append(self, key):
                 "Example: if you have self.elem1, call self.add_to_state('elem1')."
             )
         else:
-            t = getattr(self.plan, key)
-            if self.is_valid_type(t):
+            obj = getattr(self.plan, key)
+            if isinstance(obj, (FrameworkTensor, FrameworkLayerModule)):
                 self.keys.add(key)
```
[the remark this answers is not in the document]
LaRiffle: No, because the setattr called implicitely when defining your class already handled this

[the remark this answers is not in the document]
LaRiffle: Because this is only used in classes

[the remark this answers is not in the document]
LaRiffle: will make it clear

syft/messaging/plan.py
```diff
@@ -220,50 +356,42 @@ class Plan(ObjectStorage, torch.nn.Module):
...
+        # General kwargs
+        id: Union[str, int] = None,
+        owner: "sy.workers.BaseWorker" = None,
+        tags: List[str] = None,
+        description: str = None,
     ):
+        owner = owner or sy.local_worker
+        AbstractObject.__init__(self, id, owner, tags, description, child=None)
         ObjectStorage.__init__(self)
         torch.nn.Module.__init__(self)
```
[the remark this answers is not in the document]
LaRiffle: Well spotted

syft/messaging/plan.py
LaRiffle: Yes maybe... 
But I would need to clone the whole plan to do so. maybe it would be more elegant indeed, I need to think about it

syft/messaging/plan.py
LaRiffle: indeed

syft/messaging/plan.py
LaRiffle: No, keys are only really used for func2plan for people to create state elements and getting them through a nice name, but you're right I will remove them to transform the dict to a simple list like in https://pytorch.org/tutorials/beginner/examples_autograd/two_layer_net_custom_function.html

syft/workers/base.py
LaRiffle: removed! well spotted

syft/messaging/plan.py
LaRiffle: Should not be torch specific, but indeed was still depending on torch a bit, I'm working to get it completely indenpenedent

syft/messaging/plan.py
LaRiffle: Any suggestion here?
What would you write if tf to do this?

syft/messaging/plan.py
LaRiffle: how so?

syft/messaging/plan.py
Jasopaum: Oh yes, I had missed that part where we removed owner from the plan, etc.!
How will that work when we use distributed workers for real? Where will plans be?

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -631,22 +642,28 @@ def float_prec_(self):
...
         else:
             del self.child
             self.set_(tensor)
+            self.is_wrapper = False
         return self
 
     float_precision_ = float_prec_
 
-    def fix_prec(self, *args, storage="auto", field_type="int100", **kwargs):
+    def fix_prec(self, *args, storage="auto", field_type="int100", no_wrap: bool = False, **kwargs):
```
midokura-silvia: could you add a docstring to document the parameters?

test/message/test_plan.py
```diff
@@ -530,6 +554,65 @@ def forward(self, x):
     hook.local_worker.is_client_worker = True
 
 
+def test_binding_fix_precision_plan(hook):
+    """Here we make sure the attributes of a plan are still binded to state elements when calling fix_precision"""
```
midokura-silvia: binded => bound

test/message/test_plan.py
```diff
@@ -530,6 +554,65 @@ def forward(self, x):
...
+    weight_id = plan.fc1.weight.id
+    hook.local_worker.get_obj(weight_id).float_prec_()
+
+    assert (plan.fc1.weight - original_weight) < 10e-2
+
+    hook.local_worker.is_client_worker = True
+
+
+def test_binding_encrypted_plan(hook, workers):
+    """Here we make sure the attributes of a plan are still binded to state elements when calling fix_prec + share"""
```
midokura-silvia: bound

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: Sure!

syft/messaging/plan.py
LaRiffle: Same thing, nothing changes: at building (which always happens locally) you alway send evrything to the plan just like if it was a Virtual worker, then you can run a simulated computation, and that plan/worker stores all commands, which gonna make the readable_plan.
This readable_plan is included in the state attributes and is then sent to real remote workers for remote execution

syft/messaging/plan.py
```diff
@@ -74,210 +65,330 @@ class State(object):
     sure they are provided to remote workers who are sent the Plan.
     """
 
-    def __init__(self, plan):
-        self.keys = set()
+    def __init__(self, owner, plan=None, keys=None, state_ids=None):
+        self.owner = owner
         self.plan = plan
+        self.keys = keys or set()
```
Jasopaum: Are keys still used? If so, what is the difference between them and state_ids?

syft/messaging/plan.py
```diff
@@ -1,63 +1,54 @@
...
         self.args_shape = args_shape
-        self.verbose = verbose
-        if state is not None:
-            self.state = state
-            self.include_state = True
-        else:
-            self.state = {}
-            self.include_state = False
+        self.state_tensors = state or tuple()
+        # include_state is used to distinguish if the initial plan a function or a class:
```
Jasopaum: Typo: an "is" is missing

syft/messaging/plan.py
LaRiffle: Nope I forgot to remove the last occurences, thank you :)

syft/generic/frameworks/types.py
jvmancuso: There is now a tf.Module that layers inherit from

syft/messaging/plan.py
jvmancuso: There's not really a tf equivalent for torch.no_grad that I'm aware of. The new GradientTape is used to turn autograd on, which is a clear departure from torch.

It seems like we might be in a situation where Plans require a shared base class that the different frameworks extend. Plans are kind of a first class citizen in TF because of tf.function. This probably requires more than a PR discussion so let's defer it to later

LaRiffle: Like a generic Tensor which can be implemented by PyTorch & TF ?

syft/generic/frameworks/types.py
LaRiffle: Perfect

syft/messaging/plan.py
jvmancuso: We kind of already have that in TorchTensor & TensorFlowTensor -- I'm referring to a generic Plan

syft/messaging/plan.py
jvmancuso: iirc, `state_dict` is how torch returns stateful objects within a container (e.g. variable tensors from an optimizer or nn.Module)

syft/generic/frameworks/types.py
jvmancuso: Only concern is that torch.nn.Modules have quite a bit more functionality by default -- tf.Module is literally just a nested collection of variables & submodules with a name/namescope

syft/messaging/plan.py
LaRiffle: The Plan is already 99% generic
But it makes a few assumptions on Tensors / Modules

syft/messaging/plan.py
LaRiffle: That code is outdated; now is used a tuple

syft/generic/frameworks/types.py
LaRiffle: The only thing I ask from a Module is the ability to return the list of pairs ('name_param', param), do you have this in TF?

syft/messaging/plan.py
jvmancuso: I'm talking about a 100% generic Plan that torch/tf can both extend as needed to handle their different interfaces. This is in line with how we've handled other core abstractions, e.g. `FrameworkHook`, `FrameworkAttributes`, the pluggable `hook_args`, `NativeTensor`, etc.

jvmancuso: IMO these core abstractions should be kept as modular as possible, which means deciding on a generic interface and letting implementations extend it as needed

LaRiffle: I know, I'm working in that direction too :)
But I'm not sure what to restrict to PyTorch, as 99% is already generic

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -631,22 +633,38 @@ def float_prec_(self):
...
+        """
+        Convert a tensor or syft tensor to fixed precision
+
+        Args:
+            args (tuple): args to transmit to the fixed precision tensor
```
robert-wagner: nit could you denote these as *args and **kwargs in the docstring so it is clear they are of arbitrary length

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -689,13 +706,21 @@ def fix_prec(self, *args, storage="auto", field_type="int100", **kwargs):
                     "do not provide internal_type if data does not need LargePrecisionTensor to be stored"
                 )
                 del kwargs["internal_type"]
-            return syft.FixedPrecisionTensor(*args, **kwargs).on(self).enc_fix_prec()
+            fpt_tensor = syft.FixedPrecisionTensor(*args, **kwargs)
+            fpt_tensor.child = self
+            fpt_tensor = fpt_tensor.fix_precision()
```
robert-wagner: Is there a better way to do this? This seems like it could be a one liner

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -689,13 +706,21 @@ def fix_prec(self, *args, storage="auto", field_type="int100", **kwargs):
...
+
+        if not no_wrap:
+            fpt_tensor = fpt_tensor.wrap()
+
+        return fpt_tensor
 
     fix_precision = fix_prec
 
     def fix_prec_(self, *args, **kwargs):
```
robert-wagner: Could we add a docstring for this?

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -742,27 +765,37 @@ def share(
                 syft.AdditiveSharingTensor(
                     field=field, crypto_provider=crypto_provider, owner=self.owner
                 )
-                .on(self.copy())
-                .child.init_shares(*owners)
+                .on(self.copy(), wrap=False)
```
robert-wagner: While we're here, have you checked to see if this pattern is anywhere else in the code? it seems like a good time to fix them all

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -740,16 +740,27 @@ def handle_func_command(cls, command):
         return response
 
     def share(self, *owners, field=None, crypto_provider=None):
+        """
+        Forward the .share() command to the child tensor, and reconstruct a new
+        FixedPrecisionTensor since the command is not inplace and should return
+        a new chain
```
robert-wagner: Since we are adding a docstring can we describe the parameters

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -740,16 +740,27 @@ def handle_func_command(cls, command):
...
                 must be the same as the one of the original tensor"
-        self.child = self.child.share(
+        tensor = FixedPrecisionTensor(owner=self.owner, **self.get_class_attributes())
+
+        tensor.child = self.child.share(
             *owners, field=field, crypto_provider=crypto_provider, no_wrap=True
         )
+        return tensor
+
+    def share_(self, *args, **kwargs):
```
robert-wagner: Same as above for the docstring

syft/generic/object.py
```diff
@@ -35,10 +35,7 @@ def __init__(
                 a chain of tensors
         """
         self.owner = owner
-        if id is None:
-            self.id = sy.ID_PROVIDER.pop()
-        else:
-            self.id = id
+        self.id = id or sy.ID_PROVIDER.pop()
```
robert-wagner: could we make this a bit more verbose to be `id if id else sy.ID_PROVIDER.pop()` to me that is much more clear than or but still is a 1 liner and should be the same speed

syft/generic/pointers/pointer_plan.py
```diff
@@ -43,18 +47,53 @@ def __init__(
...
+        else:
+            self._locations = new_location
+
+    @property
+    def id_at_location(self):
+        n_ids = len(self._ids_at_location)
+        if n_ids != 1:
+            return self._ids_at_location
+        else:
+            return self._ids_at_location[0]
```
robert-wagner: For all of these would it cause problems if they all returned a list of length 1 and then we just set up the rest of the code to be able to handle that? Throwing this out there, not actually sold on it yet

syft/generic/pointers/pointer_plan.py
```diff
@@ -43,18 +47,53 @@ def __init__(
...
+
+    @id_at_location.setter
+    def id_at_location(self, new_id_at_location):
+        if isinstance(new_id_at_location, (int, str)):
```
robert-wagner: since we're already doing a type check, could we throw an error if it is not an accepted type?

syft/generic/pointers/pointer_plan.py
```diff
@@ -43,18 +47,53 @@ def __init__(
...
         remote plan
         """
-        result_ids = [sy.ID_PROVIDER.pop()]
+        if len(self._locations) == 1:
+            location = self.location
+        else:
+            location = args[0].location
 
-        response = self.request_execute_plan(self.location, result_ids, *args)
+        result_ids = [sy.ID_PROVIDER.pop()]
```
robert-wagner: since this is plural should pop happen more than once?

syft/generic/pointers/pointer_plan.py
```diff
@@ -131,3 +181,37 @@ def detail(worker: AbstractWorker, tensor_tuple: tuple) -> "PointerPlan":
 
     def wrap(self):
         return self
+
+    def __str__(self):
+        """Returns a string version of this pointer.
```
robert-wagner: since we are adding a docstring could we add a human readable def of what we expect

syft/messaging/protocol.py
```diff
@@ -80,8 +80,8 @@ def deploy(self, *workers):
 
         self._resolve_workers(workers)
 
-        for worker, plan in self.plans:
-            plan.send(worker)
+        for i, (worker, plan) in enumerate(self.plans):
```
robert-wagner: could we do this as list comprehension instead eg `self.plans = [(worker, plan.send(worker)) for (worker, plan) in self.plans]` (Im not 100% sure this works)

syft/serde/native_serde.py
```diff
@@ -66,12 +66,11 @@ def _detail_collection_list(worker: AbstractWorker, my_collection: Tuple) -> Col
 
     # Step 1: deserialize each part of the collection
     for part in my_collection:
+        detailed = serde._detail(worker, part)
         try:
-            pieces.append(
-                serde._detail(worker, part).decode("utf-8")
-            )  # transform bytes back to string
+            pieces.append(detailed.decode("utf-8"))  # transform bytes back to string
```
robert-wagner: How often do we expect this to fail? if it is often this should be an if statement rather than a try execpt

syft/workers/base.py
```diff
@@ -356,7 +356,7 @@ def send(
         worker = self.get_worker(worker)
 
         if hasattr(obj, "create_pointer") and not isinstance(
-            obj, (sy.Plan, sy.Protocol)
+            obj, (sy.Protocol,)
```
robert-wagner: could you make this not a tuple if you are not passing multiple values

syft/workers/base.py
```diff
@@ -556,6 +558,12 @@ def register_obj(self, obj: object, obj_id: Union[str, int] = None):
         if not self.is_client_worker:
             super().register_obj(obj, obj_id=obj_id)
 
+    def de_register_obj(self, obj: object, obj_id: Union[str, int] = None):
+        """De-registers the specified object with the current worker node.
```
robert-wagner: could you add expected parameters

syft/messaging/plan.py
```diff
@@ -74,210 +65,326 @@ class State(object):
...
+                command[i] = to_ids[from_ids.index(item)]
+            elif isinstance(item, (list, tuple)):
+                command[i] = Command.replace_ids(command=item, from_ids=from_ids, to_ids=to_ids)
+        return type_obj(command)
 
-            if self.is_missing_grad(t):
-                self.create_grad_objects(t)
 
-            setattr(self.plan, key, t.fix_precision(*args, **kwargs))
+class Procedure(object):
```
robert-wagner: Can procedure also become its own file?

syft/messaging/plan.py
```diff
@@ -74,210 +65,326 @@ class State(object):
...
-        # Info about the plan stored
-        self.plan = list()
-        self.readable_plan = readable_plan if readable_plan is not None else []
-        self.state = State(self)
+        self.name = name or self.__class__.__name__
+        self.owner = owner
+
+        # Info about the plan stored via the state and the procedure
+        self.procedure = procedure or Procedure(readable_plan, arg_ids, result_ids)
+        self.state = state or State(owner=owner, plan=self, state_ids=state_ids)
```
robert-wagner: Could we switch this up to not be or it is unclear on reading?

syft/messaging/plan.py
```diff
@@ -74,210 +65,326 @@ class State(object):
...
-
-        # Info about the plan stored
-        self.plan = list()
-        self.readable_plan = readable_plan if readable_plan is not None else []
-        self.state = State(self)
+        self.name = name or self.__class__.__name__
+        self.owner = owner
+
+        # Info about the plan stored via the state and the procedure
+        self.procedure = procedure or Procedure(readable_plan, arg_ids, result_ids)
```
robert-wagner: Same as below

syft/messaging/plan.py
```diff
@@ -74,210 +65,326 @@ class State(object):
...
-        self.id = sy.ID_PROVIDER.pop() if id is None else id
-        self.name = self.__class__.__name__ if name == "" else name
-        self._owner = sy.local_worker if owner is None else owner
-        self.verbose = verbose
-
-        # Info about the plan stored
-        self.plan = list()
-        self.readable_plan = readable_plan if readable_plan is not None else []
-        self.state = State(self)
+        self.name = name or self.__class__.__name__
```
robert-wagner: same as below. That being said the more I look at this the more it is growing on me

LaRiffle: I'm in love with this "new" syntax :D

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: Looks good!

syft/messaging/plan.py
jvmancuso: This method `set_` seems like a good candidate, as it's called `assign` in TF and would normally be reserved for single variables. In TF this operation would probably be left to the user to perform manually. Another is just below this, `create_grad_if_missing`. Another is referred to in [this comment](https://github.com/OpenMined/PySyft/pull/2627/files#r330910284) -- this is precisely what we'd like to avoid in a generic Plan, since the API should be agnostic to which framework is using/extending it.

In general, it just seems useful to keep in mind that we'll have the following correspondences on our end: `Plan` == `tf.function`, and `Protocol` == `tfe.Protocol`. If we try to maintain parity for both frameworks in a single class, we'll end up handicapping the class, so I think we can avoid that by extending a TorchPlan from this Plan in the future.  It's probably out of scope for this PR, but something worth keeping in mind.

syft/messaging/plan.py
jvmancuso: +1 to the `x or X()` syntax, this is pretty pythonic imo

syft/generic/pointers/pointer_plan.py
LaRiffle: The thing is that I wan the main usecase (as a single pointer) to be unchanged, while enabling another behaviour as a multi pointer

syft/generic/pointers/pointer_plan.py
LaRiffle: I'll change the type check to be simply against the args being a list/tuple or not, it will prevent making assumption about the format of an id

syft/generic/pointers/pointer_plan.py
LaRiffle: Good point, theoretically this is here to allow multiple ids in the result ids but in 95% of the case you only have one. We have a fallback mechanism for simple operations to populate this list when there are more than 1 object to return, but currently this hasn't been studied for plans. By default we always use only one, because we build pointers for each id and we don't want stale pointer

syft/serde/native_serde.py
LaRiffle: I corrected it all over the file

syft/messaging/plan.py
```diff
@@ -74,210 +65,326 @@ class State(object):
     sure they are provided to remote workers who are sent the Plan.
     """
 
-    def __init__(self, plan):
-        self.keys = set()
+    def __init__(self, owner, plan=None, state_ids=None):
```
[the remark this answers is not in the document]
LaRiffle: Yes I made a plan module with a plan.py / state.py / procedure.py

syft/messaging/plan.py
```diff
@@ -74,210 +65,326 @@ class State(object):
...
+        Perform a check of consistency on the elements ids.
+        """
+        tensors = []
+        for state_id in self.state_ids:
+            tensor = self.owner.get_obj(state_id)
+            assert tensor.id == state_id
+            tensors.append(tensor)
+        return tensors
+
+    def clone_state(self) -> Dict:
```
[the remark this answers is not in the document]
LaRiffle: let's call it clone_state_dict then!

syft/messaging/plan.py
```diff
@@ -74,210 +65,326 @@ class State(object):
...
-            if isinstance(ptr, torch.nn.Module):
-                for param in ptr.parameters():
-                    id_at_location.append(param.id_at_location)
-            else:
-                id_at_location.append(ptr.id_at_location)
+        for tensor in self.tensors():
+            self.create_grad_if_missing(tensor)
+            tensor.send_(location, **kwargs)
 
-        return id_at_location
```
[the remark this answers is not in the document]
LaRiffle: nope it only makes inplace calls to send we don't need any response

syft/messaging/plan.py
```diff
@@ -74,210 +65,326 @@ class State(object):
...
+    def fix_precision_(self, *args, **kwargs):
```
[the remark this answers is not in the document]
LaRiffle: I'd rather not to: until we have a reliable code I would like the user to manually define the appropriate transformations (like send in inputs to the right worker, transform to fixed prec of additive sharing, etc)
Otherwise some bugs might be hard to find!

syft/messaging/plan.py
LaRiffle: Done

syft/generic/frameworks/types.py
jvmancuso: closest thing would be `tf.Module.variables` and `tf.Module.trainable_variables`, but neither of those would get you what you want directly. I'd probably end up doing `named_params = [(v.name, v) for v in module.trainable_variables]` if I needed something similar despite it not being totally equivalent

syft/messaging/plan.py
LaRiffle: I totally agree, however should we do this in this PR or make another one?
Plan were already notoriously not supposed to be in the generic module "as is", so maybe this is a separate topic I would say
@jvmancuso you have been doing amazing job on separating generic from framework specific stuff and you have a vision both on TF & TH, would you be interested to follow up on this?

PR #2630, opened 2019-09-26
Added tutorial for encrypted linear regression and fixed some errors in the LR implementation

andrelmfarias:
With this PR I added a tutorial to show the correct usage of the BloomRegressor (Linear Regression with SMPC).

I also fixed some errors in the computation of standard errors and p-values.


andrelmfarias:
> Maybe you could detail a bit the role of the HBC worker which might not be well understood by other people than the crypto team

I added its formal definition, do you think I should try to explain it a bit more?

Issue #2632, opened 2019-09-27
Make PyTorch completely optional.


jvmncs:
Currently, pytorch-dependent syft code is mostly separate from the core syft code (in that most core syft code does not need to import torch at all), but pytorch isn't yet truly optional.

This issue will be complete once the following have been completed and merged:
- [ ] `pip install syft` no longer installs the `torch` package as a dependency.
- [ ] `import syft` completes successfully without `torch` installed
    - This involves removing all current dependencies on torch outside of the `syft/frameworks/torch` subdirectory (potentially by making that code generic or moving it into that subdirectory when appropriate)
- [ ] `pytest .` does not fail when pytorch is not installed as a dependency (likely with heavy use of the `pytest.skipif(not dependency_check.torch_available)` decorator).

Once this work is completed, we can begin migrating torch-dependent syft code to [PySyft-PyTorch](https://github.com/OpenMined/PySyft-PyTorch).

matthew-mcateer:
"truly optional"? I'm interpreting this as meaning building up the core syft code to be as independent of pytorch as possible, and setting up an optional pytorch backend much like `keras` does with `tensorflow` and `theano, right?

jvmancuso:
Yes, more or less. There is some context missing from this issue, most of it can be found in [this RFC](https://github.com/OpenMined/rfcs/blob/master/20190808-modular-pysyft-frameworks.md), and more specifically [here](https://github.com/OpenMined/rfcs/blob/master/20190808-modular-pysyft-frameworks.md#packaging).

Worth noting that "building up the core syft code to be independent" is already complete.  What's left are the two of these checkboxes, and more specifically the latter because I'm about to fix the former.

jvmancuso:
>Worth noting that "building up the core syft code to be independent" is already complete

This comment of mine is not true -- there was a bug hiding some extra dependencies, so I've updated the top-level comment to reflect the extra work.

PR #2636, opened 2019-10-01
Make installation friendlier for alt framework users

jvmancuso:
Right now, it's really easy to install PySyft if you're following along with the Udacity work & dependencies (PyTorch & TF Encrypted Keras), but it's not so great if you're experimenting with other frameworks (i.e. TensorFlow).  This PR fixes that, but breaks the old pattern of installing syft.

- `pip install syft` no longer installs scikit-learn, torch, or tf-encrypted.  It now installs a less-than-minimal set of dependencies (less than because currently tests will not pass if pytorch is not installed, see #2632).
- To reproduce the current dependency setup, use `pip install syft[udacity]` (or `pip install -r pip-dep/requirements_udacity.txt; python setup.py install`)
- Moved all pip requirements to a `pip-dep/` directory.
- Split torch and tf-encrypted dependencies out into `pip-dep/requirements_udacity.txt`.
- Added a DependencyError exception to the sandbox & `syft.frameworks.torch.linalg` for when people try to load sklearn/scipy without the dependency.


Makefile
```diff
@@ -1,23 +1,27 @@
...
+	. venv/bin/activate; pip install -Ur $(REQ_DIR)/requirements.txt; pip install -Ur $(REQ_DIR)/requirements_dev.txt; pip install -Ur $(REQ_DIR)/requirements_udacity.txt; python setup.py install
 	touch venv/bin/activate
 
 install_hooks: venv
 	venv/bin/pre-commit install
 
 notebook: venv
 	(. venv/bin/activate; \
-		python setup.py install; \
+		python setup.py install udacity; \
```
robert-wagner: I like that udacity is the default (That being said could we make this a parameter which gets passed to the make target since I think I may actually be the only person who uses this target and would like to be able to use other configurations)

Issue #2608, opened 2019-09-10
get_obj should raise a syft exception


mari-linhares:
Currently `worker.get_obj` raises a KeyError exception we should raise a syft specific exception like "ObjectNotFoundException".

This helps to avoid mistaking a regular/native python KeyError exception with exceptions raised by get_obj.

https://github.com/OpenMined/PySyft/blob/750bd2c3987a570ef1dc8554b933696d416724c7/syft/generic/object_storage.py#L84

bhadreshpsavani:
I will work on this issue

PR #2644, opened 2019-10-05
Added custom ObjectNotFoundException class

sukhadj:
This PR is associated with issue #2608. It is able to raise a syft specific exception "ObjectNotFoundException" rather than native python "KeyError".
For this, a customized class `ObjectNotFoundException` - derived from `Exception` class is written.
When object id is not found in object list of, `ObjectNotFoundException` is raised instead of `KeyError`.

P.S.: This is my first PR.


sukhadj:
Closing this PR as Exception class is to be added in syft.exception.

PR #2648, opened 2019-10-08
Fix tutorial part 7

### LINKED ISSUE #2642: Part 07 example has abnormally high loss
**Describe the bug**
In the `/example` folder, Part 07 has a very high loss amount, even after 10 epochs.

**To Reproduce**
1. Run Jupyter Notebook
2. Go to the 7th example and run the notebook all the way through.  The last code blog where you train the model creates an abnormally high loss rate... in the thousands.

**Expected behavior**
I'd expect the loss to be much lower.

**Screenshots**
Screen Shot 2019-10-04 at 11 57 02 AM

**Desktop (please complete the following information):**
 - OS: MacOS 10.14.6

midokura-silvia:
I modified the optimizer to use Adam instead of SGD and the convergence is going nicely. 
Also increased the batch size as the estimation of the linear classifier will be more stable across batches. 
Loss calculation had a broadcasting problem as well.

fixes #2642


examples/tutorials/Part 07 - Federated Learning with Federated Dataset.ipynb
```diff
@@ -106,7 +106,10 @@
     "\n",
     "# Build the FederatedDataset object\n",
     "dataset = sy.FederatedDataset(datasets)\n",
-    "print(dataset.workers)"
+    "print(dataset.workers)\n",
+    "optimizers = {}\n",
+    "for worker in dataset.workers:\n",
+    "    optimizers[worker] = th.optim.Adam(params=model.parameters(),lr=1e-2)"
```
LaRiffle: Why do you have to do this?
You're giving the same references to model parameters to all optimizers so the result should be the same, or not? (I don't know how Adam works)

midokura-silvia: I don't remember exactly the reason, but currently in PySyft having one joint optimizer across locations doesn't work with optimizers that use momentum (including Adam). 
And I switched to Adam as searching on internet I noticed that other people had problems getting the training to converge with SGD as well.

LaRiffle: Oh yes you're right, the pb is with momentum!

PR #2653, opened 2019-10-09
Added Pysyft specific ObjectNotFoundError Exception

sukhadj:
The issue is associated with #2608. 
A new custom `ObjectNotFoundError` is added in `syft.exceptions` which also generates the necessary message when provided with **object id** and **worker object**.
This exception is raised when object id key is not found on given virtual worker.


midokura-silvia:
Hi @sukhadj. Could you please fix the unit test test_virtual.py::test_obj_not_found?

You need to modify lines 204-207: 
```
    try:
        y = x + x
    except KeyError as e:
        assert "If you think this tensor does exist" in str(e)
```
try something as follows:
```
with pytest.raises(ObjectNotFoundError):
    y = x + x
```
The coverage error should go away then as well.

sukhadj:
@midokura-silvia Yes. I will make the necessary changes in the unit test test_virtual.py::test_obj_not_found? Thanks.

PR #2658, opened 2019-10-11
DASH implementation (linear regression for genetics)

andrelmfarias:
Implemented Distributed Association Scan Hammer (DASH) algorithm. This algorithm is used in the context of Linear Regression for genetics. It corresponds to section 4 of [Jonathan Bloom's paper](https://arxiv.org/pdf/1901.09531.pdf).

TODO:

- [x] Implement inversion of upper triangular matrix in SMPC using forward/back substitution: To be discussed, it requires several divisions...
- [x] Implement tests


test/torch/linalg/test_operations.py
```diff
@@ -115,3 +116,20 @@ def test_qr_mpc(hook, workers):
...
+    crypto_prov = workers["james"]
+
+    torch.manual_seed(0)  # Truncation might not always work so we set the random seed
+    n_cols = 3
+    n_rows = 3
+    R = torch.triu(torch.randn([n_rows, n_cols]))
+    invR = R.inverse()
+
+    R_sh = R.fix_precision(precision_fractional=6).share(bob, alice, crypto_provider=crypto_prov)
+    # invR_sh = DASH._inv_upper(R_sh)
```
LaRiffle: Just mentioning as a reminder that the end of the test is commented out.

andrelmfarias: Oh thank you, I had completely forgotten about it!

iamtrask: *bump*

iamtrask: _bump_

syft/frameworks/torch/linalg/lr.py
```diff
@@ -284,3 +285,247 @@ def _compute_pvalues(self):
             self.pvalue_intercept = 2 * t.cdf(-abs(tstat_intercept), self._dgf)
         else:
             self.pvalue_intercept = None
+
+
+class DASH:
+    """
+    Distributed Association Scan Hammer (DASH) algorithm based on Jonathan Bloom's algorithm.
+    It uses Secured Multi-Party Computation at combine phase.
```
LaRiffle: Secured -> Secure

syft/frameworks/torch/linalg/lr.py
```diff
@@ -284,3 +285,247 @@ def _compute_pvalues(self):
...
+    While the training is performed in SMPC, the final regression coefficients
+    are public at the end.
+
+    Reference: Section 2 of https://arxiv.org/abs/1901.09531
+
+    Args:
+        crypto_provider: a BaseWorker providing crypto elements for ASTs such as
```
LaRiffle: ASTs -> AdditiveSharingTensors (used for SMPC)

syft/frameworks/torch/linalg/lr.py
```diff
@@ -284,3 +285,247 @@ def _compute_pvalues(self):
...
+            Beaver triples
+        hbc_worker: The "Honest but Curious" BaseWorker
```
LaRiffle: Can you put a few lines explaining the role ob the hbc worker?

syft/frameworks/torch/linalg/lr.py
```diff
@@ -284,3 +285,247 @@ def _compute_pvalues(self):
...
+        self._check_ptrs(X_ptrs, C_ptrs, y_ptrs)
+
+        # Check if each y is a 2-dim or 1-dim tensor, unsqueeze it if it's 1-dim
+        for i, y in enumerate(y_ptrs):
+            if len(y.shape) < 2:
+                y_ptrs[i] = y.unsqueeze(1)
+
+        self.workers = self._get_workers(X_ptrs)
+
+        # Computing aggregated pairwise dot products remotelly
```
LaRiffle: remotelly -> remotely

syft/frameworks/torch/linalg/lr.py
```diff
@@ -284,3 +285,247 @@ def _compute_pvalues(self):
...
+                )
+
+            # Check if x, c and y are pointers
+            if not (
+                isinstance(x.child, PointerTensor)
+                and isinstance(c.child, PointerTensor)
+                and isinstance(y.child, PointerTensor)
+            ):
+                raise TypeError(
+                    "Some tensors are not pointers, please provided a wrapped Pointer Tensor"
```
LaRiffle: Can you fuse this if statement and the previous together (to make the code shorter)
`all(map(lambda t: t.has_child() and isinstance(t.child, PointerTensor), (x, c, y)))`

syft/frameworks/torch/linalg/lr.py
```diff
@@ -284,3 +285,247 @@ def _compute_pvalues(self):
...
+                )
+
+            # Check if both are in the same worker
+            if not (x.child.location == c.child.location and x.child.location == y.child.location):
```
LaRiffle: I think `(x.child.location == c.child.location == y.child.location)` is allowed

syft/frameworks/torch/linalg/lr.py
```diff
@@ -284,3 +285,247 @@ def _compute_pvalues(self):
...
+
+    @staticmethod
+    def _get_workers(ptrs):
+        """
+        Method that returns the pool of workers in a tuple
+        """
+        workers = set([])
+        for ptr in ptrs:
+            workers.add(ptr.child.location)
+        return tuple(workers)
```
LaRiffle: `return tuple(map(lambda p: p.child.location, ptrs))`

syft/frameworks/torch/linalg/lr.py
```diff
@@ -284,3 +285,247 @@ def _compute_pvalues(self):
...
+        Xy_ptrs = []
+        yy_ptrs = []
+        CX_ptrs = []
+        Cy_ptrs = []
+        for x, c, y in zip(X_ptrs, C_ptrs, y_ptrs):
+            XX_ptrs.append(x.t() @ x)
+            Xy_ptrs.append(x.t() @ y)
+            yy_ptrs.append(y.t() @ y)
+            CX_ptrs.append(c.t() @ x)
+            Cy_ptrs.append(c.t() @ y)
```
LaRiffle: nit, you can avoid doing x.t() & c.t() twice

syft/frameworks/torch/linalg/lr.py
```diff
@@ -284,3 +285,247 @@ def _compute_pvalues(self):
...
+        Performs the inversion of a right upper matrix (2-dim tensor) in MPC by
+        solving the linear equation R * R_inv = I with backward substitution.
+        """
+        R_inv = torch.zeros_like(R)
+        N = R.shape[0]
+
+        # Identity Matrix
+        I = torch.zeros_like(R)
+        for i in range(N):
+            I[i, i] += 1
```
LaRiffle: You can use `torch.eye` I think

syft/frameworks/torch/linalg/lr.py
```diff
@@ -284,3 +285,247 @@ def _compute_pvalues(self):
...
+            X_ptrs, C_ptrs, y_ptrs
+        )
+
+        # Compute remote QR decompositions
+        R_ptrs = self._remote_qr(C_ptrs)
+
+        # Secred share tensors between hbc_worker, crypto_provider and a random worker
+        # and compute aggregates. It corresponds to the Combine stage of DASH's algorithm
+        idx = random.randint(0, len(self.workers) - 1)
+        XX_shared = sum(self._share_ptrs(XX_ptrs, idx)).sum(dim=0)
```
LaRiffle: You doing the sum twice here

syft/frameworks/torch/linalg/lr.py
```diff
@@ -284,3 +285,247 @@ def _compute_pvalues(self):
...
+        Xy_shared = sum(self._share_ptrs(Xy_ptrs, idx))
```
LaRiffle: Why wouldn't you do the sum before sharing the values?

syft/frameworks/torch/linalg/lr.py
```diff
@@ -284,3 +285,247 @@ def _compute_pvalues(self):
...
+        R_inv = self._inv_upper(R)
+
+        Qy = R_inv.t() @ Cy_shared
+        QX = R_inv.t() @ CX_shared
+
+        denominator = XX_shared - (QX ** 2).sum(dim=0)
+        # Need the line below to perform inverse of a number in MPC
+        inv_denominator = ((0 * denominator + 1) / denominator).squeeze()
+
+        coef_shared = (Xy_shared - QX.t() @ Qy).squeeze() * inv_denominator
```
LaRiffle: Can't you spare the line `inv_denominator = ...` by doing
```python
coef_shared = (
 (Xy_shared - QX.t() @ Qy) / denominator
).squeeze()
```

syft/frameworks/torch/linalg/lr.py
```diff
@@ -284,3 +285,247 @@ def _compute_pvalues(self):
...
+        idx = random.randint(0, len(self.workers) - 1)
+        XX_shared = sum(self._share_ptrs(XX_ptrs, idx)).sum(dim=0)
+        Xy_shared = sum(self._share_ptrs(Xy_ptrs, idx))
+        yy_shared = sum(self._share_ptrs(yy_ptrs, idx))
+        CX_shared = sum(self._share_ptrs(CX_ptrs, idx))
+        Cy_shared = sum(self._share_ptrs(Cy_ptrs, idx))
+        R_shared = torch.cat(self._share_ptrs(R_ptrs, idx), dim=0)
+
+        # QR decomposition of R_shared
+        _, R = qr(R_shared, norm_factor=self.total_size ** (1 / 2))
```
LaRiffle: small notation issue: you go from R_shared to R, which could imply R is not shared anymore (same for R_inv)

syft/frameworks/torch/linalg/lr.py
andrelmfarias: I tried to give an explanation of its role in both algorithms here and also used the definition of "Honest bu Curious" I found [here](https://security.stackexchange.com/a/202537).

What do you think?

syft/frameworks/torch/linalg/lr.py
andrelmfarias: The idea here is to build an identity matrix with the same workers, crypto_provider and precision fractional as `R`. 

For that, I would need whether a `torch.eye_like` (which unfortunately doesn't exist) or I am obligated to do like that to have something clean.

syft/frameworks/torch/linalg/lr.py
andrelmfarias: If you look at the [notebook with DASH implementation](https://github.com/jbloom22/DASH/blob/master/dash.ipynb) I added to Bloom's repo at his request, you will see that when we build the XXs matrices privately in the Compress step, we do a sum over the rows and at the Combine step we sum the XXs.
Here I am just doing the commutated operation, first I sum the XXs, then I sum the result over the row.
So the sum here is not redundant, it's needed.

syft/frameworks/torch/linalg/lr.py
andrelmfarias: Because I need all the values in the same 3 workers for the SecureNN 3-party computation (crypto_provider, hbc_worker and a random worker from the pool).

Before `self._share_ptrs` the tensors in the `XX_ptrs` list are not secret shared with the same workers.

syft/frameworks/torch/linalg/lr.py
andrelmfarias: Finally, I decided to do it in the same order as in the notebook implementation, so now the sum over the axis 0 is being done at the call of the method `_remote_dot_products`. I have double-checked and the results are the same, but it's better to be in line with the notebook implem.

syft/frameworks/torch/linalg/lr.py
andrelmfarias: if you look at the computations of `coef_shared` and `sigma2_shared` I need to do 2 divisions by the same denominator.

By computing `inv_denominator` and using multiplication instead of division to compute `coef_shared` and `sigma2_shared` I optimize the time execution by having only one division overall.

syft/frameworks/torch/linalg/lr.py
LaRiffle: Ok all good in that case

syft/frameworks/torch/linalg/lr.py
LaRiffle: 👍

syft/frameworks/torch/linalg/lr.py
LaRiffle: Oh yes alright, thanks for pointing this out!

syft/frameworks/torch/linalg/lr.py
LaRiffle: That's right, my bad!

PR #2659, opened 2019-10-11
Add support for exp and log (with approx) on additive shared tensors

### LINKED ISSUE #2537: Exponential implementation in SMPC
**Suggestion to implement exp in SMPC**

The interest of this is to allow computation of functions which derive of exponentials like sigmoid.

Let's say you share the tensor x in x0 and x1. You want to compute exp(x).
But exp(x) = exp(x0 + x1) = exp(x0) * exp(x1).
So you can compute locally exp(x0) and exp(x1), and **additively share** these values, perform an encrypted mul and then you obtain shares of exp(x0) * exp(x1) = exp(x).

LaRiffle:
This PR extends the set of operations supported in Additive Sharing using approximations

- [x] Exponential 
- [x] Logarithm
- [x] Inverse
- [x] Sigmoid

Fix #2537


syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -182,8 +183,8 @@ def __iadd__(self, other):
     def sub(self, _self, other):
         """Subtracts a fixed precision tensor from another one.
         """
-        if isinstance(other, int):
-            scaled_int = other * self.base ** self.precision_fractional
+        if isinstance(other, (int, float)):
+            scaled_int = int(other * self.base ** self.precision_fractional)
```
robert-wagner: Should this be an int or a long?

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -443,6 +447,70 @@ def matmul(self, *args, **kwargs):
     __matmul__ = matmul
     mm = matmul
 
+    # Approximations:
+    def exp(self, iterations=8):
```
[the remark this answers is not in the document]
robert-wagner: Same with both below functions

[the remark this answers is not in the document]
jvmancuso: These should also have citations, since the inspiration comes from CrypTen. I'd say it's not substantial enough to require that the license be copied, so a simple online comment + link should do

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -202,6 +203,9 @@ def sub(self, _self, other):
 
     __sub__ = sub
 
+    def __rsub__(self, other):
```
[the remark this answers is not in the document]
LaRiffle: Not exactly because `__rsub__(self, other)` is called after `other - self` failed, typically because other was an integer. So we need to reverse arguments.

syft/frameworks/torch/tensors/interpreters/precision.py
LaRiffle: float can be supported too

syft/frameworks/torch/tensors/interpreters/precision.py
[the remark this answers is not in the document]
LaRiffle: Indeed, but I think this is a good temporary place until polynomial is ready to receive them and we have a clear strategy to integrate Poly & FixedP.
Also, I've added references to the repo where I analyse, test and adapt those approximations suggested by crypten.

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -443,6 +451,73 @@ def matmul(self, *args, **kwargs):
...
+
+        Args:
+            iterations (int): number of iterations for limit approximation
+
+        Ref: https://github.com/LaRiffle/approximate-models
+        """
+        return (1 + self / 2 ** iterations) ** (2 ** iterations)
+
+    def sigmoid(self):
+        """
```
andrelmfarias: Shouldn't we replace the sigmoid approximation by the mathematical definition using the approximated `exp`? Or do you think we lose precision by doing that?

I think we should at least compare the differences between both...

andrelmfarias:
Don't you think we should also add some efficiency tests in `test/efficiency_tests`, as I did for `tanh` and `sigmoid`?

syft/frameworks/torch/tensors/interpreters/precision.py
LaRiffle: Definitely! I'll make a comparison as part of my lib to see what's best, thansk for the suggestion

LaRiffle:
Yes, good idea!

syft/frameworks/torch/tensors/interpreters/precision.py
andrelmfarias: You're welcome!

Maybe also do a time performance comparison, to see what are the trade-offs between both

iamtrask:
Super excited about this!

syft/frameworks/torch/tensors/interpreters/precision.py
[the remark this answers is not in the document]
iamtrask: Thank you for doing that @LaRiffle and @jvmancuso

test/torch/tensors/test_precision.py
```diff
@@ -326,6 +384,118 @@ def test_torch_dot(workers):
...
+            )
+            r_sh = t_sh.exp()
+            r = r_sh.get().float_prec()
+            t = t.exp()
+            diff = (r - t).abs()
+            norm = (r + t) / 2
+            cumsum += diff / (tolerance * norm)
+
+        cumsum /= 10
+        print(cumsum)
```
Jasopaum: This can be removed :)

test/torch/tensors/test_precision.py
```diff
@@ -326,6 +384,118 @@ def test_torch_dot(workers):
...
+    Test the approximate exponential with different tolerance depending on
+    the precision_fractional considered
+    """
+    alice, bob, james = workers["alice"], workers["bob"], workers["james"]
+
+    fix_prec_tolerance = {3: 20 / 100, 4: 5 / 100, 5: 5 / 100}
+
+    for prec_frac, tolerance in fix_prec_tolerance.items():
+        cumsum = torch.zeros(5)
+        for i in range(10):
```
Jasopaum: Are you doing this because the approximation results vary significantly depending on the sharing?
If so, do you know when "bad" cases occur and how bad they are?

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -443,6 +454,104 @@ def matmul(self, *args, **kwargs):
...
+                    sigmoid(x) = 1 / (1 + exp(-x))
+                "maclaurin": Use the Maclaurin / Taylor approximation, with polynomial
+                    interpolation of degree 5 over [-8,8]
+                    NOTE: This method is faster but not as precise as "exp"
+                    Ref: https://mortendahl.github.io/2017/04/17/private-deep-learning-with-mpc/#approximating-sigmoid
+        """
+        one = self * 0 + 1
+
+        if method == "exp":
+            result = one / (1 + (self * -1).exp())
```
Jasopaum: I think we might have talked about it at some point but should you use `.inverse()` here?

test/torch/tensors/test_precision.py
```diff
@@ -93,9 +81,33 @@ def test_methods_for_linear_module(method, parameter):
...
+    # Method syntax
+    x = torch.tensor([0.1, 0.2, 0.3]).fix_prec()
+
+    y = x + x
+
+    assert (y.child.child == torch.LongTensor([200, 400, 600])).all()
+    y = y.float_prec()
+
+    assert (y == torch.tensor([0.2, 0.4, 0.6])).all()
+
```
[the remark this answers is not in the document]
LaRiffle: I will remove it

test/torch/tensors/test_precision.py
LaRiffle: Yes
No not really... :/

syft/frameworks/torch/tensors/interpreters/precision.py
LaRiffle: Correct!

Issue #2654, opened 2019-10-09
REFACTOR: Serde


cereallarceny:
### Overview
I would like to propose that we establish official standards for Serde in a way that's semi-agnostic to language. I've noticed a good many inconsistencies along the way of making my own version of Serde for syft.js. I'll mention them here as best as I can remember, but I will consider this issue to be a "catch-all" epic for all related Serde issues. This document will be a living document and will avoid problems that are related to language incompatibility (i.e. Javascript doesn't have a concept of `Tuple`, `Dict`, etc.).

### Proposed Changes
1. **Remove `msg_type` from all `Message` classes**
This is also being discussed here: https://github.com/OpenMined/PySyft/issues/2649.  I would also advocate that if we remove the `msg_type` field from the `Message` classes that we should also consider removing part or all of the `codes.py` file as it will no longer be needed.
Addressed in https://github.com/OpenMined/PySyft/pull/2687

2. **Eliminate parentheses clouding**
There are many times where parentheses are added somewhat randomly. In reality, they are not random, but rather a false attempt to create a `Tuple`. I've noticed many times where using `()` to create a `Tuple` inside of an existing set of parens fails to simplify.  This leads to very confusing circumstances like the simplification of an `Operation`: `(31, (1, ((6, CONTENT, OF, TUPLE))))` where the third opening paren is redundant.  I've also found that this is usually resolved when using `tuple()` to create a `Tuple` instead of using parens (`()`). 

3. **Choose between using a `List` or a `Tuple` consistently**
I've seen examples like in `plan.State` you have the `state_ids` being stored as a `List`, while in `plan.Procedure` you have `arg_ids` and `return_ids` being stored as a `Tuple`.  Why?  I'd say we should decide on using either the `List` or the `Tuple` for every circumstance unless the use-case is overwhelmingly undesired.  My vote is that we use the `Tuple` because immutability will allow us to better find type errors in PySyft. You don't get this luxury with a `List`.

    **Skipping number 4 for now since we can use the strategy presented by @iamtrask below within syft.js**

4. ~~**Cut down on the amount of classes we're using**~~
[...]

PR #2660, opened 2019-10-12
Serde Façade

mccorby:
Alternative to current `serde` process.

The main idea is for PySyft to comply with a set of models while preserving the current implementation instead of following the opposite direction as we have been doing so far (PySyft defining what these models look like).

* By decoupling the current serde process in PySyft from the workers and making it go through the façade will allow PySyft to evolve without affecting the clients.

* PySyft should provide an implementation for the set of neutral models that are common to all clients.

* There is an overhead in the process as sending and receiving messages but it's a temporary price to pay while all clients and PySyft itself are aligned with a solution.

* Note that this PR does not define what the common models will look like.

Note: I am aware that once Protobuf implementation (or any other protocol) is in place will make this PR obsolete but by having a façade we can start implementing those other protocols step by step without requiring a one or nothing.

### DIFF
--- a/syft/generic/object.py
+++ b/syft/generic/object.py
@@ -4,6 +4,7 @@ from typing import List
 
 import syft as sy
 from syft.generic.frameworks.hook import hook_args
+from syft.serde import SerDe, SyftSerDe
 
 
 class AbstractObject(ABC):
@@ -20,6 +21,7 @@ class AbstractObject(ABC):
         tags: List[str] = None,
         description: str = None,
         child=None,
+        serde: SerDe = SyftSerDe,
     ):
         """Initializer for AbstractTensor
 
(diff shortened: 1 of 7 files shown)


cereallarceny:
@mccorby Did you already read through the Serde refactor I put together?  We're actually already working on some of the ideas you've established here.  Great to see we're on the same page!

https://github.com/OpenMined/PySyft/issues/2654

mccorby:
> @mccorby Did you already read through the Serde refactor I put together? We're actually already working on some of the ideas you've established here. Great to see we're on the same page!
> 
> #2654

Yes, of course I read it but I'm not 100% sure so I want to give it a try to what I have in mind (that takes me too long to unfold with words ) even if that just gets me to the same place you are.

cereallarceny:
Whatever you'd like!  I've already had most of the items that I listed
approved for development and have someone working on them now.  Let me know
if you're interested in joining efforts!

mccorby:
I'm closing this as I couldn't get where I wanted to.

PR #2662, opened 2019-10-12
More detailed install instructions on readme.

Titan-C:
I experienced problems during the installation. Conda enviroments confuse/conflict my linux system and I can't install zstd.

Here I recommend to use plain python virtualenviroments. Particularly since that is what is used on the makefile.


README.md
```diff
@@ -12,14 +12,29 @@ PySyft has also been explained in video form by [Siraj Raval](https://www.youtub
...
 ```
 
+Another alternative is to use a plain python virtual environment. That is
+our preferred environment for development purposes. We provide a direct
+install instructions in our makefile. Jupyter Notebooks are not a
+dependency, are quite a handy tool.
+
+```bash
+make venv
+pip install jupyter
```
robert-wagner: Jupyter should be added to the dev requirements

README.md
```diff
@@ -12,14 +12,29 @@ PySyft has also been explained in video form by [Siraj Raval](https://www.youtub
...
+Prompt](https://docs.anaconda.com/anaconda/user-guide/getting-started/) to
+work from the command line.
 
 ```bash
 conda create -n pysyft python=3
 conda activate pysyft # some older version of conda require "source activate pysyft" instead.
 conda install jupyter notebook
 ```
 
+Another alternative is to use a plain python virtual environment. That is
```
robert-wagner: can we refer to these as python venvs ? That being said I could be convinced otherwise (plain python virtual environement sounds weird to me)

Titan-C: Changed in both places where "plain" was used.

PR #2676, opened 2019-10-18
tiny import deletion

jvmancuso:
There was a hanging import in syft/generic/tensor.py that was bothering me (high likelihood of causing an unnecessary circular import).

As an aside, it'd be cool to have some incrementally more comprehensive PEP linting in the `pyproject.toml` for black


LaRiffle:
What do you mean exactly with "As an aside, it'd be cool to have some incrementally more comprehensive PEP linting in the pyproject.toml for black"?

PR #2677, opened 2019-10-19
Upgrade to PyTorch 1.3

LaRiffle:
This PR is to have compatibility with Torch 1.3

:warning: **Note that one line was changed in the Udacity course** :warning:

ℹ️ I also patched the jit.trace problem since it doesn't look like it will be fixed in PyTorch any soon (it has been including in the documentation: https://pytorch.org/docs/stable/jit.html#torch.jit.script )


test/workers/test_websocket_worker.py
```diff
@@ -231,77 +231,75 @@ def test_websocket_worker_multiple_output_response(hook, start_remote_worker):
...
-    if PRINT_IN_UNITTESTS:  # pragma: no cover
-        print("Evaluation result before training: {}".format(result))
-
-    assert len_dataset == 30
-    assert (hist_target == [10, 10, 10]).all()
-
-    remote_proxy.close()
-    remote_proxy.remove_worker_from_local_worker_registry()
-    # process_remote_worker.terminate()
+# This test seems broken: how is the dataset sent to the remote worker? See issue #2686
```
midokura-silvia: The issue with this test is as follows: It depends on a separately started WebsocketServerWorker. It was not possible to start it from the unit test. But if you start the websocket server worker separately the test runs fine. Could you put a pytest.skip flag around the test? I would prefer that to just commenting it out.

syft/serde/torch_serde.py
```diff
@@ -294,6 +294,7 @@ def _detail_torch_size(worker: AbstractWorker, size: Tuple[int]) -> torch.Size:
         torch.device: (_simplify_torch_device, _detail_torch_device),
         torch.Size: (_simplify_torch_size, _detail_torch_size),
         torch.jit.ScriptModule: (_simplify_script_module, _detail_script_module),
+        torch._C.Function: (_simplify_script_module, _detail_script_module),
```
midokura-silvia: Nice workaround!

test/workers/test_websocket_worker.py
midokura-silvia: there are issues with the asyncio event loop which make this case a bit more difficult.

PR #2684, opened 2019-10-19
Added `String` and `StringPointer` types

AlanAboudib:
I added two new types, `String` and `StringPointer` types. They have have the same API as native 'str' python objects.

In order to use them, here is an example:

```
from syft.generic.string import String

string = String('hello world') # The 'owner' will now be automatically set to syft.local_worker

string_pointer = string.send(bob)
```


AlanAboudib:
@LaRiffle @iamtrask 

Can anyone review this PR please. It seems that Travis is not happy, but the errors are not related to the part I am pushing (or is it?), any thoughts?

LaRiffle:
You might have changed some stuff in the detailers which makes the while thing break: in particular from the failing tests I would say the line `detailer = detailers[idx_or_path]` is the one which triggers the fails,
maybe take a test which is failing _but is quite simple_ and try to understand what didn't work out and why it failed. This is what I do usually when I have tests failing

robert-wagner:
Hey, could you add an explanation of why we need this

iamtrask:
"I also made some changes to serde.py in the _detail() function to enable sending and receiving objects of classes that PySyft does not know."

This seems very risky. Does this mean that I could literally send any object I want to a remote worker and manipulate it?

iamtrask:
I should mention - we explicitly want to prevent people from being able to send types to remote workers which those workers haven't explicitly decided to support.

AlanAboudib:
@iamtrask My bad, I have added  feature to allow sending any object to a remote worker indeed. However, not in this PR. It was in the previous version of this PR, but i was causing errors, so I did the PR without this feature. Just with `String` and `StringPointer`. The `serde.py` file is left untouched.

However, the feature to send any object was added to allow sending SyferText objects such as the `Tokenizer` object. Registering external classes in `serde.py` would require PySyfert to import SyferText, that is why I opted to added that risky feature.

However, we should probability think of a different way of allowing external objects to be sent. An authorization mechanism ??

syft/generic/pointers/string_pointer.py
```diff
@@ -0,0 +1,127 @@
...
+from typing import Union
+
+from syft.generic.string import String
+from syft.generic.pointers.object_pointer import ObjectPointer
+from syft.workers.base import BaseWorker
+
+import abc
+
+
+class PointerClassMaker(abc.ABCMeta):
```
LaRiffle: It  strikes me that this part should be integrated in the hook file, in order to perform the hook a single time and put all the hook mechanisms at the same place.

syft/generic/string.py
```diff
@@ -0,0 +1,301 @@
...
+
+            # Get only magic attributes that are not
+            # in the 'ignored_attr' set
+            if attr not in ignored_attr:
+
+                dct[attr] = cls.method_hook(cls, attr)
+
+        return super(BuiltinTypeWrapper, cls).__new__(cls, clsname, bases, dct)
+
+    def _args_adaptor(self: object, args: Tuple[object]):
```
LaRiffle: You might also want to use what's done in hook_args as well which does unwrap the args (_args_adaptor) and wrap the response back (_wrap_return_value), although it might be complicated to adapt. Normally for tensors it's a one-line change in hook_args.py

syft/generic/string.py
```diff
@@ -0,0 +1,301 @@
...
+        description = simplify_str(string.description)
+
+        return (simple_core, string.id, tags, description)
+
+    @staticmethod
+    def detail(worker: BaseWorker, simple_obj: Tuple):
+        """
+           Create an object of type String from the reduced representation in `simple_obj`.
+
+           Parameters
```
LaRiffle: Use google style for docstings

syft/generic/string.py
```diff
@@ -0,0 +1,301 @@
...
+            str_kwargs["object"] = object
+
+        if encoding:
+            str_kwargs["encoding"] = encoding
+
+        if errors:
+            str_kwargs["errors"] = errors
+
+        # Create a str instance as the 'child' attribute
+        self.core = str(**str_kwargs)
```
LaRiffle: If this is relevant, it might be more appropriate to call it `.child` to be consistent with the tensors api

iamtrask:
> @iamtrask My bad, I have added feature to allow sending any object to a remote worker indeed. However, not in this PR. It was in the previous version of this PR, but i was causing errors, so I did the PR without this feature. Just with `String` and `StringPointer`. The `serde.py` file is left untouched.
> 
> However, the feature to send any object was added to allow sending SyferText objects such as the `Tokenizer` object. Registering external classes in `serde.py` would require PySyfert to import SyferText, that is why I opted to added that risky feature.
> 
> However, we should probability think of a different way of allowing external objects to be sent. An authorization mechanism ??

Not necessarily - SyferText could simply register more objects to serde when it's imported (so that they can be received), and the assumption would be that the worker would have to have also imported SyferText in order to receive commands.

AlanAboudib:
@iamtrask I discussed the issue @vvmnnnkv this week and we actually agree with that. This should be straight forward now

iamtrask:
Excellent!

iamtrask:
Hey - an odd request. As a UX thing, do you think it would be possible to actually hook string instead of having a separate class?

Aka, instead of doing

```
from syft.generic.string import String

string = String('hello world', owner = hook.local_worker)

string_pointer = string.send(bob)
```

You'd just do

```
string_pointer = 'hello world'.send(bob)
```

This just seems a lot cleaner, no?

AlanAboudib:
@iamtrask 

I have thought of this but it turned out that you cannot apply `setattr` on builtin types in python. Any thoughts?

iamtrask:
> @iamtrask
> 
> I have thought of this but it turned out that you cannot apply `setattr` on builtin types in python. Any thoughts?

http://clarete.li/forbiddenfruit/ might be able to help here :)

syft/generic/string.py
AlanAboudib: @LaRiffle Done

syft/generic/string.py
AlanAboudib: Done

syft/generic/pointers/string_pointer.py
AlanAboudib: done

AlanAboudib:
@LaRiffle @iamtrask Would you please take a look at this PR, I added all of the required changes. Phewww

By the way, 

You can now instantiate a `String` object without explicitly indicating the owner. As with tensor types, the owner will be set to `syft.local_worker` by default:

```
from syft.generic.string import String

text = String('Hello PySyft String')

```

syft/generic/frameworks/hook/hook.py
```diff
@@ -271,6 +273,73 @@ def _hook_multi_pointer_tensor_methods(self, tensor_type):
...
+                "__gt__",
+                "__lt__",
+                "__ne__",
+                "__len__",
+                "__getitem__",
+                "__str__",
+                "__repr__",
+                "__format__",
+                "lower",
+                "upper",
```
LaRiffle: Here is a short list of all str method in python, you might want to include more like title or capitalize maybe depending on your usecases: https://www.w3schools.com/python/python_ref_string.asp

syft/generic/frameworks/hook/hook.py
```diff
@@ -271,6 +273,73 @@ def _hook_multi_pointer_tensor_methods(self, tensor_type):
...
+                # Create the hooked method
+                new_method = self._get_hooked_string_method(attr)
+
+                # Add the hooked method
+                setattr(String, attr, new_method)
+
+    def _hook_string_pointer_methods(self):
+
+        # Set of methods to hook/override for 'str' to String
+        attr_include = set(
```
LaRiffle: this list is duplicate with the other one, try to only define it in one place for consistency purpose

syft/generic/frameworks/hook/hook.py
```diff
@@ -567,3 +636,121 @@ def overloaded_attr(self, *args, **kwargs):
...
+        for arg in args:
+
+            # If 'arg' is an object of type String
+            # replace it by and 'str' object
+            if isinstance(arg, String):
+                new_args.append(arg.child)
+            else:
+                new_args.append(arg)
+
+        return new_args
```
LaRiffle: I think I already mentioned it but this may fit in the hook_args logic

syft/generic/frameworks/hook/hook.py
```diff
@@ -567,3 +636,121 @@ def overloaded_attr(self, *args, **kwargs):
...
+            Operate the hooking
+            """
+
+            owner = self.owner
+            location = self.location
+            id_at_location = self.id_at_location
+
+            # Create a 'command' variable  that is understood by
+            # the send_command() method of a worker.
+            command = (attr, id_at_location, args, kwargs)
```
LaRiffle: usually it is expected to put "self" and not "self.id_at_location"

syft/generic/string.py
```diff
@@ -0,0 +1,173 @@
...
+    def __init__(
+        self,
+        object: object = None,
+        encoding: str = None,
+        errors: str = None,
+        id: int = None,
+        owner: BaseWorker = None,
+        tags: List[str] = None,
+        description: str = None,
+    ):
```
LaRiffle: Add a dosctring to explain the kwargs (like encoding, errors, etc) which is very useful!

syft/generic/string.py
```diff
@@ -0,0 +1,173 @@
...
+            return str.encode("utf-8") if str else None
+
+        # Encode the string into a bytes object
+        simple_core = simplify_str(string.child)
+
+        tags = (
+            [simplify_str(tag) if isinstance(tag, str) else tag for tag in string.tags]
+            if string.tags
+            else None
+        )
```
LaRiffle: can't you do just `tags = sy.serde._simplify(string.tags)`?

syft/generic/string.py
```diff
@@ -0,0 +1,173 @@
...
+        simple_core, id, tags, description = simple_obj
+
+        def detail_bstr(b_str):
+            return str(b_str, encoding="utf-8") if b_str else None
+
+        # It appears that all strings are converted to bytes objects
+        # after deserialization, convert them back to strings
+        tags = (
+            [detail_bstr(tag) if isinstance(tag, bytes) else tag for tag in tags] if tags else None
+        )
```
LaRiffle: smae here
can't you do just `tags = sy.serde._detail(worker, string.tags)`?

syft/generic/frameworks/hook/hook.py
AlanAboudib: I added this list as a class set variable of `String`, so we can get the list by typing `String.methods_to_hook`

syft/generic/frameworks/hook/hook.py
AlanAboudib: Done

syft/generic/string.py
AlanAboudib: Done

syft/generic/string.py
AlanAboudib: Done

syft/generic/string.py
AlanAboudib: Done

syft/generic/frameworks/hook/hook.py
AlanAboudib: That is right, Done!

iamtrask:
Heads up - merge conflicts

AlanAboudib:
conflict settled

PR #2687, opened 2019-10-20
Remove msg_type form codes.py

### LINKED ISSUE #2531: Better integration of Message class throughout codebase

### LINKED ISSUE #2649: Remove msg_type for all Message classes
In Serde we're using codes like 31, 32, 33 to describe `Message`-based classes, but then we're also keeping track of `msg_type`, which creates strange simplification strings:

`(31, (1, CONTENT_GOES_HERE))`
`31` means `Operation` in Serde
`1` means `Operation` in `codes.py` where `msg_type`'s are defined

This is super confusing for reading Serde strings.  To use the example above, `1` also means `List` in Serde.  So it could be read as "an `Operation` which contains a `List` and other content".

Let's remove all instances of `msg_type` and `MSGTYPE` around Syft.

LaRiffle:
Fixes Part 1. of #2654
Fix #2649
Fix #2531


cereallarceny:
I would love to get this merged @robert-wagner @iamtrask @LaRiffle - it's one of the Serde issues.

PR #2690, opened 2019-10-22
WIP: enable cuda processing in PySyft

midokura-silvia:
Unhooking the tensor.data attributes enables .to(device) to work correctly. 
However, this seems to create some problems.
Thus, this pull request needs some more work to make the unit tests pass again. Any help would be greatly appreciated.

### DIFF
--- a/syft/frameworks/torch/hook/hook.py
+++ b/syft/frameworks/torch/hook/hook.py
@@ -295,16 +295,12 @@ class TorchHook(FrameworkHook):
         # torch.nn.Parameter.__repr__ = hooked__repr__
 
         # Hook .data to handle chain assignment when needed
-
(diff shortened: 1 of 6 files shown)


syft/frameworks/torch/hook/hook.py
```diff
@@ -529,6 +523,7 @@ def _transfer_methods_to_native_tensor(cls, tensor_type: type, syft_type: type):
             "__ge__",
             "__lt__",
             "__le__",
+            "data",
```
LaRiffle: I don't think you have to do this, you're already directly redefining `torch.nn.Parameter.data`

midokura-silvia:
Closing PR as it tries to merge to master (and not dev)

PR #2692, opened 2019-10-22
Add python context to base worker

### LINKED ISSUE #2689: Allow worker registration of objects within python context
Tons of test and code samples using plans are currently like this:

```python
# my awesome code
hook.local_worker.is_client_worker = True
```

```python
with self.owner.registration_enabled():
    # my awesome code
```

Enabling registration all the time :O

murarugeorgec:
**Changes:**
* Add decorator such that we could have a python ctx by calling a method
* Changed one of the tests that were using the *is_client_worker* member explicitly.

LE: Fixes #2689

### DIFF
--- a/syft/workers/base.py
+++ b/syft/workers/base.py
@@ -1,4 +1,6 @@
 from abc import abstractmethod
+from contextlib import contextmanager
+
 import logging
 from typing import Callable
 from typing import List
@@ -201,6 +203,14 @@ class BaseWorker(AbstractWorker, ObjectStorage):
 
         """
         raise NotImplementedError  # pragma: no cover
+
+    @contextmanager
+    def registration_enabled(self):
+        self.is_client_worker = False
+        try:
+            yield self
+        except:
+            self.is_client_worker = True
 
     def remove_worker_from_registry(self, worker_id):
         """Removes a worker from the dictionary of known workers.
--- a/syft/workers/base.py
+++ b/syft/workers/base.py
@@ -209,7 +209,7 @@ class BaseWorker(AbstractWorker, ObjectStorage):
         self.is_client_worker = False
         try:
             yield self
-        except:
+        finally:
             self.is_client_worker = True
 
     def remove_worker_from_registry(self, worker_id):
--- a/test/torch/tensors/test_precision.py
+++ b/test/torch/tensors/test_precision.py
@@ -31,13 +31,11 @@ def test_encode_decode(workers, parameter):
 
 
 def test_fix_prec_registration(hook):
-    hook.local_worker.is_client_worker = False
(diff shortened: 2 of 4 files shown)


test/torch/tensors/test_precision.py
```diff
@@ -30,14 +30,23 @@ def test_encode_decode(workers, parameter):
     assert (x == torch.tensor([0.1, 0.2, 0.3])).all()
 
 
-def test_fix_prec_registration(hook):
-    hook.local_worker.is_client_worker = False
+@pytest.mark.parametrize("use_context", [False, True])
+def test_fix_prec_registration(hook, use_context):
+    if not use_context:
```
murarugeorgec: Q: Should I add new tests to validate that the context manager is working correctly? (and leave the old tests how they were?)
Another way is the one I added here - have a **use_context** parameter and depending on that use the python context or not (I do not like this method that much because it is hard to read).
Third method would be to duplicate the tests and in one test use the parameter explicitly and in another to use the context.

jvmancuso:
Can you explain a bit further what this PR is doing? This seems like a departure from past/current design and I'd like to understand if there are implications elsewhere in the code

test/torch/tensors/test_precision.py
LaRiffle: Thanks!
So you can add a specific test to verify it works as expected, like this one
```
def <...>_test(...):
    assert hook.local_worker.is_client_worker == True
    with hook.local_worker.registration_enabled():
        hook.local_worker.is_client_worker == False
    assert hook.local_worker.is_client_worker == True

Comment (LaRiffle) [reply test/torch/tensors/test_precision.py]: And then change all the occurences of the tests where you have the old syntax `assert hook.local_worker.is_client_worker == True` to use context instead

File: syft/workers/base.py [hunk 2]
```diff
@@ -202,6 +204,14 @@ def _recv_msg(self, message: bin):
         """
         raise NotImplementedError  # pragma: no cover
 
+    @contextmanager
+    def registration_enabled(self):
+        self.is_client_worker = False
+        try:
+            yield self
+        except:
```
LaRiffle: I think it's `finally` right?

LaRiffle:
> Can you explain a bit further what this PR is doing? This seems like a departure from past/current design and I'd like to understand if there are implications elsewhere in the code

It's just a small sugar to simplify our life, see issue https://github.com/OpenMined/PySyft/issues/2689

syft/workers/base.py
murarugeorgec: Sure

murarugeorgec:
@LaRiffle ready for review again

Issue #2678, opened 2019-10-19
<worker>.torch


iamtrask:
**Is your feature request related to a problem? Please describe.**
At present we cannot initialize tensors on a remote machine. Whenever we want to initialize something, we first initialize it locally and then call .send() to move it to the remote machine. This is *extremely* wasteful in some cases such as.
```
# BEGIN boilerplate
import syft as sy
import torch as th
hook = sy.TorchHook(th)
bob = sy.VirtualWorker(hook, id="bob")
# END boilerplate

#BEGIN code relevant to this github issue
x = th.rand(3,3).send(bob) # this i the wasteful part
#END
```

**Describe the solution you'd like**
Instead I'd like to be able to run

(skipping boilerplate)

```
x = bob.th.rand(3,3)
```
or alternatively
```
x = bob.torch.rand(3,3)
```

**Describe alternatives you've considered**

```
x = bob.rand(3,3)
```

**Additional context**
bob.torch should have the full pytorch API available for us to call, all of which simply executes on the remote machine.

gmuraru:
I can take it - it would be a challenge, but will try to do it :)

PR #2695, opened 2019-10-23
Remove random tensor generators from hook exclusions

genekogan:
This removes `rand`, `randint`, `randn`, `randn_like`, and `randperm` from the list of torch functions excluded from hook. 

Unless there is any reason specific reason not to allow these, there are use cases where it would be nice to be able to do remote random number generation. I ran into this issue trying to make a [variational autoencoder with syft](https://gist.github.com/genekogan/53eab74f004e36f6e3db5f182678c94d), as it's necessary to generate random numbers during the forward pass.


test/torch/linalg/test_operations.py
```diff
@@ -115,3 +115,16 @@ def test_qr_mpc(hook, workers):
 
     # Check if QR == t
     assert ((Q @ R - t).abs() < 1e-2).all()
+
+
+def test_remote_random_number_generation(hook, workers):
```
robert-wagner: While we’re here, could you add a test for all of them? In practice you should be able to parameterize it

genekogan: As separate functions or all within `test_remote_random_number_generation`?

LaRiffle: In the same function is good :)

genekogan: Have added tests for the rest of the functions.

genekogan: @robert-wagner I realize now my last commit is not doing what was asked. The problem is that [you can't do remote tensor generation](https://github.com/OpenMined/PySyft/issues/2678) at the moment. So the ones I added are just executed locally despite being done so in the sent model.

PR #2702, opened 2019-10-25
Add automatic publishing to pypi

robert-wagner:


robert-wagner:
Account needs to be added as a maintainer to the pypi package for this to work

Issue #2693, opened 2019-10-23
Have Notebooks build as part of the test/CI process


fdroessler:
**Is your feature request related to a problem? Please describe.**
In Slack one of the most common request for help or fixes is that the tutorial notebooks don't work. Either some new feature results in the need for the Notebook to be updated or other measures to be taken.

**Describe the solution you'd like**
Have the notebooks be build as part of the CI and testing pipeline to flag up early when something stops working.

**Describe alternatives you've considered**
Remain in status quo where experienced users need to answer the same questions over and over again in Slack.

I'll be working myself on this issue. I wanted to put up an issue for that as a reminder and as a reference for others.

youben11:
Great idea! But how do you plan to flag malfunctioning notebooks?

fdroessler:
I was planning on combining papermill with pytest and write them as tests that run in an isolated filesystem. That way I think it should be possible to get websocket notebooks tested too. By using papermill it should also allow for reducing the execution time by parameterising the notebooks (like epochs etc) although in that case some parameters will be set in the first cell. It's an idea so far I'll see how far I get.

fdroessler:
@youben11 @iamtrask I have submitted a __very__ raw [WIP] PR just to check if where I am going is in line with your ideas. Let me know your initial thoughts on the approach.

karlhigley:
Notebooks tests are now running as part of the test/CI process. 👍

PR #2703, opened 2019-10-27
[WIP] Notebook test during CI

fdroessler:
This is the PR for #2693 as a reference here.


fdroessler:
@kamathhrishi Great points thanks so much for looking at this. Yes, ultimately I think something like a `glob.glob(*.ipynb)` or similar would be good. However, some handcrafting will be necessary as some notebooks need a certain environment (workers, Websocketserver etc.) I have just included here those that are running straightforward atm to provide a minimal example.

Yes you are totally right in that it is hard to know that notebooks are part of the test, it should probably be included in the Contributer guide, I will make a note of that.

Not sure I understand your last point about the checkout the changes. For the notebooks I included I have cleared the kernel and outputs before including them in the tests. Do you mean that or something else?

LaRiffle:
> Not sure I understand your last point about the checkout the changes. For the notebooks I included I have cleared the kernel and outputs before including them in the tests. Do you mean that or something else?

I think @kamathhrishi is asking to checkout the files from the git commit, so that your PR doesn't have 18 files changed with most of them just being changed because you re-run the notebooks :)

fdroessler:
Yeah, that is a good point. However, a lot of the notebooks have weird metadata still included which makes it hard to run them automated so most of the changes that I did is to restart the kernels so the notebooks are clean.

iamtrask:
How's this coming?

fdroessler:
Thanks for following up, I'll have some time today and tomorrow to work on this.

fdroessler:
@iamtrask What is your view on all the data files that are necessary for the notebooks to run? I have the ones that don't need the data files running but not sure how to proceed with the others. A separate git lfs or downloading them when the fixture is set-up? What are your thoughts on this?

fdroessler:
I had some problems with my repo so I needed to set it up new. Will start a new PR. Sorry for any inconveniences

PR #2709, opened 2019-10-28
Private Tensors

IonesioJunior:
- Create Private Tensors
  - [x] Convert Torch Tensor into Private Tensors
  - [x] Serialize Private Tensors
  - [x] Allow/Disallow access to tensor values.
  - [x] Forward tensor methods.
  - [x] Perform Arithmetic Operations.
  - [x] Unit tests.


IonesioJunior:
We have three use cases to Private Tensors operations:

**First**: Perform a tensor operation using only itself.
```
x = th.tensor([1,2,3,4,5])

# Wrap a public tensor using a private layer
private_x = x.private_tensor(allowed_users=[SSLCertificate, SSKey, UserAuth])

result_tensor = private_x.t()
```
Should **result_tensor** inherit private_x credentials?




**Second**: Perform a tensor operation with a public tensor.
```
x = th.tensor([1,2,3,4,5])
y = th.tensor([2,2,2,2,2])

// Wrap a public tensor using a private layer
private_x = x.private_tensor(allowed_users=[SSLCertificate, SSKey, UserAuth])

result_tensor = private_x + y
```
Should **result_tensor** be converted to a public tensor or inherit **private_x** credentials?

**Third**: Perform a tensor operation with another private tensor with different credentials.
```
x = th.tensor([1,2,3,4,5])
y = th.tensor([2,2,2,2,2])

// Wrap a public tensor using a private layer
private_x = x.private_tensor(allowed_users=[SSLCertificate, SSKey])
private_y = y.private_tensor(allowed_users=[SSKey, UserAuth])

result_tensor = private_x + y
```

Should **result_tensor** inherit the intersection between **private_x** and **private_y** (Logic AND) or the union between them (Logic OR)?

IonesioJunior:
@iamtrask, @LaRiffle, and @robert-wagner could we discuss about these use cases?

iamtrask:
Certainly! Setup a calendar invite.

test/torch/tensors/test_private.py
```diff
@@ -0,0 +1,316 @@
...
+    assert not private_x.allowed_to_get(unregistered_user)
+
+
+def test_get_method(workers):
+    bob = workers["bob"]
+    x_tensor = torch.Tensor([1, 2, 3])
+
+    private_x = x_tensor.private_tensor(allowed_users=("User"))
+
+    private_x_pointer = private_x.send(bob)
```
iamtrask: Thinking about this, .send() should have all the same permissions as .get(), causing this to fail. Otherwise I could just try to call .send(me) to get around .get() blocking me.

test/torch/tensors/test_private.py
```diff
@@ -0,0 +1,316 @@
...
+            self.password = password
+
+        def __eq__(self, other):
+            if isinstance(other, UserAuthMockup):
+                return self.login == other.login and self.__password == other.password
+            else:
+                return False
+
+    registered_user = UserAuthMockup("user", "password")
+    registered_certificate = "encoded_certificate"
```
iamtrask: what is the certificate?

syft/generic/frameworks/hook/hook.py
```diff
@@ -407,6 +419,73 @@ def overloaded_native_method(self, *args, **kwargs):
 
         return overloaded_native_method
 
+    @classmethod
+    def _get_hooked_private_method(cls, method_name):
```
iamtrask: Why was this necessary to put here? Is this not the same as https://github.com/OpenMined/PySyft/blob/f39a93d085b74f16724633fd12a28c893a724c59/syft/generic/frameworks/hook/hook.py#L365

IonesioJunior: In the final of `_get_hooked_private_method`, we're setting pointers to the result tensor parent(s) and saving the command/operation used to generate it.

IonesioJunior: But thinking about this, It doesn't seem the best way to implement it.

IonesioJunior: Do you have any suggestion?

test/torch/tensors/test_private.py
IonesioJunior: Yes, It makes sense! I'll fix it!

test/torch/tensors/test_private.py
IonesioJunior: It's about SSL/TLS Certificates. We can implement an interface to authenticate companies by their Certificates (like HTTPS/WSS protocols), receiving their encoded public keys (string).

IonesioJunior: Should I wrap it into a mockup class?

syft/generic/frameworks/hook/hook.py
iamtrask: ah nvm - yours looks good. Yes we could probably compress the code some more but this is readable so I'm happy

test/torch/tensors/test_private.py
iamtrask: epic! wrapping in a mockup class might be nice but we can come back to it when the time comes

test/torch/tensors/test_private.py
IonesioJunior: Done!

test/torch/tensors/test_private.py
IonesioJunior: Done!

test/torch/tensors/test_private.py
iamtrask: You're a rockstar.

Issue #2562, opened 2019-08-27
resnet50 not supported when torch 1.0.1 / 1.1 is hooked 


mari-linhares:
```
import torch
import syft
hook = syft.TorchHook(torch)
import torchvision.models as models

def make_model(num_out_classes: int):
    """Load a resnet50 and add a new head to it."""
    model = models.resnet50(pretrained=True) 
    num_ftrs = model.fc.in_features
    model.fc = torch.nn.Linear(num_ftrs, num_out_classes)
    return model

model = make_model(7)
model(torch.zeros(torch.Size([1, 3, 224, 224])))
```

Error:

```
...
~/PySyft/syft/frameworks/torch/hook/hook.py in overloaded_func(*args, **kwargs)
    745                 handle_func_command = TorchTensor.handle_func_command
    746 
--> 747             response = handle_func_command(command)
    748 
    749             return response

~/PySyft/syft/frameworks/torch/tensors/interpreters/native.py in handle_func_command(cls, command)
    336             # in the execute_command function
    337             if isinstance(args, tuple):
--> 338                 response = eval(cmd)(*args, **kwargs)
    339             else:
    340                 response = eval(cmd)(args, **kwargs)

~/PySyft/syft/frameworks/torch/tensors/interpreters/native.py in <module>

AttributeError: module 'torch._C._nn' has no attribute 'native_avg_pool2d'
```
---
torch version 1.0.1 and 1.1
syft dev version

andrelmfarias:
Weird... I don't think there is a `native_avg_pool2d` in Pytorch, only `avg_pool2d`. Did you try to load the model without hooking torch?

Maybe the hook changes the name of the method...

marcusvlc:
@mari-linhares  For some reason, the command created by the native.py file for the avg_pool2d layer looks like this:

syft.local_worker.hook.torch._C._nn.native_avg_pool2d

While the max_pool2d looks like this:

syft.local_worker.hook.torch.nn.functional.native_max_pool2d

I am still trying to understand why this particular layer is being called from torch._C and not torch.nn, but for practical testing purposes, I just created an if to change avg_pool2d's command to conform to that of max_pool2d, and it worked normally. I was able to load networks like ResNet and MobileNet.

To fix it, simply open the native.py file and change the construction of the cmd variable to the following:

if (cmd.split (".") [- 1] == 'avg_pool2d'):
                cmd = 'syft.local_worker.hook.torch.nn.functional.native_avg_pool2d'
[...]

PR #2713, opened 2019-10-30
Making changes to avoid errors on avgpool layer

marcusvlc:
The error arises when a network with AvgPooling layer is used. I realized that the command created for this layer used the path _C_nn and not nn.functional as well as other similar layers, such as MaxPooling.

The problem was reported in this issue:

https://github.com/OpenMined/PySyft/issues/2562

My initial solution was to switch all commands that are using the _C_nn path to nn.functional.


marcusvlc:
@robert-wagner I created a solution using try catch to avoid changing the string with each iteration. If it is not good, could you suggest another approach?

iamtrask:
Hey @marcusvlc - would absolutely love to merge this. Any chance you can add a test per @robert-wagner 's request?

marcusvlc:
@iamtrask Done! 

I had to put some extra checks as the command variable was being used with the wrong path (_C._nn) in several calls. I also added a test that performs a foward on a network with several different layers, including AvgPooling.

While I was designing the tests, I noticed that a pointtensor's .size () function is returning unexpected values, is this normal? I am saying this because most pre-imported models from the torchvision library use this function to access the positions of a tensor.

iamtrask:
thank you for this PR - it's a really strong contribution which lots of folks will use!

Issue #2710, opened 2019-10-29
Single effective send of plan to worker


LaRiffle:
**Describe the solution you'd like**
Improve Plan with refs to pointerPlan to not send more than 1 time the plan —> else return the same pointerPlan

This reduces the bandwidth usage and doesn't change the UX

LaRiffle:
@murarugeorgec is willing to take this one

LaRiffle:
Fixed by #2716

PR #2716, opened 2019-11-05
Add cache for location2pointerplan

murarugeorgec:
Solves #2710 
Add a cache mechanism for the Plan class.
1-to-1 mapping between a location and a possible pointer plan that was sent to a worker.
Should reduce the bandwidth usage when the same plan might be sent multiple times to the same worker.

### DIFF
--- a/pip-dep/requirements_dev.txt
+++ b/pip-dep/requirements_dev.txt
@@ -7,7 +7,6 @@ pre-commit>=1.16.1
 pyopenssl>=19.0.0
 pyshark>=0.4.2.4
 pytest>=4.5.0
-scipy
 setuptools>=41.0.0
 sphinx-markdown-builder>=0.4.1
 Sphinx>=2.1.1
--- a/pip-dep/requirements_dev.txt
+++ b/pip-dep/requirements_dev.txt
@@ -7,6 +7,7 @@ pre-commit>=1.16.1
 pyopenssl>=19.0.0
 pyshark>=0.4.2.4
 pytest>=4.5.0
+scipy
 setuptools>=41.0.0
 sphinx-markdown-builder>=0.4.1
 Sphinx>=2.1.1
--- a/test/message/test_plan.py
+++ b/test/message/test_plan.py
@@ -962,7 +962,7 @@ def test_cached_plan_send(workers):
     pointers = plan_abs.get_pointers()
 
     assert len(pointers) == 1
-    assert plan_bob_ptr1 == plan_bob_ptr2
+    assert plan_bob_ptr1 is plan_bob_ptr2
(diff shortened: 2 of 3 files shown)


pip-dep/requirements_dev.txt
```diff
@@ -6,7 +6,9 @@ jupyter>=1.0.0
 pre-commit>=1.16.1
 pyopenssl>=19.0.0
 pyshark>=0.4.2.4
+pytest-asyncio
```
murarugeorgec: Q: Should we add versions to this?
I got a warning for this and for the scipy I got an error saying that I did not have it on my system.
Q2: Is this a good place for them?

test/message/test_plan.py
```diff
@@ -93,7 +93,7 @@ def _(data):
 
 
 def test_raise_exception_when_sending_unbuilt_plan(workers):
-    me, bob = workers["me"], workers["bob"]
```
murarugeorgec: Q: Removed me since it was not used

test/message/test_plan.py
```diff
@@ -948,3 +948,33 @@ def plan_double_abs(x):
...
+
+def test_cached_plan_send(workers):
+    bob = workers["bob"]
+
+    @sy.func2plan(args_shape=[(1,)])
+    def plan_abs(data):
+        return data.abs()
+
+    plan_bob_ptr1 = plan_abs.send(bob)
+    plan_bob_ptr2 = plan_abs.send(bob)
```
LaRiffle: Check also that plan_bob_ptr1 and plan_bob_ptr2 are the same object :) [maybe the id / id_at_location, etc]

LaRiffle: Check also that plan_bob_ptr1 and plan_bob_ptr2 are the same object :) [maybe the id / id_at_location, etc]

pip-dep/requirements_dev.txt
```diff
@@ -7,6 +7,7 @@ pre-commit>=1.16.1
 pyopenssl>=19.0.0
 pyshark>=0.4.2.4
 pytest>=4.5.0
+scipy
```
LaRiffle: Do we need it?

test/message/test_plan.py
murarugeorgec: I removed the ```==``` from line 965 and added an ```is``` (it should check the id of the two pointers). Should this be enough or I should also check the identity for the members?

pip-dep/requirements_dev.txt
murarugeorgec: I think we do not need it since ```travis``` did not complain without it.

Issue #2637, opened 2019-10-01
handle_command_function doesn't use @overloaded.module for pure tensors


LaRiffle:
**Describe the bug**
In native.py lives the changes made to the torch tensor
In particular, like other syft tensors, it has the following decorator @overloaded.module
```python
    @staticmethod
    @overloaded.module
    def torch(module):
        def roll(tensor, shifts, **kwargs):
            int_shifts = int(shifts.item())
            return torch.native_roll(tensor, int_shifts, **kwargs)

        module.roll = roll
```
and a class method handle_func_command which is supposed to call it to change some torch function behaviours (here torch.roll)
But this is not the case for pure torch tensors, you can check it by yourself
handle_func_command shoudl be fixed accordingly.

-----

EDIT: So basically if you run torch.roll(x) with x pure torch tensor, the overloading I've written **will** be applied,
However if now I run it on a wrapper (any torch.tensor with a child), it **won't** be applied.

sukhadj:
I am interested to work on this issue.

sukhadj:
@LaRiffle Can you please give an example of `handle_func_command` call? And also what is the significance of `handle_fun_command`? I am a little confused. Thank you.

LaRiffle:
handle_command_function is called automatically when you call a torch function. Depending on the type of the args provided to the torch function, the `handle_command_function` of the appropriate class will be called (you can check that there is indeed one per Tensor type class).

The mechanism which handles this is in hook.py line 425

`native.py` is intended to extend the torch Tensor. But recall that torch Tensor can be used either to perform pure torch operation or as wrappers of syft tensors (in which case they are empty and act just as a small shell)
`handle_command_function`  should always check if the function called could not be replaced by an overloaded one, defined in the "@staticmethod"  def torch of the same class.
`handle_command_function` in native.py should make sure that overloaded functions are used in both cases (as native tensors or wrappers)
Now, this check is done in
```python
# Check that the function has not been overwritten
            try:
                # Try to get recursively the attributes in cmd = "<attr1>.<attr2>.<attr3>..."
                command = cls.rgetattr(cls, cmd)
                return command(*args, **kwargs)
            except AttributeError:
                pass
```
so you just have to put it out of the wrapper/non wrapper disjunction materialized by the `try / except PureFrameworkTensorFoundError` to have it work in bother cases

murarugeorgec:
[...]

PR #2722, opened 2019-11-07
Changed handle_command_function to use @overloaded.module for syft tensors

sukhadj:
This PR solves #2637 . 
We check in handle_func_command if the function is overloaded or not. (the check was already existed for PureFrameworkTensor).
Thanks @LaRiffle for the help!!


syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -292,6 +296,14 @@ def handle_func_command(cls, command):
...
                 return args_type.handle_func_command(command)
 
+            # Check that the function is not overloaded
+            try:
+                # Try to get recursively the attributes in cmd = "<attr1>.<attr2>.<attr3>..."
+                command = cls.rgetattr(cls, cmd)
+                return command(*args, **kwargs)
+            except AttributeError:
+                pass
+
```
LaRiffle: I think you can remove the occurence of the same code below at line 316.

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -246,7 +246,11 @@ def roll(tensor, shifts, **kwargs):
             int_shifts = int(shifts.item())
             return torch.native_roll(tensor, int_shifts, **kwargs)
 
+        def add(a, b):
+            return a + b
+
         module.roll = roll
+        module.add = add
```
LaRiffle: This will be only for the test, when you're happy with the PR remove these lines

sukhadj: Ok!!

syft/frameworks/torch/tensors/interpreters/native.py
sukhadj: Hey @LaRiffle, the code above is important because else it doesn't handle case where we pass shift parameter as a tensor.
For example,
```python
x = torch.tensor([1.0, 2.0, 3, 4, 5])
index = torch.tensor([-1.0])
result = torch.roll(x, index)
```
exits with
```python
TypeError: roll(): argument 'shifts' (position 2) must be tuple of ints, not Tensor
```
Do we have any alternative or this is fine?

LaRiffle: I think you should put that block out of the 
```
<here>
try:
...
except PureFrameworkTensorFoundError
```
it should work better!

LaRiffle: This way you can indeed remove the duplicate and only keep a single occurence

sukhadj: Hey,
That too results in an error 
```python 
RuntimeError: Error, Please consider calling ".get" method instead of ".item" method, so you can be safely getting the item you need.
```
when we pass syft tensor (wrapper) as shift parameter.

This might be because the ```handle_command_function``` calls correct overloaded function but arguments needs to be resolved first (wrapper to be converted to the framework tensor) before passing to the overloaded function.

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -294,13 +294,21 @@ def handle_func_command(cls, command):
 
             # build the new command
             new_command = (cmd, None, new_args, new_kwargs)
+            # Check that the function is not overloaded - for syft tensors
+            try:
+                # Try to get recursively the attributes in cmd = "<attr1>.<attr2>.<attr3>..."
+                command = cls.rgetattr(cls, new_command)
```
andrelmfarias: The 2nd arg of `rgetattr` should be a string with a structure like so: `attr1.attr2.attr3...`
But here you are feeding the function with `new_command` which is a tuple.

Shouldn't you be feeding it with cmd instead?

sukhadj: That's correct. Thanks for bringing this to notice :).
I have replaced ```new_command``` with ```cmd```. It was raising `AttributeError` everytime. 
Now it's not working for one case though when we pass **wrapper tensor as shifts argument**. It's raising the exception `RuntimeError: Error, Please consider calling ".get" method instead of ".item" method, so you can be safely getting the item you need.`

iamtrask:
```
test/torch/tensors/test_additive_shared.py         508      5    99%   411-416
```

Looks like these lines aren't getting called in the testing suite. Can we either remove them or ensure taht they get run? (this is what's causing OpenCI to fail)

iamtrask:
> ```
> test/torch/tensors/test_additive_shared.py         508      5    99%   411-416
> ```
> 
> Looks like these lines aren't getting called in the testing suite. Can we either remove them or ensure taht they get run? (this is what's causing OpenCI to fail)

Bump! Would really love to merge this PR!

gmuraru:
@sukhadj are you still working on this?

sukhadj:
@gmuraru the branch no longer exists. We can close this PR. I'll open a new PR with changes 🤘

Issue #2708, opened 2019-10-28
Fix Beaver triple generation


LaRiffle:
**Describe the bug**
The local worker can read the beaver triple values at generation time (see syft/frameworks/torch/crypto/beaver.py) which is a security issue if it also owns shares. These shares should only be known by the crypto_provider.
In addition it calls 3 times .share which could be optimized by stacking values and calling a single time .share before unstacking. (see torch.stack)

LaRiffle:
@murarugeorgec is willing to take this one

PR #2728, opened 2019-11-10
Beaver triple gen

gmuraru:
Fix for #2708

What changed:
* Flatten a, b, c before concatenating - this was needed since we can not simply stack the tensors if they have different dimensions


syft/frameworks/torch/crypto/beaver.py
```diff
@@ -27,7 +27,13 @@ def request_triple(
     a = torch.randint(field, a_size)
     b = torch.randint(field, b_size)
     c = cmd(a, b)
-    a_shared = a.share(*locations, field=field, crypto_provider=crypto_provider).child
-    b_shared = b.share(*locations, field=field, crypto_provider=crypto_provider).child
-    c_shared = c.share(*locations, field=field, crypto_provider=crypto_provider).child
+
+    res = torch.cat((a.view(-1), b.view(-1), c.view(-1)))
```
murarugeorgec: @LaRiffle those should be moved to the cryptoProvider?

syft/frameworks/torch/crypto/beaver.py
```diff
@@ -27,7 +27,13 @@ def request_triple(
...
+
+    shares = res.share(*locations, field=field, crypto_provider=crypto_provider).child
+
+    a_shared = shares[: a.numel()].reshape(a_size)
+    b_shared = shares[a.numel() : -c.numel()].reshape(b_size)
+    c_shared = shares[-c.numel() :].reshape(c.shape)
+
```
midokura-silvia: Is there a unit test testing this code?

gmuraru: There are already tests that are testing the ```request_triple``` method - should I add more?

midokura-silvia: if it is already tested, it's fine, thanks.

gmuraru:
I will continue working on the first item from @LaRiffle options and then come back to this issue.

iamtrask:
> I will continue working on the first item from @LaRiffle options and then come back to this issue.

Looks like this is unblocked - great work on https://github.com/OpenMined/PySyft/pull/2741!

syft/exceptions.py
```diff
@@ -273,11 +273,11 @@ def __init__(self, obj_id, worker):
             + " which does not exist!!! "
         )
         message += (
-            "Use .send() and .get() on all your tensors to make sure they're"
+            "Use .send() and .get() on all your tensors to make sure they're "
```
gmuraru: Not related to the issue. When a string is split on multiple lines like this the printed resulted string would be all the chunks concatenated - need to add space between words :)

PR #2735, opened 2019-11-15
Enable CUDA processing in PySyft

midokura-silvia:
continued from #2690

Note that Travis is not testing the cuda part as cuda is (currently) not available.


iamtrask:
This is indeed possible - but I wouldn't want to require this to run PySyft's tests by default. https://stackoverflow.com/questions/3087361/gpu-emulator-for-cuda-programming-without-the-hardware

andrelmfarias:
I know from a PyTorch contributor it is possible to test in GPU with CircleCI, but we would need a special plan for it.

PR #2739, opened 2019-11-16
Cleanup directory names

iamtrask:
With syft.frameworks.torch, the directory names have becomee disorganized
and don't seem to follow any set standard. Also, some are simply too long
and cumbersome to be typing out all the time (such as differential_privacy).

I have replaced long names with their shorthand options, federated leearning
to simply fl, differential_privacy to dp. I also changed crypto to mpc
since all of the algorithms contained within the folder exclusively
related to secure multi-party computation. If and when we implement
a wider varitey of crypto algorithms, we can reconsider this name. However,
at present, I would like for us to use a more descriptive directory name.


test/torch/federated/test_dataloader.py
```diff
@@ -1,14 +1,14 @@
 import torch as th
 import syft as sy
-from syft.frameworks.torch import federated
```
lc0: we can still do 
```python
import federated as fl
```

lc0:
shorter imports look better, but what about explicit is being better than implicit? 

I am afraid, that long term `fl`, `dp` and so on, not going to make it easier for users to use the library

iamtrask:
My hunch is that these shorthands are becoming industry standards (MPC for sure already is, DP as well, FL and HE well on their way)

iamtrask:
Alternatively we could offer both

PR #2740, opened 2019-11-17
PaillierTensor and Arg Type Matching (experimental)

iamtrask:
This PR adds support for Paillier homomorphic encryption by wrapping the python-paillier library (https://python-paillier.readthedocs.io/en/develop/)

It also adds experimental support for performing operations between two tensors which don't have the same tensor chains.

Details in commit messages.

TODO: serde - but we shouldn't block this PR for it.


syft/frameworks/torch/tensors/interpreters/paillier.py
```diff
@@ -0,0 +1,296 @@
...
+        """This method will encrypt each value in the tensor using Paillier
+        homomorphic encryption.
+
+        Args:
+            *public_key a public key created using
+                syft.frameworks.torch.he.paillier.keygen()
+        """
+
+        inputs = self.child.flatten().tolist()
+        new_child = sy.pool().map(public_key.encrypt, inputs)
```
mihaitodor: I'm probably out of my depth here, since I don't yet know much about how this library works and how it's used, but this can be made considerably faster by precomputing the results of [obfuscate](https://github.com/data61/python-paillier/blob/955f8c0bfa9623be15b75462b121d28acf70f04b/phe/paillier.py#L596) (maybe in the background, in a separate thread / service) and then passing them to the `r_value` parameter of [encrypt](https://github.com/data61/python-paillier/blob/955f8c0bfa9623be15b75462b121d28acf70f04b/phe/paillier.py#L144). Even so, [this operation](https://github.com/data61/python-paillier/blob/955f8c0bfa9623be15b75462b121d28acf70f04b/phe/paillier.py#L136) should be precomputed as well, but it looks like the library was not designed that way.

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -812,3 +813,28 @@ def combine(self, *pointers):
         ps.append(self)
 
         return syft.combine_pointers(*ps)
+
+    def encrypt(self, public_key):
+        """This method will encrypt each value in the tensor using Paillier
+        homomorphic encryption.
+
+        Args:
+            *public_key a public key created using
```
LaRiffle: docstring style Google

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -812,3 +813,28 @@ def combine(self, *pointers):
...
+        x2 = PaillierTensor().on(x)
+        x2.child.encrypt_(public_key)
+        return x2
+
+    def decrypt(self, private_key):
+        """This method will decrypt each value in the tensor, returning a normal
+        torch tensor.
+
+        Args:
+            *private_key a private key created using
```
LaRiffle: docstring style Google

syft/frameworks/torch/tensors/interpreters/paillier.py
```diff
@@ -0,0 +1,296 @@
...
+        """
+        super().__init__(id=id, owner=owner, tags=tags, description=description)
+        print("creating paillier tensor 2")
+
+    def encrypt(self, public_key):
+        """This method will encrypt each value in the tensor using Paillier
+        homomorphic encryption.
+
+        Args:
+            *public_key a public key created using
```
LaRiffle: docstring style Google

syft/frameworks/torch/tensors/interpreters/paillier.py
```diff
@@ -0,0 +1,296 @@
...
+
+        Args:
+            owner: An optional BaseWorker object to specify the worker on which
+                the tensor is located.
+            id: An optional string or integer id of the PaillierTensor.
+        """
+        super().__init__(id=id, owner=owner, tags=tags, description=description)
+        print("creating paillier tensor 2")
+
+    def encrypt(self, public_key):
```
LaRiffle: Will you use this one?

syft/frameworks/torch/tensors/interpreters/paillier.py
LaRiffle: Agreed,
is this more a python-paillier issue or a Pysyft issue?

syft/frameworks/torch/tensors/interpreters/paillier.py
```diff
@@ -0,0 +1,296 @@
...
+
+        data = np.array(new_child).reshape(self.child.shape)
+        self.child = data
+        self.pubkey = public_key
+
+    def decrypt(self, private_key):
+        """This method will decrypt each value in the tensor, returning a normal
+        torch tensor.
+
+        =Args:
```
LaRiffle: dosctring

syft/generic/frameworks/hook/hook.py
```diff
@@ -371,24 +371,63 @@ def overloaded_native_method(self, *args, **kwargs):
             """
 
             if not hasattr(self, "child"):  # means that it's not a wrapper
+
+                # if self is a natural tensor but the first argument isn't,
+                # wrap self with the appropriate type and re-run
+                if len(args) > 0 and hasattr(args[0], "child"):
```
LaRiffle: hasattr(args[0], "child") .  -> args[0].is_wrapper ?

syft/generic/frameworks/hook/hook.py
```diff
@@ -371,24 +371,63 @@ def overloaded_native_method(self, *args, **kwargs):
...
+
+                    # if we allow this for PointerTensors it opens the potential
+                    # that we could accidentally serialize and send a tensor in the
+                    # arguments
+                    if not isinstance(args[0].child, PointerTensor):
+                        self = type(args[0].child)().on(self, wrap=True)
+                        args = [args[0]]
+                        return overloaded_native_method(self, *args, **kwargs)
```
LaRiffle: Rename
arg = args[0]

Simplify
```
args = [args[0]]
return overloaded_native_method(self, *args, **kwargs)
```
`return overloaded_native_method(self, arg, **kwargs)`
should be the same

syft/generic/frameworks/hook/hook.py
```diff
@@ -371,24 +371,63 @@ def overloaded_native_method(self, *args, **kwargs):
...
+                            self.child, PointerTensor
+                        ):
+                            # TODO: add check to make sure this isn't getting around a security class
+
+                            _args = list()
+                            _args.append(type(self)().on(args[0], wrap=False))
+                            for a in args[1:]:
+                                _args.append(a)
+
+                            args = _args
```
LaRiffle: can you simplify this, like above?

syft/frameworks/torch/tensors/interpreters/paillier.py
```diff
@@ -0,0 +1,296 @@
...
+
+    # Module & Function overloading
+
+    # We overload two torch functions:
+    # - torch.add
+    # - torch.nn.functional.relu
+
+    @staticmethod
+    @overloaded.module
+    def torch(module):
```
LaRiffle: Clean the docstring as well if you can

syft/frameworks/torch/tensors/interpreters/paillier.py
```diff
@@ -0,0 +1,296 @@
...
+
+        # if self is not encrypted and args[0] is encrypted
+        if isinstance(self.child, th.Tensor):
+            out.child = self.child.numpy().dot(args[0].child)
+
+        # if self is encrypted and args[0] is not encrypted
+        else:
+            out.child = self.child.dot(args[0])
+
+        return out
```
LaRiffle: I think mm = __ matmul __ 
so you can add:
`__matmul__ == mm` to add support for `x @ y`

syft/frameworks/torch/tensors/interpreters/paillier.py
mihaitodor: I'd say both, given the limited design of `python-paillier`, but it depends on the intended usage of PySyft. If PySyft needs to be as fast as possible when calling the python-paillier `encrypt` function, then this is one way to speed it up considerably, at the expense of pre-computing randomisers separately in advance. Of course, this would have to be implemented in PySyft somehow.

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: I was wondering if `encrypt` shoudln't be the entry point both for HE and MPC, you'd have a method argument which is equal to "he" or "mpc" and depending on which you call the he logic you wrote or the .share (and possibly the .fix_prec as well) ; you'd add the other options for encryption (public_key, or workers, or cryptoprovider also)

syft/frameworks/torch/tensors/interpreters/paillier.py
iamtrask: That's a great future work item! :)

syft/frameworks/torch/tensors/interpreters/native.py
iamtrask: I dunno - I like the distinction between encrypting something and sharing something. If I'm reading code, it's a lot easier to read the difference between the words ".share() and .encrypt() than .encrypt(bobs_key, proto=he) and .encrypt(bobs_worker, alices_worker, proto=mpc). The first you know immediately the purpose fo the function. The second you don't find out until the end (classic example for an increase in cognitive load)

syft/frameworks/torch/tensors/interpreters/paillier.py
iamtrask: Agreed - removed

syft/generic/frameworks/hook/hook.py
iamtrask: Agreed

iamtrask: sometimes args[0] is an int

iamtrask: so gotta do hasattr

syft/generic/frameworks/hook/hook.py
iamtrask: done

syft/generic/frameworks/hook/hook.py
iamtrask: i don't think i can

PR #2741, opened 2019-11-17
Create tensor directly on worker - zeros, tensor, rand, randn

### LINKED ISSUE #2678: <worker>.torch

gmuraru:
Fixes #2678

Added hooks for ```torch.zeros```, ```torch.rand``` and ```torch.tensor``` directly into the ```BaseWorker``` such that we could do:
```
alice = sy.VirtualWorker(id="alice", hook=hook)
x_ptr = alice.tensor([1,2,3]) <- create the tensor [1,2,3] at alice and return a pointer
```

### Tests
[x] Added unit tests for VirtualWorker, BaseWorker and WebsocketClient
[x] Unit tests for whitelisted and not whitelisted methods
[x] Add a simple way to extend the ```remote``` call with other frameworks (like ```tensorflow```)


iamtrask:
Great to see you working on this project!

LaRiffle:
Great job, ping me when you're ready for a review!

test/workers/test_base.py
```diff
@@ -95,3 +95,30 @@ def test_enable_registration_with_ctx(hook):
     with hook.local_worker.registration_enabled():
         hook.local_worker.is_client_worker == False
     assert hook.local_worker.is_client_worker == True
+
+
+def test_send_command_whitelist(hook):
```
gmuraru: Q: @LaRiffle should I also copy those two tests in the ```test_websocket_worker```

LaRiffle: would be a good idea!
just an info:
prefer to do:
```
def test_send_command_whitelist(workers):
    bob  = workers["bob"]
    ...
```

syft/workers/base.py
```diff
@@ -163,6 +154,20 @@ def __init__(
                 # self is the to-be-created local worker
                 self.add_worker(self)
 
+        if hook is None:
+            self.framework = None
+        else:
+            # TODO[jvmancuso]: avoid branching here if possible, maybe by changing code in
+            #     execute_command or command_guard to not expect an attribute named "torch"
+            #     (#2530)
+            self.framework = hook.framework
```
gmuraru: Had to move those here because of the following scenario.
1. In the tests we spin up the ```Websocket Server``` and ```Websocket Client```
2. Create one client and get a ```ConnectionRefused```
3. Create the second client and for this one ```line 149``` will run, which will override the ```remote``` attribute.
4. ```Remote``` contains the ```torch``` attribute (is a ```Torch``` instance - class that can be found in the hooks for the torch framework)  and the ```Torch``` instance contains an attribute ```worker``` that is a weak reference to ```self```.
The problem is that remote would get copied into the new ```WebsocketClient``` and we will have a reference to an item that should be garbage collected + ```ws``` attribute from ```WebsocketClient``` would be ```None``` (since we got a ```Connection Refused``` with the first client) --- there will be generated an error caused by the function ```_forward_message``` because ```self.remote.torch.worker.ws``` <--- the one that was refused is```None```

iamtrask:
This PR is looking really quite good to me. Fantastic work George! You're really spinning up on the codebase quite quickly!

iamtrask:
Would love to add .randn as well, but no need to block this PR to modify the list of supported methods (it's the generic ability that's the hard part)

PR #2755, opened 2019-11-24
[MRG] Notebook test

fdroessler:
Automated testing of Notebooks using Papermill. Reopened this after some Repo problems.


fdroessler:
Currently the following notebooks are not running:
- `Part 10 - Federated Learning with Secure Aggregation.ipynb`
Reason: 
Details see: https://openmined.slack.com/archives/C6EEFN3A8/p1572184769145600
But it seems that PyTorch does not implement log2 for LongTensors and this causes the following error:
```
~/git/PySyft/syft/frameworks/torch/hook/hook_args.py in <lambda>(i)
     32 backward_func = {
     33     TorchTensor: lambda i: i.wrap(),
---> 34     torch.Tensor: lambda i: i.wrap(),
     35     torch.nn.Parameter: lambda i: torch.nn.Parameter(data=i),
     36     AutogradTensor: lambda i: AutogradTensor(data=i).on(i, wrap=False),

AttributeError: 'numpy.ndarray' object has no attribute 'wrap'
```

- `Part 13b - Secure Classification with Syft Keras and TFE - Secure Model Serving.ipynb` and `Part 13c - Secure Classification with Syft Keras and TFE - Private Prediction Client.ipynb`
Reason:
It seems like the model sharing using syft keras for some reasons blocks such that the serving of the model in Part 13b can't be done thus Part 13c can't be run either.

fdroessler:
In the advanced folder the following notebooks are currently not tested:

- `Federated Recurrent Neural Network.ipynb`
Reason:
I am not 100% sure but my suspicion is that the `model.copy()` command does not do what the notebook thinks it does. So the in the evaluation cell (last cell) the model is received once via `model.copy().get()` and in the second occurrence it does not exist anymore and it throws an error that the object with that ID does not exist anymore.

- `Build your own tensor type (advanced).ipynb`
Rason:
This notebook requires modification of the source code. I am not sure what the best way of testing this is and also it is not 100% clear if the notebook is up-to-date.

fdroessler:
Update:
For the notebook:
- `Federated Recurrent Neural Network.ipynb` I now modified the notebook in order to first `get` the models and then send `copies` of them to workers. This seems to fix the issue from before but I am not sure if I am just working around a bug or if this has actually changed.

Maybe @DanyEle could have a look?

DanyEle:
Hi @fdroessler . Thank you for spotting the issue out! Indeed, model.get() extracts the model from the remote worker, so calling .get() once again will raise an error, whereas .copy().get() does not raise such an error when using it multiple times. Did this behaviour change? That was my reason for using .copy().get().

Your fix sounds good, BTW!

midokura-silvia:
@fdroessler Your PR should merge to the branch dev and not to master. 
I have the impression that papermill is more powerful than treon, so let's go ahead with papermill.

fdroessler:
Great I'll clean this up tomorrow and finish it.

fdroessler:
@iamtrask @midokura-silvia There might be something strange going on. Not sure how it is supposed to be but I think because the Virtualworkers in the hook stay for a test session it means I can't create new Websocketworkers with the same name. I am now removing them from the hook before adding new ones. Will see if that creates issues down the line. Tried to use a different hook but I think it will not create a new one but just provide me with the existing hook?

fdroessler:
I'll open a couple of issues with reference to missing tests or notebooks that need fixing.

LaRiffle:
I just have a question related to the modifications of the tutorials: they are many of them, but most of them are output deletion for example, are they needed?

fdroessler:
Not necessarily. I prefer it this (cleaned) to avoid major merge conflicts when content changes but am happy to revert these changes if you prefer the content to be included.

lc0:
> I just have a question related to the modifications of the tutorials: they are many of them, but most of them are output deletion for example, are they needed?

Sometimes, such output allows users to read notebooks without downloading. By stripping output we might loose ease of read. What do you think?

examples/tutorials/Part 06 - Federated Learning on MNIST using a CNN.ipynb
```diff
@@ -1,5 +1,18 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "tags": [
+     "parameters"
+    ]
+   },
+   "outputs": [],
+   "source": [
+    "epochs = 10"
```
lc0: shall we make it as a global variable to make a bit more transparent?

test/notebooks/test_notebooks.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def test_notebooks_basic(isolated_filesystem):
+    """Test Notebooks in the tutorial root folder."""
+    notebooks = glob.glob("*.ipynb")
+    for notebook in notebooks:
+        if Path(notebook).name in exclusion_list_basic:
+            continue
+        res = pm.execute_notebook(
+            notebook, "/dev/null", parameters={"epochs": 1, "n_test_batches": 5}, timeout=300
```
lc0: this is great, that we can just pass such parameter! 👍

test/notebooks/test_notebooks.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+        assert isinstance(res, nbformat.notebooknode.NotebookNode)
+
+
+def test_notebooks_advanced(isolated_filesystem):
```
lc0: Why do you need to separate these? besides the path

arturomf94:
Hey, I found that the following tests are failing. I think it might have to do with this PR.
```python
test/notebooks/test_notebooks.py::test_notebooks_basic FAILED

test/notebooks/test_notebooks.py::test_notebooks_advanced FAILED

test/notebooks/test_notebooks.py::test_fl_sms FAILED
```
Apologies in advance if this due to my specific installation, but these are the only failing tests so thought it might be useful to comment on here.

test/notebooks/test_notebooks.py
fdroessler: good question, I did initially because the tests where hard to debug and taking long. But I was thinking to work with a more general approach soon because that way new notebooks are discovered and don't have to be added manually.

fdroessler:
> Hey, I found that the following tests are failing. I think it might have to do with this PR.
> 
> ```python
> test/notebooks/test_notebooks.py::test_notebooks_basic FAILED
> 
> test/notebooks/test_notebooks.py::test_notebooks_advanced FAILED
> 
> test/notebooks/test_notebooks.py::test_fl_sms FAILED
> ```
> 
> Apologies in advance if this due to my specific installation, but these are the only failing tests so thought it might be useful to comment on here.

Could you be a bit more specific? Where are they failing, in a PR or just locally?

arturomf94:
Yes - I followed the instructions outlined [here](https://github.com/OpenMined/PySyft/blob/master/CONTRIBUTING.md). Specifically, it failed during the [installation](https://github.com/OpenMined/PySyft/blob/master/CONTRIBUTING.md#installing-pysyft-after-cloning-repository). This was the outcome of running `python setup.py test` locally as indicated [here](https://github.com/OpenMined/PySyft/blob/master/INSTALLATION.md#6-test-your-installation).

fdroessler:
I replied on Slack.

test/notebooks/test_notebooks.py
lc0: I though it's because you wanna give different timeout or so, but I do not think it's currently the case :)

PR #2759, opened 2019-11-26
[WIP]Support negative value for shares

### LINKED ISSUE #2705: Transform shares to support negative values
For AdditiveSharing and FixedPrecision, if the field is 2**64, the values of the sahres should lie in `[-2**63, 2**63-1]` and not in `[0, 2**64]`. It allows for automatic wrapping for LongTensor and `field=2**63`.

Yugandhartripathi:
Closes #2705 
Changes:
 - Updated Tutorial 9 basic example to support negative shares
 - Changed generate_shares in additive_shared.py to generate in range(-field/2,field/2 - 1) instead of (0,field)


iamtrask:
CC: @LaRiffle

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -198,7 +198,7 @@ def generate_shares(secret, n_workers, field, random_type):
         random_shares = [random_type(secret.shape) for _ in range(n_workers - 1)]
 
         for share in random_shares:
-            share.random_(field)
+            share.random_(int(-field / 2), int(field / 2) - 1)
```
gmuraru: Q: Should we give an warning if the field values has an odd value?
In that case the random interval would be smaller.

Yugandhartripathi: I don't think we need a warning since we are working in Finite Field of 2 elements i.e field will always have a value of the form 2^m.

gmuraru: In that case it should be fine :)

Issue #2510, opened 2019-08-15
Add documentation to the existing code base


kakirastern:
**Is your feature request related to a problem? Please describe.**
It would be great, as suggested by @iamtrask, to add documentation to the existing code base so that the code would be easier to understand, especially to new users. 

**Describe the solution you'd like**
Ideally this tasks should be broken down into smaller PR's focusing on the different parts of the code base, such as the `serde` utility, or the tests. 

**Additional context**
Would be a great way to contributors to gain familiarity with the `PySyft` code base.

iamtrask:
Note that MANY people can pick up this issue. We will likely never close it. 

Steps:

1) fork the repo into your own github
2) clone your fork down to your local filesystem
3) look for a part of the codebase that isn't documented
4) read the code and figure out what's going on (this is the hard part)
5) add inline documentation using the correct style
6) commit and push your code to your fork
7) create a pull request back into the main PySyft (which you can do using th big green button on Github"

libratiger:
> Note that MANY people can pick up this issue. We will likely never close it.
> 
> Steps:
> 
> 1. fork the repo into your own github
> 2. clone your fork down to your local filesystem
> 3. look for a part of the codebase that isn't documented
> 4. read the code and figure out what's going on (this is the hard part)
> 5. add inline documentation using the correct style
> 6. commit and push your code to your fork
> 7. create a pull request back into the main PySyft (which you can do using th big green button on Github"

since the hard part is step 4. Is there any detail arch for this part?

PR #2763, opened 2019-12-01
added inline documentation 

abogaziah:
added inline documentation as was suggested in issue #2510 to keras test file


abogaziah:
@LaRiffle done!

iamtrask:
Well done! Thank you for contributing documentation - we need more PRs like yours!

Issue #2707, opened 2019-10-28
Deactivate Plans in Plans


LaRiffle:
Plans should not be executed as Plans when they are included in another Plans. This allows to easily extend Plan or to cut Plans in small modules which can also be run individually as Plans.

Examples:
```
@func2plan
def foo1(x):
     return x + 1

@func2plan
def foo2(x):
     y = foo1(x)
     return y * 2
```

Idea: create a variable global at worker scale when a plan is being built to force other plans to run as normal functions.

gmuraru:
@LaRiffle tried the following code - in the ```Plans``` notebook:

```
@func2plan()
def plan_double_abs(x):
    x = x + x
    x = torch.abs(x)
    return x

@func2plan()
def plan_test(x):
    f = plan_double_abs(x)
    return f*2
```
and from what I checked,  if I run ```plan_test.build(torch.tensor([1., -2.]))``` it will simply call the build method for the ```plan_test``` (will not call ```build``` for ```plan_double_abs``` even thought it is called inside ```plan_test```)
and after that I ran:
```
pointer_plan = plan_test.send(device_1) # device_1 is a VirtualWorker
pointer_to_data = device_1.search('input_data')[0]
pointer_to_result = pointer_plan(pointer_to_data)
```

Also, added a print message in the ```execute_commands``` methpd from the ```Plan``` class:
```
for message in self.procedure.operations:
    print("message ", message)
    bin_message = sy.serde.serialize(message, simplified=True)
    _ = self.owner.recv_msg(bin_message)
```
The messages generated are:
```
message  (33, ((6, ((5, (b'__add__',)), (25, (98142542040, 78527494689, b'device_1', None, (11, (2,)), True)), (6, ((25, (98142542040, 78527494689, b'device_1', None, (11, (2,)), True)),)), (0, ()))), (70741324193,)))
message  (33, ((6, ((5, (b'torch.abs',)), None, (6, ((25, (25994186521, 70741324193, b'device_1', None, None, True)),)), (0, ()))), (91046882485,)))
message  (33, ((6, ((5, (b'__mul__',)), (25, (14581124267, 91046882485, b'device_1', None, None, True)), (6, (2,)), (0, ()))), (82141830512,)))
```
From what it seems ```__add__``` and ```torch.abs``` are from ```plan_double_abs``` and the ```__mul__``` from the ```plan_test``` (from the return).

What I think is happening is that the ```plan_double_abs``` plan is not building the plan inside.
[...]

PR #2791, opened 2019-12-07
Deactivate plans in plans

gmuraru:
### Description
Solves #2707.

Building a plan that has already a built plan should directly call the function - take into consideration ```*args and **kwargs```.

For doing this, we keep in the owner a state, which would indicate if a plan is already building or not. If a plan is already building, then directly call the function.

### Later EDIT
Made this PR to support plans in plans - even if the inner plan is already built (with a state) - for that I had to add more information that is sent to the worker (the states for the first plans)

### DIFF
--- a/syft/frameworks/torch/hook/hook.py
+++ b/syft/frameworks/torch/hook/hook.py
@@ -32,7 +32,6 @@ from syft.messaging.plan import Plan
 from syft.messaging.promise import Promise
 
 from syft.exceptions import route_method_exception
-from syft.exceptions import TensorsNotCollocatedException
 
 
 class TorchHook(FrameworkHook):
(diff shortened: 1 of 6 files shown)


syft/frameworks/torch/hook/hook.py
```diff
@@ -32,7 +32,6 @@
 from syft.messaging.promise import Promise
 
 from syft.exceptions import route_method_exception
-from syft.exceptions import TensorsNotCollocatedException
```
gmuraru: I think this was not needed

test/message/test_plan.py
```diff
@@ -1046,3 +1046,158 @@ def plan_abs(data):
...
+        def plan_2_1(data, state):
+            (bias,) = state.read()
+            return -2 * plan_3(data) + bias
+
+        @sy.func2plan()
+        def plan_1(data):
+            res = plan_2_1(data)
+            return plan_2_2(res)
+
+        # (-2 * (x + tensor(3) + 42) + tensor(2) + 1331)
```
gmuraru: I can remove this - It was only a simple way that I thought of to show what each plan/function is doing

PR #2799, opened 2019-12-09
Add backward to the ambiguous meth list in hook_args

LaRiffle:


syft/frameworks/torch/hook/hook_args.py
```diff
@@ -47,15 +47,16 @@
 
 ambiguous_methods = {
```
iamtrask: I don't know what this list is for. Can we add a comment?

gmuraru: I think we can take the comment from ```hook_args.py```?

```
 # Methods or functions whose signature changes a lot and that we don't want to "cache", because
 # they have an arbitrary number of tensors in args which can trigger unexpected behaviour
 ambiguous_methods = set()
 ambiguous_functions = set()
```

iamtrask: perfect

LaRiffle: Adressed

PR #2803, opened 2019-12-10
Add prototype of parameters() for pointer plans

LaRiffle:
This requires to fix hook_args registration

Usa case targeted:
```python
import torch
from torch import nn
from torch import optim
import syft as sy
hook = sy.TorchHook(torch)
sy.local_worker.is_client_worker = False # To register objects locally
# A Toy Dataset
data = torch.tensor([[0,0],[0,1],[1,0],[1,1.]], requires_grad=True).send(bob)
target = torch.tensor([[0],[0],[1],[1.]], requires_grad=True).send(bob)
# A Toy Model
lin = nn.Linear(2,1)
class Net(sy.Plan): 
    def __init__(self):
        super(Net, self).__init__()
        self.fc1 = nn.Linear(2,1)
    def forward(self, x):
        x = self.fc1(x)
        return x
model = Net()
opt = optim.SGD(params=model.parameters(),lr=0.1)
model.build(torch.tensor([[0,0]]))
model = model.send(bob)
def train():
    # Training Logic
    for iter in range(20):
        opt.zero_grad()
        # 2) make a prediction
        pred = model(data)
        loss.backward()
        # 5) change those weights
        opt.step()
        # 6) print our progress
        print(loss.get().data)
train() 
```


syft/generic/pointers/pointer_plan.py
```diff
@@ -93,6 +93,26 @@ def __call__(self, *args, **kwargs):
 
         return response
 
+    def parameters(self) -> List:
+        """Return a list of pointers to the plan parameters"""
+
+        assert (
+            len(self._locations) == 1
+        ), ".parameters() for PointerPlan with > 1 locations is currently not implemented."
+        # TODO implement this feature using MultiPointerTensor
```
gmuraru: Q: I have an idea, but I do not know if it is a good one or not. What do you think of linking issues directly in the code? (like create the issue and put the issue number as comment)

LaRiffle: Definitely a good idea :)

matthiaslau: I don‘t think it is a good idea to link github issues in code, this should be separated. Not every part could be matched to one issue over time. And imagine e.g. moving to gitlab ;-).

karlhigley: Yeah, I don't think every part could be linked to an issue (that's what the history and `git blame` are for anyway), but it might be helpful in `TODO` comments to be able to link to an issue with further info and discussion so that stuff doesn't have to live in the comments. I have no strong opinions here, just wanted to chime in to say that I can see potential value. 🤷‍♂

matthiaslau:
This PR is not fixing the problem yet. The use case mentioned above is still creating the optimizer with the plan parameters. When moving the optimizer creation after the plan was sent to Bob to use the PlanPointer this fails as optimizer cannot be created on TensorPointers :-/. I don‘t have a good solution to this yet (see slack).

syft/generic/pointers/pointer_plan.py
iamtrask: I 100% support TODO: <link to issue>

iamtrask:
> This PR is not fixing the problem yet. The use case mentioned above is still creating the optimizer with the plan parameters. When moving the optimizer creation after the plan was sent to Bob to use the PlanPointer this fails as optimizer cannot be created on TensorPointers :-/. I don‘t have a good solution to this yet (see slack).

Should we merge?

LaRiffle:
Ok this should fix it!

Issue #2792, opened 2019-12-07
Translate our tutorials to Romanian


gmuraru:
"At present, our tutorial notebook series is only available in English, but it would be very desirable for people to be able to learn PySyft in their native language, such as Romanian.

For this PR, copy the current notebooks and translate the inline text to Romanian, placing the notebooks in a "romanian" folder within the "tutorials" folder." (Andrew Trask, s/Hindi/Romanian)

https://github.com/OpenMined/PySyft/tree/master/examples/tutorials

PR #2806, opened 2019-12-10
Add tutorial 1 in Romanian language

gmuraru:
Fixes (partially) #2792 .

Translate the first tutorial to Romanian.


fdroessler:
Hi, I have been working on the test suit for all notebooks. I have just updated it but the PR is not merged yet. However, I have now moved all the translations into a subfolder so maybe you could do the same?

As in move the `romanian` folder into a subfolder of tutorials:
`examples/tutorials/translations/romanian/`

That way once both of our PRs are merged, yours should be automatically included in the test suit.

PR #2807, opened 2019-12-10
Fixed installation command in Windows

DanyEle:
Simply removed the 'udacity' keyword in the installation, which is not currently recognized by the installation procedure.


INSTALLATION.md
```diff
@@ -157,7 +157,7 @@ First, you need to enter the PySyft folder you cloned earlier:
 
 Then
 
-`python setup.py install udacity`
+`python setup.py install`
```
darkmatter18: You noticed it good. I also thought udacity is not needed here.

DanyEle:
@LaRiffle  Since it's just one line, I guess we can merge it?

PR #2812, opened 2019-12-11
Fix serde inconsistencies

vvmnnnkv:
This PR follows discussion in issue #2654 and tries to make serde simplification consistent across all types:
* Fix missing simplifications, e.g. all str should be simplified/detailed consistently by _simplify_str/_detail_str
* Use same simplified data structure for all types: `(CODE, (DATA,))` that you can decode as `code, data = simplified_tuple`
* Simplify list that wraps operations in Procedure to avoid "magic parens"

There's a case when worker_id in Procedure's operations needs to be updated from int to str type. Procedure had to be fixed for that, given that str is now simplified.


syft/messaging/plan/procedure.py
```diff
@@ -74,16 +74,30 @@ def update_ids(
             from_worker: The previous worker that built the plan.
             to_worker: The new worker that is running the plan.
         """
+
+        # We operate on simplified content of Operation, hence all values should be simplified
+        from_workers_simplified = None
+        to_workers_simplified = None
+        if from_worker and to_worker:
```
gmuraru: Q: We will always have ```from_worker``` and ```to_worker```. Is there any case only one is ```!= None```?

syft/messaging/plan/procedure.py
```diff
@@ -74,16 +74,30 @@ def update_ids(
...
+            from_workers_simplified = [sy.serde._simplify(None, from_worker)]
+            to_workers_simplified = [sy.serde._simplify(None, to_worker)]
+
+        from_ids_simplified = None
+        to_ids_simplified = None
+        if len(from_ids) and len(to_ids):
```
gmuraru: Q: The same case as above for here

syft/serde/native_serde.py
```diff
@@ -152,12 +172,14 @@ def _simplify_dictionary(worker: AbstractWorker, my_dict: Dict) -> Tuple:
     pieces = list()
     # for dictionaries we want to simplify both the key and the value
     for key, value in my_dict.items():
-        pieces.append((serde._simplify(worker, key), serde._simplify(worker, value)))
+        pieces.append(
+            (serde._simplify(worker, key), serde._simplify(worker, value) if not shallow else value)
```
gmuraru: Q: Can't we simply send shallow as a parameter to ```serde_simplify```

syft/serde/native_serde.py
```diff
@@ -334,9 +355,9 @@ def _detail_ndarray(
...
 
     """
-    res = numpy.frombuffer(arr_representation[0], dtype=arr_representation[2]).reshape(
-        arr_representation[1]
-    )
+    arr_shape = serde._detail(worker, arr_representation[1])
+    arr_dtype = serde._detail(worker, arr_representation[2])
+    res = numpy.frombuffer(arr_representation[0], dtype=arr_dtype).reshape(arr_shape)
 
     assert type(res) == numpy.ndarray
```
gmuraru: Q: Shouldn't we remove this assert? We should get an assert if ```numpy.frombuffer``` will fail, right?

syft/serde/torch_serde.py
```diff
@@ -326,30 +310,25 @@ def _detail_torch_parameter(worker: AbstractWorker, param_tuple: tuple) -> torch
...
+    return (obj.save_to_buffer(),)
 
 
-def _detail_script_module(worker: AbstractWorker, script_module_bin: str) -> torch.jit.ScriptModule:
+def _detail_script_module(
+    worker: AbstractWorker, script_module_bin: Tuple
+) -> torch.jit.ScriptModule:
     """"Strategy to deserialize a binary input using Torch load"""
-    script_module_stream = io.BytesIO(script_module_bin)
+    script_module_stream = io.BytesIO(script_module_bin[0])
```
gmuraru: Q: We should get here with a ```len(script_module_bin)``` of size 1, right? I am thinking if we should add an assert of not.

test/test_serde_full.py
```diff
@@ -1318,7 +1328,7 @@ def compare(detailed, original):
 # syft.frameworks.torch.tensors.interpreters.promise.PromiseTensor
 def make_promisetensor(**kwargs):
     pt = syft.Promise.FloatTensor(shape=torch.Size((3, 4)))
-    pt.tags = ["tag1"]
+    pt.tag("tag1")
     pt.description = "I promise"
```
gmuraru: Q: Shouldn't we do the same for ```description```? Maybe ```pt.describe("I promise")```.

Also, when is ok to use ```tag``` vs ```tags```?
Should ```tags``` be used for TorchTensors (since we know we are working with tensors) and ```tag``` for ```AbstractObject``` (and we do not know the underlying object?)?

syft/messaging/plan/procedure.py
vvmnnnkv: I think replacement doesn't make sense when any of them are `None`.
This was existing code, I just moved it out of the loop to avoid making simplification each time.

syft/messaging/plan/procedure.py
vvmnnnkv: Same as above :)

syft/serde/torch_serde.py
vvmnnnkv: Yes. I haven't seen asserts in similar cases, e.g. in `_detail_str`, `_detail_range`, `_detail_ndarray`, etc.
Do you think they should be added?

test/test_serde_full.py
vvmnnnkv: Thanks! I've updated to `.describe()` everywhere.
I think I shouldn't have used `.tags` and `.description` properties directly at all. Just didn't know about them 🤷‍♂

syft/serde/torch_serde.py
gmuraru: I think it is ok how it is now.

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -987,7 +987,12 @@ def simplify(worker: AbstractWorker, tensor: "AdditiveSharingTensor") -> tuple:
         # Don't delete the remote values of the shares at simplification
         tensor.set_garbage_collect_data(False)
 
-        return (tensor.id, tensor.field, tensor.crypto_provider.id, chain)
+        return (
+            tensor.id,
```
LaRiffle: You add a simplifier for the object_pointer id, but not here
Any reason for this?

syft/generic/pointers/pointer_protocol.py
```diff
@@ -92,15 +92,19 @@ def get(self, deregister_ptr: bool = True):
     @staticmethod
     def simplify(worker: AbstractWorker, ptr: "PointerPlan") -> tuple:
 
-        return (ptr.id, ptr.id_at_location, ptr.location.id, ptr.garbage_collect_data)
+        return (
+            ptr.id,
```
LaRiffle: same remark here. Actually I'm not sure ids should be simplified as they are integers and this only slow down the process

syft/serde/torch_serde.py
```diff
@@ -246,21 +243,8 @@ def _detail_torch_tensor(worker: AbstractWorker, tensor_tuple: tuple) -> torch.T
...
-            if isinstance(tag, bytes):
-                tag = tag.decode("utf-8")
-            tags[i] = tag
-        tensor.tags = tags
-
-    if description is not None:
-        if isinstance(description, bytes):
-            description = description.decode("utf-8")
-        tensor.description = description
+    tensor.tags = syft.serde._detail(worker, tags)
```
LaRiffle: :+1:

syft/generic/pointers/pointer_protocol.py
vvmnnnkv: Also confused by this.
In some places (e.g. PointerPlan, PointerProtocol, MultiPointerTensor) id is defined as `Union[str, int]` in constructor, in some places (e.g. FixedPrecisionTensor, LoggingTensor) it's mentioned in doc comment that it's str or int. But in parent classes AbstractTensor, AbstractObject, and sy.ID_PROVIDER.pop() that generates those id's, it seems to be just int.

I can make serde to consistently treat ID as int or int/str everywhere, but there might be a reason why it's different for different classes?

karlhigley: Just chiming in to say that I have encountered the dual nature of ids as both integers and strings, and I am also confused by this. Would be easier to handle if it were one or the other, but if it needs to be both, would be great to make that consistent (someday and probably not in this PR.)

vvmnnnkv: OK, I've made serde follow constructor (or parent constructor) arg type definition or documentation:
torch.nn.Parameter - int
torch.Tensor - int
AdditiveSharingTensor - int or str
FixedPrecisionTensor - int or str
CRTPrecisionTensor - int
LoggingTensor - int or str
MultiPointerTensor - int or str
Plan - int or str
Protocol - int *
PointerTensor - int or str
PointerPlan - int or str
PointerProtocol - int or str
ObjectWrapper - int
ObjectPointer - int or str
TrainConfig - int or str
AutogradTensor - int *
PrivateTensor - int or str
PromiseTensor - int or str

*serde supports int or str, didn't change that

PR #2813, opened 2019-12-11
Splitnn tutorial

H4LL:
These are two new tutorials being added to the splitNN set. These include multilayer split neural network and folded split neural network. A SplitNN class has also been defined.


fdroessler:
Was really looking forward to this functionality. Thanks a lot. BTW you also included the data in your commit. Maybe remove it because it is quite large

H4LL:
> Was really looking forward to this functionality. Thanks a lot. BTW you also included the data in your commit. Maybe remove it because it is quite large.
> 
> Also is the `Tutorial 1 - SplitNN Introduction.ipynb` the same as `SplitNN Introduction.ipynb`?

Thanks, I've been trying to get it to work for a while. It is the same, yeh. Just a different title to match in with the rest. Also, taken out the MNIST data, cheers!

PR #2816, opened 2019-12-12
Update pytorch requirement to >=1.3

### LINKED ISSUE #2730: PyTorch 1.3 installation: No matching distribution found
ERROR: Could not find a version that satisfies the requirement torch==1.3 (from syft[udacity]) (from versions: 0.1.2, 0.1.2.post1, 0.1.2.post2)
ERROR: No matching distribution found for torch==1.3 (from syft[udacity])

### LINKED ISSUE #2767: not able to install pysyft for pytorch version 1.3
Screenshot (3)

refactormyself:
This will resolve #2730 and close #2767 too. So that a direct edit of requirement.txt will no more be needed. The procedure will work as it is explained.


sandeepwww:
even for 1.3.1, I am seeing 
```
No local packages or working download links found for torch==1.3.1
```

iamtrask:
So one issue with this though is that when PyTorch releases a new version (without our heads up) ALL of our unit tests can start failing out of no-where. This can be very stressful for the dev teams, lol. 

Not 100% against it but the reason we fix to specific versions is that our hooking logic is prone to break version to version. We're quite tightly coupled to PyTorch even now.

refactormyself:
> even for 1.3.1, I am seeing
> 
> ```
> No local packages or working download links found for torch==1.3.1
> ```

@sandeepwww, I guess you probably have but just to confirm.
Did you double-check your PyTorch version? 
I had similar error until I explicitly did `conda install pytorch==1.3.1`

refactormyself:
> So one issue with this though is that when PyTorch releases a new version (without our heads up) ALL of our unit tests can start failing out of no-where. This can be very stressful for the dev teams, lol.
> 
> Not 100% against it but the reason we fix to specific versions is that our hooking logic is prone to break version to version. We're quite tightly coupled to PyTorch even now.

So then we fix it to `torch==1.3.1` right?

vvmnnnkv:
Maybe `torch~=1.3.0` can do the job?
It's equivalent of `>=1.3.0, ==1.3.*`.
patch-release should be backward-compatible and safe for update?

pip-dep/requirements.txt
```diff
@@ -5,7 +5,7 @@ lz4>=2.1.6
 msgpack>=0.6.1
 numpy>=1.16.0
 tblib>=1.4.0
-torch==1.3
+torch>=1.3
```
vvmnnnkv: Perhaps, use `torch==1.3.*` or `torch~=1.3.0` (which is probably same) to use latest torch version only from 1.3 stream.
However, I'd check pytorch updates policy to ensure they don't break things in patch-releases :)

karlhigley:
CI tests indicate that we're coupled to exactly PyTorch 1.3.0 (due to TorchVision.)

PR #2817, opened 2019-12-12
Fix bug with inplace operations on FPT

Jasopaum:
Fixed a strange bug when using `/=` and `*=` operations on FixedPrecisionTensor


test/torch/tensors/test_precision.py
```diff
@@ -279,6 +279,23 @@ def test_torch_div(workers):
     assert (z == torch.tensor([[3.0, 4.1], [1.0, 0.0]])).all()
 
 
+def test_inplace_operations():
```
gmuraru: Q: Could you also specify that those are arithmetic operations?
You could also have inplace bitwise operations, but I do not think if they are tested.
@iamtrask, @LaRiffle is there any requirements for bitwise operations?

iamtrask: Hard for me to ever say no to more tests! ;)

PR #2823, opened 2019-12-16
Update requirements.txt

iamtrask:
Update syft-proto name and point to PyPI instead of Github.


karlhigley:
This seems good, but there's no diff...Maybe these changes already made it into master?

PR #2805, opened 2019-12-10
[MRG] Improve notebook tests

fdroessler:
This PR is to further improve the notebook tests by adding the following:

- [x]  Add some notes to CONTRIBUTING.md
- [x]  Include translation notebooks in tests
- [x]  Add tests that checks if all notebooks are either ignored or tested
- [x]  Speed up test suit

PR #2827, opened 2019-12-16
fix: fix test_fl_sms on test_notebooks.py

arturomf94:
Currently it could fail with os.chdir since changing directory doesn't alter import paths, it changes the directory for opening files, according to [this SO answer](https://stackoverflow.com/questions/23619595/pythons-os-chdir-function-isnt-working). Instead, it should import the module with `sys.path.append` before using `os.chdir`.

The corresponding tests were failing on my machine (Linux) but works fine with this fix.

See original PR: https://github.com/OpenMined/PySyft/pull/2805


test/notebooks/test_notebooks.py
```diff
@@ -120,9 +120,10 @@ def test_fl_with_trainconfig(isolated_filesystem, start_remote_server_worker_onl
 
 def test_fl_sms(isolated_filesystem):
     sys.path.append("advanced/Federated SMS Spam prediction/")
-    os.chdir("advanced/Federated SMS Spam prediction/")
```
gmuraru: I think the best solution would be not to call ```os.chdir``` at all. I see [here](https://stackoverflow.com/questions/23619595/pythons-os-chdir-function-isnt-working) - the link you provided - that it has different behaviors on different operating systems.
What I am thinking is to add the ```preprocess.py``` path to ```sys``` and then simply run the notebook (how it is done now).

This will mean that on line 137, notebook will be ```p_name / notebook```.

What do you think?

arturomf94: Just for the record... Tried this with @gmuraru but it doesn't work so the best solution is the current one on the PR.

gmuraru:
@arturomf94 this was still necessary?

arturomf94:
Well it's better to have it this way, but currently it wouldn't change anything, since this test is currently skipped due to it failing during the Travis build because of memory issues.

gmuraru:
I can merge this after the travis tests are passing - the solution for all the problems will be that someone will start looking into the "SMS Tests"

PR #2836, opened 2019-12-18
Extract compression module from serde to simplify re-use

### LINKED ISSUE #2834: Extract compression from `serde` to separate module
Compression and serialization are currently intertwined with each other in the same module.

In order to pave the way for alternate serialization approaches (e.g. Protobuf), it would be helpful to extract compression out of `serde.py` and into its own module.

- _Combine all serialization approaches in the same module._ Seems potentially messy.
- _Put each serialization approach in its own module and duplicate compression._ Would work, but harder to maintain.

**Additional context**

karlhigley:
This makes the compression and de-compression easier to use again in a
second serialization approach (e.g. Protobuf.) Resolves #2834.

### DIFF
--- a/syft/serde/__init__.py
+++ b/syft/serde/__init__.py
@@ -2,5 +2,5 @@ from syft.serde.serde import *
 from syft.serde.torch_serde import *
 from syft.serde.serde import _simplify
 from syft.serde.serde import _detail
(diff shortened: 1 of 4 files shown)


karlhigley:
Rebased on master to resolve conflicts.

PR #2838, opened 2019-12-18
fix: fix problem with notebooks tests due to outdated library

arturomf94:
nbconvert seems to be causing problems due to it being an outdated
version.

See: https://github.com/nteract/papermill/issues/125


arturomf94:
It seems like this does not solve the problem with Travis, so this PR will be closed for now.

PR #2843, opened 2019-12-19
Extract a `msgpack` module from `serde`

karlhigley:
This sets up a situation where additional serialization methods
(like Protobuf) can each have their own module within `serde`.


syft/exceptions.py
```diff
@@ -174,9 +174,9 @@ def simplify(worker: "sy.workers.AbstractWorker", e):
         except AttributeError:
             attributes = {}
         return (
-            sy.serde._simplify(worker, tp.__name__),
-            sy.serde._simplify(worker, traceback_str),
-            sy.serde._simplify(worker, attributes),
+            sy.serde.msgpack.serde._simplify(worker, tp.__name__),
```
karlhigley: I don't love these long chains of module names, but my thinking is that it makes clear just how far into the internals of another module this code is reaching.

syft/serde/__init__.py
```diff
@@ -1,6 +1,7 @@
 from syft.serde.serde import *
-from syft.serde.torch_serde import *
-from syft.serde.serde import _simplify
-from syft.serde.serde import _detail
-from syft.serde.compression import _compress
-from syft.serde.compression import _decompress
+from syft.serde.msgpack.torch_serde import *
```
karlhigley: I don't think is needed any more, since it's imported in `msgpack/__init__.py` below.

PR #2849, opened 2019-12-20
Cleanup Protocol

TTitcombe:
Cleans-up `protocol.py` a little bit. In particular:
* Arrange imports to PEP8
* Add doc strings and Args, Returns, Raises where missing
* Adds a couple of inline comments
* Adds missing typehints / corrects some type hints

Overall, errors found when running `mypy syft` have decreased

This PR makes a small contribution to #2510


iamtrask:
Well done sir! Thank you for jumping into this!

PR #2850, opened 2019-12-21
Change `Operation` constructor to take named args (instead of tuple)

karlhigley:
This change sacrifices some generality in order to make the actually-in-use contents more explicit and visible. It's a bit awkward in the current code, but should make it easier to construct Protobuf schemas with appropriate types for the parameters. Some of the awkwardness comes from maintaining backwards compatibility for msgpack serialization, some of it comes from the structure of the code that sends commands. It may still be possible to push these named arguments "farther out" into the code that sends commands, which could either be added to this PR or saved for a later one.

### DIFF
--- a/syft/messaging/message.py
+++ b/syft/messaging/message.py
@@ -126,10 +126,10 @@ class Operation(Message):
         # call the parent constructor - setting the type integer correctly
         super().__init__()
 
-        self.command_name = cmd_name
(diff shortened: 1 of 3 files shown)


karlhigley:
Same! Hoping that the small but repeated costs of accessing more instance variables here are offset by skipping simplification, but we'll have to measure and find out.

PR #2856, opened 2019-12-25
Added `String` type to `msgpack/serde.py` to allow serialization

AlanAboudib:
I added `String` to the `OBJ_SIMPLIER_AND_DETAILERS` list in `msgpack/serde.py` to allow serialization of `String`.

I also added some unit tests for serializing `String` objects and other `str` hooked methods.


AlanAboudib:
@LaRiffle @iamtrask Would you please take a look at this?

test/test_serde_full.py
```diff
@@ -1020,6 +1020,27 @@ def compare(detailed, original):
     ]
 
 
+# syft.generic.string.String
+def make_string(**kwargs):
+
+    return [
+        {
+            "value": syft.generic.string.String(
+                "Hello World", id=1234, tags=["tag1", "tag2"], description="description"
```
vvmnnnkv: It seems to be more correct to use `.tag()` and `.describe()` methods to set tags and description.
And when using .tag(), it will be simplified as `set` instead of `list`.

AlanAboudib: Thank you @vvmnnnkv for the review. Actually, `String` type inherits `syft.generic.object.AbstractObject`. That latter has `tags` and `description` as `__init__` arguments, so I followed the same `__init__` method prototype for uniformity. Do you suggest that  I remove `tags` and `description` for `__init__`?

vvmnnnkv: Well, there's some type inconsistency with tags.
`syft.generic.object.AbstractObject` constructor stores `tags` as is (in your case as a `list`).
However, `syft.generic.object.AbstractObject.tag()` assumes `self.tags` is a `set` (and makes it a `set` if it's not initialized yet).
So if you set tags to be a `list` from `__init__`, this will probably make further usage of `.tag()` method on that tensor impossible because `list` doesn't have `.add` method.

Also, simplified object will be different because `list` and `set` are simplified differently.
Given that tutorials use `.tag()` method, I'd rather use it instead of passing tags in constructor. Or update what you're passing to `set`. 

Additional issue you'll see is that `set` doesn't preserve order, so if you specify 2 tags (`{"tag1", "tag2"}`) it can be simplified as `(CODE[set], ((CODE[str], (b"tag1",)), (CODE[str], (b"tag2",))))` or `(CODE[list], ((CODE[str], (b"tag2",)), (CODE[str], (b"tag1",))))` and comparison of simplified objects may fail (flaky). See here https://github.com/OpenMined/PySyft/blob/master/test/test_serde_full.py#L104 how to tackle this (or just use 1 tag).

karlhigley: Might be worth sorting the results of set simplification to iron out the flakiness?

vvmnnnkv: @karlhigley in serde itself? :)

karlhigley: @vvmnnnkv That's what I was thinking. Unless the list of tags is enormous, should be relatively cheap. I don't think there's a standard Python data structure that would move the cost of ordering to the write (rather than read) side, but there could be one I don't know about.

AlanAboudib: Thank you @vvmnnnkv and @karlhigley for the helpful review.

I changed `tags` to a set instead of a list and used a custom comparison function as in `make_set`. This should be good now

AlanAboudib: Would you please review the new commit? Thank you

PR #2863, opened 2019-12-27
Move msgpack serde tests into a sub-module and extract helpers

karlhigley:
This is a slight reorganization of the tests to open up the possibility of parallel structure and shared code between msgpack and Protobuf serde tests. No functional changes intended.


karlhigley:
This is a weird test failure:
```
=================================== FAILURES ===================================

_____________________________ test_serde_coverage ______________________________

    def test_serde_coverage():

        """Checks all types in serde are tested"""

        for cls, _ in msgpack.serde.simplifiers.items():

            has_sample = cls in samples

>           assert has_sample is True, "Serde for %s is not tested" % cls

E           AssertionError: Serde for <class 'test.message.test_plan.test_plan_built_on_class.<locals>.Net'> is not tested

E           assert False is True

test/serde/msgpack/test_serde_full.py:92: AssertionError
```
Passes locally, seems like that class must get dynamically added to the serializers by the Plan tests. Not quite sure how to debug that.

karlhigley:
The reason moving the tests breaks them is because it changes the order in which the tests are run. Previously the serde tests ran before the plan tests; with this change, they run after the plan tests. So this exposed an issue that already existed but wasn't yet caught by the tests. I'll see what I can do about that.

test/serde/msgpack/test_msgpack_serde_full.py
```diff
@@ -0,0 +1,144 @@
...
+samples[syft.messaging.message.IsNoneMessage] = make_isnonemessage
+samples[syft.messaging.message.GetShapeMessage] = make_getshapemessage
+samples[syft.messaging.message.ForceObjectDeleteMessage] = make_forceobjectdeletemessage
+samples[syft.messaging.message.SearchMessage] = make_searchmessage
+samples[syft.messaging.message.PlanCommandMessage] = make_plancommandmessage
+
+samples[syft.exceptions.GetNotPermittedError] = make_getnotpermittederror
+samples[syft.exceptions.ResponseSignatureError] = make_responsesignatureerror
+
+# Dynamically added to msgpack.serde.simplifiers by some other test
```
vvmnnnkv: This one can be removed now (was exact same problem you had with Plan, but with BaseWorker, I just added it in the samples list to avoid test failure)

karlhigley: Workers have `force_full_simplification`, so we must serialize them intentionally at some point, right? I don't really understand when/why that happens though.

PR #2865, opened 2019-12-27
 Set up Protobuf serde and tests for Python `NoneType`

karlhigley:
This lays the foundation for incrementally adding Protobuf serialization
and handles the simplest possible case, which is an empty message
(usually sent as a response to another message.)

Depends on #2863 and OpenMined/syft-proto#18.


karlhigley:
(Tests fail because OpenMined/syft-proto#18 hasn't yet been merged/released.)

syft/serde/protobuf/serde.py
```diff
@@ -0,0 +1,356 @@
...
+        # from, including `type(obj)`. We can skip the type of the
+        # object because we already tried this in the
+        # previous step.
+        classes_inheritance = inspect.getmro(type(obj))[1:]
+
+        for inheritance_type in classes_inheritance:
+            if inheritance_type in bufferizers:
+                # Store the inheritance_type in bufferizers so next time we see this type
+                # serde will be faster.
+                bufferizers[current_type] = bufferizers[inheritance_type]
```
vvmnnnkv: Should we also have `inherited_simplifiers_found` here?

karlhigley: Yes! Good catch.

PR #2868, opened 2019-12-29
fix: targets in makefile were faulty

Rishav1:
Update the makefile to install requirements specific to target's needs. Also added requirements_notebooks to setup tests_require because `python setup.py test` was failing.


iamtrask:
I attempted merge conflicts

Issue #2797, opened 2019-12-09
Send (or not) Autograd Tensor


LaRiffle:
**Is your feature request related to a problem? Please describe.**
Currently, sending/getting Autograd tensors cause an error because grad_fn is added to the args serialize which is not possible.

**Alternative solutions**
A. Either we want to serialize AutogradTensor and we A.1 remove the grad_fn arg or A.2 add serialization for all the grad_fn components or B. We don't want AutogradTensor to be sent across the wire and we raise an explicit error

**Describe the solution you'd like**
We need to serialize grad_fn for AutogradTensor. To do so, one needs to understand how grad_fn is built up in PySyft and how the components should be serialized is an easy way (= try avoiding modifying syft/serde too much)

This triggers weird errors like: https://github.com/OpenMined/PySyft/issues/2666

LaRiffle:
Related to: https://github.com/OpenMined/PySyft/issues/2137

iamtrask:
We should allow autograd tensor to be sendable across the wire, but we should serialize the functions using a string version of them (with specific ones whitelisted) instead of serializaing any function at all.

CC: @karlhigley this is related to what you were talking about.

sukhadj:
Can I work on this?

adityaprabhat01:
is this issue still open?

karlhigley:
Looks like this got resolved in #2871.

sachin-101:
@LaRiffle Please close this issue if it is resolved.

PR #2871, opened 2019-12-30
Make grad_fn serializable

sukhadj:
Solves #2797 
This PR makes grad_fn serializable.


karlhigley:
For testing, you can add examples in the [serde test helpers](https://github.com/OpenMined/PySyft/blob/master/test/serde/serde_helpers.py) and add `GradFunc` to the list of classes tested for round-trip serialization [here](https://github.com/OpenMined/PySyft/blob/master/test/serde/msgpack/test_msgpack_serde_full.py#L80).

pip-dep/requirements.txt
```diff
@@ -1,13 +1,13 @@
 Flask>=1.0.2
 flask_socketio>=3.3.2
-syft-proto>=0.1.0a1
+git+git://github.com/sukhadj/syft-proto@master#egg=syft-proto
```
karlhigley: Does this depend on changes in `syft-proto`? If not, we should probably leave the requirement version specification.

sukhadj: Yes. I have added GradFunc proto code in proto.json
https://github.com/sukhadj/syft-proto/blob/88b27455905264e9d3e11950936ab2b7e5493aed/proto.json#L159

karlhigley: Ah, okay, cool. Could you open a `syft-proto` PR for that change?

sukhadj: Yes

karlhigley:
@sukhadj This has a few small conflicts with master. Could you resolve them so we can get the CI tests running?

pip-dep/requirements.txt
```diff
@@ -4,11 +4,12 @@ syft-proto>=0.1.0a1.post2
 lz4>=2.1.6
 msgpack>=0.6.1
 numpy>=1.16.0
+phe>=1.4.0
```
LaRiffle: This file shouldn't be changed I think

pip-dep/requirements_dev.txt
```diff
@@ -9,4 +9,4 @@ pyshark>=0.4.2.4
 pytest>=4.5.0
 setuptools>=41.0.0
 sphinx-markdown-builder>=0.4.1
-Sphinx>=2.1.1
\ No newline at end of file
+Sphinx>=2.1.1
```
LaRiffle: smae

syft/frameworks/torch/tensors/interpreters/autograd.py
```diff
@@ -338,11 +338,13 @@ def detail(worker: AbstractWorker, tensor_tuple: tuple) -> "AutogradTensor":
...
             preinitialize_grad=preinitialize_grad,
-            grad_fn=grad_fn,
             # local_autograd=local_autograd,
             data=chain,  # pass the de-serialized data
             tags=syft.serde.msgpack.serde._detail(worker, tags),
             description=syft.serde.msgpack.serde._detail(worker, description),
         )
 
+        # Assigning grad_fn here. grad_fn doesn't get assigned when passed in constructor
+        tensor.grad_fn = syft.serde.msgpack.serde._detail(worker, grad_fn)
```
LaRiffle: Would you be willing to fix the `AutogradTensor.__init__`  to fix this? Should be very quick

syft/frameworks/torch/tensors/interpreters/gradients_core.py
```diff
@@ -38,6 +46,70 @@ def __call__(self, grad):
     def __repr__(self):
         return self.__class__.__name__
 
+    def __setattr__(self, name, value):
+        self.__dict__[name] = value
+
+        # add attributes of grad function to _parameters list
+        # essential for serialization and deserialization
+        if not name in ["next_functions", "result", "_parameters"]:
```
LaRiffle: Maybe this is not restrictive enough, can you add a check on the type (should be a Tensor I think)

syft/frameworks/torch/tensors/interpreters/gradients_core.py
```diff
@@ -38,6 +46,70 @@ def __call__(self, grad):
...
+
+        # Add class name of the grad_fn object at the begining.
+        # Essential while deserializing the object
+        cls = syft.serde.msgpack.serde._simplify(worker, grad_fn.__class__.__name__)
+        grad_fn_params = [cls]
+
+        for param in grad_fn._parameters:
+
+            if hasattr(param, "child"):
+                param = syft.serde.msgpack.serde._simplify(worker, param.child)
```
LaRiffle: Why do you transmit param.child and not param?

syft/frameworks/torch/tensors/interpreters/gradients_core.py
```diff
@@ -38,6 +46,70 @@ def __call__(self, grad):
...
+         Args:
+            gradfn_tuple: a tuple containing all the simplified attributes of a
+            grad function
+
+        Returns:
+            A correct gradient function object
+
+        """
+        # Extract the class name from tuple
+        cls = syft.serde.msgpack.serde._detail(worker, gradfn_tuple[0])
```
LaRiffle: I think that you can all `syft.serde.msgpack.serde._detail` directly on the list, which would make your syntax lighter:
smthg like this
`cls_name, *grad_fn_params = syft.serde.msgpack.serde._detail(worker, gradfn_tuple)`

pip-dep/requirements.txt
sukhadj: That must have occurred while resolving the commits
It has been restored to the master. :)

syft/frameworks/torch/tensors/interpreters/autograd.py
sukhadj: Fixed.

syft/frameworks/torch/tensors/interpreters/gradients_core.py
sukhadj: So the problem here is there might be a case where attribute is not a Tensor (consider power in PowBackward function). 
Can you explain a case for which this might not be restrictive enough?

syft/frameworks/torch/tensors/interpreters/gradients_core.py
sukhadj: My intuition was here we need to push every element of the chain in the tuple (like we do AutogradTensor). But that will be taken care of.
Hence the code is changed and now transmitting param instead of param.child.

syft/frameworks/torch/tensors/interpreters/gradients_core.py
sukhadj: Hey, I have changed the functions. Please review :)

karlhigley:
Most of the CI test failures are due to a version mismatch between `syft-proto` and `PySyft` which will be resolved when #2876 is merged. This PR still needs serialization tests for `GradFunc` though.

sukhadj:
@karlhigley I will add tests for `GradFunc` soon :smiley:

sukhadj:
@karlhigley @LaRiffle Should we add tests for each gradient function? 
Or will the test for one class (any or specific) suffice?

karlhigley:
@sukhadj I don’t mean to hassle you about tests, I’m just excited about this one! 😅

I’d pick a gradient function that exercises as much of the code as possible and use that as your test example.

syft/frameworks/torch/tensors/interpreters/gradients_core.py
LaRiffle: Oh you're right!
My only concern is that if people add new internal attributes to GradFunc, then they will need to add them to this list. But maybe this is ok for the moment. Just change the list to be a set :)

sukhadj: Done :+1:

sukhadj:
@karlhigley I believe we currently have some problems mul and pow. So will go ahead with simple `AddBackward` for now :smile:

sukhadj:
Hey @karlhigley @LaRiffle , the tests are failing due to inconsistent serialization (and one because of syntax error :sweat_smile: ).
I have opened the issue #2882. Any thoughts? :smiley:

sukhadj:
There seems to be one more problem here.
The comparison of _attributes is not working. Will be working on that :100:

karlhigley:
It's interesting that the round-trip serialization test passes, but the simplification test fails. Makes me wonder if the simplification test assertion has the correct expected value. 🤔

sukhadj:
> It's interesting that the round-trip serialization test passes, but the simplification test fails. Makes me wonder if the simplification test assertion has the correct expected value. thinking

Hey, so currently comparison function we employ just compares if class name matches but not the attributes. I'm working on comparison of  _attributes. Not sure if we should just check equality of list :smiley:

sukhadj:
Hey @LaRiffle @karlhigley need review on this

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -984,6 +984,15 @@ def set_garbage_collect_data(self, value):
         for _, share in shares.items():
             share.garbage_collect_data = value
 
+    def get_garbage_collect_data(self):
```
karlhigley: This seems like a good addition, but doesn't look like it actually gets called anywhere yet. Did I miss it?

sukhadj: It was used in how previously garbage_collect_data was handled.

syft/frameworks/torch/tensors/interpreters/gradients_core.py
```diff
@@ -38,6 +46,62 @@ def __call__(self, grad):
...
+        # Detail and extract the class name and attributes from the tuple
+        grad_fn_attrs = []
+        cls, *grad_fn_attrs = syft.serde.msgpack.serde._detail(worker, gradfn_tuple)
+
+        if cls == "GradFunc":
+            return GradFunc(*grad_fn_attrs)
+
+        cls = getattr(gradients, cls)
+
+        return cls(*grad_fn_attrs)
```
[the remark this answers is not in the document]
sukhadj: Done

syft/frameworks/torch/tensors/interpreters/autograd.py
```diff
@@ -40,7 +40,10 @@ def __init__(
             self.grad = data * 0
         else:
             self.grad = None
+
         self.grad_fn = None
+        if "grad_fn" in kwargs.keys():
+            self.grad_fn = kwargs["grad_fn"]
```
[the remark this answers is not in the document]
sukhadj: Yess :smile:

karlhigley:
🎉

PR #2875, opened 2019-12-31
Add sk-learn to notebook docker image

Nolski:
[Tutorial 5](https://github.com/OpenMined/PySyft/blob/master/examples/tutorials/Part%2005%20-%20Welcome%20to%20the%20Sandbox.ipynb) and presumably some others require sk-learn in order to work correctly. This was causing some errors to run with the default notebook docker image.

This patch adds the `sandbox` option to installing syft for this the notebook docker image.

### DIFF
--- a/docker-images/pysyft-notebook/Dockerfile
+++ b/docker-images/pysyft-notebook/Dockerfile
@@ -5,7 +5,7 @@ ENV WORKSPACE /workspace
 # Setup workspace environment
 RUN apt-get update && apt-get install -y gcc
 RUN conda install jupyter notebook
-RUN pip install --no-cache-dir syft[udacity]
+RUN pip install --no-cache-dir syft[udacity,sandbox]
 
 # Create jupyter notebook workspace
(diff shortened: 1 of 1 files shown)


docker-images/pysyft-notebook/Dockerfile
```diff
@@ -5,7 +5,7 @@ ENV WORKSPACE /workspace
 # Setup workspace environment
 RUN apt-get update && apt-get install -y gcc
 RUN conda install jupyter notebook
-RUN pip install --no-cache-dir syft[udacity]
+RUN pip install --no-cache-dir syft[udacity, sandbox]
```
youben11: I think it shouldn't be a space there

gmuraru:
@youben11 do you think we should add the tensorflow requirements in this PR or are we ok with merging and closing this?

youben11:
For sake of clarity, I think this PR should only add the sklearn requirement

PR #2877, opened 2020-01-02
Add Protobuf serialization for `Object` and `Operation` messages, `AST` tensor

karlhigley:
This adds serialization for all the types required to make round-trip serialization of `AdditiveSharingTensor` work (at least to the extent that serialization tests pass.)

Depends on OpenMined/syft-proto#24 and #2876.

### DIFF
--- a/syft/messaging/message.py
+++ b/syft/messaging/message.py
@@ -10,8 +10,10 @@ All Syft message types extend the Message class.
 import syft as sy
 from syft.workers.abstract import AbstractWorker
 
-
 from syft_proto.messaging.v1.message_pb2 import ObjectMessage as ObjectMessagePB
+from syft_proto.messaging.v1.message_pb2 import OperationMessage as OperationMessagePB
+from syft_proto.types.syft.v1.operation_pb2 import Operation as OperationPB
+from syft_proto.types.syft.v1.arg_pb2 import Arg as ArgPB
 
 
 class Message:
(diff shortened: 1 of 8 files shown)


karlhigley:
(Tests won’t pass until the corresponding `syft-proto` PR gets merged.)

karlhigley:
@vvmnnnkv Wanted to get something of reasonable complexity serialized, so I tackled this, but haven’t forgotten about your previous comments. Happy to do some refactoring to handle id getting/setting, avoid unnecessary `CopyFrom`s, etc (either in this PR or as a separate PR.)

karlhigley:
Getters and setters worked out well. 👍 

These notebook tests are killing me though. They take a long time and it's hard to tell why they're failing.

karlhigley:
I haven't been able to replicate these test failures locally, so going to try restarting the build.

syft/serde/protobuf/serde.py
```diff
@@ -215,8 +219,13 @@ def serialize(
 
     protobuf_obj = _bufferize(worker, obj)
 
-    if type(obj) == type(None):
+    obj_type = type(obj)
```
vvmnnnkv: Sorry, I've missed this in the beginning, but why do we need to fit everything into SyftMessagePB?
This limits `serialize` to types allowed in SyftMessagePB

syft/messaging/message.py
```diff
@@ -195,6 +200,114 @@ def detail(worker: AbstractWorker, msg_tuple: tuple) -> "Operation":
...
+            sy.serde.protobuf.proto.get_protobuf_id(pb_id)
+            for pb_id in protobuf_obj.operation.return_ids
+        ]
+
+        operation_msg = Operation(command, owner, tuple(args), kwargs, tuple(return_ids))
+
+        return operation_msg
+
+    @staticmethod
+    def _bufferize_args(worker: AbstractWorker, args: list) -> list:
```
vvmnnnkv: These look like general purpose methods that can be placed somewhere along with `get_protobuf_id`?

syft/messaging/message.py
```diff
@@ -195,6 +200,114 @@ def detail(worker: AbstractWorker, msg_tuple: tuple) -> "Operation":
...
+            protobuf_args.append(Operation._bufferize_arg(worker, arg))
+        return protobuf_args
+
+    @staticmethod
+    def _bufferize_arg(worker: AbstractWorker, arg: object) -> ArgPB:
+        protobuf_arg = ArgPB()
+        try:
+            setattr(protobuf_arg, "arg_" + type(arg).__name__, arg)
+        except:
+            getattr(protobuf_arg, "arg_" + type(arg).__name__).CopyFrom(
```
vvmnnnkv: Hmm, what will happen if type is not present in Arg... will it just raise AttributeError twice?
Maybe explicit check for supported type (based on ArgPB) would be better?

syft/serde/protobuf/serde.py
karlhigley: In order to properly deserialize a Protobuf message, you need to know the type. Since we don’t know which message to expect, the `SyftMessage` wrapper type makes the message type always the same, which makes deserialization easy. If we didn’t have it, we could try to parse an incoming message as each type in sequence and use whichever worked, but that would be kinda messy.

syft/messaging/message.py
karlhigley: Yeah, they maybe could be. Arg is only used in Operations so far. I don’t have a strong opinion either way.

syft/messaging/message.py
karlhigley: I think that’s what will happen, yeah. I’m not entirely sure how to do an explicit check for supported types here, but it sounds like a reasonable thing to do.

PR #2880, opened 2020-01-03
Add useful error message for failing notebook

iamtrask:


karlhigley:
Coverage tests fail because the exception case isn't exercised. 🤕

karlhigley:
I've been cherry-picking this into my branches when notebook tests are failing, so it does seem useful, but I can't think of a good way to make sure that the exception case gets handled without causing a test failure. I guess we could maybe create an intentionally broken notebook and assert that it throws an exception?

karlhigley:
Superseded by #2919, which parametrizes the notebook tests and helps `pytest` give better error messages.

PR #2883, opened 2020-01-04
Rework Plans/Procedures to store non-simplified Operation objects

karlhigley:
This will make it easier to serialize them to multiple formats (like msgpack and Protobuf.)


syft/messaging/plan/procedure.py
```diff
@@ -74,74 +74,58 @@ def update_ids(
...
-                operation = Procedure.replace_operation_ids(
-                    operation, from_ids_simplified, to_ids_simplified
-                )
+            # replace ids in the owner
+            owner = operation.cmd_owner
+            if owner is not None:
+                if owner.id in from_ids:
+                    owner.id = to_ids[from_ids.index(owner.id)]
+
+                if owner.id_at_location in from_ids:
```
LaRiffle: I think owner have no `id_at_location`

syft/messaging/plan/procedure.py
```diff
@@ -74,74 +74,58 @@ def update_ids(
...
-                operation = Procedure.replace_operation_ids(
-                    operation, from_workers_simplified, to_workers_simplified
-                )
-
-            if from_ids_simplified and to_ids_simplified:
-                operation = Procedure.replace_operation_ids(
-                    operation, from_ids_simplified, to_ids_simplified
-                )
+            # replace ids in the owner
+            owner = operation.cmd_owner
```
LaRiffle: What is the type of `operation` now? I can't find the definition here it must be in the unchanged files

karlhigley: It's an `Operation` object.

syft/messaging/plan/procedure.py
karlhigley: The owner can be a `PointerTensor`, so it can have an `id_at_location` property, but may not always. As written, this code may fail with non-pointer tensors. 🤔

karlhigley:
Superseded by #2912 and #2910. Closing, but will revisit.

PR #2884, opened 2020-01-04
Comment out numpy serde

vvmnnnkv:
Just want to check if tests pass if we just remove numpy serde.
I couldn't find if it's really used anywhere.


karlhigley:
Seems like this has served its purpose, but can re-open if needed.

PR #2893, opened 2020-01-10
Update fedavg algorithm to support non-IID setting

codergan:
Update the fedavg algorithm in pysyft, support niid setting


syft/frameworks/torch/fl/utils.py
```diff
@@ -29,60 +29,67 @@ def extract_batches_per_worker(federated_train_loader: sy.FederatedDataLoader):
     return batches
 
 
-def add_model(dst_model, src_model):
+def add_model(dst_model, src_model,dst_no_data,src_no_data):
```
karlhigley: `src_num_data` and `dst_num_data` might be clearer here; I thought these were booleans at first.

codergan: Thank you for the advice, I changed here and also the tot_num_data

karlhigley:
@codergan Looks like CI turned up a formatting issue. Do you have the [pre-commit hook](https://github.com/OpenMined/PySyft/blob/master/CONTRIBUTING.md#setting-up-pre-commit-hook) set up?

codergan:
> @codergan Looks like CI turned up a formatting issue. Do you have the [pre-commit hook](https://github.com/OpenMined/PySyft/blob/master/CONTRIBUTING.md#setting-up-pre-commit-hook) set up?

Hi, I am following the instruction and got this error when setiing the pre-commit hook:
fatal error: too many errors emitted, stopping now [-ferror-limit=]
20 errors generated.
error: Setup script exited with error: command 'gcc' failed with exit status 1
make: *** [venv/bin/activate] Error 1
May I know what do I suppose to do? And the error of the check seems to be error on formatting, do you know what is the reason for that?

karlhigley:
The CI tests check that the `black` code formatter wouldn’t change any of the files in the commit. It normally runs during the pre-commit hook and fixes any formatting issues for you. I can’t tell from the log section you shared what the problem might be—I suspect the issues to address are farther up in the output. Seems like some kind of compilation error though?

syft/frameworks/torch/fl/utils.py
```diff
@@ -29,60 +29,67 @@ def extract_batches_per_worker(federated_train_loader: sy.FederatedDataLoader):
...
-    model = model_list[0]
-    for i in range(1, nr_models):
-        model = add_model(model, model_list[i])
-    model = scale_model(model, 1.0 / nr_models)
+
+    total_num_data=0
+    model=None
+    for i in range(len(models)):
+        model = add_model(model, models[i],total_num_data,data_num[i])
+        model = scale_model(model, 1.0 / (total_num_data+data_num[i]))
```
TTitcombe: Would it not be simpler to scale the model outside the loop and in `add_model` only send the `data_num[i]` parameter? At the moment you're scaling the model by `total_num_data` and then multiplying it by `total_num_data` the next time you add a model

syft/frameworks/torch/fl/utils.py
```diff
@@ -29,60 +29,67 @@ def extract_batches_per_worker(federated_train_loader: sy.FederatedDataLoader):
     return batches
 
 
-def add_model(dst_model, src_model):
+def add_model(dst_model, src_model,dst_num_data,src_num_data):
```
LaRiffle: Run `black` to clean the python code (here put spaces between the args)

syft/frameworks/torch/fl/utils.py
```diff
@@ -29,60 +29,67 @@ def extract_batches_per_worker(federated_train_loader: sy.FederatedDataLoader):
...
     dict_params = dict(params)
     with torch.no_grad():
-        for name, param in dict_params.items():
+        for name, _ in dict_params.items():
             dict_params[name].set_(dict_params[name].data * scale)
     return model
 
 
-def federated_avg(models: List[torch.nn.Module]) -> torch.nn.Module:
+def federated_avg(models,data_num):
```
LaRiffle: Keep the typing info and add yours

syft/frameworks/torch/fl/utils.py
```diff
@@ -29,60 +29,67 @@ def extract_batches_per_worker(federated_train_loader: sy.FederatedDataLoader):
...
-def add_model(dst_model, src_model):
+def add_model(dst_model, src_model,dst_num_data,src_num_data):
     """Add the parameters of two models.
-
         Args:
+
             dst_model (torch.nn.Module): the model to which the src_model will be added.
             src_model (torch.nn.Module): the model to be added to dst_model.
+            dst_num_data: the number of data within dst_model.
+            src_num_data: the number of data within src_model.
```
LaRiffle: Add typing info
I'd suggest changing the name `dst_num_data` is not very clear neither `the number of data` -> number of data items? size of the dataset?

syft/frameworks/torch/fl/utils.py
```diff
@@ -29,60 +29,67 @@ def extract_batches_per_worker(federated_train_loader: sy.FederatedDataLoader):
...
-    for i in range(1, nr_models):
-        model = add_model(model, model_list[i])
-    model = scale_model(model, 1.0 / nr_models)
+
+    total_num_data=0
+    model=None
+    for i in range(len(models)):
+        model = add_model(model, models[i],total_num_data,data_num[i])
+        model = scale_model(model, 1.0 / (total_num_data+data_num[i]))
+        total_num_data=total_num_data+data_num[i]
```
LaRiffle: tiny: use +=

syft/frameworks/torch/fl/utils.py
LaRiffle: I agree, this code looks correct, but the new feature you're including here might be an opportunity to refactor this file to make it more intuitive. If you're willing to do this thats would be great! Otherwise I'm afaird the next person willing to improve the features will break smthg because it doesn't fully understand what's going on.

syft/frameworks/torch/fl/utils.py
codergan: I change it to data_amount

codergan:
Hello guys, sorry that recently is a bit busy. I just changed accordingly and plz let me know if there is any problem @LaRiffle @karlhigley

karlhigley:
@codergan It looks like one of the notebooks has a call to `federated_avg` that needs to be updated to use the `data_amount` parameter. Is it possible to make that param optional?

codergan:
> @codergan It looks like one of the notebooks has a call to `federated_avg` that needs to be updated to use the `data_amount` parameter. Is it possible to make that param optional?

I just make data_num default to be None. And create an list with ones to be data_num in this situation

tvmsouza:
Hi guys, nice to meet you. I think that these features are very important for the framework, so I'm very interested in it. I saw this implementation of FEDAVG is a little bit different from that presented on (McMahan at all 2016) https://arxiv.org/abs/1602.05629. Maybe i can be wrong, and the implementation can have some motivation to be like the way it is. However, it is something that I notice, and I want to share it with you.

I saw that in current implementation the first worker to be agregated is multiplied by the number of samples two times and when the model is scaled, this multiplication is not take in consideration, on the division/scale of the model. So the first worker receive more relevance than the second in some cases:

if models == []:
        return None
    if data_amount == None:
        data_amount = [1] * len(models)
    model = add_model(None, models[0], 0, data_amount[0])
    **total_num_data = data_amount[0]**
    for i in range(1, len(models)):
        model = add_model(**model**, models[i], **total_num_data**, data_amount[i])
        model = scale_model(model, 1.0 / (**total_num_data + data_amount[i]**))
        total_num_data += data_amount[i]
    return model

Using the same code base from @karlhigley i tried to implement the version of macmahan cited on paper:

def federated_avg(models, data_amount = None) :

    if models == []:
        return None
    total_data = sum(data_amount)
    data_scale = [num_samples/total_data for num_samples in data_amount]
    model = add_model_corrected(models[0], models[1] ,  data_scale[0], data_scale[1])
    for i in range(2, len(models)):
        model = add_model_corrected(model, models[i], 1.0, data_scale[i])
    return model

The total number of samples of all workers in a round is calculated before and the data_scale factor of each model is also calculated by the division (num_samples/total_data) to scale each model while executing the model addition process.

I think that the results are different between the two implementations. 

Thank you for your attention.
Best regards.

PR #2896, opened 2020-01-12
Create python implementation of torch.nn.Conv2d and torch.nn.AvgPool2d

iamtrask:
For the same reasons we have python implementations of LSTM and GRU, I have created a basic implementation of Conv2d and AvgPool2d

It does not include all functionality, but it does include assertions to cause errors for settings which are not supported.


Jasopaum:
We already have our own conv2d implementation, it’s not in the nn module but in the FixedPrecisionTensor class, I think. It was implemented before the rnn so it’s not at the right place (I know we should have moved it before).
Maybe it can be improved but you can at least start from there 🙂

iamtrask:
Hi @Jasopaum - upon further review of that class it does indeed look like a convolutional layer exists there. However, unbeknownst to me the class has an enormous amount of functionality which is unrelated to FixedPrecision representations. I suspect that this has been done for coding convenience and possibly justified for performance reasons. If so, this is done at the expense of code re-use and good organization. 

At present, I will not refer to it and have created a Github issue describing extensive refactoring required (https://github.com/OpenMined/PySyft/issues/2897). Once the refactoring is completed, we can adjust the classes created in this PR to reference wherever the new functionality exists.

Jasopaum:
I do agree that a refactor would do some good! :+1:

.travis.yml
```diff
@@ -40,7 +40,7 @@ before_script:
   - mypy syft || echo "Type errors found, continuing build..."
 script:
   - coverage run -m pytest test
-  - coverage report --omit="syft/grid/*" --ignore-errors --fail-under 95 -m
+  - coverage report --omit="syft/grid/*","syft/frameworks/torch/nn/rnn.py" --ignore-errors --fail-under 95 -m
```
karlhigley: Needed to make the rest of the PR pass coverage requirements?

syft/frameworks/torch/nn/conv.py
```diff
@@ -0,0 +1,111 @@
...
+    torch.nn.Conv2d and so we must have python ports available for all layer types
+    which we seek to use.
+
+    Note that this module has been tested to ensure that it outputs the exact output
+    values that the main module outputs in the same order that the main module does.
+
+    However, there is often some rounding error of unknown origin, usually less than
+    1e-6 in magnitude.
+
+    This module has not yet been tested with GPUs but should work out of the box.
```
karlhigley: Might be worth capturing an issue for this.

test/generic/test_string.py
```diff
@@ -61,3 +62,8 @@ def test_string_methods():
     assert String("Hello %s") % "PySyft" == string
 
     assert str(string) == "Hello PySyft"
+
+    x = String("Hello PySyft")
+    bob = sy.VirtualWorker(id="bob", hook=sy.hook)
+    out = x.send(bob)
+    assert isinstance(out, sy.generic.pointers.string_pointer.StringPointer)
```
karlhigley: This looks like a good thing to test! I might have missed how it's related to the rest of the PR though. Is it used here somewhere?

test/torch/nn/test_conv.py
```diff
@@ -0,0 +1,35 @@
...
+    model2.weight = model.weight
+    model2.bias = model.bias
+
+    data = th.rand(10, 1, 28, 28)
+
+    out = model(data)
+
+    out2 = model2(data)
+
+    assert (out2 - out).abs().max() < 0.000001
```
karlhigley: This could use either [`pytest.approx`](https://docs.pytest.org/en/latest/reference.html#pytest-approx) or [`numpy.allclose`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html) to make the comparison more readable.

test/torch/nn/test_pool.py
```diff
@@ -0,0 +1,33 @@
...
+    pool = nn.AvgPool2d(2)
+    pool2 = nn2.AvgPool2d(2)
+
+    data = th.rand(10, 1, 8, 8)
+
+    model_out = model(data)
+    out = pool(model_out)
+    out2 = pool2(model_out)
+
+    assert (out - out2).abs().max() < 0.000001
```
karlhigley: Could use either [`pytest.approx`](https://docs.pytest.org/en/latest/reference.html#pytest-approx) or [`numpy.allclose`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html) again here.

.travis.yml
iamtrask: Yes - it looks like because there were no tests in that directory at all that it wasn't being included in code coverage. When I added tests for some things in the directory, it then had to fail for the rest. It's out of scope for this PR to test that rnn right now, but we can get to it eventually.

test/generic/test_string.py
iamtrask: It's not related - i was trying to figure out why code coverage was failing. I thought it might just be a rounding error or something so I tested an untested piece of code.

.travis.yml
iamtrask: created an issue to capture this

https://github.com/OpenMined/PySyft/issues/2926

test/torch/nn/test_conv.py
iamtrask: Will fix before merging.

iamtrask: Fixed

test/torch/nn/test_pool.py
iamtrask: Fixed

syft/frameworks/torch/nn/conv.py
iamtrask: https://github.com/OpenMined/PySyft/issues/2927

PR #2905, opened 2020-01-14
[WIP] BigInt Tensor

iamtrask:
In the codebase we often conflate fixed precision and large precision functionality into combined "hybrid tensors". This is a part of an initiative to reverse this by creating a BigInt tensor. FixedPrecision should then optionally inherit from this tensor if it needs to represent a fixed precision representation that cannot fit into Torch's largest integer types.


karlhigley:
Optionally inherit? Like, two different classes with one that inherits and one that doesn't, or...?

iamtrask:
Sorry - "inherit" was the wrong term. 

A user of FixedPrecisionTensor could do so using either IntTensor, LongTensor, or BigIntTensor as the underlying representation, as in:

Wrapper -> FixedPrecisionTensor -> IntTensor
Wrapper -> FixedPrecisionTensor -> LongTensor
Wrapper -> FixedPrecisionTensor -> BigIntTensor

are all viable candidates.

PR #2907, opened 2020-01-14
Remove LPT and CRT

### LINKED ISSUE #2872: Remove Numpy from LargePrecisionTensor
We cannot use LargePrecisionTensor for remote executions or within plans if it depends on numpy. 

Implement the same functionality without using numpy. In other words, remove the numpy import from this file.

https://github.com/OpenMined/PySyft/blob/b0b8a13524150002524c15c6c283b4b14de62d40/syft/frameworks/torch/tensors/interpreters/large_precision.py

Hooking numpy is also an option, but it's not one we want to consider at the moment because it adds a great deal of complexity to intentionally use multiple frameworks.

mccorby:
LPT and CRT make extensive use of `numpy` which makes it not fit for cross-platform scenarios.

This PR removes LargePrecisionTensor and CRT from the codebase as new alternatives are on its way.

Closes #2872


mccorby:
> May I ask you to also remove CRTTensor? It is not used too... :/

Let me check with Jason. @Jasopaum

mccorby:
> Thanks for the changes! Can you remove numpy fo the requirements to validate the change? :)

I can still see `numpy` used in notebooks, pate and other places in the project.

LaRiffle:
> > Thanks for the changes! Can you remove numpy fo the requirements to validate the change? :)
> 
> I can still see `numpy` used in notebooks, pate and other places in the project.
Ok cool!

PR #2910, opened 2020-01-14
Refactor Plan using PlaceHolder

LaRiffle:
This PR proposes a new way to build Plan without make a mock send to smthg which resembles a worker. It shows that different strategies exist to "log" all the commands executed within a function. That we don't need to deal with ids anymore if we use dynamic Placeholder objects.

Note that almost nothing changes for the end user
Note also that you have to update syft_proto to support Placeholder


karlhigley:
Rebasing this on master to bring it back to a place where we can auto-update with the "merge master" button.

LaRiffle:
Is it necessary to force push?
I'd like to avoid this as much as possible, it make me sad :( :D

karlhigley:
Master would neither merge cleanly, nor give me the opportunity to resolve merge conflicts, so that was the best way I could figure out to do it.

PR #2918, opened 2020-01-17
Split tutorial notebook translation test into separate tests per lang

karlhigley:
The single translation test was taking more than 10m to execute, which was leading Travis CI to infer that our tests were frozen and terminate the builds.


karlhigley:
We can improve on this farther by splitting each notebook out into a separate test, but this is sufficient to fix the tests that are now failing on master so I'm going to merge to unblock everything else.

PR #2913, opened 2020-01-16
NumpyTensor

iamtrask:
I have created a new tensor type called NumpyTensor which will allow us to combine all tensors which utilize numpy tensors under the hood into a single abstraction. 

I have initialized it with some basic functionality and future pull requests can extend it with more.

PR #2922, opened 2020-01-18
Update hook.py

vishalsubbiah:
Based on comment here: https://github.com/OpenMined/PySyft/pull/2913#discussion_r367878631 
`HookedTensor` is missing a `numpy_tensor` argument.
Submitting the fix.

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/hook.py
+++ b/syft/frameworks/torch/tensors/interpreters/hook.py
@@ -8,7 +8,7 @@ class HookedTensor(AbstractTensor):
     method hooked without having to add it to the hook.py file.
     """
 
-    def __init__(self, owner=None, id=None, tags=None, description=None, verbose=False):
+    def __init__(self, numpy_tensor=None, owner=None, id=None, tags=None, description=None, verbose=False):
         """Initializes a HookedTensor.
 
         Args:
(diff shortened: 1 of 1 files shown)


karlhigley:
Thanks for the fix! Looks like the build is saying there’s a code formatting issue of some kind. Set up the [commit hooks](https://github.com/OpenMined/PySyft/blob/master/CONTRIBUTING.md) and they’ll format the code for you when make a commit.

vishalsubbiah:
@karlhigley I fixed the formatting in the second commit. The build is still running for it.  Thanks for the link to the commit hooks, that will help in the future.

syft/frameworks/torch/tensors/interpreters/hook.py
```diff
@@ -8,7 +8,9 @@ class HookedTensor(AbstractTensor):
     method hooked without having to add it to the hook.py file.
     """
 
-    def __init__(self, owner=None, id=None, tags=None, description=None, verbose=False):
+    def __init__(
+        self, numpy_tensor=None, owner=None, id=None, tags=None, description=None, verbose=False
```
karlhigley: I think we actually want to make the reverse change here; the Numpy code got moved over to [NumpyTensor](https://github.com/OpenMined/PySyft/blob/be9e0d894e7fe09cf22f93a3a27ad42a3903a037/syft/frameworks/torch/tensors/interpreters/numpy.py), so we should probably remove both the `numpy_tensor` parameter and the docstring explanation of it.

vishalsubbiah: oh okay. fixed it now.

karlhigley:
@vishalsubbiah Thanks for the PR!

PR #2928, opened 2020-01-18
Run PR tests on various Pythons

karlhigley:
Includes Python 3.6 and 3.7. Could be run on more Pythons, but that's the intersection of versions with PySyft and PyTorch support.


karlhigley:
Two issues here:
- The test that checks that all the notebooks have been tested should exclude the translation notebooks in the normal test run, but include the check in the translation test run.
- The efficiency tests run slightly longer in Github Actions, which pushes them over the threshold for acceptable test execution time. Might need to lengthen the acceptable time, or skip these tests in CI.

iamtrask:
I did not mean to merge this! The button changed with all the github actions or something - I don't know how this happened. Reverting.

PR #2930, opened 2020-01-20
Bump pytorch to 1.4

gmuraru:
Update to torch 1.4 .

I think we should move to the new version in case we add more functionality/solve more bugs -- we need to make sure that all our changes are according to the new torch changes.

In this PR it was introduced a new serialization/deserialization for ```torch.memory_format```. It was needed because it is used as a parameter in the ```__deep_copy__``` method (it was introduced in pytorch 1.4 - the change can be found [here](https://github.com/pytorch/pytorch/blob/master/torch/nn/parameter.py#L32) -- before this it was a simple ```clone``` without any parameter).

We also need to update the ```syft-proto``` file -- the PR is [here](https://github.com/OpenMined/syft-proto/pull/29)


pip-dep/requirements.txt
```diff
@@ -5,10 +5,10 @@ msgpack>=0.6.1
 numpy>=1.16.0
 phe>=1.4.0
 scipy>=1.4.1
-syft-proto>=0.1.0a1.post23
+git+git://github.com/gmuraru/syft-proto@fd93a01#egg=syft-proto
```
gmuraru: This is needed only for testing

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -661,7 +661,7 @@ def attr(self, attr_name):
 
         return attr_val
 
-    def clone(self):
+    def clone(self, *args, **kwargs):
```
gmuraru: We need to make sure that we take into account the arguments sent to clone

syft/serde/torch/serde.py
```diff
@@ -29,6 +29,16 @@
 TORCH_STR_DTYPE = {name: cls for cls, name in TORCH_DTYPE_STR.items()}
 
 
+TORCH_MFORMAT_ID = {
+    torch.channels_last: 1,
+    torch.contiguous_format: 2,
+    torch.contiguous_format: 3,
```
LaRiffle: There is two identcal keys here

syft/frameworks/torch/tensors/interpreters/native.py
gmuraru: Ahhh..yes :( I should

syft/serde/torch/serde.py
gmuraru: Yep. Tested again with
```
>>> for attr in dir(torch):
...     a = getattr(torch, attr)
...     if isinstance(a, torch.memory_format):
...             print(a)
...```
And the result is:
``` 
torch.channels_last
torch.contiguous_format
torch.preserve_format
```

File: syft/frameworks/torch/tensors/interpreters/native.py (hunk 2 above)
Comment (gmuraru) [reply syft/frameworks/torch/tensors/interpreters/native.py]: This is needed because of [this] (https://github.com/pytorch/pytorch/blame/master/torch/nn/parameter.py#L32) - it was added 3 months ago.

File: syft/frameworks/torch/tensors/interpreters/native.py [hunk 4]
```diff
@@ -661,17 +661,17 @@ def attr(self, attr_name):
...
         """
         Clone should keep ids unchanged, contrary to copy
         """
-        cloned_tensor = self.native_clone()
+        cloned_tensor = self.native_clone(args, kwargs)
```
LaRiffle: You should keep the stars: `*args, **kwargs`

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -661,17 +661,17 @@ def attr(self, attr_name):
...
         cloned_tensor.id = self.id
         cloned_tensor.owner = self.owner
         cloned_tensor.is_wrapper = self.is_wrapper
 
         if self.has_child():
-            cloned_tensor.child = self.child.clone()
+            cloned_tensor.child = self.child.clone(args, kwargs)
```
LaRiffle: same here

PR #2936, opened 2020-01-20
Migrate CI tests to Github Actions

karlhigley:
This also splits up the test suite so that the notebook translation tests are only run when there's a change to the tutorials in order to speed up the builds. The (untranslated) basic and advanced tutorial notebook tests are run in every build, so we'll still know if changes to the rest of the code base break the notebooks. Making changes to the tutorials to fix issues caused by other parts of the code base will then trigger the translation tests.


karlhigley:
It seems like the efficiency tests run with fairly variable execution time in Github Actions. I could bump the execution time threshold to make them pass, but that seems to defeat the point of the tests. Alternately, I could skip them in the CI builds? Thoughts on this @iamtrask?

LaRiffle:
@karlhigley unfortunately I think we must get rid of the efficiency test, github travis or actions are not a good place to test for efficiency. This should be made elsewhere with a dedicated server to have a clear idea on how this is evolving in time

karlhigley:
@LaRiffle I'll skip the efficiency tests in the GH Actions then. They seemed to pass consistently on Travis, so maybe we can continue running them there.

karlhigley:
🎉

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+        sudo apt-get update
+        python -m pip install --upgrade pip
+        hash -r
+        pip3 install -r pip-dep/requirements.txt
+        pip3 install -r pip-dep/requirements_dev.txt
+        pip3 install -r pip-dep/requirements_udacity.txt
+        pip3 install -r pip-dep/requirements_notebooks.txt
+        pip3 install scikit-learn
+        python setup.py install
+    - name: Lint, format, and type-check
```
vvmnnnkv: Not sure this will help to speed up things much, but what if we run linting tools deps installation?
Turnaround for formatting errors would be faster, in theory :)

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+        flake8 . --count --exit-zero --statistics --select=E,F,W,C90
+        # Run type checker in CI such that type errors can be identified and gradually addressed.
+        # Once all existing type errors are addressed, the `|| echo` after the mypy call can be removed.
+        mypy syft || echo "Type errors found, continuing build..."
+    - name: Test with pytest
+      run: |
+        coverage run -m pytest test --deselect=test/notebooks/test_notebooks.py::test_notebooks_basic_translations --deselect=test/notebooks/test_notebooks.py::test_all_translation_notebooks --deselect=test/efficiency
+    - name: Coverage reports
+      run: |
+        coverage report --omit="syft/grid/*" --ignore-errors --fail-under 95 -m
```
vvmnnnkv: `--omit="syft/grid/*"` is removed in travis config, but left here - is that intentionally?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,55 @@
+name: Run tests
+
+on: [push]
```
vvmnnnkv: This should be also triggered on `pull_request` event so it works on PRs?
And `push` event should probably [exclude](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#excluding-branches-and-tags) `master` branch - so we don't run tests on master after PR is merged?

vvmnnnkv:
Not in the scope of this PR:
This seems like something we could try for speeding up dependencies step: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows
I can make separate issue if you think it worth's trying out.

.github/workflows/run-tests.yml
LaRiffle: Keeping test when merging on master is good to make sure that nothing got wrong, and to have the ✔︎sign on our last commit which is a good security when someone reaches out saying master is broken ;)

LaRiffle: It already works on PRs using `push`, from what I experienced on another project

.github/workflows/run-tests.yml
LaRiffle: I agree!

LaRiffle: You might need to keep  `pip3 install -r pip-dep/requirements_dev.txt` however

.github/workflows/run-tests.yml
vvmnnnkv: @LaRiffle hmm, that's strange. Technically this is a push to a forked repo.

LaRiffle: Oh yes on a forked repo it might not work :/
But on branches from OpenMined/Pysyft it should (but they are not a majority)

.github/workflows/run-tests.yml
karlhigley: Ah, forgot to remove it here. It’s been moved into setup.cfg.

.github/workflows/run-tests.yml
karlhigley: I was considering running linting as a completely separate action, since the first failure will cancel other actions in progress. Let's leave that for a future issue/PR though?

karlhigley:
@vvmnnnkv Caching dependencies sounds like a good optimization. Could you create an issue for it?

karlhigley:
Looks like the Github Action builds are roughly ten minutes faster than the Travis builds. Leaving the Travis builds enabled for now, until we have some confirmation that the GH Actions work as expected and catch relevant failures.

PR #2940, opened 2020-01-21
fix 3 bugs in the tutorial Part06-07

Rachelxuan11:
in Part07: boston_data is a tuple. it cannot be index with str type as ['alice']. we should index with boston_data[0]['alice'] to avoid "TypeError: tuple indices must be integers or slices, not str"

in Part06: 2 strange Runtime error about cuda. 
1) "RuntimeError: Expected object of device type cuda but got device type cpu for argument #1  'self' in all to _th_set_" can be fixed with "torch.set_default_tensor_type(torch.cuda.FloatTensor)  # fix ERROR: Runtime error when move the model to device. "
2) "RuntimeError: Caught RuntimeError in DataLoader worker process 0. RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method" can be fixed with "torch.multiprocessing.set_start_method('spawn') "


Rachelxuan11:
> Hey, your changes in the notebooks are not visible :)

Rachelxuan11:
Hi, sorry about that... Now the difference is shown in ReviewNB. Could you please check it again?

examples/tutorials/Part 06 - Federated Learning on MNIST using a CNN.ipynb
```diff
@@ -112,7 +112,10 @@
     "\n",
     "device = torch.device(\"cuda\" if use_cuda else \"cpu\")\n",
     "\n",
-    "kwargs = {'num_workers': 1, 'pin_memory': True} if use_cuda else {}"
+    "kwargs = {'num_workers': 1, 'pin_memory': True} if use_cuda else {}\n",
+    "\n",
+    "torch.set_default_tensor_type(torch.cuda.FloatTensor)\n",
+    "torch.multiprocessing.set_start_method('spawn')"
```
LaRiffle: If it's only for cuda I'd make it optional base on the boolean `use_cuda`

Rachelxuan11: Yes, I just tested on CPU. These errors are only for a cuda device.

Issue #2947, opened 2020-01-22
Incorporate versioneer to packaging automation


systemshift:
As mentioned in  #2929.
To streamline publishing syft to pypi, versioneer would help automate things further.

systemshift:
@karlhigley please assign me the issue.

imskr:
@karlhigley I would like to work on this, if no one is working.

karlhigley:
@imskr There's already [a PR for it](https://github.com/OpenMined/PySyft/pull/2949), which I've been holding off on merging until I get a better handle on how we want to do versioning across the various Syft repos.

imskr:
Okay.

PR #2949, opened 2020-01-22
add versioneer

systemshift:
solves #2947.
feedback welcome.


versioneer.py
```diff
@@ -0,0 +1,1857 @@
+# Version: 0.18
```
robert-wagner: So we can just add this as a dependency rather than copying the source here

systemshift: so versioneer.py should also be added to gitignore as well?

karlhigley: We should probably also take that approach of adding `versioneer` as a dependency in `syft-proto` (which currently has the source included in the root of the repo.)

vvmnnnkv: Adding versioneer.py source to repo is how it's supposed to be installed: https://github.com/warner/python-versioneer/blob/master/INSTALL.md
Unless I misunderstood it :)
Perhaps this is to enable getting versions w/o installing any deps upfront.

vvmnnnkv:
See files usually updated during version bump:
https://github.com/OpenMined/PySyft/pull/2943/files
I guess `docs/conf.py` should be updated to use versioneer while `syft/version.py` file and `[bumpversion]` section in `setup.cfg` can be removed.

versioneer.py
systemshift: I second @vvmnnnkv, versioneer the package has only one usage `versioneer install` which adds versioneer.py.
I tried to work around it with #2949 to avoid having the file being in the repo. But now things break in regards of style.

.travis.yml
```diff
@@ -40,7 +40,7 @@ before_script:
   - mypy syft || echo "Type errors found, continuing build..."
 script:
   - coverage run -m pytest test
-  - coverage report --omit="syft/grid/*","syft/frameworks/torch/nn/rnn.py" --ignore-errors --fail-under 95 -m
+  - coverage report --omit="syft/grid/*","syft/frameworks/torch/nn/rnn.py", "versioneer.py" --ignore-errors --fail-under 95 -m
```
karlhigley: I think the space before `"versioneer.py"` is causing problems with the coverage checks, which may be what's making Travis weird for this PR.

systemshift: It does not seem to be the case.

karlhigley:
Resolved a conflict in `travis.yml`, which was caused by moving the list of files ignored by coverage checks into `setup.cfg`. I think `versioneer.py` will need to be added there.

systemshift:
> Resolved a conflict in `travis.yml`, which was caused by moving the list of files ignored by coverage checks into `setup.cfg`. I think `versioneer.py` will need to be added there.

looks like the merge you done got travis working. but the commit I just added still shows the same issue.
Also, now the new branch conflict cannot be solved by me, it shows I don't have write access(it was not the case before when conflicts show up)

PR #2951, opened 2020-01-22
Exorcise a Python `eval` and transmute into the lesser demon `getattr`

karlhigley:


iamtrask:
Exorcisms!

karlhigley:
@LaRiffle Ah, good catch. What a difference [a year makes](https://github.com/OpenMined/PySyft/pull/1814), huh? 😄

karlhigley:
I'm going to leave further optimization of this code for a later issue, since this PR is about fixing a security hole.

LaRiffle:
Ok good, maybe just `cmd = cmd.split(".")`  because this is a one line change, and we'll do the rest later

PR #2959, opened 2020-01-24
WIP: interim commit for feedback

Prtfw:
PR for feedback:

current state of notebook as screenshot https://share.getcloudapp.com/z8un95GO

(don't worry ... I will rebase / squash / clean up before merge ... )

one weird thing is I get this error when specifying a shape in the float tensor like 
`r = test.promise_tensor().FloatTensor(shape = (3,3))` (see notebook or notebook screenshot for details)
https://share.getcloudapp.com/2NurQx57

if I don't specify shape it works.

If don't set https://github.com/OpenMined/PySyft/pull/2959/files#diff-c98d0a76eff05791989ddd31f5dcc7a6R13 "shape=None" by default ... even if I specify shape in the float tensor I get "missing args for shape" error.

### DIFF
--- a/examples/protocol_exp/PromiseTensor and Protocol Roadmap.ipynb
+++ b/examples/protocol_exp/PromiseTensor and Protocol Roadmap.ipynb
@@ -25,6 +25,8 @@
     "import syft as sy\n",
     "import torch as th\n",
     "\n",
+    "#!pip install -e . ## install local editable copy\n",
+    "\n",
     "%load_ext autoreload\n",
     "%autoreload 2"
    ]
@@ -52,6 +54,7 @@
     }
    ],
    "source": [
+    "# verify is using local editable copy\n",
     "!pip show syft"
    ]
   },
--- a/examples/protocol_exp/PromiseTensor and Protocol Roadmap.ipynb
+++ b/examples/protocol_exp/PromiseTensor and Protocol Roadmap.ipynb
@@ -57,7 +57,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -96,20 +96,13 @@
    "execution_count": null,
    "metadata": {},
    "outputs": [],
-   "source": []
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
    "source": [
-    "president."
+    "'''experimental code'''"
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [
     {
@@ -131,7 +124,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 5,
+   "execution_count": 4,
    "metadata": {},
    "outputs": [
     {
@@ -139,7 +132,7 @@
      "output_type": "stream",
      "text": [
       "60 torch.FloatTensor\n",
-      "(Wrapper)>[PromiseTensor(me:94591929895) -future-> FloatTensor -blocking-> 0 plans]\n",
+      "(Wrapper)>[PromiseTensor(me:1462691177) -future-> FloatTensor -blocking-> 0 plans]\n",
       "<VirtualWorker id:test #objects:1>\n"
      ]
     }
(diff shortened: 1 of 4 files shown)


syft/workers/base.py
```diff
@@ -532,6 +537,33 @@ def send_command(
...
+        ''' this works but can't pass tensor type '''
+        # pt  = sy.Promise.FloatTensor(shape=(3,3))
+        # print(pt)
+        # ref = pt.send(self)
+
+
+        '''
+        this gets error cannot serialize
+        see screen shot of stack trace:
+        https://share.getcloudapp.com/OAuLdOm6
```
karlhigley: I haven't seen this particular error before, but I can say that I've been wrestling with serialization a lot, so I'm not surprised that it doesn't work as expected right off the bat. It might be useful to print and/or use the debugger to examine the contents of `simple_objects` on line 280. It should only contain Python primitive like lists, tuples, integers etc, and if there's other stuff in there that might be what's causing the issue.

examples/protocol_exp/PromiseTensor and Protocol Roadmap.ipynb
```diff
@@ -0,0 +1,467 @@
+{
```
karlhigley: <p>I think making the method <code>move</code>  do something other than moving the tensor is going to be a surprising and non-intuitive API design choice. Can we keep the existing semantics of <code>move</code>  and add a second method that extends the graph across workers?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenMined/PySyft/pull/2959/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

karlhigley: <p>Could we use different example roles/people here?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenMined/PySyft/pull/2959/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

karlhigley:
Based on the current state of #2910, we might need to do some thinking together about what a Promise is and how it relates to a Placeholder. The current implementation of Promises is based on Plans (which it probably shouldn't be) and `Promise.keep()` reaches way into the internals of Plans (which it definitely shouldn't.) Sorting that out will probably make `PromiseTensors` and `Protocols` easier to make progress on.

syft/workers/base.py
Prtfw: good call, I will look into that!

examples/protocol_exp/PromiseTensor and Protocol Roadmap.ipynb
Prtfw: ack(acknowledged!) @iamtrask ^ your thoughts?

Prtfw: ack(acknowledged!) => will change (although this notebook is mostly for my own debugging purposes... probably won't get actually merged)

Prtfw:
> Based on the current state of #2910, we might need to do some thinking together about what a Promise is and how it relates to a Placeholder. The current implementation of Promises is based on Plans (which it probably shouldn't be) and `Promise.keep()` reaches way into the internals of Plans (which it definitely shouldn't.) Sorting that out will probably make `PromiseTensors` and `Protocols` easier to make progress on.

I haven't gotten that far :)  
I'm still wrapping my head around how promise tensor currently interacts with plans. And if I can have it interact with (a new?) protocol class instead (and I am guessing both plans and protocols will likely have similar interfaces)... looking up "placeholders" now.

examples/protocol_exp/PromiseTensor and Protocol Roadmap.ipynb
iamtrask: Sounds good to me (and should be an easy change).

This does beg the question though, what does .move() do in this context? I suppose it would move the PromiseTensor without extending the graph?

syft/workers/base.py
iamtrask: I haven't seen this error either. Let's try Karl's idea.

karlhigley:
This was a good start to the discussion. 👍 

Closing since we're in the process of redesigning and writing an RFC for protocols.

Issue #2670, opened 2019-10-17
Send & Get datasets


LaRiffle:
We can currently send and get tensors, we would like to have the ability to directly fetch datasets with all the data inputs and targets in an easy way.

## Things to do:
- [x] Change `sy.BaseDataset` to inherit from `AbstractObject` to have tagging and description and to be registrable by workers: this allows datasets to be referenced for the search functionality
- [x] Extend `BaseDataset` with a `simplify` and `detail` method (see other objects with a simplify and detail method like the `FixedPrecisionTensor` to understand the behaviour): this allows to send and receive datasets. (+ reference BaseDataset in serde.py just like FixedPrecisionTensor). **You need to update syft-proto as well**
- [x] Support pointing to BaseDataset: create a new type of Pointer for BaseDataset which inherits from `ObjectPointer` (get inspiration for example from `PointerProtocol`)
- [x] Add the .send() and .get() functionality to BaseDataset (for .send() you can get inspiration from `Protocol` and for .get() `native.py`)
- [x] Allow search of datasets in .search to fetch remote datasets: make sure the `BaseWorker.search` method correctly retrieve datasets

Each of this point can be addressed in a separate PR if needed.

NovaVic:
Hi Theo,
I can take this one.
I see similar coding structure (with tag, description, simplify, detail methods etc.) in protocol.py file - so can to some extent imagine what the "things to do" are referring to but I think I still need more context as a new person on this.
Thanks in advance Theo!

NovaVic:
Hi Theo @LaRiffle, Thanks for the updated description - that definitely helps! And yes I feel it would be better manageable if I can address change related to each of the points as separate PR.

You can assign it to me. Thanks Theo!

LaRiffle:
Hi @NovaVic, any updates on how the project is going? :)

Guillem96:
@LaRiffle any updates on this issue? If you want I can give it a try

LaRiffle:
Please go ahead :)

Guillem96:
@LaRiffle When I try to send a Dataset I get this error:

`syft.exceptions.UndefinedProtocolTypeError: syft.frameworks.torch.fl.dataset.BaseDataset is not defined in the protocol file`

Should I modify syst proto package too?

LaRiffle:
> @LaRiffle When I try to send a Dataset I get this error:
> 
[...]

PR #2960, opened 2020-01-24
Send & Get Datasets

abogaziah:
first milestone in the issue #2670


syft/frameworks/torch/fl/dataset.py
```diff
@@ -22,8 +24,19 @@ class BaseDataset:
...
+    def __init__(
+        self,
+        data,
+        targets,
+        transform=None,
+        id: int = None,
+        owner=None,
+        tags: List[str] = None,
+        description: str = None,
+        child=None,
```
LaRiffle: for conciseness just put `**kwargs` and report them in `super().__init__(**kwargs)`

abogaziah: done

syft/generic/pointers/pointer_dataset.py
```diff
@@ -0,0 +1,47 @@
...
+            id_at_location=id_at_location,
+            owner=owner,
+            garbage_collect_data=garbage_collect_data,
+            id=id,
+            tags=tags,
+            description=description,
+        )
+
+    def get(self):
+        dataset = self.location.get_obj(self.id_at_location)
```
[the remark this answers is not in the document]
abogaziah: Ok, I changed it to use the inherited .get() functionality of ObjectPointer

syft/workers/base.py
```diff
@@ -951,7 +951,7 @@ def search(self, query: Union[List[Union[str, int]], str, int]) -> List[PointerT
                 # set garbage_collect_data to False because if we're searching
                 # for a tensor we don't own, then it's probably someone else's
                 # decision to decide when to delete the tensor.
-                ptr = obj.create_pointer(garbage_collect_data=False, owner=sy.local_worker).wrap()
+                ptr = obj.create_pointer(garbage_collect_data=False, owner=sy.local_worker, location=self).wrap()
```
LaRiffle: Why did you need to change this `location` argument?

abogaziah: because if the location wasn't passed it would assume the location is the local worker, which isn't correct. the search result would be always me | id -> me | id_at_location

Issue #2927, opened 2020-01-18
Small rounding errors in handcrafted Conv2d and Pool2d Layers


iamtrask:
**Describe the bug**
There is a small rounding error in these layers, as discussed in https://github.com/OpenMined/PySyft/pull/2896/files/fa21f480c8bfe4196dac2f911cd938db7ff96f03#diff-369e971f29cf5226c9b64f75072095ac

**To Reproduce**
Run the following unit tests

https://github.com/OpenMined/PySyft/blob/39d2be0aeb4b5aa0413b6b8696cccb7fb960ab0d/test/torch/nn/test_conv.py#L7

https://github.com/OpenMined/PySyft/blob/39d2be0aeb4b5aa0413b6b8696cccb7fb960ab0d/test/torch/nn/test_pool.py#L6

**Expected behavior**
Ideally there would be 0 rounding error, but this might be a PyTorch issue not us.

arshjot:
I investigated this a bit and found that we need to do a direct summation instead of calculating sum of sums. The following modifications are required:
* [`conv.py`](https://github.com/OpenMined/PySyft/blob/acdc96828cebae674817d238f68e0e81dfd41071/syft/frameworks/torch/nn/conv.py#L94-L95) - change `.sum(3).sum(3)` to `.sum((3, 4))`
* [`pool.py`](https://github.com/OpenMined/PySyft/blob/acdc96828cebae674817d238f68e0e81dfd41071/syft/frameworks/torch/nn/pool.py#L65) - change `.sum(2).sum(2)` to `.sum((2, 3))`

By making the above changes, we get exactly the same output for both the pooling implementations. However, a rounding error is still present for convolution possibly due to floating-point arithmetic as indicated [here](https://discuss.pytorch.org/t/pytorch-conv2d-vs-numpy-reference-different-outcomes-rounding-error-or-mistake/8921).

karlhigley:
@arshjot Thanks for looking into it! Could you submit a PR with those changes?

arshjot:
Yes, I'll be happy to do so!

iamtrask:
Incredible work!

iamtrask:
As reported - there is still rounding error in the Conv2d implementation. This issue should stay open until it is resolved. I have spent a few more hours looking into it and haven't found anything yet.

LaRiffle:
That's strange, it looks like the PR solved it and chnaged the test form approx to exact check didn't it?

arshjot:
[...]

PR #2964, opened 2020-01-26
Modify conv implementation and corresponding test to get exact match

arshjot:
Solves #2927 by changing the convolution implementation and disabling mkldnn while running the test. Test for handcrafted convolution also modified to check for exact match.

### DIFF
--- a/test/torch/nn/test_conv.py
+++ b/test/torch/nn/test_conv.py
@@ -4,15 +4,15 @@ import torch.nn as nn
 import pytest
 
 
-# Disable mkldnn to avoid rounding errors due to difference in implementation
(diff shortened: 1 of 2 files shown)


test/torch/nn/test_conv.py
```diff
@@ -4,6 +4,10 @@
 import pytest
 
 
+# Disable mkldnn to avoid rounding errors due to difference in implementation
+th._C._set_mkldnn_enabled(False)
```
gmuraru: MAYBE: Wouldn't be better to unset the ```mkldnn``` and then set it to ```True``` again in the function.

arshjot: I agree, I will change it

PR #2965, opened 2020-01-26
[WIP] Add alice/bob example with store

gmuraru:
I tested on the local machine and it works...it seems that when I move to two separate machines I get the following error - need to investigate it:
```
<torch.distributed.TCPStore object at 0x7fd6cbff5b58>
  File "bob_work.py", line 14, in <module>
    bob.connect_to_mpc(1, 2)
  File "/home/george/sy/PySyft/syft/workers/base.py", line 241, in connect_to_mpc
    crypten.init(rank, world_size)
  File "/home/george/sy/CrypTen/crypten/__init__.py", line 23, in init
    init_ttp=crypten.mpc.ttp_required()
  File "/home/george/sy/CrypTen/crypten/communicator/__init__.py", line 32, in _init
    cls.initialize(rank, world_size, init_ttp=init_ttp)
  File "/home/george/sy/CrypTen/crypten/communicator/distributed_communicator.py", line 106, in initialize
    cls.instance = DistributedCommunicator(init_ttp=init_ttp, store=store)
  File "/home/george/sy/CrypTen/crypten/communicator/distributed_communicator.py", line 65, in __init__
    rank=self.rank,
  File "/home/george/miniconda2/envs/pysyft/lib/python3.7/site-packages/torch/distributed/distributed_c10d.py", line 410, in init_process_group
    timeout=timeout)
  File "/home/george/miniconda2/envs/pysyft/lib/python3.7/site-packages/torch/distributed/distributed_c10d.py", line 478, in _new_process_group_helper
    timeout=timeout)
RuntimeError: [/pytorch/third_party/gloo/gloo/transport/tcp/pair.cc:761] connect [127.0.1.1]:29577: Connection refused
```
For those changes I had to modify the initialization method to keep a store variable in the ```DistributedCommunicator```

Testing scenario:
1. One worker on one laptop and one worker on another.
2. One worker will be the master : will have set the MASTER_IP as 127.0.0.1 and the other one will have the MASTER_IP: IP_ADDRESS_OF_THE_MASTER
3. Run the workers and they should set/get the values in the data tables.

I think this is not the correct approach taking into consideration the discussion from the Slack Pytorch channel - they suggest us to use send/recv to send parameters across machine

### DIFF
--- a/crypten_int_gm/crypten/crypten/__init__.py
+++ b/crypten_int_gm/crypten/crypten/__init__.py
@@ -1,0 +1,368 @@
+#!/usr/bin/env python3
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import crypten.communicator as comm
+import crypten.mpc  # noqa: F401
+import crypten.nn  # noqa: F401
+import torch
+
+# other imports:
+from . import debug
+from .cryptensor import CrypTensor
+from .mpc import ptype
+
+
+def init(rank=0, world_size=1):
+    comm._init(use_threads=False,
+               rank=rank,
+               world_size=world_size,
+               init_ttp=crypten.mpc.ttp_required()
+            )
+    if comm.get().get_rank() < comm.get().get_world_size():
+        _setup_przs()
+        if crypten.mpc.ttp_required():
+            crypten.mpc.provider.ttp_provider.TTPClient._init()
+
+
+def init_thread(rank, world_size):
+    comm._init(use_threads=True, rank=rank, world_size=world_size)
+    _setup_przs()
+
+
+def uninit():
+    return comm.uninit()
+
+
+def is_initialized():
+    return comm.is_initialized()
+
+
+# the different private type attributes of an mpc encrypted tensor
+arithmetic = ptype.arithmetic
+binary = ptype.binary
+
+
+def print_communication_stats():
+    comm.get().print_communication_stats()
+
+
+def reset_communication_stats():
+    comm.get().reset_communication_stats()
+
+
+# Set backend
+__SUPPORTED_BACKENDS = [crypten.mpc]
+__default_backend = __SUPPORTED_BACKENDS[0]
+
+
+def set_default_backend(new_default_backend):
+    """Sets the default cryptensor backend (mpc, he)"""
+    global __default_backend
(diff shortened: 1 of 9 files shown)


crypten_int_gm/crypten/crypten/__init__.py
```diff
@@ -0,0 +1,368 @@
+#!/usr/bin/env python3
```
gmuraru: This is a change in crypten - I simply pass the rank and the world size in init

crypten_int_gm/crypten/crypten/communicator/distributed_communicator.py
```diff
@@ -0,0 +1,284 @@
+#!/usr/bin/env python3
```
gmuraru: The same is available for the ```DistributedCommunicator``` I keep as member the self.store table.

syft/frameworks/mpc/crypten/__init__.py
```diff
@@ -0,0 +1,15 @@
+import crypten
```
gmuraru: Those are needed to have a simple method to put (key, value) objects in the store

syft/workers/base.py
```diff
@@ -221,6 +225,27 @@ def registration_enabled(self):
         finally:
             self.is_client_worker = True
 
+    def connect_to_mpc(self, rank, world_size):
+        import os
+        default_args = {
+             "DISTRIBUTED_BACKEND": "gloo",
```
youben11: Shouldn't this be BACKEND?

syft/workers/base.py
```diff
@@ -221,6 +225,27 @@ def registration_enabled(self):
...
+             "MASTER_IP": "127.0.0.1",
+             "MASTER_PORT": 9000,
+             "WORLD_SIZE": world_size,
+             "RANK": rank,
```
youben11: You should also add "RENDEZVOUS" as "env://" so that init_method is set appropriately as stated in the [docs](https://pytorch.org/docs/stable/distributed.html#environment-variable-initialization)

gmuraru: Currently I use the ```store```, ```init_method``` is mutually exclusive with ```store```

syft/workers/base.py
gmuraru: I took this from the ```crypten/communicator/distributed_communicator.py``` and it also seems that on line 35 in the ```DistributedCommunicator``` the check is done for ```distributed_backend```.upper()

PR #2972, opened 2020-01-29
Add Protobuf serialization for Placeholders, Plans, and States

karlhigley:
Depends on OpenMined/syft-proto#33.


syft/messaging/message.py
```diff
@@ -270,7 +294,20 @@ def unbufferize(worker: AbstractWorker, protobuf_obj: "OperationMessagePB") -> "
...
+
+        if return_placeholders:
+            if len(return_placeholders) == 1:
+                operation_msg = Operation(
+                    command, owner, tuple(args), kwargs, return_placeholders[0]
+                )
+            else:
+                operation_msg = Operation(command, owner, tuple(args), kwargs, return_placeholders)
+        else:
+            operation_msg = Operation(command, owner, tuple(args), kwargs, tuple(return_ids))
```
karlhigley: This is a workaround for the overloading of `return_ids` to also contain placeholders some of the time, which doesn't work well in Protobuf, because members of a `oneof` field can't be repeated. In order to remove the need for this, I think some refactoring of the `Operation` class is probably in order, but I'd probably defer it to a later PR.

syft/messaging/plan/plan.py
```diff
@@ -578,6 +580,100 @@ def detail(worker: AbstractWorker, plan_tuple: tuple) -> "Plan":
...
+        protobuf_plan.name = plan.name
+        protobuf_plan.tags.extend(plan.tags)
+
+        if protobuf_plan.description:
+            protobuf_plan.description = plan.description
+
+        if type(plan.placeholders) == type(dict()):
+            placeholders = plan.placeholders.values()
+        else:
+            placeholders = plan.placeholders
```
karlhigley: I'm not 100% sure that `plan.placeholders` can be anything other than a `dict` here. `replace_with_placeholders()` seems to have multiple potential return types, but I'm not sure they actually get stored anywhere. Possible this `if/else` can be simplified to just assume that it will receive `dict`.

syft/messaging/plan/plan.py
```diff
@@ -578,6 +580,100 @@ def detail(worker: AbstractWorker, plan_tuple: tuple) -> "Plan":
...
+        Returns:
+            plan: a Plan object
+        """
+
+        worker._tmp_placeholders = {}
+        id = sy.serde.protobuf.proto.get_protobuf_id(protobuf_plan.id)
+
+        operations = []
+        for operation in protobuf_plan.operations:
+            op_msg = OperationMessagePB()
```
karlhigley: Why are we creating Protobuf objects in `unbufferize`? The Protobuf schemas maintain a distinction between operations (without regard to where they might be executed) and operation messages (which contain operations being sent somewhere.) That distinction doesn't exist in Python yet, which means that `_unbufferize` doesn't know how to handle raw `OperationPB` objects yet. I think this should be cleaned up, but maybe as part of further refactoring the `Operation` class.

syft/messaging/plan/state.py
```diff
@@ -137,3 +140,62 @@ def detail(worker: AbstractWorker, state_tuple: tuple) -> "State":
...
+        protobuf_placeholders = [
+            sy.serde.protobuf.serde._bufferize(worker, placeholder)
+            for placeholder in state.state_placeholders
+        ]
+        protobuf_state.placeholders.extend(protobuf_placeholders)
+
+        state_tensors = []
+        for tensor in state.tensors():
+            protobuf_tensor = sy.serde.protobuf.serde._bufferize(worker, tensor)
+            state_tensor = StateTensorPB()
```
karlhigley: `StateTensorPB` exists to smooth over the multiple possible types in `state.tensors()`, which may include Torch parameters in addition to various other tensor types. It has no corresponding class in Python.

syft/messaging/message.py
karlhigley: I was initially thinking of adding a separate field to the `Operation` class for placeholders, so that either ids or placeholders could be present (but hopefully not both.) The more I think about it though, the more I wonder if it would make sense to introduce separate classes for `AdHocOperations` and `PlannedOperations` (or something like that.)

syft/messaging/plan/plan.py
karlhigley: @LaRiffle Is the structure of `plan.placeholders` always a `dict` now?

syft/messaging/plan/state.py
karlhigley: @LaRiffle Might be helpful to make a list of what tensor types can be included in the state. I've included Torch tensors and parameters so far, but pretty sure there are more than that, right?

karlhigley:
"Bufferize" was @vvmnnnkv's brainstorm. (I think he was joking, but I liked it.)

Issue #2981, opened 2020-01-30
Support field of 2**64


Yugandhartripathi:
**Is your feature request related to a problem? Please describe.**
Too many modulo operations involved in basic operations which costs significant computation time and resources.

**Describe the solution you'd like**
AST and FPT should work in finite field of 2**64 and since LongTensor automatically wraps in that range we no longer need to explicitly call modulo operations at every stage.

**Describe alternatives you've considered**
NA

**Additional context**
NA

Yugandhartripathi:
We can close this one now :smile:

karlhigley:
🎉

PR #2982, opened 2020-01-30
Increase field to 2**64

Yugandhartripathi:
Issue: #2981
General:
- [x] Documentation fixes (remove my debugging notes)
- [x] Remove CRTTensor
- [x] Remove LargePrecisionTensor
- [x] Tutorial Part 9 fix in all languages
- [ ] Optimisations

FPT:
- [x] Get FPT operations working for 64 bit and 32 bit field
- [x] Add dtype arg (long: 64bit field and int: 32bit field)
- [x] Typecasting and test for dtype arg
- [x] Add tests for ops on 32 bit field

AST:
- [x] support dtype arg
- [x] Add test for dtype and ops on 32 bit field ~(only div left due to snn issues)~
- [x] Get AST ops working for 64/32 bit field
- [x] Refactor cat and stack
~- spdz mul issues~
~- securenn div issues~
~- securenn relu_deriv issues for comparison ops~
Turns out there is some issue with `.send` that's causing above 3 to fail so just need to find it
Problem: 2**64 field arg was causing overflow during serialisation

Serialisation:
- [x] Msgpack hacky solution: convert field from int->str before serialization and str->int after deserialization 
- [x] Protobuf support for 2**64 values
- [x] Add dtype in protobuf serialisation
- [x] syft_proto changes to support above changes

Beaver:
- [x] Negative share support for triples

SecureNN:
- [x] Update Qbits and remove modulo that can be handled by native tensor
- [x] Investigate and handle behavior of shares with L-1,p field values

Investigation Report:
Problem: Need some way to support custom fields like old times. And for that need custom modulo function to handle negative share values properly.
Solution:
AST:
- [x] custom field support in AST using dtype
- [x] custom modulo function for negative custom field support in AST
- [x] AST test for dtype="custom" and field=67 and field=2**64-1 ~(private_mul spdz dtype mismatch issue)~
- [x] Tests for custom dtype(with field=67 and field=2**64-1)

SecureNN:
- [x] - 32 bit support (any operations between torch.IntTensor and torch.LongTensor give result as torch.LongTensor and that's a problem also initialise Q_bits based on field value instead of default 64)
- [x] Tests for 32 bit ops (need to resolve recursion error in share_convert and msb output shares always in 2**64 issue)


syft/__init__.py
```diff
@@ -64,13 +64,11 @@
...
 from syft.frameworks.torch.tensors.decorators.logging import LoggingTensor
 from syft.frameworks.torch.tensors.interpreters.additive_shared import AdditiveSharingTensor
-from syft.frameworks.torch.tensors.interpreters.crt_precision import CRTPrecisionTensor
 from syft.frameworks.torch.tensors.interpreters.autograd import AutogradTensor
 from syft.frameworks.torch.tensors.interpreters.precision import FixedPrecisionTensor
 from syft.frameworks.torch.tensors.interpreters.numpy import create_numpy_tensor as NumpyTensor
 from syft.frameworks.torch.tensors.interpreters.private import PrivateTensor
-from syft.frameworks.torch.tensors.interpreters.large_precision import LargePrecisionTensor
-from syft.execution.placeholder import PlaceHolder
+from syft.frameworks.torch.tensors.interpreters.placeholder import PlaceHolder
```
karlhigley: This actually should be imported from `syft.execution.placeholder`.

Yugandhartripathi: Ok yeah, I messed it up while resolving merge conflict probably. I will fix it.

Yugandhartripathi:
Well that's interesting it's failing for `test/crypto/test_snn->test_share_convert` on 3.7. It's passing that test every time on local though. I will investigate it further.

As for the other test `test/serde/protobuf/test_protobuf_serde_full.py::test_serde_roundtrip_protobuf[AdditiveSharingTensor]` I think it is dependent on merge of this [PR](https://github.com/OpenMined/syft-proto/pull/48).

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -156,13 +195,7 @@ def get(self):
...
                 shares.append(share)
 
-        res_field = sum(shares) % self.field
-
-        gate = res_field.native_gt(self.field / 2).long()
-        neg_nums = (res_field - self.field) * gate
-        pos_nums = res_field * (1 - gate)
-        result = neg_nums + pos_nums
-
+        result = self.modulo(sum(shares), self.field) if self.dtype == "custom" else sum(shares)
```
Jasopaum: Maybe add a small comment to say that the modulo is performed automatically with torch.int32 and torch.int64 values (if I'm not wrong ^^)

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -721,8 +764,8 @@ def unbind(tensor_shares, **kwargs):
 
         module.unbind = unbind
 
-        @overloaded.function
-        def stack(tensors_shares, **kwargs):
+        def cat_and_stack(tensors_shares, cmd, **kwargs):
```
Jasopaum: nit: maybe the part where the if is needed could be performed in `cat` and `stack` directly?

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -38,10 +40,30 @@ def __init__(
...
+                    self.dtype = "int"
+                    self.field = 2 ** 32
+                else:
+                    self.dtype = "long"
+                    self.field = 2 ** 64
+            else:
+                # Invalid args dtype and field
+                raise ValueError(
+                    "Unsupported arg value for dtype. Use dtype='long' or dtype='int'."
+                )
```
Jasopaum: Just a question: why can we have a custom field for AST but not for FPT?

syft/serde/msgpack/serde.py
```diff
@@ -375,6 +373,14 @@ def deserialize(binary: bin, worker: AbstractWorker = None) -> object:
     return _deserialize_msgpack_simple(simple_objects, worker)
 
 
+def _simplify_field(obj):
```
Jasopaum: Could you explain what you're doing here?

syft/serde/msgpack/serde.py
```diff
@@ -446,6 +454,13 @@ def _simplify(worker: AbstractWorker, obj: object, **kwargs) -> object:
         return obj
 
 
+def _detail_field(typeCode, val):
```
Jasopaum: Maybe write `str(2^64)` instead of "18446744073709551616" so that it's clearer what is means

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -145,6 +172,18 @@ def backward(self, *args, **kwargs):
...
+        mask_pos = x > ((field - 1) // 2)
+        mask_neg = x < -(field // 2)
+        if mask_pos.any():
+            mask_pos = mask_pos.long()
+            return self.modulo(x - (mask_pos * field), field)
+        elif mask_neg.any():
+            mask_neg = mask_neg.long()
+            return self.modulo(x + (mask_neg * field), field)
+        else:
+            return x.type(self.field_to_torch_dtype)
```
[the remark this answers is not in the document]
Yugandhartripathi: Yes, for automatic modulo based on native tensor dtype.

syft/frameworks/torch/tensors/interpreters/precision.py
Yugandhartripathi: Initial plan was to not provide an option for custom field at all for both FPT and AST. But there are certain use cases in securenn that demand field to be 67 and 2**64-1(for dtype long) and 2**32-1(for dtype int). 
So had to introduce custom dtype option that should only be used internally. But we don't need anything of the sort for FPT.

syft/serde/msgpack/serde.py
Yugandhartripathi: msgpack can't pack python integers larger than 2**64-1 and causes an error like: "Python int too large and can't be converted to C unsigned long long". So we need to convert all values larger than that into string for successful serialisation.

syft/serde/msgpack/serde.py
Yugandhartripathi: Oh yeah, sure. I will change it

syft/frameworks/torch/tensors/interpreters/additive_shared.py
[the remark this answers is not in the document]
Jasopaum: I just find it a bit weird that this function is applied to x, takes a field and uses self for the dtype (which should be possible to infer from the field, am I wrong?).

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -38,10 +40,30 @@ def __init__(
         """
         super().__init__(id=id, owner=owner, tags=tags, description=description)
 
-        self.field = field
         self.base = base
         self.precision_fractional = precision_fractional
         self.kappa = kappa
+        if dtype == "long":
```
iamtrask: In the future, (long after this PR is merged), if we wanted to represent something bigger than "long" - how would we do so?

(Hint: please mention something about a separate BigIntTensor abstraction tensor type  - not extending FixedPrecisionTensor with the ability to represent even larger precisions.)

iamtrask: Please write a github issue for future work here before merging

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -51,6 +52,34 @@ def __init__(
...
+                raise ValueError("Field cannot be None for custom dtype")
+            self.field = field
+        elif dtype == "long":
+            self.dtype = dtype
+            self.field = 2 ** 64
+        elif dtype == "int":
+            self.dtype = dtype
+            self.field = 2 ** 32
+        else:
+            warnings.warn("Use dtype instead of field")
```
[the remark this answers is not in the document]
Yugandhartripathi: Okay, I will check and add appropriate message for that.

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -65,14 +94,12 @@ def __init__(
...
-        self.field = (2 ** securenn.Q_BITS) if field is None else field  # < 63 bits
-        self.n_bits = (
-            n_bits if n_bits is not None else max(8, round(math.log(self.field, 2)))
-        )  # < 63 bits
+        self.n_bits = round(math.log(self.field, 2))
         # assert 2 ** self.n_bits == self.field
         self.crypto_provider = (
             crypto_provider if crypto_provider is not None else sy.hook.local_worker
         )
+        self.field_to_torch_dtype = torch.int32 if self.field <= 2 ** 32 else torch.int64
```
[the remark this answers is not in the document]
Yugandhartripathi: Yeah that would be much better. I will rename it

syft/frameworks/torch/tensors/interpreters/additive_shared.py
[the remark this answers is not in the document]
Yugandhartripathi: Yes, that does seem like a roundabout way of doing it and it can cause if not issues then confusion in the future. I will modify it to derive torch_dtype from the field itself here.

syft/frameworks/torch/tensors/interpreters/additive_shared.py
Yugandhartripathi: Yes you are right. I will add the comment.

syft/frameworks/torch/tensors/interpreters/additive_shared.py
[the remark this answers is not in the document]
Jasopaum: And then, if self is not used, you can make this a static method :)

syft/frameworks/torch/mpc/beaver.py
```diff
@@ -8,6 +8,7 @@ def request_triple(
     crypto_provider: AbstractWorker,
     cmd: Callable,
     field: int,
+    dtype: str,
```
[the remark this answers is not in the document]
Yugandhartripathi: Sure, I am on it.

syft/frameworks/torch/tensors/interpreters/additive_shared.py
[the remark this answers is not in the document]
Yugandhartripathi: Yes that will be for the best. As I can revert generate_shares to be static too then.

syft/frameworks/torch/tensors/interpreters/precision.py
Yugandhartripathi: Okay, I am on it.

syft/frameworks/torch/tensors/interpreters/additive_shared.py
Yugandhartripathi: Yes, done.

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -51,6 +52,34 @@ def __init__(
         """
         super().__init__(id=id, owner=owner, tags=tags, description=description)
 
+        self.child = shares
+        if dtype == "custom":
+            self.dtype = dtype
+            if field == None:
```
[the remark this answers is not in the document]
Yugandhartripathi: ok, done

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -51,6 +52,34 @@ def __init__(
...
+        elif dtype == "long":
+            self.dtype = dtype
+            self.field = 2 ** 64
+        elif dtype == "int":
+            self.dtype = dtype
+            self.field = 2 ** 32
+        else:
+            warnings.warn("Use dtype instead of field")
+            # Since n mod 0 is not defined
+            if type(field) == int and field > 0:
```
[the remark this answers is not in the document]
Yugandhartripathi: Yeah, done

syft/serde/msgpack/serde.py
```diff
@@ -455,7 +455,7 @@ def _simplify(worker: AbstractWorker, obj: object, **kwargs) -> object:
 
 
 def _detail_field(typeCode, val):
-    if typeCode == 5 and val == "18446744073709551616":
+    if typeCode == 5 and val == str(2 ** 64):
```
karlhigley: What does `typeCode == 5` signify here? Which type does `5` mean?

Yugandhartripathi: It's the proto.json defined code for type str

Jasopaum: Do you need to hardcode it? Isn't there a way to have the code from the type?

Yugandhartripathi: I haven't explored that possibility yet. I will look into it.

karlhigley: I think you can use something [like this](https://github.com/OpenMined/PySyft/blob/ba804732961722d9e9b41e19fd580691bf5ebf42/test/serde/msgpack/test_msgpack_serde.py#L51).

Yugandhartripathi:
If the review request here for writing team is not automated then I am sorry I must have clicked on the option by mistake.

syft/serde/msgpack/serde.py
Yugandhartripathi: Done :+1:

karlhigley:
It's automated to request Writing Team review when the translations change. 👍

syft/serde/msgpack/serde.py
```diff
@@ -69,7 +68,7 @@
 from syft.messaging.message import SearchMessage
 from syft.messaging.message import PlanCommandMessage
 from syft.messaging.message import WorkerCommandMessage
-from syft.serde import compression
+from syft.serde import compression, msgpack
```
karlhigley:
```suggestion
from syft.serde import compression
from syft.serde import msgpack
```

karlhigley:
I notice that the tests take ~35 minutes to run on this PR, compared to ~17 minutes on the latest commit on `master`. I'm not sure if that's because this PR adds some slow tests, or if it maybe slows down existing tests. I'll defer to @OpenMined/crypto-team to assess potential performance impacts re: MPC, but I'm hoping we can find a way to speed the test suite back up somewhat.

Yugandhartripathi:
I have added some extensive tests here for testing dtype arg compatibility with each and every operation I almost doubled the size of every test function in AST and FPT and also for securenn. I can remove some parts of it I added for catching corner cases as they might be a little overkill now that dtype arg is working well.
As for the performance impact: I don't think(or atleast I hope) these changes are not making tests slower because a major point for removing explicit modulos and using native tensor types for wrapping was to get performance improvement from removing repeated modulos in every function.

syft/serde/msgpack/serde.py
Yugandhartripathi: Any reasons for doing it this way?

karlhigley: It makes the diffs of future PRs cleaner if one of those needs to change, which makes rebasing/cherry-picking/merging easier.

karlhigley:
I’d guess that SecureNN is probably the slow part, right?

Yugandhartripathi:
Securenn tests does take more time than other tests. But it should still be faster or equal to the master branch's version of the same tests. The only thing serving to the contrary is double the number of tests every securenn test is duplicated and ran once for dtype=long and once for dtype=int.

Yugandhartripathi:
Ok so removing tests didn't work out. Tests are still taking roughly double the time of master. Also the advanced/encrypted_lr tutorial notebook is failing due to timeout which again signifies slow down of MPC ops. I don't see any changes that should lead to that much slowdown overall. 
@LaRiffle Is it possible that it is simply taking more time because of the increased field size as we have all ops now happening in field 2**64 by default?

LaRiffle:
I've done a small benchmark, apparently your code is faster than for example my FSS PR on `test/torch` (320s vs 300s on my laptop)

syft/serde/msgpack/serde.py
```diff
@@ -450,6 +459,13 @@ def _simplify(worker: AbstractWorker, obj: object, **kwargs) -> object:
         return obj
 
 
+def _detail_field(typeCode, val):
+    if typeCode == msgpack.proto_type_info(str).code and val == str(2 ** 64):
```
LaRiffle, line 463: cache `str(2 ** 64):` in a variable

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -288,33 +304,17 @@ def mul_and_div(self, other, cmd):
...
-            )
+            # The comparison is different if new_other is a torch tensor or an AST
+            sgn_other = (new_other > 0).type(self.torch_dtype)
             pos_other = new_other * sgn_other
-            neg_other = (
-                (self.field - new_other) * (1 - sgn_other)
-                if isinstance(new_other, torch.Tensor)
-                else new_other * (sgn_other - 1)
-            )
+            neg_other = new_other * (sgn_other - 1)
```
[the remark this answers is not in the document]
Yugandhartripathi: I left it there initially because we were using sign related info below in some cases. 
But now upon reconsideration it is clear that it is just redundant with signed shares and should be removed to save computation.

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -721,40 +771,34 @@ def unbind(tensor_shares, **kwargs):
 
         module.unbind = unbind
 
-        @overloaded.function
-        def stack(tensors_shares, **kwargs):
+        def cat_and_stack(tensors_shares):
```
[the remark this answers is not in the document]
Yugandhartripathi: Done

syft/frameworks/torch/mpc/securenn.py
```diff
@@ -361,12 +402,10 @@ def share_convert(a_sh):
         ]
     )
     eta_sh = eta_p_sh + (1 - j) * eta_pp - 2 * eta_pp * eta_p_sh
-
     # 10)
     theta_sh = beta - (1 - j) * (alpha + 1) + delta_sh + eta_sh
-
     # 11)
-    y_sh = -theta_sh + a_sh + u_sh
+    y_sh = (0 * theta_sh) + a_sh + u_sh
```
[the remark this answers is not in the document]
Yugandhartripathi: Yes, I am still not entirely clear on the maths behind it but as we discussed it seems a simple add operation with zero shares in L-1 field gets the job done here. 
I will comment unused parts out and open a separate issue for further inspection into this.

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -145,6 +174,20 @@ def backward(self, *args, **kwargs):
         we just ignore the call."""
         pass
 
+    @staticmethod
+    def modulo(x, field):
```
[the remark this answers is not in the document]
Yugandhartripathi: x can be either AST, PointerTensor or simple torch.Tensor so I don't think we can annotate type for the function.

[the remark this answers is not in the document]
Yugandhartripathi: It is definitely expensive but we currently only need it for custom field ops that are in securenn.

syft/serde/msgpack/serde.py
Yugandhartripathi: Done :+1:

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -65,14 +96,12 @@ def __init__(
...
-        self.field = (2 ** securenn.Q_BITS) if field is None else field  # < 63 bits
-        self.n_bits = (
-            n_bits if n_bits is not None else max(8, round(math.log(self.field, 2)))
-        )  # < 63 bits
+        self.n_bits = round(math.log(self.field, 2))
         # assert 2 ** self.n_bits == self.field
         self.crypto_provider = (
             crypto_provider if crypto_provider is not None else sy.hook.local_worker
         )
+        self.torch_dtype = torch.int32 if self.field <= 2 ** 32 else torch.int64
```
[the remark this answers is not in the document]
Yugandhartripathi: Sure

syft/frameworks/torch/mpc/spdz.py
```diff
@@ -39,14 +42,16 @@ def spdz_mul(cmd: Callable, x_sh, y_sh, crypto_provider: AbstractWorker, field:
...
         j = sy.MultiPointerTensor(children=[j1, j0])
     else:
         j = sy.MultiPointerTensor(children=[j1] + list(j0.child.values()))
 
     delta_b = cmd(delta, b)
     a_epsilon = cmd(a, epsilon)
-
-    return delta_epsilon * j + delta_b + a_epsilon + a_mul_b
+    res = delta_epsilon * j + delta_b + a_epsilon + a_mul_b
+    res = res.type(field_to_dtype)
```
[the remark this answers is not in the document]
Yugandhartripathi: Yes, I believe it solves the problem where spdz output was causing recursion depth limit exceeded issues in `modulo` function for custom field value ops.

syft/frameworks/torch/mpc/securenn.py
```diff
@@ -321,37 +333,66 @@ def share_convert(a_sh):
...
+            (
+                a_tilde_shares[workers[0].id].copy().get()
+                + a_tilde_shares[workers[1].id].copy().get()
+            )
+            > ((L - 1) // 2)
+        )
+        + (
+            (
+                a_tilde_shares[workers[0].id].copy().get()
+                + a_tilde_shares[workers[1].id].copy().get()
```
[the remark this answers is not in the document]
Yugandhartripathi: Done :+1:

syft/frameworks/torch/mpc/securenn.py
```diff
@@ -14,30 +15,31 @@
...
 
 
-def decompose(tensor):
+def decompose(tensor, field):
     """decompose a tensor into its binary representation."""
-    n_bits = Q_BITS
-    powers = torch.arange(n_bits)
+    field_to_dtype = torch.int64 if field > 2 ** 32 else torch.int32
+    n_bits = round(math.log(field, 2))
+    powers = torch.arange(n_bits, dtype=field_to_dtype)
```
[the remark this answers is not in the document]
Yugandhartripathi: Done

syft/frameworks/torch/mpc/securenn.py
```diff
@@ -321,37 +333,66 @@ def share_convert(a_sh):
...
+    u_sh = _shares_of_zero(1, L - 1, "custom", crypto_provider, *workers)
 
     # 2)
     a_tilde_sh = a_sh + r_sh
     a_shares = a_sh.child
-    beta0 = ((a_shares[workers[0].id] + r_shares[workers[0].id]) >= L).long()
-    beta1 = ((a_shares[workers[1].id] + r_shares[workers[1].id]) >= L).long()
+    beta0 = (
+        ((a_shares[workers[0].id] + r_shares[workers[0].id]) > ((L - 1) // 2))
+        + ((a_shares[workers[0].id] + r_shares[workers[0].id]) < -(L // 2))
```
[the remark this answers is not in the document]
Yugandhartripathi: Can't I just do this with memorize decorator on a field_max and field_min methods?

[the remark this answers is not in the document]
Yugandhartripathi: We call this modulo method only for custom dtype which is only for internal use so I know it will only get 3 field values (2**32 -1, 2**64 -1, 67). Which can be memorized to speed it up .

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -145,6 +174,20 @@ def backward(self, *args, **kwargs):
         we just ignore the call."""
         pass
 
+    @staticmethod
+    def modulo(x, field):
+        torch_dtype = torch.int64 if field > 2 ** 32 else torch.int32
+        mask_pos = x > ((field - 1) // 2)
+        mask_neg = x < -(field // 2)
```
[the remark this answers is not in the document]
Yugandhartripathi: I made an auxilliary function for this alongside max and min with memorize decorator.

syft/frameworks/torch/tensors/interpreters/precision.py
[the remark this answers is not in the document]
Yugandhartripathi: Update: Removing this is causing shape mismatch assertion error when linalg calls securenn.division. Need to investigate further. Keeping this logic in place for now.

syft/frameworks/torch/mpc/securenn.py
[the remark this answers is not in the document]
LaRiffle: I agree! But as the  `(L - 1) // 2`  isn't that slow, you won't get substantial gain using memorize like this. I would suggest keeping those as attributes for the AST (well property to be exact) and compute them them only when required (possibly using memrize this time). That's not the thing which will bring huge speed up but it will help :)

syft/frameworks/torch/mpc/securenn.py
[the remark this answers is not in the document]
Yugandhartripathi: Commented everything out. Now it's just u_sh and a_sh and it works.

syft/frameworks/torch/mpc/securenn.py
```diff
@@ -188,7 +222,7 @@ def private_compare(x_bit_sh, r, beta):
     c_else = (l1_mask * c_ie1) + ((1 - l1_mask) * c_igt1)
 
     # Mask for the case r == 2^l −1
-    r_mask = (r == (L - 1)).long()
+    r_mask = (r == ((L // 2) - 1)).long()
```
LaRiffle, line 225: don't forget to cache this one

syft/frameworks/torch/mpc/spdz.py
```diff
@@ -39,14 +42,16 @@ def spdz_mul(cmd: Callable, x_sh, y_sh, crypto_provider: AbstractWorker, field:
     delta_epsilon = cmd(delta, epsilon)
 
     # Trick to keep only one child in the MultiPointerTensor (like in SNN)
-    j1 = torch.ones(delta_epsilon.shape).long().send(locations[0], **no_wrap)
-    j0 = torch.zeros(delta_epsilon.shape).long().send(*locations[1:], **no_wrap)
+    field_to_dtype = torch.int32 if field <= 2 ** 32 else torch.int64
```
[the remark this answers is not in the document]
LaRiffle: don't forget to cache this one

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -65,10 +101,7 @@ def __init__(
         else:
             self.child = None
 
-        self.field = (2 ** securenn.Q_BITS) if field is None else field  # < 63 bits
-        self.n_bits = (
-            n_bits if n_bits is not None else max(8, round(math.log(self.field, 2)))
-        )  # < 63 bits
+        self.n_bits = round(math.log(self.field, 2))
```
LaRiffle, line 104: Cache this :)

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -505,12 +552,20 @@ def _public_mul(self, shares, other, equation):
...
+                        if self.dtype == "custom"
+                        else cmd(share, other) + zero_shares[worker]
+                    )
                 return res
             else:
                 return {
-                    worker: (cmd(share, other) % self.field) for worker, share in shares.items()
+                    worker: (
+                        self.modulo(cmd(share, other), self.field)
+                        if self.dtype == "custom"
```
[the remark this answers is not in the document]
LaRiffle: ping here, this would make your code simpler :)

syft/frameworks/torch/tensors/interpreters/precision.py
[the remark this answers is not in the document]
LaRiffle: > shape mismatch

That's intriguing!

syft/frameworks/torch/tensors/interpreters/additive_shared.py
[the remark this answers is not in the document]
Yugandhartripathi: Oh, I missed this one. I am on it :+1:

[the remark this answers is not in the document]
Yugandhartripathi: Done

syft/frameworks/torch/tensors/interpreters/additive_shared.py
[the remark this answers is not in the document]
Yugandhartripathi: Refactored modulo as non-static now we can use self.torch_dtype directly and for min/max using property

syft/frameworks/torch/mpc/securenn.py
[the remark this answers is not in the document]
Yugandhartripathi: I added the property for min_value and max_value but I am not sure how to cache it into private variable after computing first time from there. Need help.

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -139,6 +139,14 @@ def shape(self):
         for share in self.child.values():
             return share.shape
 
+    @property
+    def min_value(self):
+        return -(self.field // 2)
+
+    @property
+    def max_value(self):
+        return (self.field - 1) // 2
```
LaRiffle, line 148:
```suggestion
    # declare in the init  self._min_value  = None and same for max
    @property
    def min_value(self):
        if self._min_value is None:
            self._min_value = -(self.field // 2)
        return self._min_value

    @property
    def max_value(self):
        # same here
```

Yugandhartripathi: Thanks, done :+1:

Yugandhartripathi:
Failing on `test/serde/simplify` as detailed here #3359

iamtrask:
Absolutely fantastic work @Yugandhartripathi !!!

syft/frameworks/torch/mpc/securenn.py
```diff
@@ -305,68 +347,85 @@ def share_convert(a_sh):
...
+    # )
+    # eta_sh = eta_p_sh + (1 - j) * eta_pp - 2 * eta_pp * eta_p_sh
     # 10)
-    theta_sh = beta - (1 - j) * (alpha + 1) + delta_sh + eta_sh
-
+    # theta_sh = beta - (1 - j) * (alpha + 1) + delta_sh + eta_sh
     # 11)
-    y_sh = -theta_sh + a_sh + u_sh
+    # NOTE:
+    # It seems simple operation with shares in L-1 field is enough to conver a_sh from L to L-1
```
knexator, line 426: No it isn't! This commit introduces subtle bugs in everything that depends on convert share and thus most of SecureNN

Simply try:
```
x = torch.tensor([-2, -1, 0, 1, 2]).share(alice, bob, dtype="custom", field=32, crypto_provider=crypto).child
relu_deriv(x).get() #gives garbage results 
```

Please read carefully the papers; if a mathematician has introduced tons of weird operations, it's probably for some reason.

karlhigley: @knexator Could you create an issue for this?

Yugandhartripathi: @knexator Well we did investigate it quite a bit. But you are right I missed out on it maybe tests I used weren't strict enough. Anyway let's open an issue and work to resolve it quickly.

Yugandhartripathi: It was left as comments for precisely this reason so that it can be further investigated

Yugandhartripathi: Also dtype="custom" is not intended for user. I think we need someway to make sure that user can't use custom dtype to share the tensors while at the same time it can be used internally for secureNN ops.

Yugandhartripathi: #3389

Yugandhartripathi: @knexator It's alright, no worries :slightly_smiling_face: . Let me know if you need help wrt any changes made here.

PR #2985, opened 2020-01-31
Disable `msgpack` serialization of raw `Message` objects

### LINKED ISSUE #2514: Automatic Typing for Default Message Detailer
At present, the default detailer for Message (an abstract type) returns an object of type Message instead of returning an object of the correct type.

Use the msg_tuple[0] to look up the correct type.

Return NotImplementedException instead and require all Message types to have their own detailers.

karlhigley:
Serializing generic messages won't work well in Protobuf, and this
capability isn't currently used in the existing code base, so let's
disable it.

Resolves #2514.

### DIFF
--- a/syft/messaging/message.py
+++ b/syft/messaging/message.py
@@ -58,6 +58,11 @@ class Message:
         """
         This function takes the attributes of a Message and saves them in a tuple.
         The detail() method runs the inverse of this method.
+
+        This method is provided as a convenience to make it easier to serialize
+        simple message type sub-classes, but other code should never rely on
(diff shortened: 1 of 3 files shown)


karlhigley:
Hmm, this is tricky. I still want to test `Message.simplify()`, but if it's included in the list of classes that have simplifiers and detailers, then the serialization coverage test expects it to be in the round-trip tests, which won't work without a real detailer.

karlhigley:
@vvmnnnkv Is there a reasonable way to test that the detailer raises an error in the round-trip serialization tests?

PR #2989, opened 2020-01-31
Add a failing test for JIT tracing a complicated Syft plan

karlhigley:
No assertions yet because the test errors out before any assertions could be checked.


karlhigley:
@vvmnnnkv I think you found a way to JIT trace complicated plans, so closing this draft PR.

Issue #2758, opened 2019-11-26
Part 10: Federated Learning with Encrypted Gradient Aggregation (Bug found)


Krypton3:
Hi, 

I was testing this tutorial with my own dataset. It's working fine till secure encrypted aggregation. However, I have found this error with my dataset. I have also followed the tutorial fully. It also shows the same error for secure encrypted aggregation. 

The error is the following:

> AttributeError                            Traceback (most recent call last)
> <ipython-input-72-431a5695a2e8> in <module>()
>       5 for epoch in range(args.epochs):
>       6     print(f"Epoch {epoch + 1}")
> ----> 7     train(epoch)
>       8     # test()
>       9 
> 
> 8 frames
> /usr/local/lib/python3.6/dist-packages/syft/frameworks/torch/hook/hook_args.py in <lambda>(i)
>      32 backward_func = {
>      33     TorchTensor: lambda i: i.wrap(),
> ---> 34     torch.Tensor: lambda i: i.wrap(),
>      35     torch.nn.Parameter: lambda i: torch.nn.Parameter(data=i),
>      36     AutogradTensor: lambda i: AutogradTensor(data=i).on(i, wrap=False),
> 
> AttributeError: 'numpy.ndarray' object has no attribute 'wrap'

@iamtrask  @LaRiffle 
I would appreciate an explanation. Thank you.

alejandrosame:
Hi there!

@iamtrask reported in the Slack group that the issue is due to the import of NumPy to execute log2 like shown in https://github.com/OpenMined/PySyft/blob/master/syft/frameworks/torch/tensors/interpreters/native.py#L741. 

Since NumPy can only run locally, it cannot be used for Federated Learning yet. The solutions for this error would be to either:
* Hook NumPy as another PySyft framework.
* Implement log2 in PyTorch.

Cheers!

MarksASP95:
Hey, I'm not sure how to apply the solutions you provided, could you be more specific?

MarksASP95:
This notebook works on version 0.1.13a1.

alejandrosame:
Hi @MarksASP95!

Both solutions involve development. I'm currently working and documenting the NumPy integration [on this issue](https://github.com/OpenMined/PySyft./issues/2771).

PR #2990, opened 2020-02-01
fix attributeerror in tutorial part10

Rachelxuan11:
**Problem**
Run tutorial Part10 and got the same error as this issue #2758. 
This error is shown as follows:

> <ipython-input-6-c4374ada1b4c> in train_enc()
     17             spdz_params = list()
     18             for remote_index in range(len(compute_node)): # for each client
---> 19                 spdz_params.append(params[remote_index][param_i].copy().fix_precision().share(bob, alice, crypto_provider=james).get()) # AttributeError: 'numpy.ndarray' object has no attribute 'wrap', if .get() is used here. This could be avoided by model.get() first, and then .share() the copy to a list of clients. This obeys the actually protocol of encrypted model averaging, but fixes this bug.

**Solution**
user `model.copy().get().fix_precision().share(...)` instead of `model.copy().fix_precision().share(...).get()` could avoid this error


examples/tutorials/Part 10 - Federated Learning with Secure Aggregation.ipynb
```diff
@@ -197,7 +197,7 @@
     "        if batch_idx % args.log_interval == 0:\n",
     "            loss = loss.get()\n",
     "            print('Train Epoch: {} [{}/{} ({:.0f}%)]\\tLoss: {:.6f}'.format(\n",
-    "                epoch, batch_idx * data.shape[0], len(train_loader),\n",
+    "                epoch, batch_idx * data.shape[0], len(train_loader) * data.shape[0],\n",
     "                       100. * batch_idx / len(train_loader), loss.item()))\n",
```
Rachelxuan11: a small change

examples/tutorials/Part 10 - Federated Learning with Secure Aggregation.ipynb
```diff
@@ -221,7 +221,6 @@
     "    for data, target in test_loader:\n",
     "        output = model(data)\n",
     "        test_loss += F.mse_loss(output.view(-1), target, reduction='sum').item() # sum up batch loss\n",
-    "        pred = output.data.max(1, keepdim=True)[1] # get the index of the max log-probability\n",
```
Rachelxuan11: redundant for this example regression task

examples/tutorials/Part 10 - Federated Learning with Secure Aggregation.ipynb
```diff
@@ -458,7 +457,7 @@
     "        for param_i in range(len(params[0])):\n",
     "            spdz_params = list()\n",
     "            for remote_index in range(len(compute_nodes)):\n",
-    "                spdz_params.append(params[remote_index][param_i].copy().fix_precision().share(bob, alice, crypto_provider=james).get())\n",
+    "                spdz_params.append(params[remote_index][param_i].copy().get().fix_precision().share(bob, alice, crypto_provider=james)\n",
```
Rachelxuan11: avoid the AttributeError

gmuraru:
It seems that there is an error here (it is lalso present on master, but also in this PR). Do you have some time to investigate :)
```
~/contrib/tests/PySyft/syft/generic/frameworks/hook/hook_args.py in <lambda>(i, **kwargs)
    742         if isinstance(r, (list, tuple))  # if the rule is a list or tuple.
    743         # Last if not, rule is probably == 1 so use type to return the right transformation.
--> 744         else lambda i, **kwargs: register_tensor(i, **kwargs)
    745         for a, r in zip(response, rules)  # And do this for all the responses / rules provided
    746     ]

~/contrib/tests/PySyft/syft/generic/frameworks/hook/hook_args.py in register_tensor(tensor, owner, response_ids)
    710             and each id is pop out when needed.
    711     """
--> 712     tensor.owner = owner
    713     try:
    714         tensor.id = response_ids.pop(-1)

AttributeError: 'numpy.ndarray' object has no attribute 'owner'
```

Issue #2622, opened 2019-09-22
Operate autograd tensors with normal tensors


LaRiffle:
**Describe the bug**
See ambiguity in Issue #2618, where we need to add `requires_grad = True` to input data to get the computation working.
We should do as we did for operating FixPrecisionTensor with AdditiveSharedTensor

See code of #2618 for details.

Issue #2937, opened 2020-01-20
Flaky test in test_precision.py


loopylangur:
**Describe the bug**
The test `test_torch_sigmoid_approx` in `test/torch/tensors/test_precision.py` fails intermittently (4 out of 30 times that i tried) with the following error:

```python
>               assert (diff / (tolerance * norm)) < 1  
E               assert (tensor(0.0585) / (0.05 * tensor(0.9880))) < 1  
```
Does this indicate a bug or a dependency issue? I can help contribute to a fix if suggested.

**To Reproduce**
The error can be reproduced using the following seed:
```
def test_torch_sigmoid_approx(workers):
       torch.manual_seed(7893584826795948387)
       .....
```

**Expected behavior**
The test should always pass

**Screenshots**
NA

**Desktop (please complete the following information):**
Environment:
```
Ubuntu 16.04
python 3.7
numpy==1.17.4
syft-tensorflow==0.1.0
tensorflow==1.15.0
torch==1.3.0
torchvision==0.4.1 
```

**Additional context**
Full error log:
```
Falling back to insecure randomness since the required custom op could not be found for the installed version of TensorFlow. Fix this by compiling custom ops. Missing file was 'python3.7/site-packages/tf_encrypted/operations/secure_random/secure_random_module_tf_1.15.0.so'
WARNING:tensorflow:From python3.7/site-packages/tf_encrypted/session.py:24: The name tf.Session is deprecated. Please use tf.compat.v1.Session instead.

=============================================================================================== test session starts ===============================================================================================
platform linux -- Python 3.7.6, pytest-5.3.2, py-1.8.1, pluggy-0.13.1 -- bin/python
cachedir: .pytest_cache
rootdir: PySyft, inifile: setup.cfg
collecting ... collected 1 item

test/torch/tensors/test_precision.py::test_torch_sigmoid_approx FAILED

==================================================================================================== FAILURES =====================================================================================================
____________________________________________________________________________________________ test_torch_sigmoid_approx ____________________________________________________________________________________________

workers = {'alice': <VirtualWorker id:alice #objects:580>, 'bob': <VirtualWorker id:bob #objects:544>, 'charlie': <VirtualWorker id:charlie #objects:0>, 'james': <VirtualWorker id:james #objects:81>, ...}

    @assert_time(max_time=40)
    def test_torch_sigmoid_approx(workers):
        """
        Test the approximate sigmoid with different tolerance depending on
        the precision_fractional considered
        """
[...]

PR #2998, opened 2020-02-02
No need for requires_grad for input data

gmuraru:
Fixes issue #2622


karlhigley:
The test itself passes when run independently from the rest of the test suite, so I suspect there's some sort of coupling happening through shared state. I couldn't identify exactly what the problem is though. (My first guess was workers or the hook, but decoupling them didn't seem to help. Maybe RNG state?)

gmuraru:
There is an issue regarding the ```test_precision``` -- (here)[https://github.com/OpenMined/PySyft/issues/2937] -- basically it is what you said @karlhigley . For the moment we can bump the error threshold a little bit.

gmuraru:
Increased the error threshold such that this should also fix #2937

Issue #2986, opened 2020-01-31
Send arguments to crypten workers for computation


LaRiffle:
We'd like to sent tagged tensor to syft workers
and then send the tag to the crypten parties, so that they can retrieve the data by asking the syft worker. This way we avoid using file systems.

PR #3003, opened 2020-02-03
Add tag for sending tensors to crypten

gmuraru:
Added ```self``` such that we can reference the worker in the ```toy_func``` and use the tag to work with the specific data.

A possible fix for #2986

### DIFF
--- a/syft/frameworks/crypten/__init__.py
+++ b/syft/frameworks/crypten/__init__.py
@@ -1,4 +1,46 @@
+import torch
+import syft
 from syft.frameworks.crypten.context import toy_func, run_party
+import crypten.communicator as comm
+import crypten
 
 
-__all__ = ["toy_func", "run_party"]
+def load(tag: str, src: int, id_worker: int):
+    if comm.get().get_rank() == src:
+        worker = syft.local_worker.get_worker(id_worker)
+        result = worker.search(tag)[0].get()
+
+        # file contains torch.tensor
+        if torch.is_tensor(result):
+            # Broadcast load type
+            load_type = torch.tensor(0, dtype=torch.long)
+            comm.get().broadcast(load_type, src=src)
+
+            # Broadcast size to other parties.
+            dim = torch.tensor(result.dim(), dtype=torch.long)
+            size = torch.tensor(result.size(), dtype=torch.long)
+
+            comm.get().broadcast(dim, src=src)
+            comm.get().broadcast(size, src=src)
+            result = crypten.mpc.MPCTensor(result, src=src)
+    else:
+        # Receive load type from source party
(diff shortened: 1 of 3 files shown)


syft/frameworks/crypten/context.py
```diff
@@ -79,12 +82,30 @@ def _send_party_info(worker, rank, msg, return_values):
...
 
 
-def toy_func():
-    # Toy function to be called by each party
-    alice_t = crypten.cryptensor([73, 81], src=0)
-    bob_t = crypten.cryptensor([90, 100], src=1)
-    out = bob_t.get_plain_text()
-    return out.tolist()  # issues with putting torch tensors into queues
+def toy_func(*args, **kwargs):
+    # !! This might work only because we have virtual workers
```
youben11: We shouldn't perform workers operations on crypten functions I think. We should try at best to make those operations transparent by handling them via function decorators or our own crypten.load function

gmuraru: I will look at the ```crypten.load``` how it works and try something from there

syft/frameworks/crypten/context.py
```diff
@@ -79,12 +83,12 @@ def _send_party_info(worker, rank, msg, return_values):
     return_values[rank] = response.contents
 
 
-def toy_func():
-    # Toy function to be called by each party
-    alice_t = crypten.cryptensor([73, 81], src=0)
-    bob_t = crypten.cryptensor([90, 100], src=1)
-    out = bob_t.get_plain_text()
-    return out.tolist()  # issues with putting torch tensors into queues
+def toy_func(*args, **kwargs):
```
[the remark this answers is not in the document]
gmuraru: Done :)

syft/frameworks/crypten/__init__.py
```diff
@@ -1,4 +1,46 @@
...
+
+        # file contains torch.tensor
+        if torch.is_tensor(result):
+            # Broadcast load type
+            load_type = torch.tensor(0, dtype=torch.long)
+            comm.get().broadcast(load_type, src=src)
+
+            # Broadcast size to other parties.
+            dim = torch.tensor(result.dim(), dtype=torch.long)
+            size = torch.tensor(result.size(), dtype=torch.long)
```
LaRiffle: I'm curious why you had to provide the dim + the size: is this because the receving part:
```
# Receive size from source party
            dim = torch.empty(size=(), dtype=torch.long)
            comm.get().broadcast(dim, src=src)
```
should have a fixed size in advance?

youben11: That's because only the src party knows the dimensions and size of the tensor to be encrypted, and the shares should be created accordingly, this this information should be sent to all the other parties via brodcast.

PR #3012, opened 2020-02-05
trigger github action notebook checks

systemshift:
As @karlhigley pointed out to me, github actions only run 2 checks for tests. But does not run the other 2 checks when changes are applied to notebooks.


.github/workflows/run-translation-tests.yml
```diff
@@ -3,7 +3,9 @@ name: Run notebook translation tests
 on:
   push:
     paths:
-    - 'examples/tutorials/**'
+    - "examples/tutorials/**"
+  pull_request:
```
karlhigley: I think this `pull_request` section also needs to specify the paths.

systemshift: > Thanks for taking this on! Looks like it’s running the translation tests on this PR, which it shouldn’t. I think it just needs to specify the paths for both pushes and pull requests.

Yes, I removed the changes I made to the notebook and it still triggers the checks. Will try and change that.

systemshift: I will revert the last commit if everything is fine.

> Looks like it’s running the translation tests on this PR, which it shouldn’t.
`examples/tutorials/**` includes all files in tutorials. If you want only the notebook files then it should be `example/tutorials/*.ipynb`

karlhigley: I think we do want everything in that directory (recursively) to trigger the translation tests. Does `**` cover that? Do we need `**/*`?

karlhigley:
It looks like there's something off with the notebook tests action. It's running all the tests (including the notebooks/translations), but should only run the notebook tests.

systemshift:
> It looks like there's something off with the notebook tests action. It's running all the tests (including the notebooks/translations), but should only run the notebook tests.

running `pytest test test/notebooks/test_notebooks.py::test_notebooks_basic_translations` on my local machine also runs all tests. I don't think this has to do with CI, perhaps the actual test units we have call the main test suite in some way.

.github/workflows/run-translation-tests.yml
systemshift: no, `**` should work and include everything in `tutorials/`

PR #3013, opened 2020-02-05
Remove travis

systemshift:
From a discussion with @karlhigley , now that the project has migrated to github actions, travis still takes a long time to run. Leading to errors on checks even if everything passes, which prevents us from merging good pull requests.

PR #3018, opened 2020-02-06
update comment in test units

systemshift:
Once  #3013 is merged to master, we should test the new test unit that was ignored.


karlhigley:
This test still fails in the GH Action with the error:
```
FileNotFoundError: [Errno 2] No such file or directory: './data/inputs.npy'
```

Issue #2528, opened 2019-08-19
Update Automatically the Docker Hub using GitHub Actions


meandmymind:
**Original issue** 
Docker container https://hub.docker.com/r/openmined/pysyft-notebook is 2 months old. Some parts of the project has changed a lot (e.g. Plan), so tutorials don't work as expected.

**Updated**
Same error found in #2939

We should update the container automatically when we merge to master using github actions

Example of this is done here:
https://github.com/arkhn/fhir-pipe/blob/master/.github/workflows/main.yml

linamnt:
Hi I would like to work on this if no one else has started.

PR #3030, opened 2020-02-07
Automatically update Docker Hub using GitHub Actions (on push to master)

linamnt:
Automatically update Docker Hub image at [https://hub.docker.com/r/openmined/pysyft-notebook](https://hub.docker.com/r/openmined/pysyft-notebook) on push to Master (using GitHub Actions).

Addresses https://github.com/OpenMined/PySyft/issues/2528

Requires:
- [ ] DOCKER_LOGIN & DOCKER_PASSWORD need to be added by Project Maintainers as secrets in repo settings

Questions:
- not sure how to explicitly test that this works, so I'm not sure if there is a way and I'm missing that test...
- Python version set to '3.x', is this okay? 
    - In the test workflows, both 3.6 and 3.7 are specifically tested
    - in the publish to pypi, 3.x is used
- Did not include publishing to GitHub Packages as per [the example](https://github.com/arkhn/fhir-pipe/blob/master/.github/workflows/main.yml) listed in https://github.com/OpenMined/PySyft/issues/2528. Do you want this feature?

### DIFF
--- a/.github/workflows/update-docker-image.yml
+++ b/.github/workflows/update-docker-image.yml
@@ -1,0 +1,38 @@
+name: Update Docker Image
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  publish:
+    name: Build and publish
+    runs-on: ubuntu-latest
+
+    env:
+      IMAGE: openmined/pysyft-notebook
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: '3.x'
+
+      - name: Build image
+        run: docker build -t $IMAGE -f docker-images/pysyft-notebook/Dockerfile .
+
+      - name: Tag images
+        run: |
+          docker tag $IMAGE $IMAGE:latest
+          docker tag $IMAGE $IMAGE:$GITHUB_SHA
+
+      - name: Publish to docker hub
+        run: |
+          echo "${{ secrets.DOCKER_PASSWORD }}" |  docker login -u ${{ secrets.DOCKER_LOGIN }} --password-stdin
+          docker push $IMAGE:latest
+          docker push $IMAGE:$GITHUB_SHA
+
+


vvmnnnkv:
Another way is to connect docker hub with PySyft repo and setup build(s). It will rebuild images automatically on repo update. The build may be faster than in GH Actions because docker hub leverages build cache.
https://docs.docker.com/docker-hub/builds/

linamnt:
On a related note, GH Actions does support caching, which could eventually support Docker caching natively, but it's definitely not there yet in any elegant form, despite being able to maybe wrangle some working solution together with the available actions (see https://github.com/actions/cache/issues/31)

karlhigley:
@linamnt Thanks for your work on this! Finally got it merged, and it looks like there's an issue with the process of building the Docker image on GH. It's failing with the error:
```
Step 8/10 : COPY ./entrypoint.sh /
COPY failed: stat /var/lib/docker/tmp/docker-builder735825847/entrypoint.sh: no such file or directory
##[error]Process completed with exit code 1.
```
Any idea what that's about or how to address it?

linamnt:
Seems we need to pass the entrypoint.sh that goes along with the Dockerfile, but I think it needs to passed as it's not in the current working directory. I can work on it later today if it's not a rush.

vvmnnnkv:
Docker context is pysyft root while `entrypoint.sh` referenced in Dockerfile is located in `docker-images/pysyft-notebook/` folder. Build command can be updated to specify correct context path:
`docker build -t $IMAGE -f docker-images/pysyft-notebook/Dockerfile docker-images/pysyft-notebook`
Alternatively, Dockerfile can be updated to specify `entrypoint.sh` location from pysyft root.

karlhigley:
That sounds reasonable. Happy to merge a fix when you a get a chance to work on it. No rush.

linamnt:
I made another PR here: https://github.com/OpenMined/PySyft/pull/3066 with the solution by @vvmnnnkv

karlhigley:
@linamnt Thanks for the PR! Will merge once the tests pass.

Issue #2775, opened 2019-12-04
Translate our tutorials to Spanish


iamtrask:
At present, our tutorial notebook series is only available in English, but it would be very desirable for people to be able to learn PySyft in their native language, such as Spanish.

For this PR, copy the current notebooks and translate the inline text to Spanish, placing the notebooks in a "espanol" folder within the "tutorials" folder.

https://github.com/OpenMined/PySyft/tree/master/examples/tutorials

arturomf94:
I'll take this issue :)

iamtrask:
All yours!

iamtrask:
Heads up @arturomf94 - @ricardopretelt has translated Part 1!

arturomf94:
Thanks @iamtrask. Indeed, we've been coordinating in order to avoid double work :)

iamtrask:
Excellent!

Issue #2782, opened 2019-12-04
Translate our tutorials to your native language!


iamtrask:
At present, our tutorial notebook series is only available in English, but it would be very desirable for people to be able to learn PySyft in their native language

For this PR, copy the current notebooks and translate the inline text to a language you know, placing the notebooks in a <language-name> folder within the "tutorials" folder.

https://github.com/OpenMined/PySyft/tree/master/examples/tutorials

Rules of Thumb:
- Only translate English which is in the notebook or generated from code in the notebook. Any English which is coming from the PySyft library itself (such as error messages) should remain in English.

ricardopretelt:
I can start right now with the translation to spanish as it is my native languange, if anyone else wants to contribute can contact me

iamtrask:
Make sure to collaborate with these folks too - submit one tutorial at a time via pull reequesets to minimize the chance of double work. https://github.com/OpenMined/PySyft/issues/2775

ricardopretelt:
ok thanks, already coordinated with @arturomf94

wonderit:
@iamtrask 
Could you please open an issue for Korean also? As a fan of this project, I would like to participate in this issue.

ricardopretelt:
@wonderit I think that anyone can open the issue, just for convenience use the same format as the others and the labels translation, good first issue and feature

fdroessler:
In order to avoid cluttering of the "root" tutorials directory my suggestion would be to have all translations in a`examples/tutorials/translations/` subdirectory. I have done this for some of the notebook tests already and suggested it in the PRs for Korean, Espanol and Romanian. If this is not opposed by the maintainers I would make this a more general suggestion?

iamtrask:
Sounds great!

ricardopretelt:
I have the 3rd tutorial translated to spanish, but we need the previous parts merged with the new updates to do a new PR avoiding version conflicts

adilnaut:
I will translate to Russian, dm me if you want to collaborate!

Metrix1010:
I can translate into French although not my native language (I'd be using nbTranslate).
DM me to collaborate.

I need some clarification for this task.

Are we:
1. Creating a hard copy of the tutorials folder translated into a specified language?
2. Updating the main installation file to require each user to install nbTranslate to self select language?
3. Both?

Thanks :)

KyloRen1:
[...]

PR #3031, opened 2020-02-07
Translations to Spanish of Tutorials Part 10 and 11

arturomf94:
This comes as a response to the initiative to translate all the
PySyft tutorials, as it is desirable to learn PySyft in our native
language.

As requested, every part of the tutorial is done via pull request
separately to:

 - Track better the changes in every document.

 - Avoid double work.

Resolves: #2775
See also: #2782


arturomf94:
There is a problem with the tutorial part 10 due to https://github.com/OpenMined/PySyft/pull/2990. 

It can be noted in [`test_notebooks.py`](https://github.com/OpenMined/PySyft/blob/master/test/notebooks/test_notebooks.py) that this notebook is excluded from the tests on the original English version.

PR #3033, opened 2020-02-08
Cleanup PATE code

TTitcombe:
This PR adds `moment` and `noise_eps` values to the print statement when sensitivity cannot be calculated. Currently, the console will be spammed with an identical statement if you have lots of incompatible noise/moment values; at least now the statements demonstrate that different calculations are taking place.

This PR also:
* Adds type hints
* Formats docstrings
* Remove commented-out code and print statements

I've added a notice to the top of the file to make clear that changes have been made to the code, as required by Apache 2 license.


syft/frameworks/torch/dp/pate.py
```diff
@@ -241,9 +254,6 @@ def perform_analysis(teacher_preds, indices, noise_eps, delta=1e-5, moments=8, b
 
     eps_list_nm = (total_log_mgf_nm - math.log(delta)) / l_list
 
-    # print("Epsilons (Noisy Max): " + str(eps_list_nm))
-    # print("Smoothed sensitivities (Noisy Max): " + str(total_ss_nm / l_list))
```
LaRiffle: do you want to rm this?

syft/frameworks/torch/dp/pate.py
```diff
@@ -254,10 +264,7 @@ def perform_analysis(teacher_preds, indices, noise_eps, delta=1e-5, moments=8, b
 
     ss_eps = 2.0 * beta * math.log(1 / delta)
     ss_scale = 2.0 / ss_eps
-    # print("To get an " + str(ss_eps) + "-DP estimate of epsilon, ")
-    # print("..add noise ~ " + str(ss_scale))
-    # print("... times " + str(total_ss_nm / l_list))
-    # print("Epsilon = " + str(min(eps_list_nm)) + ".")
```
LaRiffle: same for this?

syft/frameworks/torch/dp/pate.py
TTitcombe: I think so. If i've understood it correctly, the original tensorflow code (https://github.com/tensorflow/privacy/blob/master/research/pate_2017/analysis.py#L272) used this before adding noise to estimate epsilon, which is why it prints "to get this epsilon, use this noise". This code is formulated slightly differently as we supply the noisy counts to the function. So even if the code wasn't commented-out it wouldn't make complete sense, as we've already added the noise

PR #3037, opened 2020-02-09
Fix potential medium/high severity security holes

systemshift:
In order to # #3036 to merge on master, some minor edits need to be made and remove the few warnings that show up.


karlhigley:
It looks like `literal_eval` is causing a bunch of test failures and hanging the build. 😢

karlhigley:
If that turns out to be difficult to fix, I think you can label the line with `#nosec` to exclude it from the scan.

karlhigley:
I think the reason `literal_eval` doesn't work here is that `Plan` is actually relying on `eval` to execute commands, not just convert strings to literal data structures. The commands that get evaluated by the `Plan` tests are:

- 'torch.nn.functional.linear'
- 'torch.nn.functional.relu'
- 'torch.nn.functional.log_softmax'

I made a [change](https://github.com/OpenMined/PySyft/pull/2951/files) recently that addressed another `eval`, and maybe we could take a similar approach here.

karlhigley:
Since commands could also be executed on other frameworks (i.e. not PyTorch) though, maybe the best thing to do for now is just label this line so the security checks ignore it.

PR #3048, opened 2020-02-11
Fix broken asynchronous federated learning tutorial

midokura-silvia:
The methods federated_client.evalute() and federated_client.async_fit() are now called via the new Baseworker.execute_worker_function() function. Before it relied on Baseworker.execute_command() whose behaviour changed and is meant to be used for Tensor functions only. 

The pullrequest needs a modification in syft-proto.


syft/workers/base.py
```diff
@@ -467,6 +467,28 @@ def execute_command(self, message: tuple) -> PointerTensor:
...
+                """
+        command_name, (args, kwargs, return_ids) = message
+
+        command_name = command_name
+        response = getattr(self, command_name)(*args, **kwargs)
+        #  TODO [midokura-silvia]: send the tensor directly
+        #  TODO this code is currently necessary for the async_fit method in websocket_client.py
+        import torch
+
+        if isinstance(response, torch.Tensor):
```
LaRiffle: You can put isinstance syft.FrameworkTensor I think here and not have to import torch

syft/workers/base.py
```diff
@@ -1004,9 +1024,7 @@ def message_pending_time(self, seconds: Union[int, float]) -> None:
         self._message_pending_time = seconds
 
     @staticmethod
-    def create_message_execute_command(
-        command_name: str, command_owner=None, return_ids=None, *args, **kwargs
-    ):
+    def create_message_execute_command(command_name: str, return_ids=None, *args, **kwargs):
```
LaRiffle: This method name is a bit confusing, it's very generic, while we are performing a specific ExecuteWorkerFunctionMessage order.
Maybe it could be renamed `create_message_execute_worker_function` or smthg like this?

PR #3056, opened 2020-02-15
Italian Translation for Tutorial 01

DanyEle:
Added basic Italian translation for Tutorial 01 with some text from Andrea Provino's blog.


DanyEle:
@karlhigley Any idea why so many checks are failing? All I did was adding one more file without modifying any existing files.

karlhigley:
@DanyEle The test that's failing is `test_github_workflows_exist_for_all_languages`, which checks that each language has a GH workflow that runs the translated notebooks for that language when they change. You can find an example workflow file [here](https://github.com/OpenMined/PySyft/blob/master/.github/workflows/tutorials-espa%C3%B1ol.yml).

karlhigley:
@DanyEle Testing out a change to the CI jobs that will bypass the workflow file requirement.

DanyEle:
Already looks better. Now only the tutorials test for build 3.7 are failing with the following error:

Test with pytest:
```
##[error]Process completed with exit code 128.
Run git diff --name-only master examples/tutorials/translations/ > ./test/notebooks/git-diff.txt
fatal: ambiguous argument 'master': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
##[error]Process completed with exit code 128.
```

karlhigley:
@arturomf94 is working on this. We'll get it merged, one way or another.

DanyEle:
Looks alright now! All checks passing.

arturomf94:
@DanyEle looks like it passed, but for the wrong reasons. What were trying to do is to test only the notebooks that have been changed during the build. You can have a look at the changes here https://github.com/OpenMined/PySyft/pull/3085 hopefully everything will run ok with that.

karlhigley:
@DanyEle We finally got the translation tests running correctly, so I'll merge this when the tests pass.

PR #3057, opened 2020-02-16
Function Secret Sharing

LaRiffle:
### About
This PR proposes an implementation and some refinements of Function Secret Sharing.

Function Secret Sharing (FSS) (introduced with great details [here](https://eprint.iacr.org/2018/707.pdf) and [here](https://eprint.iacr.org/2019/1095) is a technique which allows to compute over secret shared data. Details of this cool method will come in a future blog post.

Most interestingly, this is the first internal use of Plans to improve efficiency of crypto protocols by reducing the number of interactions, which is made necessary because the great benefit of FSS over other MPC protocols like SPDZ is the small number of communication rounds.


karlhigley:
@LaRiffle If the Worker, Hook, and Plan parts of this PR are complete (or nearly so), could we split them out into a separate PR that we can go ahead and merge? Want to be working from the latest/greatest in those parts of the code as we make additional improvements.

LaRiffle:
@karlhigley yes I agree! But I’m not sure how  to do it haha

karlhigley:
@LaRiffle Your commits are pretty granular, so you can probably:
* Create a second branch pointing at the same commit
* Do an interactive rebase
* Remove the commits related to FSS and the experimental notebooks

Whatever is left would be the general improvements to the hooks, plans, protocols, etc.

LaRiffle:
Ok I've 🍒 picked the most of them, the PR is here -> https://github.com/OpenMined/PySyft/pull/3080

syft/messaging/message.py
```diff
@@ -133,6 +133,7 @@ def __init__(self, name, target, args_, kwargs_, return_ids):
         super().__init__()
 
         self.action = ComputationAction(name, target, args_, kwargs_, return_ids)
+        self.return_value = return_value
```
karlhigley: What's the type of `return_value` here and what is it for? Should be added to the docstring.

syft/workers/base.py
```diff
@@ -462,7 +471,11 @@ def execute_command(self, message: tuple) -> PointerTensor:
                 response = hook_args.register_response(
                     command_name, response, list(return_ids), self
                 )
-                return response
+                if return_value or isinstance(response, (int, float, bool, str)):
```
karlhigley: Does this mean I can set `return_value` to `False` and still get a response? That seems surprising.

syft/workers/base.py
```diff
@@ -511,7 +524,11 @@ def execute_plan_command(self, message: tuple):
         return command(*args)
 
     def send_command(
-        self, recipient: "BaseWorker", message: tuple, return_ids: str = None
+        self,
+        recipient: "BaseWorker",
+        message: tuple,
+        return_ids: str = None,
+        return_value: bool = False,
```
karlhigley: How does the sender decide whether or not to request a return value?

syft/workers/base.py
```diff
@@ -179,6 +179,11 @@ def __init__(
                 self.tensorflow = self.framework
                 self.remote = Remote(self, "tensorflow")
 
+        # storage object for crypto primitives
+        self.crypto_store = PrimitiveStorage(owner=self)
```
karlhigley: 👍 

We should make `ObjectStorage` work like this too (eventually.)

test/torch/mpc/test_fss.py
```diff
@@ -1,58 +1,64 @@
...
 import syft
 import torch as th
 
 from syft.frameworks.torch.mpc.fss import DPF, DIF, n
 
 
-def test_DPF():
+@pytest.mark.parametrize("op", ["eq", "le"])
+def test_fss_klass(op):
+    klass = {"eq": DPF, "le": DIF}[op]
```
karlhigley: The PEP8 standard for avoiding name collisions with reserved words like `class` to append an underscore like `class_`.

syft/workers/base.py
LaRiffle: @karlhigley This is a way to do a combined remote op + get in a single communication round
Classical workflow in PySyft requires two separate calls, and because in work on low interaction crypto protocols I suggested this optimization.
However, this should only be valid when .get() is valid ie it shouldn't provide extra rights
I have probably left a security issue here... But this idea is not supposed to allow people to do weird stuff!

syft/frameworks/torch/mpc/__init__.py
```diff
@@ -0,0 +1,15 @@
+protocol_store = {}
+
+
+def crypto_protocol(protocol_name):
```
Jasopaum, line 4: Could you add some docstring to say how and when it's used?

syft/generic/pointers/pointer_tensor.py
```diff
@@ -305,8 +299,7 @@ def remote_send(
         message = TensorCommandMessage.communication(
             self.id_at_location, self.location.id, [destination.id], kwargs
         )
-        self.owner.send_msg(message=message, location=self.location)
-        return self
+        return self.owner.send_msg(message=message, location=self.location)
```
Jasopaum, line 302: Why are we changing that?

syft/generic/pointers/pointer_tensor.py
```diff
@@ -281,12 +281,6 @@ def move(self, destination: AbstractWorker, requires_grad: bool = False):
 
         ptr = self.remote_send(destination, requires_grad=requires_grad)
 
-        # We make the pointer point at the remote value. As the id doesn't change,
-        # we don't update ptr.id_at_location. See issue #3217 about this.
-        # Note that you have now 2 pointers on different locations pointing to the
-        # same tensor.
-        ptr.location = destination
-
         return ptr
```
Jasopaum, line 284: Maybe we already answered this question but what is the difference between remote_send and move now?
Because I see written
```
def move():
     return self.remote_send()
```

test/torch/pointers/test_pointer_tensor.py
```diff
@@ -323,10 +323,11 @@ def test_move(workers):
 
     alice.clear_objects()
     bob.clear_objects()
-    x = torch.tensor([1.0, 2, 3, 4, 5]).send(bob)
+    t = torch.tensor([1.0, 2, 3, 4, 5])
+    x = t.send(bob)
```
Jasopaum, line 327: Can't we do `torch.tensor([1.0, 2, 3, 4, 5]).send(bob)` anymore?

syft/frameworks/torch/mpc/primitives.py
```diff
@@ -0,0 +1,160 @@
...
+
+        Args:
+            type_op: eq, comp, or xor_add
+            n_instances: how many primitives to retrieve. Comparison is pointwise so this in
+                convenient
+            remove: if true, pop out the primitive. If false, only read it. Read mode is
+                needed because we're working on virtual workers and they need to gather
+                a some point and then re-access the keys.
+        """
+        type_op = {"eq": "fss_eq", "comp": "fss_comp", "xor_add": "xor_add_couple"}[type_op]
```
Jasopaum, line 43: Why is this translation needed? Couldn't we use directly "fss_eq" everywhere?

syft/frameworks/torch/mpc/primitives.py
```diff
@@ -0,0 +1,160 @@
...
+                # [:1] ~ [slice(1)]
+                # [1:] ~ [slice(1, None)]
+                # [:, :, :1] ~ [slice(None)] * 2 + [slice(1)]
+                n_dim = len(prim.shape)
+                get_slice = [slice(None)] * (n_dim - 1) + [slice(n_instances)]
+                remaining_slice = [slice(None)] * (n_dim - 1) + [slice(n_instances, None)]
+
+                keys.append(prim[get_slice])
+                if remove:
+                    primitive_stack[i] = prim[remaining_slice]
```
Jasopaum, line 61: Could you specify the form of the primitive_stacks somewhere in the docstring?
Because I don't know if it could be simpler or if I just miss something ^^

syft/frameworks/torch/mpc/primitives.py
```diff
@@ -0,0 +1,160 @@
...
+        """
+        Include primitives in the store
+
+        Args:
+            types_primitives: dict {crypto_type: str: primitives: list}
+        """
+        for crypto_type, primitives in types_primitives.items():
+            assert hasattr(self, crypto_type), f"Unknown crypto primitives {crypto_type}"
+
+            current_primitives = getattr(self, crypto_type)
```
Jasopaum, line 114: What about having a dict attribute instead of using getattr?

test/torch/mpc/test_crypto_store.py
```diff
@@ -0,0 +1,21 @@
...
+
+def test_primitives_usage(workers):
+    me, alice, bob, crypto_provider = (
+        workers["me"],
+        workers["alice"],
+        workers["bob"],
+        workers["james"],
+    )
+    me.crypto_store.provide_primitives(["fss_eq"], [alice, bob], n_instances=6)
+    _ = alice.crypto_store.get_keys("fss_eq", 2, remove=False)
```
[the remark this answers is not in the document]
LaRiffle: Done

syft/generic/object_storage.py
```diff
@@ -21,6 +22,10 @@ def __init__(self):
         # This is an index to retrieve objects from their tags in an efficient way
         self._tag_to_object_ids = defaultdict(set)
 
+    @property
+    def _tensors(self):
+        return dict(filter(lambda obj: isinstance(obj[1], FrameworkTensor), self._objects.items()))
```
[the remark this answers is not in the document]
LaRiffle: I agree!

syft/generic/pointers/pointer_tensor.py
LaRiffle: Because remote_get is a remote operation which shouldn't change the object you're working on; it's still the same pointer

LaRiffle: Oh no we're doing the opposite!

LaRiffle: Will fix this

syft/generic/pointers/pointer_tensor.py
LaRiffle: There isn't much difference actually, except the one you just mention: it's the same remote operation but you get as an answer not the same pointer

test/torch/pointers/test_pointer_tensor.py
LaRiffle: You can but I need to keep a ref to `t` :)

syft/frameworks/torch/mpc/primitives.py
LaRiffle: I've added it :)

syft/frameworks/torch/mpc/primitives.py
LaRiffle: Not sure it makes a big difference right?

test/torch/pointers/test_pointer_tensor.py
Jasopaum: Oops, indeed 👍

syft/frameworks/torch/mpc/primitives.py
Jasopaum: Not at all, it's just that I don't like `getattr` very much ^^
But after having read it again, I'm fine with it

syft/frameworks/torch/mpc/primitives.py
```diff
@@ -0,0 +1,169 @@
...
+            for i, prim in enumerate(primitive_stack):
+                # We're selecting on the last dimension of the tensor because it's simpler for
+                # generating those primitives in crypto protocols
+                # [:] ~ [slice(None)]
+                # [:1] ~ [slice(1)]
+                # [1:] ~ [slice(1, None)]
+                # [:, :, :1] ~ [slice(None)] * 2 + [slice(1)]
+                n_dim = len(prim.shape)
+                get_slice = [slice(None)] * (n_dim - 1) + [slice(n_instances)]
+                remaining_slice = [slice(None)] * (n_dim - 1) + [slice(n_instances, None)]
```
Jasopaum, line 66: The `torch.narrow()` method didn't do the job?

syft/frameworks/torch/mpc/primitives.py
```diff
@@ -0,0 +1,169 @@
...
+            type_op: fss_eq, fss_comp, or xor_add_couple
+            n_instances: how many primitives to retrieve. Comparison is pointwise so this in
+                convenient
+            remove: if true, pop out the primitive. If false, only read it. Read mode is
+                needed because we're working on virtual workers and they need to gather
+                a some point and then re-access the keys.
+        """
+        primitive_stack = getattr(self, type_op)
+
+        available_instances = len(primitive_stack[0]) if len(primitive_stack) > 0 else -1
```
Jasopaum, line 52: Would len(primitive_stack) be always the same for each type? (For instance always 3 for beaver)
In this case, we could init self.beaver to ([], [], []), what do you think?

syft/frameworks/torch/mpc/primitives.py
```diff
@@ -0,0 +1,169 @@
...
+
+        Args:
+            crypto_types: type of primitive (fss_eq, etc)
+            workers: recipients for those primitive
+            n_instances: how many of them are needed
+            **kwargs: any parameters needs for the primitive builder
+
+        Returns:
+
+        """
```
Jasopaum, line 94: Not sure I followed everything but this method never uses already created primitives, does it?
Do you think it could call `get_keys()` and `get_keys()` would return or create keys if there are not enough of them?

syft/generic/utils.py
```diff
@@ -0,0 +1,10 @@
+class memorize(dict):
```
Jasopaum, line 1: Could you add a bit of doc here?

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -368,10 +369,9 @@ def handle_func_command(cls, command):
 
         return response
 
-    def _get_response(cmd, args, kwargs):
-        """
-            Return the evaluation of the cmd string parameter
-        """
+    @staticmethod
+    @memorize
```
Jasopaum, line 373: What is the reason for this to have changed?

syft/frameworks/torch/mpc/primitives.py
```diff
@@ -0,0 +1,169 @@
...
+            "beaver": self.build_triples,
+            "xor_add_couple": self.build_xor_add_couple,
+        }
+
+    def get_keys(self, type_op, n_instances=1, remove=True):
+        """
+        Return FSS keys primitives #TODO
+
+        Args:
+            type_op: fss_eq, fss_comp, or xor_add_couple
```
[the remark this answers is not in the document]
LaRiffle: Will come in a later PR :)
First i'd like to have automatic generation of the preprocessing, and much more (including my FSS numpy PR which is almost pushed)

syft/frameworks/torch/mpc/primitives.py
```diff
@@ -0,0 +1,169 @@
...
+            n_instances: how many primitives to retrieve. Comparison is pointwise so this in
```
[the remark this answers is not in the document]
LaRiffle: I've added details here!

syft/frameworks/torch/mpc/primitives.py
LaRiffle: <3 So much better! <3

syft/frameworks/torch/mpc/primitives.py
LaRiffle: It might vary depending on the implementation. For example depending of the bit precision, (now 32), the number of elements in the FSS keys would change with the current implem

syft/frameworks/torch/mpc/primitives.py
LaRiffle: This method is here to feed the different crypto stores, so shat get-keys will work.
It's here to decorrelate preprocessing and online phase: provide_primitives should be called way before the computation really happens, to have a fats online phase

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: Just a small optimization nothing essential here

syft/frameworks/torch/mpc/fss.py
```diff
@@ -0,0 +1,401 @@
...
+        forward_func=xor_add_convert_1, owner=worker, tags=["#xor_add_1"], is_built=True
+    )
+    worker.register_obj(xor_add_plan)
+    xor_add_plan = sy.Plan(
+        forward_func=xor_add_convert_2, owner=worker, tags=["#xor_add_2"], is_built=True
+    )
+    worker.register_obj(xor_add_plan)
+
+
+def request_run_plan(worker, plan_tag, location, return_value, args=tuple(), kwargs=dict()):
```
Jasopaum, line 65: Random thought here.
Isn't it possible to use pointer to plans to request a run?

LaRiffle: Yes you can, but you need an extra round of communication to get your pointer

PR #3059, opened 2020-02-17
Split tutorial translation tests into separate Github Workflows

karlhigley:
The translation tests are taking too long to run which is blocking the translation PRs. This PR breaks the translation tests into separate Github Workflow Actions, so that they can be run only when notebooks in the relevant language are changed.


arturomf94:
This looks reasonable. I was working on a a potential solution on [this branch](https://github.com/arturomf94/PySyft/tree/fix-translation-tests). Only the changed translations would be tested, but don't know whether this would work yet. What do you think?

karlhigley:
@arturomf94 Something like that could also work. It would even be a little smoother to start translations in a new language, because you wouldn't have to create a new GH Workflow file. One downside is that it looks like the translation tests wouldn't run locally.

arturomf94:
@karlhigley that's right. Perhaps there can be a local test and a remote test (this one). So, the remote one wouldn't really be testing any notebooks locally because no diff?

PR #3060, opened 2020-02-18
Only test translation notebooks if they have been changed

arturomf94:
Solves the same problem as https://github.com/OpenMined/PySyft/pull/3059 but only tests the notebooks if they have actually been changed.


karlhigley:
The test coverage checks are failing because lines 41-43 of test_notebooks.py aren’t run, which I think is because there is no git diff file.

arturomf94:
Weird. I've included an empty file, just to check if it was indeed that.

PR #3067, opened 2020-02-19
Remove decode instruction from tags/description

IonesioJunior:
Tags and descriptions are no longer received as binary values during the `pointer_tensor.detail` method.
Therefore, we can remove the decoding instructions.
These changes solve  the [PyGrid issue](https://github.com/OpenMined/PyGrid/issues/473)


hdodenhof:
To add a little context: this is caused by some [breaking changes](https://github.com/msgpack/msgpack-python#major-breaking-changes-in-msgpack-10) in `msgpack` 1.0 which was released yesterday.

IonesioJunior:
Thank you @hdodenhof ! ❤️

PR #3069, opened 2020-02-19
(TESTING FROM FORK) Update GH checkout action to v2

karlhigley:


karlhigley:
Seems to work both on PRs from branches on OpenMined/PySyft directly and on branches from forks. 👍

Issue #3007, opened 2020-02-04
Add latency simulation to Virtual workers


LaRiffle:
**Describe the solution you'd like**
It would be great to add an option to simulate latency on virtual workers while sending message.
Basically just add a small sleep in the `_recv_msg` or `_send_msg` virtual worker methods.

jefersonf:
What do you mean by latency simulation? It would be an new option to a _VirtualWorker_ like `add_latency=100`. Like the following example.
```
...
import time

class VirtualWorker(BaseWorker, FederatedClient):
    def _send_msg(self, message: bin, location: BaseWorker, latency: int = 0) -> bin:
        time.sleep(latency)
        return location._recv_msg(message)
```

Vishal-Ramesh2:
I'd like to implement this

BBloggsbott:
I'd like to implement this

jefersonf:
Hi @BBloggsbott, I tried to tackle this. And I've actually started on, if you would like to give some help please take a look at this [branch](https://github.com/jefersonf/PySyft/tree/latency_simulation/syft/).

In fact, the following case is working so far:

```
import torch
import syft as sy
hook = sy.TorchHook(torch)

alice = sy.VirtualWorker(id="alice", hook=hook)
bob  = sy.VirtualWorker(id="bob", hook=hook)
x = torch.tensor([1,2,3])

# alice waits 2 seconds before sending 'x' to bob.
alice.send(x, bob, pending_time=2)

```

BBloggsbott:
@jefersonf Why not just use the change you mentioned [here](https://github.com/OpenMined/PySyft/issues/3007#issuecomment-582028097)? I feel only the virtual workers will need to simulate latency.

jefersonf:
> @jefersonf Why not just use the change you mentioned [here](https://github.com/OpenMined/PySyft/issues/3007#issuecomment-582028097)? I feel only the virtual workers will need to simulate latency.

At first I thought so too, but I think in cases where a tensor is sent to a worker like in the following:

```
torch.tensor([1,2,3]).send(alice, pending_time=2)
```
the same behavior should be hold.

Vishal-Ramesh2:
Ah. Okay.

How can I help with this?

BBloggsbott:
Ah. Okay.

How can I help with this?

PR #3070, opened 2020-02-19
Add message pending time option to virtual workers

jefersonf:
As mentioned in #3007. It would be useful simulate latency while sending messages between workers. So I add an option to set a `message_pending_time` in the creation of a VirtualWorker.

You can see an example of usage below or in the [_test_send_msg_](https://github.com/jefersonf/PySyft/blob/msg_pending_time/test/workers/test_virtual.py#L18-L47).

```
...
alice = syft.VirtualWorker(
    hook=torchHook, id="alice", message_pending_time=1.5, verbose=True
)
bob = syft.VirtualWorker(hook=torchHook, id="bob")

obj = torch.Tensor([1, 2, 3])

# In this example, 2 messages will be sent and each one will take 1.5 seconds
alice.send(obj, bob)
```

Logs Output:

```
worker <VirtualWorker id:alice #objects:0> sending (ObjectMessage tensor([1, 2, 3])) to <VirtualWorker id:bob #objects:5>
pending time of 1.5 seconds to send message...
worker <VirtualWorker id:alice #objects:0> sending (ForceObjectDeleteMessage 4444064520) to <VirtualWorker id:bob #objects:6>
pending time of 1.5 seconds to send message...
[PointerTensor | alice:9312967354 -> bob:36492502540]
```
<br/>
Resolves: #3007


syft/workers/virtual.py
```diff
@@ -1,9 +1,17 @@
+from time import sleep
+from typing import Union
```
karlhigley: Is there actually a Union type in this file?

jefersonf: Fixed!

PR #3074, opened 2020-02-20
Syft load based ONLY on rank

### LINKED ISSUE #3043: Create a syft.load(tensor, src=worker_id) functionality
This should be equivalent to worker.search, but is supposed to be retrieving information _locally_, when src matches the current worker. (behaviour is supposed to be close to crypten.load)
As is, this function is supposed not to fail and to always return something  (even random) when using in a plan at building.
Details on implementation are not all solved yet.

gmuraru:
Fixes #3043 
Send the party information in the init message such that we need to keep track only of the worker id.

### DIFF
--- a/syft/frameworks/crypten/__init__.py
+++ b/syft/frameworks/crypten/__init__.py
@@ -1,17 +1,28 @@
 import torch
 import syft
+
 from syft.frameworks.crypten.context import toy_func, run_party
+from syft.exceptions import WorkerNotFoundException
+
 import crypten.communicator as comm
 import crypten
 
 
-def load(tag: str, src: int, id_worker: int):
-    if comm.get().get_rank() == src:
-        worker = syft.local_worker.get_worker(id_worker)
-        result = worker.search(tag)[0].get()
+def load(tag: str, id_worker: int):
 
-        # file contains torch.tensor
+    src = syft.local_worker.get_rank_from_id(id_worker)
+
+    try:
+        worker = syft.local_worker.get_worker(id_worker, fail_hard=True)
+
+        # It should be a tensor and not a pointer
+        results = worker.search(tag)
+        assert len(results) == 1
+
+        result = results[0]
+
         if torch.is_tensor(result):
+
             # Broadcast load type
             load_type = torch.tensor(0, dtype=torch.long)
             comm.get().broadcast(load_type, src=src)
@@ -23,7 +34,10 @@ def load(tag: str, src: int, id_worker: int):
             comm.get().broadcast(dim, src=src)
             comm.get().broadcast(size, src=src)
             result = crypten.mpc.MPCTensor(result, src=src)
-    else:
+        else:
+            raise TypeError("Unrecognized load type on src")
+
+    except WorkerNotFoundException:
         # Receive load type from source party
         load_type = torch.tensor(-1, dtype=torch.long)
         comm.get().broadcast(load_type, src=src)
@@ -38,7 +52,6 @@ def load(tag: str, src: int, id_worker: int):
             result = crypten.mpc.MPCTensor(torch.empty(size=tuple(size.tolist())), src=src)
         else:
             raise TypeError("Unrecognized load type on src")
-    # TODO: Encrypt modules before returning them
 
     return result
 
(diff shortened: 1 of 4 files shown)


syft/workers/base.py
```diff
@@ -970,8 +1012,13 @@ def search(self, query: Union[List[Union[str, int]], str, int]) -> List[PointerT
                 # set garbage_collect_data to False because if we're searching
                 # for a tensor we don't own, then it's probably someone else's
                 # decision to decide when to delete the tensor.
-                ptr = obj.create_pointer(garbage_collect_data=False, owner=sy.local_worker).wrap()
-                results.append(ptr)
+                if obj.owner is not self:
```
gmuraru: If the worker is the owner of the object - then simply return the object

karlhigley: This plays a bit fast and loose with types (tensors vs pointers), which works fine in Python but becomes kind of a pain to encode in the Syft protocol and deal with in other languages. If this is a convenience change that isn't required for the main purpose of the PR, I'd probably rather leave part alone.

gmuraru:
This is a PR with the changes + what was added on the Master Branch (to keep it up to date).
The only effective changes were made in:
```
syft/frameworks/crypten/__init__.py
syft/frameworks/crypten/context.py
syft/workers/base.py
```

karlhigley:
@gmuraru This is now triggering a bunch of translation tests, since it (technically) makes changes to a bunch of the translation paths. You might want to locally merge `master` into `pytorch/crypten` and push that directly to `pytorch/crypten` first, then rebase this PR on top.

syft/workers/base.py
gmuraru: This PR does not depend necessary on this (I can call ```get``` in the ```load function```).
I think the problem was more "semantics" wise - like if you have the tensor  - why you should get a pointer to that tensor and not the effective tensor?

karlhigley: To make the return type consistent. If sometimes I get one thing and sometimes I get another thing, then I have to check what was returned in order to know how to process it, which proliferates `if` statements through the code base.

gmuraru: Yep, will update this.

syft/frameworks/crypten/__init__.py
```diff
@@ -1,17 +1,28 @@
...
-def load(tag: str, src: int, id_worker: int):
-    if comm.get().get_rank() == src:
-        worker = syft.local_worker.get_worker(id_worker)
-        result = worker.search(tag)[0].get()
+def load(tag: str, src: int):
+    worker_id = syft.local_worker.get_worker_id_from_rank(src)
+    assert worker_id != -1
+
+    try:
+        worker = syft.local_worker.get_worker(worker_id, fail_hard=True)
```
youben11: we should keep crypten's paradigm of checking `src` value for differentiating the operations to be done, some issues may if the `get_worker` return a worker for a non-expected `worker_id`

syft/frameworks/crypten/context.py
```diff
@@ -133,8 +139,8 @@ def wrapper(*args, **kwargs):
             # Send messages to other workers so they start their parties
             threads = []
             for i in range(len(workers)):
-                rank = i + 1
-                msg = CryptenInit((rank, world_size, master_addr, master_port))
+                rank = i+1
```
youben11: spaces

gmuraru: Done

syft/frameworks/crypten/__init__.py
gmuraru: I modified to use the rank :)

syft/frameworks/crypten/__init__.py
```diff
@@ -1,17 +1,26 @@
...
 
 
-def load(tag: str, src: int, id_worker: int):
-    if comm.get().get_rank() == src:
-        worker = syft.local_worker.get_worker(id_worker)
-        result = worker.search(tag)[0].get()
+def load(tag: str, src: int):
+    worker = syft.local_worker.get_worker_from_rank(src)
+
+    if worker and src == worker.rank:
```
youben11: Can't we simply check either `src == comm.get().get_rank()` ?

gmuraru: Yep...you are right. I overcomplicated this :(

gmuraru:
will close this one since we will use tags for retrieving data.

PR #3075, opened 2020-02-20
Split the (English) tutorial tests into a separate Github Action

karlhigley:
The tutorial tests take about 15 minutes to run, which is approximately
equal to the running time of all the rest of the tests. Running them in
parallel should speed up the PR checks.

### DIFF
--- a/.github/workflows/tutorials.yml
+++ b/.github/workflows/tutorials.yml
@@ -4,15 +4,13 @@ on:
   push:
     branches:
     - master
-    paths:
-    - "examples/tutorials/**"
-    - "!examples/tutorials/translations/**"
+    paths-ignore:
+    - "examples/tutorials/translations/**"
 
   pull_request:
     types: [opened, synchronize, reopened]
-    paths:
-    - "examples/tutorials/**"
-    - "!examples/tutorials/translations/**"
+    paths-ignore:
+    - "examples/tutorials/translations/**"
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


.github/workflows/tutorials.yml
```diff
@@ -0,0 +1,45 @@
+name: Tutorials
+
+on:
+  push:
+    branches:
+    - master
+    paths:
+    - "examples/tutorials/**"
+    - "!examples/tutorials/translations/**"
```
karlhigley: This excludes changes to the translations from the diff paths that trigger the English tutorial tests.

.github/workflows/tutorials.yml
```diff
@@ -0,0 +1,45 @@
...
+        hash -r
+        pip3 install -r pip-dep/requirements.txt
+        pip3 install -r pip-dep/requirements_dev.txt
+        pip3 install -r pip-dep/requirements_udacity.txt
+        pip3 install -r pip-dep/requirements_notebooks.txt
+        pip3 install scikit-learn
+        python setup.py install
+    - name: Run the tutorials
+      run: |
+        pytest test/notebooks/test_notebooks.py -m "not translation"
```
karlhigley: `-m "not translation"` excludes all the tests that are marked as `translation` tests, so this runs all the notebook tests except the translation tests (which have their own action.)

setup.cfg
```diff
@@ -27,4 +27,5 @@ omit =
 	syft/frameworks/torch/nn/rnn.py
 	syft/workers/websocket_server.py
 	test/efficiency/*
+	test/notebooks/*
```
karlhigley: Since the notebook tests no longer run in the same action as the rest of the tests, they also need to be excluded from the test coverage checks.

PR #3076, opened 2020-02-20
[WIP] Crypten POC for plans

gmuraru:


test/crypten/test_context.py
```diff
@@ -13,12 +15,21 @@ def test_context(workers):
...
-    @run_multiworkers([alice, bob], master_addr="127.0.0.1")
+    @sy.func2plan()
+    def toy_func():
+        alice_tensor = crypten.load("crypten_data", 1, "alice")
+        bob_tensor = crypten.load("crypten_data", 2, "bob")
+
+        #crypt = crypten.cat([alice_tensor, bob_tensor], dim=0)
+        return (alice_tensor + bob_tensor).get_plain_text().tolist()
+
+    th.Tensor.get_plain_text = lambda x: th.tensor([1,2])
```
gmuraru: Need to rethink this

syft/frameworks/crypten/context.py
```diff
@@ -83,15 +85,7 @@ def _send_party_info(worker, rank, msg, return_values):
...
-def toy_func():
-    alice_tensor = syft_crypt.load("crypten_data", 1, "alice")
-    bob_tensor = syft_crypt.load("crypten_data", 2, "bob")
-
-    crypt = crypten.cat([alice_tensor, bob_tensor], dim=0)
-    return crypt.get_plain_text().tolist()
-
-
-def run_multiworkers(workers: list, master_addr: str, master_port: int = 15987):
+def run_multiworkers(plan, workers: list, master_addr: str, master_port: int = 15995):
```
youben11: Shouldn't the plan be transparent here? Like the user would only decorate his function and we build a plan out of his function

test/crypten/test_context.py
```diff
@@ -13,15 +15,24 @@ def test_context(workers):
...
 
-    @run_multiworkers([alice, bob], master_addr="127.0.0.1")
+    @sy.func2plan()
+    def toy_func():
+        # crypten.load is actually syft_load
+        alice_tensor = crypten.load("crypten_data", 1, "alice", size=(2, ))
+        bob_tensor = crypten.load("crypten_data", 2, "bob", size=(2, ))
+
+        result = alice_tensor + bob_tensor
+        result = result.get_plain_text()
```
youben11: The load function shouldn't fail and it would return a real `cryptensor`, right? Then it's not the fake `get_plain_text` that's called here.

syft/frameworks/crypten/context.py
gmuraru: Yep, it should. I added it like this to have a fast testing mechanism if the plans work,

test/crypten/test_context.py
gmuraru: Hmm, when the plan is building the  ```load``` should go on the Exception branch from here:
```
try:
            response = func(*args, **kwargs)
        except Exception as e:
            response = th.rand(kwargs["size"])
```
And it should be a dummy response (only to have the shape of the response) and the result should also be a ```        th.randn(self.shape)``` from ```syft/generic/tensor```

PR #3078, opened 2020-02-20
Separate execution types from messaging types

karlhigley:
Plans, Protocols, Operations, States, and so forth aren't messaging
primitives, since they can be created and used locally. They *can* be
sent in messages, but don't necessarily need to be. Drawing this
distinction lays some of the groundwork for sorting out next generation
Plans and Protocols.

Depends on OpenMined/syft-proto#44.


karlhigley:
This PR is just a packaging name change. The next step will be extracting the concept of an Operation (that gets executed on a tensor) from an OperationMessage (which gets sent around between workers.) That addresses some weirdness in the serialization and makes the execution primitives more independent from the messaging protocol.

karlhigley:
@LaRiffle More changes coming, so kinda had to propagate them. 😅

PR #3080, opened 2020-02-21
Improvements on syft Hook & Plans

LaRiffle:
### Workers

Create tag_to_object index for worker to have a faster retrieval using tags
Optimize the worker search function
Change workflow to request a remote search: don’t call worker.search directly but request_search

### Plans

Add tracer to torch.tensor to support tracing tensors declared inside a plan
Add support of comparison in plans
Add support of plans with no arguments
Add explicit error when Plan building fails
Support torch.dtype & torch.size in plans
Support multi-output operations in plans (like torch.split)
Fix tag_sort function of plans to sort placeholders using their tags

### PointerPlans

Update PointerPlan to better support execution on multipointertensor
Add tag serde to PointerPlan

### Other

Remove stale add_registration_to___init_ in torchhook
Build a dedicated hook method for Placeholder
Create a @crypto_protocol decorator to define two same operations but each using a specific protocol (like SecureNN, fss, etc)
Fix creation of AdditiveSharedTensor when shares are already provided
Fix placeholder instantiation
Remove excluded function to hook them again (like th.manual_seed, ones, range)


karlhigley:
@LaRiffle Looks like the test failures mostly have to do with serialization and tagging.

vvmnnnkv:
When input parameter is returned from plan, the output is just empty tuple.
Is that by design or :bug:?

Code example:
```
@sy.func2plan(args_shape=[(5, 5),])
def upd_test(tensor):
    tensor.add_(1.)
    return tensor

upd_test.forward = None
t = th.randn(5,5)
out = upd_test(t)
print(out)
```

syft/execution/plan.py
```diff
@@ -52,7 +52,16 @@ def __call__(self, plan_function):
...
-            plan.build(*args)
+            try:
+                plan.build(*args)
+            except TypeError as e:
+                print(
+                    "Automatic build using @func2plan failed!\nCheck that:\n"
+                    " - you have provided the correct number of shapes in args_shape\n"
+                    " - you have no simple numbers like int or float as args. If you do "
+                    "so, please consider using a tensor instead."
+                )
```
Jasopaum: Could you put this in an error message instead of simply printing it?

syft/frameworks/torch/tensors/interpreters/placeholder.py
```diff
@@ -31,10 +31,12 @@ def instantiate(self, tensor):
         Add a tensor as a child attribute. All operations on the placeholder will be also
         executed on this child tensor.
 
-        We remove wrappers is there are any.
+        We remove wrappers or Placeholders if is there are any.
         """
-        if hasattr(tensor, "child") and tensor.is_wrapper and tensor.child is not None:
+        if isinstance(tensor, PlaceHolder):
             self.child = tensor.child
+        elif hasattr(tensor, "child") and tensor.is_wrapper and tensor.child is not None:
```
Jasopaum: The if-statement looks a bit weird.
Can `tensor.is_wrapper` be True while `hasattr(tensor, "child")` is False or `tensor.child is None`?

syft/generic/frameworks/hook/hook.py
```diff
@@ -187,9 +186,9 @@ def _which_methods_should_we_auto_overload(self, tensor_type: type):
             A list of methods to be overloaded.
         """
 
-        boolean_comparators = ["__gt__", "__ge__", "__lt__", "__le__"]
+        self.boolean_comparators = ["__gt__", "__ge__", "__lt__", "__le__"]
```
Jasopaum: Maybe it would be better to put this in `__init__`?

syft/generic/object.py
```diff
@@ -56,12 +56,22 @@ def describe(self, description: str) -> "AbstractObject":
...
             self.tags = set()
 
-        for new_tag in _tags:
-            self.tags.add(new_tag)
+        # Update the owner tag index
+        for tag in tags:
+            self.tags.add(tag)
+            if self.owner is not None:
+                if self.id not in self.owner._objects:
+                    self.owner.register_obj(self)
```
Jasopaum: Are we sure we want to do this under the hood?

syft/generic/object.py
```diff
@@ -56,12 +56,22 @@ def describe(self, description: str) -> "AbstractObject":
...
+                if tag not in self.owner._tag_to_object_ids:
```
Jasopaum: nit: if `self.owner._tag_to_object_ids` was a `defaultdict(set)`, we could avoid the if

syft/generic/object_storage.py
```diff
@@ -79,35 +81,40 @@ def set_obj(self, obj: Union[FrameworkTensorType, AbstractTensor]) -> None:
...
+                else:
+                    self._tag_to_object_ids[tag].add(obj.id)
+
+    def rm_obj(self, remote_key: Union[str, int], force=False):
         """Removes an object.
 
         Remove the object from the permanent object registry if it exists.
 
         Args:
             remote_key: A string or integer representing id of the object to be
```
Jasopaum: Why is the argument called remote_key?

syft/generic/object_storage.py
```diff
@@ -126,3 +133,28 @@ def clear_objects(self, return_self: bool = True):
     def current_objects(self):
         """Returns a copy of the objects in the object storage."""
         return self._objects.copy()
+
+    def find_by_id(self, id):
+        """Local search by id"""
+        if id in self._objects:
```
Jasopaum: nit: I think you could use the `.get()` method on dicts

syft/generic/object_storage.py
```diff
@@ -126,3 +133,28 @@ def clear_objects(self, return_self: bool = True):
...
+            tag (str): exact tag searched
+
+        Return:
+            A list of results, possibly empty
+        """
+        if tag in self._tag_to_object_ids:
+            results = []
+            for obj_id in self._tag_to_object_ids[tag]:
+                obj = self.find_by_id(obj_id)
+                if obj is not None:
```
Jasopaum: Why is this check needed? When would we fill the dict with None objects?

syft/generic/object_storage.py
```diff
@@ -126,3 +133,28 @@ def clear_objects(self, return_self: bool = True):
...
+        """Local search by tag
+
+        Args:
+            tag (str): exact tag searched
+
+        Return:
+            A list of results, possibly empty
+        """
+        if tag in self._tag_to_object_ids:
+            results = []
```
Jasopaum: I may like list comprehension too much but maybe you can use one here (if it makes the code clearer)

syft/execution/plan.py
```diff
@@ -394,6 +421,21 @@ def __call__(self, *args, **kwargs):
             else:
                 return tuple(response)
 
+    def instantiate(self, placeholder, response):
```
Jasopaum: As `self` is not used here, maybe you could make this a static method

syft/execution/plan.py
```diff
@@ -682,12 +777,17 @@ def tag_sort(keyword):
...
-    # TODO is only works up to 9 return values, because comparison is done on str and '7' > '16'
+
     def extract_key(placeholder):
         for tag in placeholder.tags:
             if keyword in tag:
-                return tag
-
-        return TypeError(f"Tag '{keyword}' not found in placeholder tags:", placeholder.tags)
+                try:
+                    return int(tag.split("-")[-1])
```
Jasopaum: I see a lot of logic about tags appearing, would it be worth having a Tag class to avoid spreading it everywhere?

syft/generic/pointers/pointer_plan.py
```diff
@@ -83,6 +83,21 @@ def __call__(self, *args, **kwargs):
         Transform the call on the pointer in a request to evaluate the
         remote plan
         """
+        if len(self._locations) > 1 and isinstance(args[0], sy.MultiPointerTensor):
+            responses = {}
+            for location in self._locations:
+                child_args = list(
```
Jasopaum: I've always found list comprehensions clearer and more pythonic

syft/workers/base.py
```diff
@@ -911,62 +911,95 @@ def _fetch_protocol_remote(
...
+            A list of valid results found.
+
+        TODO Search on description is not supported for the moment
         """
         if isinstance(query, (str, int)):
             query = [query]
+        # Empty query returns all the tagged and registered values
+        elif len(query) == 0:
+            results = []
+            for obj in self._objects.values():
```
Jasopaum: Maybe I misunderstood something but why don't you use self. _tag_to_object_ids if you only want tagged objects?

syft/workers/base.py
```diff
@@ -911,62 +911,95 @@ def _fetch_protocol_remote(
...
+                if hasattr(obj, "tags") and obj.tags is not None and len(obj.tags) > 0:
+                    results.append(obj)
+            return results
+
+        results = None
+        for query_item in query:
+            results_item = set()
```
Jasopaum: Why do you create this intermediary set instead of putting everything into results?

syft/workers/base.py
```diff
@@ -911,62 +911,95 @@ def _fetch_protocol_remote(
...
+            self.register_obj(result)
+        return results
 
-        return result
+    def find_or_request(self, tag, location):
+        """
+        Allow efficient retrieval: if the tag is know locally, return the local
+        element. Else, perform a search on location
+        """
+        results = self.find_by_tag(tag)
```
Jasopaum: Not sure I fully understand for which use case this method exists.
But my remark here is: is it true that if the tag exists, it is not necessarily associated to a pointer to location?

syft/generic/frameworks/hook/hook.py
LaRiffle: Correct! I'll set it globally, but I can't put it in init as it is an abstract method

syft/generic/object.py
LaRiffle: Well, if I tag smthg I want to search for it later, so it has to be registered right?
Registration is not 100% done properly for the moment so this is kind of a fix

syft/frameworks/torch/tensors/interpreters/placeholder.py
LaRiffle: I rmed `tensor.is_wrapper`

syft/generic/object_storage.py
LaRiffle: No clue! it was changed at some poitn by someone :D

syft/generic/object_storage.py
LaRiffle: hum not sure it wouldn't keep on a single line. Unless you have an idea?

syft/generic/object_storage.py
LaRiffle: find_by_id can return None if there is a bug in the coherence of _tag_to_object_ids. It's to improve resilience

syft/generic/pointers/pointer_plan.py
LaRiffle: This makes sense!

syft/workers/base.py
LaRiffle: The tag can be used to refer to local or remote values. Use case: I download a plan using a tag. Next time I need it I use find_or_request and because I already have it I don't need to re-download it!
Does it answer your question

syft/workers/base.py
LaRiffle: good point

syft/workers/base.py
LaRiffle: Correct!

syft/execution/plan.py
LaRiffle: I agree, but I'm not sure tags are mature enough for this: here I use tags with a common prefix and number index appended. I don't know if all tags will support this features or not, so keeping this hidden in plan.py doesn't bother me.

syft/frameworks/torch/tensors/interpreters/placeholder.py
LaRiffle: Ok that's the contrary: I only keep `tensor.is_wrapper` 
The pb here is the this flag (is_wrapper) is not 100% reliable. But here it seems to be ok

syft/generic/object.py
Jasopaum: Sure! But if we raise an error when we try to tag an unregistered tensor, we'll see when the registration is broken and it'll force us to fix it

syft/generic/object_storage.py
Jasopaum:
```result = [
    self.find_by_id(obj_id)
    for obj_id in self._tag_to_object_ids[tag]
    if self.find_by_id(obj_id) is not None
]
```
But I'm not sure if it's better because I have find_by_id twice. You can keep it as is :)

syft/workers/base.py
Jasopaum: Only the first one :)
But to reformulate my second question: do we want to check that the found pointer really points to location before returning it?
And also, can't it be the case that we have the same tag for pointer to different locations?

syft/generic/object.py
LaRiffle: Haha that's right - but I know it's broken and that's it's really tricky to solve and I don't want to do it in this PR^^
May I add a TODO instead, linking to a new issue?

syft/generic/object_storage.py
LaRiffle: Will do :)

syft/workers/base.py
LaRiffle: > do we want to check that the found pointer really points to location before returning it
Excellent point I'll add this check!

syft/generic/object.py
Jasopaum: Fine by me if we add a comment to say that it should eventually be removed when everything is fixed (and if we don't forget to actually fix that) :)

PR #3083, opened 2020-02-21
Multi-worker computation graphs (SplitNN)

### LINKED ISSUE #3000: Make .send() differentiable
If you build a graph of computation across multiple workers, you'll use send or move to do so. But those are not differentiable and hence .backward() won't work except on the last worker where you call it

I'd like send() to be crossable is the opposite direction (using a reference to the original value/ location) to send back a .backward() call when needed

**Additional context**
This is super useful for SplitNN

LaRiffle:
We add an option to send() to support backpropagation on a computation graph held by several workers.

Fix #3000

Example:
```python

x = th.tensor([1., 2., 3, 4, 5], requires_grad=True)
y = (x + x)
me.register_obj(y) # registration on the local worker is sometimes buggy

y_ptr = y.send(alice, requires_grad=True)
z_ptr = y_ptr * 2

z_ptr2 = z_ptr.move(bob, requires_grad=True)
z = z_ptr2.sum()
z.backward()

print(x.grad)
#> tensor([4., 4., 4., 4., 4.])
```

TODO:
- [x] Comment & clean code
- [x] Add tests
- [x] Finish splitNN tutorials cleaning


syft/workers/base.py
```diff
@@ -669,15 +688,22 @@ def de_register_obj(self, obj: object, _recurse_torch_objs: bool = True):
 
     # SECTION: convenience methods for constructing frequently used messages
 
-    def send_obj(self, obj: object, location: "BaseWorker"):
+    def send_obj(self, obj: object, location: "BaseWorker", requires_grad=False):
```
karlhigley: Wonder if we could avoid the binary flag if the `sender` and `origin_id` were already set on the `object` by the time this method got called. 🤔

syft/workers/base.py
```diff
@@ -516,7 +535,7 @@ def execute_communication_action(self, action: CommunicationAction) -> PointerTe
 
             response = hook_args.register_response("send", response, [sy.ID_PROVIDER.pop()], self)
 
-            self.rm_obj(obj_id)
+            # self.rm_obj(obj_id) #TODO put it back
```
karlhigley: Definitely TODO this! 😄

syft/messaging/message.py
```diff
@@ -210,10 +210,12 @@ class ObjectMessage(Message):
     to do so.
     """
 
-    def __init__(self, object_):
+    def __init__(self, object_, sender=None, origin_id=None):
         """Initialize the message."""
 
         self.object = object_
+        self.object.sender = sender
+        self.object.origin_id = origin_id
```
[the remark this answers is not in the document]
karlhigley: Actually, on reading further, I think putting these properties on the object makes a lot of sense, but it might make sense to do sooner in the method call chain to avoid having to pass a flag around to decide whether or not to set them.

syft/messaging/message.py
```diff
@@ -232,7 +234,11 @@ def simplify(worker: AbstractWorker, msg: "ObjectMessage") -> tuple:
         Examples:
             data = simplify(msg)
         """
-        return (sy.serde.msgpack.serde._simplify(worker, msg.object),)
+        return (
+            sy.serde.msgpack.serde._simplify(worker, msg.object),
+            sy.serde.msgpack.serde._simplify(worker, msg.object.sender),
+            sy.serde.msgpack.serde._simplify(worker, msg.object.origin_id),
```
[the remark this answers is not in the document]
karlhigley: I think these should probably be added to the serialization for tensors.

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -58,6 +58,42 @@ class TorchTensor(AbstractTensor):
     def has_child(self):
         return hasattr(self, "child")
 
+    def trigger_origin_backward_hook(self, sender: str, origin_id: int):
```
karlhigley: This is some very neat magic! 👍

syft/generic/pointers/pointer_tensor.py
```diff
@@ -180,10 +180,10 @@ def get_class_attributes(self):
     @staticmethod
     def create_pointer(
         tensor,
-        location: AbstractWorker = None,
+        location: Union[AbstractWorker, str] = None,
         id_at_location: (str or int) = None,
         register: bool = False,
-        owner: AbstractWorker = None,
+        owner: Union[AbstractWorker, str] = None,
```
karlhigley: Would be nice if we can keep the types of these parameters simple. Does this help with another part of the PR? Feel like I might have missed something.

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -638,8 +681,18 @@ def allow(self, user=None) -> bool:
                 current_tensor = current_tensor.child
         return True
 
-    def move(self, location):
-        self.child = self.child.move(location)
+    def move(self, location: BaseWorker, requires_grad: bool = False):
+        """
```
karlhigley: Love the docstrings in this PR. ❤️

syft/workers/base.py
```diff
@@ -669,15 +688,22 @@ def de_register_obj(self, obj: object, _recurse_torch_objs: bool = True):
...
                 receive the object.
+            requires_grad: Default to False. If true, whenever the remote value of this tensor
+                will have its gradient updated (for example when calling .backward()), a call
+                will be made to set back the local gradient value.
         """
-        return self.send_msg(ObjectMessage(obj), location)
+        if requires_grad:
+            kwargs = dict(sender=self.id, origin_id=obj.id)
+        else:
+            kwargs = dict(sender=None, origin_id=None)
```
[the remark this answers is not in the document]
karlhigley: Seems like creating an `ObjectMessage` changes the state of the `object`, which is kinda surprising! Can it happen sooner?

test/torch/pointers/test_pointer_tensor.py
```diff
@@ -289,7 +289,7 @@ def test_move(workers):
 
     x.move(alice)
 
-    assert x.id_at_location not in bob._objects
+    assert x.id_at_location in bob._objects
```
karlhigley: Having read through the code above, I think I see what you're trying to do now, which is 👍 

I still find this change in the semantics of `move` surprising though. It no longer fits my intuition of what `move` means. Should we change the method name? Introduce an additional method so we can keep the original semantics of `move`? Not sure how to handle this, but I'd be very surprised if I `moved` something and it made a copy in a new location.

syft/workers/base.py
LaRiffle: I handle the requires_grad kwarg at the worker.send() level now

syft/messaging/message.py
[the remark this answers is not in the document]
LaRiffle: I've added those attributes to the tensor :)

syft/generic/pointers/pointer_tensor.py
LaRiffle: I just felt guilty and wanted to pay tribute to the 2 `owner = tensor.owner.get_worker(owner)` below :/

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: Thanks :))

LaRiffle:
My last interrogations are linked to:

https://github.com/OpenMined/PySyft/pull/3083/files?file-filters%5B%5D=.py#r393791192
I need for move not to delete the object but I hardly see how to know when to delete or not. My main concern here is with stale pointers. Id'd rather be conservative on the deletion

https://github.com/OpenMined/PySyft/pull/3083/files?file-filters%5B%5D=.py#r393797731
I don't have an elegant name for move which keeps a local copy. I've added a comment in the docstring to underline this behaviour and why it is here (to mimic .send() behaviour), but I'm happy to continue discussion about this in this PR or in another one

examples/tutorials/advanced/Split Neural Network/Tutorial 1 - SplitNN Introduction.ipynb
```diff
@@ -197,9 +197,40 @@
...
+     "text": [
+      "Epoch 0 - Training loss: 0.3258790969848633\n",
+      "Epoch 1 - Training loss: 0.23088108003139496\n"
+     ]
+    },
+    {
+     "ename": "KeyboardInterrupt",
+     "evalue": "",
+     "output_type": "error",
+     "traceback": [
```
Jasopaum: Beware, the notebook fails here

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -58,6 +58,42 @@ class TorchTensor(AbstractTensor):
...
+            """
+            The function setting back the gradient and calling backward
+
+            Args:
+                grad: the gradient tensor being set
+            """
+
+            location = self.owner.get_worker(sender)
+
+            origin_ptr = syft.PointerTensor.create_pointer(
```
Jasopaum: Not sure I followed...
You're creating a pointer from sender to receiver? And then set grad on this new pointer?

syft/generic/pointers/pointer_tensor.py
```diff
@@ -261,24 +261,47 @@ def create_pointer(
...
+        # move to local target is equivalent to doing .get()
+        if self.owner.id == destination.id:
+            return self.get()
+
+        ptr = self.remote_send(destination, requires_grad=requires_grad)
+
+        # We make the pointer point at the remote value. As the id doesn't change,
+        # we don't update ptr.id_at_location
+        # Note that you have now 2 pointers on different locations pointing to the
+        # same tensor.
```
Jasopaum: This could definitely cause issues :)
Not to be fixed in this PR but maybe we can create an issue or keep track of this in another way

syft/generic/pointers/pointer_tensor.py
```diff
@@ -261,24 +261,47 @@ def create_pointer(
...
 
-        return self
+        Args:
+            destination: the new location of the remote data
+            requires_grad: see send() for details
 
-    def remote_send(self, destination):
+        Returns:
+            A pointer to location
+        """
```
Jasopaum: So basically, move is a remote_send at the end of which the "mover" points to the new location, am I right?

syft/serde/msgpack/torch_serde.py
```diff
@@ -148,6 +148,9 @@ def _simplify_torch_tensor(worker: AbstractWorker, tensor: torch.Tensor) -> bin:
     # and here... leaving a reerence here so i can find it later
     # TODO fix pointer bug
 
+    sender = tensor.sender if hasattr(tensor, "sender") else None
```
Jasopaum: To be sure I understand, you added a sender and an origin_id attribute on an object when it's sent?
I didn't think as much as you about it but maybe it would be cleaner to add this information in the message and not add it on the object?

syft/workers/base.py
```diff
@@ -332,6 +332,7 @@ def send(
         workers: "BaseWorker",
         ptr_id: Union[str, int] = None,
         garbage_collect_data=None,
+        requires_grad=False,
```
Jasopaum: Could you add docstring about requires_grad?

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: I create a pointer from the receiver to the sender (so in the reverse order), back to the origin the tensor came from, and I ask to set back there a gradient

syft/generic/pointers/pointer_tensor.py
LaRiffle: Yes indeed!

syft/generic/pointers/pointer_tensor.py
LaRiffle: Will do:  #3217

syft/serde/msgpack/torch_serde.py
LaRiffle: Haha it was in the message but based on Karl suggestions I moved it to the object, where it seems somewhat cleaner :)

Jasopaum: Hum, I still don't think dynamically adding attributes to object is clean but if you agreed on it, it's ok for now :)

syft/frameworks/torch/tensors/interpreters/native.py
Jasopaum: Oh no, ok, the pointer points back to the origin tensor (thanks to id_at_location=origin_id).
So basically, you just want to send a message to sender to tell him to set_grad(something) on the tensor with id origin_id. Don't you think it would be better to use messages and actions directly instead of going through a pointer (which is less intuitive, I think)

syft/generic/pointers/pointer_tensor.py
karlhigley: In that case the `owner` param is a string?

test/torch/pointers/test_pointer_tensor.py
karlhigley: I wonder if we should call the new behavior something like `copy_to`. 🤔

syft/serde/msgpack/torch_serde.py
karlhigley: I'm not a huge fan of dynamically adding attributes either. My thinking was that we're essentially trying to build up a cross-worker tensor chain, and tensors already contain the information required to handle normal tensor chains, so that might be a reasonable place to keep this information too. It does seem like there's some overlap between pointers, messages/actions, and tensor chaining that's pretty fuzzy right now.

I'm not sure `sender` is actually the right name for this attribute if it's on a tensor, because that sort of implies a messaging context. Maybe `origin`? 🤔

LaRiffle: I'll go for `origin` and  set them in the "constructor" (or kindof) :)

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -55,18 +55,21 @@ class TorchTensor(AbstractTensor):
     checking AbstractTensor.
     """
 
+    origin = None
+    origin_id = None
```
Jasopaum: id_at_origin? :)

LaRiffle: Hahaha damn. **Standardisation** sir lariffle!

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: I'm made a proposition using your brand new TensorCommandMessage.computation :)

Jasopaum: Much sexier 😍

syft/generic/pointers/pointer_tensor.py
LaRiffle: Yes it can be!

test/torch/pointers/test_pointer_tensor.py
LaRiffle: `copy_to` is legit indeed

PR #3113, opened 2020-02-27
Chebyshev or Sigmoid for tanh

gmuraru:
Added the reference for the paper where the polynomial approximations are discussed.
[Here](http://tygert.com/realcrypt.pdf) is the paper (it is also added as a comment)

They talk about more functions and how to approximate those - maybe it is worth to have approximations for more functions?

Added tests and the possibility to choose between Chebyshev or Sigmoid for computing the hyperbolic tangent.

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/additive_shared.py
+++ b/syft/frameworks/torch/tensors/interpreters/additive_shared.py
@@ -349,8 +349,6 @@ class AdditiveSharingTensor(AbstractTensor):
                 crypto_provider=self.crypto_provider,
                 **no_wrap,
             ).child
-        elif isinstance(other, AdditiveSharingTensor):
-            other = other.child
         elif not isinstance(other, dict):
             # if someone passes in a constant, we cast it to a tensor, share it and keep the dict
             other = (
@@ -487,15 +485,14 @@ class AdditiveSharingTensor(AbstractTensor):
 
             if other_is_zero:
                 res = {}
+                first_it = True
 
-                it = iter(shares.items())
-                worker, share = next(it)
-                first_cmd = cmd(share, other)
-                zero_shares = self.zero(first_cmd.shape).child
-                res[worker] = (first_cmd + zero_shares[worker]) % self.field
-
-                for worker, share in it:
+                for worker, share in shares.items():
                     cmd_res = cmd(share, other)
+                    if first_it:
+                        first_it = False
+                        zero_shares = self.zero(cmd_res.shape).child
+
                     res[worker] = (cmd(share, other) + zero_shares[worker]) % self.field
                 return res
             else:
--- a/syft/frameworks/torch/tensors/interpreters/additive_shared.py
+++ b/syft/frameworks/torch/tensors/interpreters/additive_shared.py
@@ -243,12 +243,14 @@ class AdditiveSharingTensor(AbstractTensor):
 
         return sy.MultiPointerTensor(children=pointers)
 
-    def zero(self):
+    def zero(self, shape=None):
         """
         Build an additive shared tensor of value zero with the same
         properties as self
         """
-        shape = self.shape if self.shape else [1]
+
+        if shape == None:
+            shape = self.shape if self.shape else [1]
         zero = (
             torch.zeros(*shape)
             .long()
@@ -347,6 +349,8 @@ class AdditiveSharingTensor(AbstractTensor):
                 crypto_provider=self.crypto_provider,
                 **no_wrap,
             ).child
+        elif isinstance(other, AdditiveSharingTensor):
+            other = other.child
         elif not isinstance(other, dict):
             # if someone passes in a constant, we cast it to a tensor, share it and keep the dict
             other = (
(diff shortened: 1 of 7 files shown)


syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -243,12 +243,14 @@ def reconstruct(self):
 
         return sy.MultiPointerTensor(children=pointers)
 
-    def zero(self):
+    def zero(self, shape=None):
```
gmuraru: Needed this in case I have a matrix multiplication - in that case the dimension for the zero shares might be different.

test/torch/tensors/test_precision.py
```diff
@@ -467,6 +466,34 @@ def test_torch_sigmoid_approx(workers):
...
 
+def test_torch_tanh_approx(workers):
+    """
+    Test the approximate tanh with different tolerance depending on
+    the precision_fractional considered
+    """
+    alice, bob, james = workers["alice"], workers["bob"], workers["james"]
+
+    fix_prec_tolerance_by_method = {
+        "chebyshev": {3: 3 / 100, 4: 3 / 100, 5: 3 / 100},
```
gmuraru: Those might need more tweaking - currently, it works, but we might want a lower bound

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -347,6 +349,8 @@ def add(self, shares: dict, other):
                 crypto_provider=self.crypto_provider,
                 **no_wrap,
             ).child
+        elif isinstance(other, AdditiveSharingTensor):
```
LaRiffle: This is not supposed to happen: the decorator makes sure that the AST other is replaced by his .child

So if you still have an AST it means that you've operated something like [syfttype]>AST>... with the self AST

What is the usecase for this?

syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -482,11 +486,18 @@ def _public_mul(self, shares, other, equation):
...
+
+                it = iter(shares.items())
+                worker, share = next(it)
+                first_cmd = cmd(share, other)
+                zero_shares = self.zero(first_cmd.shape).child
+                res[worker] = (first_cmd + zero_shares[worker]) % self.field
+
+                for worker, share in it:
+                    cmd_res = cmd(share, other)
+                    res[worker] = (cmd(share, other) + zero_shares[worker]) % self.field
```
LaRiffle: Can you put all in the same loop  (and add a flag to notice when it's the first iteration) ?

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -413,7 +413,7 @@ def matmul(self, *args, **kwargs):
             ), "In matmul, all args should have the same precision_fractional"
 
         if isinstance(self.child, AdditiveSharingTensor) and isinstance(other.child, torch.Tensor):
-            # If we try to matmul a FPT>(wrap)>AST with a FPT>torch.tensor,
+            # If we try to matmul a FPT>(wrap)>AST with a FPT>(wrap)>torch.tensor,
```
LaRiffle: Nope, `FPT>(wrap)>torch.tensor` is forbidden!

But actually `FPT>(wrap)>AST` is not correct anymore, now we do `FPT>AST`  directly

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -422,7 +422,7 @@ def matmul(self, *args, **kwargs):
         elif isinstance(other.child, AdditiveSharingTensor) and isinstance(
             self.child, torch.Tensor
         ):
-            # If we try to matmul a FPT>torch.tensor with a FPT>(wrap)>AST,
+            # If we try to matmul a FPT>(wrap)>torch.tensor with a FPT>(wrap)>AST,
```
LaRiffle: same here

test/torch/tensors/test_precision.py
LaRiffle: That's ok I guess!

syft/frameworks/torch/tensors/interpreters/additive_shared.py
gmuraru: Sure

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -422,7 +422,7 @@ def matmul(self, *args, **kwargs):
         elif isinstance(other.child, AdditiveSharingTensor) and isinstance(
             self.child, torch.Tensor
         ):
-            # If we try to matmul a FPT>torch.tensor with a FPT>(wrap)>AST,
```
gmuraru: @LaRiffle removed the wrap

LaRiffle: Great!

PR #3115, opened 2020-02-27
Fix failing tests on master

karlhigley:
RNG state coupling between tests still needs a more thorough fix, but
this will make the tests pass for now.


iamtrask:
*crosses fingers*

PR #3124, opened 2020-03-01
Add dependency caching to Github Actions

### LINKED ISSUE #3088: Cache dependencies in CI tests
Running the tests and tutorials currently takes around 15 minutes, of which 4 minutes or so (~25%) is spent installing dependencies.

Use [GH Cache Action](https://github.com/actions/cache) to cache the dependencies and speed up the build.

- Leave it as is. Slower, but works.
- Speed up the tests and tutorials. Maybe possible, but they will inevitably get slower again over time.

**Additional context**
We'll probably want to hash all of the `pip-dep/*.txt` files in order to come up with the cache key.

imskr:
Cache dependencies in CI tests. 
This is draft PR. Need some clarification.
Fixes #3088


imskr:
@karlhigley Am I doing the right way? and do I have to add all of `pip-dep/*.txt` in keys

.github/workflows/cache.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    steps:
+    - uses: actions/cache@v1.1.2
+      id: cache
+      if: startsWith(runner.os, 'Linux')
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/pip-dep/requirements*.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
```
karlhigley: I think we can leave off this optional key. The docs say:
> restore-keys: Optional An ordered list of alternative keys to use for finding the cache if no cache hit occurred for key.

If we don't have a cache hit, it's okay to just install the dependencies again. (In fact, that's what we want.)

.github/workflows/cache.yml
```diff
@@ -0,0 +1,26 @@
...
+    branches:
+    - master
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/cache@v1.1.2
```
karlhigley: I think we'll want this step to be inserted into the `tests.yml`, `tutorials.yml`, and `translated-tutorials.yml` files (rather than in a file by itself.)

.github/workflows/cache.yml
```diff
@@ -0,0 +1,26 @@
...
+      id: cache
+      if: startsWith(runner.os, 'Linux')
```
karlhigley: All of the jobs run on Linux right now and I don't foresee us adding MacOS Github Actions any time soon, so we can probably omit the `startsWith` check.

.github/workflows/cache.yml
```diff
@@ -0,0 +1,26 @@
...
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/pip-dep/requirements*.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+    - name: Install Dependencies
+      if: steps.cache.outputs.cache-hit != 'true'
```
karlhigley: Once the caching step is added to the other yml files, their installation steps will need this check.

.github/workflows/cache.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/cache@v1.1.2
+      id: cache
+      if: startsWith(runner.os, 'Linux')
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/pip-dep/requirements*.txt') }}
```
karlhigley: I _think_ the leading `**` is unnecessary here.

.github/workflows/cache.yml
imskr: should I remove line 14 and add it into `tests.yml` , `tutorials.yml` and `translated-tutorials.yml` ?

imskr:
@karlhigley says `##[error]The template is not valid. hashFiles('/pip-dep/requirements*.txt') failed. Directory '/home/runner/work/PySyft/PySyft' is empty`

.github/workflows/tests.yml
```diff
@@ -22,6 +22,10 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
+    - uses: actions/cache@v1.1.2
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('pip-dep/requirements*.txt') }}
     - name: Install dependencies
```
karlhigley:
```suggestion
    - name: Install dependencies
      if: steps.cache.outputs.cache-hit = 'true'
```

.github/workflows/translated-tutorials.yml
```diff
@@ -28,6 +28,10 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
+    - uses: actions/cache@v1.1.2
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('pip-dep/requirements*.txt') }}
     - name: Install dependencies
```
karlhigley:
```suggestion
    - name: Install dependencies
      if: steps.cache.outputs.cache-hit = 'true'
```

.github/workflows/tutorials.yml
```diff
@@ -27,6 +27,10 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
+    - uses: actions/cache@v1.1.2
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('pip-dep/requirements*.txt') }}
     - name: Install dependencies
```
karlhigley:
```suggestion
    - name: Install dependencies
      if: steps.cache.outputs.cache-hit = 'true'
```

PR #3127, opened 2020-03-01
Run the tests in random orders

karlhigley:
Hoping to shake out weird coupling between tests based on Torch hook, worker, and RNG state.


karlhigley:
Random test ordering still needs to be disabled for the tutorials and translations, because of the way the notebook coverage tests work.

karlhigley:
Would be great to also set the `torch.manual_seed` via entrypoints in the confit file too.

Issue #3008, opened 2020-02-04
Re-work Protocols to incorporate Actions, Placeholders, Roles, and Executions


LaRiffle:
We propose a new way of building Protocol on top of Plan using PlaceHolder.

**Example of protocol**
```python
@sy.protocol(alice=RoleWorker())
def foo(x):
    y = x - 1
    y_ptr = y.send(alice)
    z_ptr = y_ptr.abs()
    return z_ptr
```

**Standard workflow:**
```python
# build constructs a tape of logs for each RoleWorker
foo.build(th.Tensor([-1]))

print(foo.plans)
#>out  {'local': Plan1, 'alice': Plan2}

foo_deployed = foo.deploy(local=james, alice=bob)

res = foo_deployed.run(th.tensor([1.5]))
print(res)
#>out ptr to bob with remote value=0.5

# redeploy: basically copy and instantiate the Roles
foo_deployed2 = foo.deploy(local=andrew, alice=dan)
``` 

`protocol.build`  will create a tape of log on each Role involved (these Role are like a worker that exist just for the build)(note that a `local` Role is created implicitly in addition to `alice`). Each tape will become a plan. Each Plan will be composed of a list of `Operation` or `ObjectMessage` exclusively. More precisely sending a tensor A -> B will add a `ObjectMessage` in tape of A. Calling now `get()`from A on a remote pointer to B will add a `ObjectMessage` on tape B to get the tensor back. (same strategy for move). During the build, ie when the tracing is on, ie when the tapes are recording, the `owner` of Placeholders (which is one of the Role worker involved) used in the Operation as arguments is used to determine in which tape goes the operation, and defines the owner of the return placeholders.

# Steps:
- Create `Role` which should have the same function with workers as the placeholder for tensors
- Add a tracing functionality on `tensor.send`: it should in particular record the origin and target Roles.
- Create a trace log (or _tape_) per worker. (_One issue here is that it's often hard to know on which worker an operation is running when dealing with virtual workers_)
- Build plans under the hook when calling `protocol.build`, from the trace with multiple tapes
- After building `tensor.send` corresponds to a remote instantiation of a placeholder.
- Placeholder should know which Operation they are involved in, so that when it's instantiated, either from local or remote command, it can trigger the Operation depending on it, provided that all the placeholder in this Operation have been instantiated.

###  Function

```python
@sy.protocol(
     alice=RoleWorker(state=(nn.Linear(2, 3), )), 
[...]

PR #3132, opened 2020-03-02
Extract ComputationAction from OperationMessage

karlhigley:
An early step toward #3008. Depends on OpenMined/syft-proto#47.


karlhigley:
Operation and OperationMessage now have separate (but related) msgpack serialization methods. The next step is to extract Operation's Protobuf serialization from the existing methods on OperationMessage into new methods on Operation.

syft/messaging/message.py
```diff
@@ -149,9 +164,9 @@ def contents(self):
...
         simplify return_ids because they are always a list of integers, meaning they're already simplified)."""
 
-        message = (self.cmd_name, self.cmd_owner, self.cmd_args, self.cmd_kwargs)
+        message = (self.action.name, self.action.operand, self.action.args, self.action.kwargs)
 
-        return (message, self.return_ids)
+        return (message, self.action.return_ids)
 
     @staticmethod
     def simplify(worker: AbstractWorker, ptr: "OperationMessage") -> tuple:
```
iamtrask: Are we planning to keep "simplify" and "detail' around long term given Protobuf?

syft/messaging/message.py
```diff
@@ -189,72 +197,32 @@ def detail(worker: AbstractWorker, msg_tuple: tuple) -> "OperationMessage":
...
-            for return_id in return_ids:
-                if type(return_id) == PlaceHolder:
-                    protobuf_op.return_placeholders.extend(
-                        [sy.serde.protobuf.serde._bufferize(worker, return_id)]
-                    )
-                else:
-                    sy.serde.protobuf.proto.set_protobuf_id(protobuf_op.return_ids.add(), return_id)
-
-        protobuf_op_msg.operation.CopyFrom(protobuf_op)
+        protobuf_op = ComputationAction.bufferize(worker, operation_message.action)
```
iamtrask: Is there an intentional design choice in "bufferize" being used functionally instead of as a method on operation_message.action?

iamtrask: (not a big deal)

syft/messaging/message.py
karlhigley: No, but they still need to work for the time being since Protobuf serialization isn’t complete yet.

syft/messaging/message.py
karlhigley: Yes, there is, but it originally comes from msgpack serialization. If I recall correctly, it’s because there’s no object to call `detail`/`unbufferize` on; the object is the output of those methods (rather than an input.) As a result, those methods have to be static, and making `simplify`/`bufferize` static also creates a parallel structure for serialization in both directions.

karlhigley:
@LaRiffle Yeah, the actions in `Plans` can only be `ComputationActions`. That restriction is captured [in the type annotations in `__init__()`](https://github.com/OpenMined/PySyft/pull/3132/files#diff-df3ce31527cd05c4bf6283bf594bde03R99), but I didn't want to make the name of the property super-long and inconvenient.

karlhigley:
(Rebased this PR on top of the Plan changes in #3080.)

PR #3137, opened 2020-03-03
Rebase Crypten

gmuraru:


syft/workers/virtual.py
```diff
@@ -1,10 +1,19 @@
...
+        """send message to worker location"""
+        if self.message_pending_time > 0:
+            if self.verbose:
+                print(f"pending time of {self.message_pending_time} seconds to send message...")
+            sleep(self.message_pending_time)
+
         return location._recv_msg(message)
 
     def _recv_msg(self, message: bin) -> bin:
+        """receieve message"""
```
LaRiffle: Typo

test/workers/test_virtual.py
```diff
@@ -25,6 +25,9 @@ def test_send_msg():
     # get pointer to local worker
     me = sy.torch.hook.local_worker
 
+    # pending time to simulate lantency (optional)
+    me.message_pending_time = 0.1
```
LaRiffle: As me is global, shouldn’t we set it back to zero at the end of the test?

test/workers/test_virtual.py
```diff
@@ -34,10 +37,14 @@ def test_send_msg():
...
 
     # Send data to bob
+    start_time = time()
     me.send_msg(ObjectMessage(obj), bob)
+    elapsed_time = time() - start_time
 
     # ensure that object is now on bob's machine
     assert obj_id in bob._objects
+    # ensure that object was sent 0.1 secs later
+    assert abs(elapsed_time - me.message_pending_time) < 0.1
```
LaRiffle: That’s confusing why not just ensure rlapsed_time > 0.1 ?

test/workers/test_virtual.py
gmuraru: It should be set

test/workers/test_virtual.py
gmuraru: You are right, will do that.

syft/workers/virtual.py
gmuraru: Done

test/workers/test_virtual.py
gmuraru: Done

PR #3139, opened 2020-03-04
Little fixes in the code

gmuraru:
Some issues found by @LaRiffle while rebasing CrypTen.

### DIFF
--- a/test/workers/test_virtual.py
+++ b/test/workers/test_virtual.py
@@ -41,10 +41,12 @@ def test_send_msg():
     me.send_msg(ObjectMessage(obj), bob)
     elapsed_time = time() - start_time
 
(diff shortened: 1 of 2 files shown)


test/workers/test_virtual.py
```diff
@@ -41,10 +41,12 @@ def test_send_msg():
     me.send_msg(ObjectMessage(obj), bob)
     elapsed_time = time() - start_time
 
+    me.message_pending_time = 0
```
gmuraru: Q: There is still a problem with this approach if any of the above lines will throw an exception, then the value for ```message_pending_time`` remains 0

LaRiffle: I agree, a python context would be better here but let's say it's ok for now

PR #3162, opened 2020-03-08
README.md update: specify python=3.7

teddykoker:
As of now, `conda create -n pysyft python=3` will create an environment with python 3.8, which will not be able to run PySift, due to its Tensorflow dependencies. By specifying `python=3.7`, we solve this issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Prompt](https://docs.anaconda.com/anaconda/user-guide/getting-started/) to
 work from the command line.
 
 ```bash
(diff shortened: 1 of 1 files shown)


gmuraru:
Should we also update the following?
```
INSTALLATION.md -- it says python3.6 or above
docker-images/pysyft-worker/Dockerfile -- it uses python:3.6-slim
```

karlhigley:
Python 3.6 is also supported, but 3.8 is not.

PR #3165, opened 2020-03-09
Plans for CrypTen

gmuraru:


syft/workers/virtual.py
```diff
@@ -7,4 +7,5 @@ def _send_msg(self, message: bin, location: BaseWorker) -> bin:
         return location._recv_msg(message)
 
     def _recv_msg(self, message: bin) -> bin:
+        """receive message"""
```
gmuraru: @LaRiffle this one and the one below are the fixes for the issues you observed

syft/frameworks/crypten/__init__.py
```diff
@@ -1,13 +1,13 @@
...
-import syft
-from syft.frameworks.crypten.context import toy_func, run_party
+import syft as sy
+from syft.frameworks.crypten.context import run_party
 import crypten.communicator as comm
 import crypten
 
-
-def load(tag: str, src: int, id_worker: int):
+def load(tag: str, src: int, id_worker: int, size: tuple):
```
youben11: We may not know the size of loaded tensors so we shouldn't have `size` as a required parameter, instead, it should be optional, and we either exchange messages in case size is not provided, or doesn't exchange messages at all if it's not.

syft/execution/plan.py
```diff
@@ -382,8 +384,17 @@ def __call__(self, *args, **kwargs):
...
                 else:
+                    if isinstance(_self.child, crypten.mpc.MPCTensor):
+                        if isinstance(_self.child, crypten.mpc.MPCTensor):
+                            _self = _self.child
+                            args = [arg.child for arg in args]
+
                     response = getattr(_self, cmd)(*args, **kwargs)
-                return_placeholder.instantiate(response.child)
+
+                if "child" in dir(response):
```
LaRiffle: maybe use `hasattr` instead?

syft/frameworks/crypten/hook/hook.py
```diff
@@ -0,0 +1,23 @@
...
+
+    @tracer(func_name=cmd_name)
+    @wraps(func)
+    def overloaded_func(*args, **kwargs):
+        assert "size" in kwargs
+
+        try:
+            response = func(*args, **kwargs)
+        except Exception as e:
+            response = th.rand(kwargs["size"])
```
LaRiffle: Why do you need to run the function? I mean is it required or it could just be always a random response ?

syft/frameworks/torch/hook/hook.py
```diff
@@ -34,6 +34,13 @@
 
 from syft.exceptions import route_method_exception
 
+from syft.frameworks.crypten import load as crypten_load
+
+try:
+    import crypten
+except e:
+    print("CrypTen not found")
```
LaRiffle: I think you can use the same strategy than for the imports or PT & TF (but this is not important)

syft/frameworks/torch/hook/hook.py
```diff
@@ -472,6 +482,19 @@ def _hook_torch_module(self):
...
 
+    # TODO -- Should do this only if we have CrypTen as a backend
+    def _hook_crypten(self):
+        from syft.frameworks.crypten.hook.hook import get_hooked_crypten_func
+
+        overload_funcs = ["load"]
+
+        native_func = getattr(crypten, "load")
+        setattr(crypten, "native_load", native_func)
+        setattr(crypten, "load", crypten_load)
```
LaRiffle: This line can be removed I guess

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -254,7 +254,7 @@ def roll(tensor, shifts, **kwargs):
     def handle_func_command(cls, command):
         """
         Operates as a router for functions. A function call always starts
-        by being handled here and 3 scenarii must be considered:
+        by being handled here and 3 scenarios must be considered:
```
LaRiffle: scenarii is ok - it's when you want to look as a star, you put the italian plural for works that have italian origins

syft/workers/base.py
```diff
@@ -415,7 +416,14 @@ def run_crypten_party(self, message: tuple):
         """
 
         rank, world_size, master_addr, master_port = message
-        return_value = run_party(toy_func, rank, world_size, master_addr, master_port, (), {})
+
+        plan = None
+        for key, val in self._objects.items():
```
LaRiffle: Why not using tags to retrieve the plan?

syft/generic/frameworks/hook/hook.py
```diff
@@ -538,8 +540,10 @@ def overloaded_native_method(self, *args, **kwargs):
                 method = getattr(new_self, method_name)
                 response = method(*new_args, **new_kwargs)
 
-                response.parents = (self.id, new_self.id)
+                if isinstance(response, crypten.mpc.MPCTensor):
+                    return response
 
+                response.parents = (self.id, new_self.id)
```
LaRiffle: Not sure I understand the role of those 3 lines for the moment!

gmuraru: Removed them

gmuraru:
Moved here #3254 so will close this one.

PR #3166, opened 2020-03-09
Add SplitNN tutorial

haofanwang:
Add SplitNN tutorials for vertically partitioned data @H4LL .


iamtrask:
Before merging - would you be willing to add a bit more explanation about SplitNN and the code you've written?

haofanwang:
Hi, @LaRiffle. I have added more descriptions and comments in the notebook. This notebook follows the style of the previous ones (tutorial 2 & 3) and is a supplement of tutorial series for SplitNN. The pros & cons have been described (tutorial 1). 

To be consistent with existing tutorials, I think you can merge this first. I will update all tutorials together using the new syntax once your PR has been merged.

H4LL:
Hey @haofanwang  thanks for the tutorial update. It's a great start! however some things I think the following would be great;

- Instead of concatenating the two images it would be good to split a single image (I don't think the two image option is as intuitive)
- In order to make this scale for an arbitrary number of vertical splits then the vertical split layer should be in its own array; [[a_1],[a_2],[b]]. We can then write loops to deal with these arrays for any number of horizontal/ veritical splits. It also makes the models object slightly cleaner as models get more complex (there is an implicit structure int the model array).
- In the description on what's going on in vertically split training environment you mention the different images as different modalities. This is great and accurate but is maybe not colourful enough to make it as clear as it could be at a glance. It might be good to also describe a real-world use-case to tell more of a story/ capture peoples imagination.

overall I'm super impressed with this, great effort!

H4LL:
Hey @haofanwang, this is a really promising start. I think we still need to work on the narrative and refactor the code to fit this functionality inside the SplitNN class. The idea is we want to be able to use this almost identically to how we would a conventional neural network. Let's work together on slack to make it just right

haofanwang:
Done. Please let me know if have any further questions. @H4LL

H4LL:
Hey,

haofanwang:
@H4LL Sure. I will remove my name from the original 3 tutorials first. Then I will PR for the tutorial 4 independently. Thanks. Closed.

PR #3168, opened 2020-03-09
Introduce communication actions

Jasopaum:
Introduce a new CommunicationAction abstraction.

Depends on OpenMined/syft-proto#50.


syft/messaging/message.py
```diff
@@ -156,17 +161,7 @@ def return_ids(self):
...
-        Some of our codebase still assumes that all message types have a .contents attribute. However,
-        the contents attribute is very opaque in that it doesn't put any constraints on what the contents
-        might be. Since we know this message is a action, we instead choose to store contents in two pieces,
-        self.message and self.return_ids, which allows for more efficient simplification (we don't have to
-        simplify return_ids because they are always a list of integers, meaning they're already simplified)."""
-
-        message = (self.action.name, self.action.target, self.action.args, self.action.kwargs)
-
-        return (message, self.action.return_ids)
+        return self.action
```
karlhigley: This might have to remain a tuple, which could either be crafted here or delegated to `self.action.contents`, depending on where we want that code to live. I'd probably rather put it on `CommandMessage`, since knowing about messaging isn't one of `Action's` responsibilities.

syft/messaging/message.py
```diff
@@ -218,9 +211,14 @@ def bufferize(worker: AbstractWorker, action_message: "CommandMessage") -> "Comm
...
         protobuf_op_msg = CommandMessagePB()
-        protobuf_op = ComputationAction.bufferize(worker, action_message.action)
 
-        protobuf_op_msg.action.CopyFrom(protobuf_op)
+        if isinstance(action_message.action, ComputationAction):
+            protobuf_op = ComputationAction.bufferize(worker, action_message.action)
+            protobuf_op_msg.computation.CopyFrom(protobuf_op)
+        elif isinstance(action_message.action, CommunicationAction):
+            protobuf_op = CommunicationAction.bufferize(worker, action_message.action)
+            protobuf_op_msg.communication.CopyFrom(protobuf_op)
```
karlhigley: As long as `ComputationAction` and `CommunicationAction` are included in the serde's list of classes with their own serialization (and it looks like they are), I think you can call `sy.serde.protobuf.serde._serialize(action_message.action)` here without having to check types.

syft/messaging/message.py
```diff
@@ -240,11 +238,18 @@ def unbufferize(worker: AbstractWorker, protobuf_obj: "CommandMessagePB") -> "Co
         Examples:
             message = unbufferize(sy.local_worker, protobuf_msg)
         """
-        detailed = ComputationAction.unbufferize(worker, protobuf_obj.action)
-
-        return CommandMessage(
-            detailed.name, detailed.target, detailed.args, detailed.kwargs, detailed.return_ids
-        )
+        if protobuf_obj.computation.ByteSize():
```
karlhigley: There's [a trick](https://github.com/OpenMined/PySyft/blob/7dc983865a8db7e51918b44bff44ac0ba82b120a/syft/execution/state.py#L191) for figuring out which type is present and retrieving the value without an `if` statement.

syft/messaging/message.py
```diff
@@ -240,11 +238,18 @@ def unbufferize(worker: AbstractWorker, protobuf_obj: "CommandMessagePB") -> "Co
...
+            detailed = ComputationAction.unbufferize(worker, protobuf_obj.computation)
+            action = ComputationAction(
+                detailed.name, detailed.target, detailed.args, detailed.kwargs, detailed.return_ids
+            )
+        elif protobuf_obj.communication.ByteSize():
+            detailed = CommunicationAction.unbufferize(worker, protobuf_obj.communication)
```
karlhigley: Once you retrieve the `action` from the Protobuf object, you should be able to call `sy.serde.protobuf.serde._unbufferize` on it and pass the result into the `CommandMessage` constructor below.

syft/workers/base.py
```diff
@@ -401,7 +402,13 @@ def send(
 
         return pointer
 
-    def execute_command(self, message: tuple) -> PointerTensor:
+    def execute_command(self, message) -> PointerTensor:
```
karlhigley:
```suggestion
    def execute_command(self, action: Action) -> PointerTensor:
```
The code below suggests this might be the appropriate type annotation, but I'm not sure if this is right.

syft/workers/base.py
```diff
@@ -401,7 +402,13 @@ def send(
...
+        if isinstance(message, ComputationAction):
+            return self.execute_computation(message)
+        else:
+            return self.execute_communication(message)
+
+    def execute_computation(self, message) -> PointerTensor:
```
karlhigley:
```suggestion
    def execute_computation(self, action: ComputationAction) -> PointerTensor:
```
If `Action` is the type, `message` might not be the right name for these parameters.

syft/workers/base.py
```diff
@@ -510,9 +529,9 @@ def send_command(
         name, target, args_, kwargs_ = message
 
         try:
-            ret_val = self.send_msg(
-                CommandMessage(name, target, args_, kwargs_, return_ids), location=recipient
-            )
+            action = ComputationAction(name, target, args_, kwargs_, return_ids)
+            message = CommandMessage(action)
+            ret_val = self.send_msg(message, location=recipient)
```
karlhigley: I think this is fine, but I still wonder about adding static pseudo-constructors like `CommandMessage.computation()` and `CommandMessage.communication()` to make snippets like this a bit more terse.

test/serde/serde_helpers.py
```diff
@@ -1333,7 +1375,7 @@ def make_computation_action(**kwargs):
     def compare(detailed, original):
         detailed_msg = (detailed.name, detailed.target, detailed.args, detailed.kwargs)
         original_msg = (original.name, original.target, original.args, original.kwargs)
-        assert type(detailed) == syft.messaging.message.ComputationAction
+        assert type(detailed) == syft.execution.computation.ComputationAction
```
karlhigley: 👍

test/serde/serde_helpers.py
```diff
@@ -1374,48 +1416,67 @@ def compare(detailed, original):
...
-        detailed_msg = (detailed.name, detailed.target, detailed.args, detailed.kwargs)
-        original_msg = (original.name, original.target, original.args, original.kwargs)
-        for i in range(len(original_msg)):
-            if type(original_msg[i]) != torch.Tensor:
-                assert detailed_msg[i] == original_msg[i]
-            else:
-                assert detailed_msg[i].equal(original_msg[i])
-        assert detailed.return_ids == original.return_ids
-        return True
+        if isinstance(detailed.action, syft.execution.computation.ComputationAction):
```
karlhigley: 👍 

Would be nice if the `compare` functions were re-usable outside the scope of the original `make_whatever` function, but that's definitely out of scope for this PR.

syft/execution/communication.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        super().__init__()
+
+        self.obj = obj
+        self.source = source
+        self.destinations = destinations
+        self.kwargs = kwargs
+
+    @property
+    def contents(self):
```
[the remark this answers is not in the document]
karlhigley: On a re-read, this is unfortunate but reasonable for now. We should eventually remove whatever code relies on the `contents` property, but that's out of scope for this PR.

syft/execution/communication.py
```diff
@@ -0,0 +1,170 @@
...
+
+        kwargs = {
+            key: CommunicationAction._unbufferize_arg(worker, kwarg)
+            for key, kwarg in protobuf_obj.kwargs.items()
+        }
+
+        return CommunicationAction(obj, source, destinations, kwargs)
+
+    @staticmethod
+    def _bufferize_arg(worker: AbstractWorker, arg: object) -> ArgPB:
```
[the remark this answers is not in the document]
Jasopaum: I wanted to put them in the `sy.serde.protobuf.serde` file but I don't understand your remark, did you said it shouldn't be there?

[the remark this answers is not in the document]
karlhigley: Ahh, sorry. That's where I would want to put them too, but that requires serialization to/from a specific Python class, and PySyft doesn't have an `Arg` class that corresponds to `ArgPB`.

syft/messaging/message.py
Jasopaum: But I also need to use different arguments in both cases so I don't see how I could avoid having an `if` statement

syft/messaging/message.py
Jasopaum: Oh, didn't see this message before answering to the previous one...

karlhigley: Ha, same! I didn't read far enough to have the whole idea in mind yet when I wrote the comment above.

syft/workers/base.py
Jasopaum: This will be clearer after your PR is ready

syft/workers/base.py
Jasopaum: Same

syft/workers/base.py
karlhigley: (Maybe YAGNI applies here though.)

Jasopaum: Haha, I was just going to do that. I do think it would be clearer to have 
```
message = CommandMessage.computation(name, target, args_, kwargs_, return_ids)
```
instead of the 2 lines above.
Do you think we should refrain from implementing that until there is a more urgent need to do so?

karlhigley: Nah, if you're onboard, let's go for it. It reads nicely and leaves the calling code cleaner. 👍

syft/execution/communication.py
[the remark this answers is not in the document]
Jasopaum: Oh, I was thinking of copy pasting this function at the end of the file but I didn't really read it before... I don't think that `_bufferize_arg` fits the philosophy of `serde/protobuf/serde.py`.
Do you have any other idea?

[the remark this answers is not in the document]
karlhigley: Oh! Yeah, moving them to the end of that file sounds good. 👍 

(You're right that they don't quite fit the philosophy that everything else adheres to. The `ArgPB` class is a consequence of trying to capture very flexible types in a static typing system. We just finished unwinding another somewhat similar case motivated by typing in Protobuf with `Operation`/`OperationMessage`, but I'm not sure if we get can rid of `Arg`.)

karlhigley:
Last thing is to update the requirements file to point at a new `syft-proto` release (once there is one.)

PR #3175, opened 2020-03-10
Avoid simplifying messages when routing/handling them

karlhigley:


syft/messaging/message.py
```diff
@@ -494,32 +571,29 @@ def __init__(self, command_name: str, message: tuple):
...
 
-        # call the parent constructor - setting the type integer correctly
-        super().__init__()
-
         self.command_name = command_name
-        self.message = message
+        self.args = args_
 
     @property
     def contents(self):
```
Jasopaum: Can this one be removed also?

syft/workers/base.py
```diff
@@ -116,11 +116,11 @@ def __init__(
         self._message_router = {
             CommandMessage: self.execute_command,
             PlanCommandMessage: self.execute_plan_command,
-            ObjectMessage: self.set_obj,
+            ObjectMessage: self.handle_object_msg,
             ObjectRequestMessage: self.respond_to_obj_req,
             ForceObjectDeleteMessage: self.rm_obj,  # FIXME: there is no ObjectDeleteMessage
             IsNoneMessage: self.is_tensor_none,
```
Jasopaum: Don't you need to modify is_tensor_none to take the message with an object id inside as parameter?

syft/messaging/message.py
```diff
@@ -420,11 +469,25 @@ class ForceObjectDeleteMessage(Message):
     # TODO: add more efficient detailer and simplifier custom for this type
     # https://github.com/OpenMined/PySyft/issues/2512
 
-    def __init__(self, contents):
-        """Initialize the message using default Message constructor.
+    def __init__(self, obj_id):
+        """Initialize the message."""
+
+        self.object_id = obj_id
```
Jasopaum: Don't you need to do something after routing this message to worker.rm_obj?

syft/messaging/message.py
```diff
@@ -54,21 +54,29 @@ def contents(self):
     def _simplify(self):
         return (self.contents,)
 
+    def __str__(self):
+        """Return a human readable version of this message"""
+        return f"({type(self).__name__} {self.contents})"
```
Jasopaum: self.contents will mostly be None because we'll use child classes, don't you think?
Do we want to keep the contents for Message?

syft/workers/base.py
```diff
@@ -116,11 +116,11 @@ def __init__(
         self._message_router = {
             CommandMessage: self.execute_command,
             PlanCommandMessage: self.execute_plan_command,
-            ObjectMessage: self.set_obj,
+            ObjectMessage: self.handle_object_msg,
             ObjectRequestMessage: self.respond_to_obj_req,
             ForceObjectDeleteMessage: self.rm_obj,  # FIXME: there is no ObjectDeleteMessage
```
Jasopaum: This method on object_storage doesn't expect a message

syft/workers/base.py
```diff
@@ -116,11 +116,11 @@ def __init__(
...
             IsNoneMessage: self.is_tensor_none,
-            GetShapeMessage: self.get_tensor_shape,
+            GetShapeMessage: self.handle_get_shape_message,
             SearchMessage: self.respond_to_search,
             ForceObjectDeleteMessage: self.force_rm_obj,
```
Jasopaum: Same: this method on object_storage doesn't expect a message

syft/messaging/message.py
karlhigley: I tried to get rid of it, but bad things happened. Each message should probably have its own string representation, I suppose.

syft/workers/base.py
karlhigley: `set_obj` doesn’t, so I added a new method that does.

syft/workers/base.py
karlhigley: Yeah, this is a second method that wraps the first one.

karlhigley: Wait, you mean `force_rm_obj`? Yeah, that one doesn’t either. Needs a wrapper method.

Jasopaum: Yep, that what I meant :)

syft/messaging/message.py
karlhigley: Fixed this by making `Message` an Abstract Base Class, removing the `contents` property, and making each sub-class responsible for its own string representation.

PR #3177, opened 2020-03-11
Added additional comments to test files

vineeth14:
Partial fix for #2510
Added docstrings to test_object_storage.py
Added inline comments

### DIFF
--- a/test/generic/test_object_storage.py
+++ b/test/generic/test_object_storage.py
@@ -7,19 +7,19 @@ def test_clear_objects():
     """
     Asserts that objects in storage are cleared
     """
-    obj_storage = object_storage.ObjectStorage()
-    """
-    obj_storage is a wrapper object to a collection of objects 
-    """
+    obj_storage = (
+        object_storage.ObjectStorage()
+    )  #  obj_storage is a wrapper object to a collection of objects
+
     x = torch.tensor(1)
     obj_storage.set_obj(x)
 
-    objs = obj_storage.current_objects()#Returns a copy of the objects in obj_storage(here:x)
+    objs = obj_storage.current_objects()  # Returns a copy of the objects in obj_storage(here:x)
 
     assert len(objs) == 1
     assert objs[x.id] == x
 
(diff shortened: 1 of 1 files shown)


test/generic/test_object_storage.py
```diff
@@ -4,24 +4,32 @@
...
 
 def test_clear_objects():
+    """
+    Asserts that objects in storage are cleared
+    """
     obj_storage = object_storage.ObjectStorage()
-
+    """
+    obj_storage is a wrapper object to a collection of objects 
+    """
```
TTitcombe: An inline comment `# ` would be fine here

test/generic/test_object_storage.py
```diff
@@ -4,24 +4,32 @@
...
     x = torch.tensor(1)
     obj_storage.set_obj(x)
 
-    objs = obj_storage.current_objects()
+    objs = obj_storage.current_objects()#Returns a copy of the objects in obj_storage(here:x)
```
TTitcombe: [`black`](https://github.com/psf/black) requires comments to begin two spaces after code. I think this is why the tests are failing

vineeth14: Oh, I'll fix that 
Thank you

vineeth14:
> These description / docstring don't seem correct

Hey, can you point out which docstring seems incorrect?

test/generic/test_object_storage.py
```diff
@@ -4,24 +4,32 @@
...
 
     objs = obj_storage.current_objects()
     assert len(objs) == 0
     assert ret_val == obj_storage
 
 
 def test_clear_objects_return_None():
+    """
+    Asserts that objects are cleared and returned as none
+    """
```
LaRiffle: For example here I think it could be:
> Check storage clear_objects method when no return is needed

More generally, the docstring should resume the main idea of the function in a clear way: it's a helper. Try to ba as clear and explicit as possible when you add dosctring ;)

PR #3182, opened 2020-03-12
Added Japanese language translated tutorial notebooks from part 1 to 12

kouohhashi:
I translated tutorial from part 1 to 12 into Japanese.


kouohhashi:
Is this because of part 10 ?
We can not run part 10 as is.

karlhigley:
You can see the error logs on the “Checks” tab. Looks like an “invalid character in identifier” error and maybe a few other things.

kouohhashi:
Thank you

kouohhashi:
How can I avoid timeout error?
it looks like timeout=300 may not be big enough for Part 06.

PR #3186, opened 2020-03-12
Turn existing formatted strings into f-strings

TTitcombe:
**This PR**:
- Ran a linter, [`flynt`](https://github.com/ikamensh/flynt), to turn formatted strings which already exist in the codebase to f-strings
- Adds `flynt` to pre-commit

f-strings are quicker than other types of formatting strings.

I can remove the pre-commit addition if needed. It may not be worth the additional overhead of running pre-commit so long as non-f strings in new PRs are caught in review.

There is no existing issue relating to this PR.

**To test**:
1. Check  than the converted strings have been formatted correctly
2. Check pre-commit compiles and works as expected

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,3 +19,13 @@ repos:
     rev: v1.1.0
     hooks:
     -   id: python-use-type-annotations
+-   repo: local
+    hooks:
+    -   id: flynt
+        name: flynt
+        entry: flynt
+        args: [--fail-on-change]
+        types: [python]
+        language: python
+        additional_dependencies:
+        -   flynt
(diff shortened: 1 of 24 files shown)


.pre-commit-config.yaml
```diff
@@ -19,3 +19,13 @@ repos:
...
+-   repo: local
+    hooks:
+    -   id: flynt
+        name: flynt
+        entry: flynt
+        args: [--fail-on-change]
+        types: [python]
+        language: python
+        additional_dependencies:
+        -   flynt
```
karlhigley: Does this need to be added to the dependencies files?

TTitcombe: Not for the pre-commit to work - it will install it automatically if needed - but perhaps it would simplify things to add flynt to dev dependencies anyway. Will do that in a bit

karlhigley:
LGTM

Let me know when it's ready to merge.

Issue #2897, opened 2020-01-12
Cleanup FixedPrecisionTensor


iamtrask:
**Describe the bug**
FixedPrecision tensor has become a catch-all for a myraid of functionality which have nothing to do with fixing precision. At present it seems to include:

0) The ability to model Fixed Precision Numbers using underlying Integer representations
1) BigInt functionality allowing for large precision
2) Convolutional functionality
3) many references to specific subclass types (such as AdditiveSharingTensor).
4) an implementation of pow() which seems unrelated to FixedPrecisionRepresentation
5) Newton-shulz inversion
6) approximation of exponentiation
7) approximation of sigmoid
8) approximation of log
9) etc...

All functionality other than (0) should be removed from this class and refactored into appropriate other classes.

iamtrask:
I will begin working on a properly decoupled BigInt tensor

LaRiffle:
That's right!
1. So actually once you are in Fixed-precision, you can't do anymore pow, exp, simoid, log, inversion the way you did previously, so you have to redefine them, and I'd say this is the good place.
2. Less important, some functions like Convolutional should indeed be put in our dedicated `nn` module where rnn already lies.
3. Having references to AdditiveSharingTensor is *super* annoying, but from our design seems hard to avoid, since we want to operate tensors with different types (FPT & AST)

Actually, we could follow the design principle of another lib and define fixed-precision not as a tensor type but as an encoder which is attached to a tensor. This sounds weird but:
- It reconciliate FPT & AST in a single type and solves 3.
- It diminishes the number of tensors types and reduces the tensor chain size
- It is more readable for non-expert users

To avoid the limitation of having to redefine the encoder for the HomomorphicTensor, etc, encoder could be supplied by a inheritance mechanism, but then @karlhigley is not going to be happy lol

In a nutshell
- I don't think 1. is that of a problem
- Issue 2. can be solved easily
- Issue 3. needs more intelligence, to make for example the mul function of FPT readable

karlhigley:
My only ask would be to supply the encoder as a constructor param with a default value (instead of as an additional parent class via inheritance.) 🙂

iamtrask:
I don't see the semantic difference between an encoder and a tensor type
(something used in our tensor chain).

karlhigley:
I'm not sure I fully understand it either, just chiming in on the point about inheritance. 😄

LaRiffle:
[...]

PR #3195, opened 2020-03-14
Add dropout support for SMPC

### LINKED ISSUE #2500: Dropout is currently not supported for SMPC

sukhadj:
fixes #2500 .
Currently using dropout function specified on ref. https://stackoverflow.com/questions/54109617/implementing-dropout-from-scratch. 
More robust implementation to be added.
This code block is yet to be tested.

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/precision.py
+++ b/syft/frameworks/torch/tensors/interpreters/precision.py
@@ -864,6 +864,10 @@ class FixedPrecisionTensor(AbstractTensor):
                             .child
                         )
 
+                        if inplace:
+                            input = input * noise
+                            return input
+
                         return input * noise
 
                     return input
--- a/syft/frameworks/torch/tensors/interpreters/precision.py
+++ b/syft/frameworks/torch/tensors/interpreters/precision.py
@@ -860,7 +860,7 @@ class FixedPrecisionTensor(AbstractTensor):
                                 binomial.sample(input.shape).type(torch.FloatTensor)
                                 * (1.0 / (1.0 - p))
                             )
-                            .fix_prec()
+                            .fix_prec(**input.get_class_attributes())
                             .child
                         )
 
--- a/syft/frameworks/torch/tensors/interpreters/precision.py
+++ b/syft/frameworks/torch/tensors/interpreters/precision.py
@@ -845,6 +845,31 @@ class FixedPrecisionTensor(AbstractTensor):
 
                 module.linear = linear
 
+                def dropout(input, p=0.5, training=True, inplace=False):
+                    """
+                    The dropout class calls functional dropout. Hence overloading functional dropout
+                    so that even works for dropout layer class.
+                    Ref:  https://stackoverflow.com/questions/54109617/implementing-dropout-from-scratch
+                    """
+                    if training:
+                        binomial = torch.distributions.binomial.Binomial(probs=1 - p)
+
+                        # we must convert the normal tensor to fixed precision before multiplication
+                        noise = (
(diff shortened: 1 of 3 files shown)


syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -845,6 +845,31 @@ def linear(*args):
...
+                    if training:
+                        binomial = torch.distributions.binomial.Binomial(probs=1 - p)
+
+                        # we must convert the normal tensor to fixed precision before multiplication
+                        noise = (
+                            (
+                                binomial.sample(input.shape).type(torch.FloatTensor)
+                                * (1.0 / (1.0 - p))
+                            )
+                            .fix_prec()
```
LaRiffle: Here put `.fix_prec(**input.get_class_attributes())`  to get the same field & precission  than te input otherwise you won't be able to multiply them in the more general cases

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -845,6 +845,31 @@ def linear(*args):
...
+                    Ref:  https://stackoverflow.com/questions/54109617/implementing-dropout-from-scratch
+                    """
+                    if training:
+                        binomial = torch.distributions.binomial.Binomial(probs=1 - p)
+
+                        # we must convert the normal tensor to fixed precision before multiplication
+                        noise = (
+                            (
+                                binomial.sample(input.shape).type(torch.FloatTensor)
+                                * (1.0 / (1.0 - p))
```
LaRiffle: Quick question here, shouldn't you multiply by `p` instead ?
I'm no dropout expert but I just read this: https://machinelearningmastery.com/dropout-for-regularizing-deep-neural-networks/

sukhadj:
> Don't forget to add a test:
> 
>     * at training you might just want to show the signal decrease
> 
>     * at testing just show the signal is unchanged

Done.
This test just checks if dropout has filtered out some inputs.
A more robust check would be comparing `(num_of_inputs_filtered / input_size)` to p. The first vale is close to p. So should we put a threshold on `((num_of_inputs_filtered / input_size) - p)` ? Thoughts on this?

syft/frameworks/torch/tensors/interpreters/precision.py
sukhadj: Done :+1:

syft/frameworks/torch/tensors/interpreters/precision.py
sukhadj: We are using inverted dropout which uses (1/(1-p)) to scale up the result.
Ref: https://www.coursera.org/lecture/deep-neural-network/dropout-regularization-eM33A

But would like to get it reviewed by someone :smiley:

LaRiffle: Excellent you're right!
https://pgaleone.eu/deep-learning/regularization/2017/01/10/anaysis-of-dropout/ confirms this

sachin-101:
@sukhadj @LaRiffle #2500 mentions implementing nn.Dropout class for PySyft.
https://github.com/OpenMined/PySyft/blob/0561a5d21c1787afeeb9753a0487f0c91d38a07d/syft/frameworks/torch/nn/rnn.py#L184
But this PR adds it to precision.py
In order to clean fixed_precision #2897 I will export dropout to nn in PR #3253 . @sukhadj  Are there any catchs I should be familiar of ?

sukhadj:
Nothing major!! Go ahead  :+1:

PR #3198, opened 2020-03-15
Standardize the names of tensor, plan, and worker command messages

karlhigley:
Depends on OpenMined/syft-proto#53.


karlhigley:
Yeah, I'm not so sure either. Possible that they could share an underlying implementation?

PR #3199, opened 2020-03-15
Improve Plans flexibility

### LINKED ISSUE #3184: Improve Plans flexibility

tudorcebere:
This PR fixes #3184 

Solved:
* added support when input tensor is also output tensor.
* added a test case when input tensor is also output tensor.

To do:
* check for the number of parameters
* create support for nested structures


karlhigley:
@tudorcebere Looks like this has a minor formatting issue! You can fix it by running `black .` in the root directory of the repo. Also highly recommend [installing the commit hooks](https://github.com/OpenMined/PySyft/blob/master/CONTRIBUTING.md#setting-up-pre-commit-hook), which will handle this for you in the future.

PR #3201, opened 2020-03-15
Extract a Role class from Plans

Jasopaum:


syft/execution/role.py
```diff
@@ -0,0 +1,387 @@
...
+            return None
+
+    def add_tensor_to_state(self, tensor):
+        placeholder = sy.PlaceHolder(id=tensor.id, owner=self.owner)
+        placeholder.instantiate(tensor)
+        self.state.state_placeholders.append(placeholder)
+        # TODO isn't it weird that state placeholders are both in state and plan?
+        self.placeholders[tensor.id] = placeholder
+
+    def fecth_placeholders_from_ids(self, obj):
```
iamtrask: typo fetch_placeholders_from_ids

Jasopaum: Thanks! Just fixed that

syft/generic/object.py
```diff
@@ -17,7 +17,7 @@ def __init__(
         self,
         id: int = None,
         owner: "sy.workers.AbstractWorker" = None,
-        tags: List[str] = None,
+        tags: Set[str] = None,
```
karlhigley: 💯

syft/execution/plan.py
```diff
@@ -582,42 +368,39 @@ def __str__(self):
...
+        #         f"_{extract_tag(arg)}" if isinstance(arg, PlaceHolder) else str(arg)
+        #         for arg in action.args
+        #     )
+        #     if action.kwargs:
+        #         line += ", " + ", ".join(f"{k}={w}" for k, w in action.kwargs.items())
+        #     line += ")\n"
+        #     out += line
+
+        # out += "    return "
+        # out += ", ".join(f"_{extract_tag(p)}" for p in self.find_placeholders("output"))
```
karlhigley: Can we remove this now? 👍

Jasopaum: I want to because I don't find it very useful to print a list of actions like that... What do you think?

PR #3205, opened 2020-03-16
Sigmoid aprox with tanh + fix sigmoid

### LINKED ISSUE #3131: Sigmoid in SMPC failing for some shapes
**Describe the bug**
 Examples to illustrate:

Working
```python
a = torch.Tensor([1,2,3])
a = a.fix_precision().share(bob, alice, crypto_provider = crypto_provider, requires_grad = True)
torch.nn.Sigmoid()(a)
```
Not working
```python
a = torch.Tensor([[1,2,3],[4,5,6]])
a = a.fix_precision().share(bob, alice, crypto_provider = crypto_provider, requires_grad = True)
torch.nn.Sigmoid()(a)
```
Error: `AssertionError: Must be batches of square matrices`

(same with `requires_grad=False`)

The error seems linked to tensor shape

gmuraru:
Fixes #3131 
Also adds the approximation for sigmoid using the tanh.

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/precision.py
+++ b/syft/frameworks/torch/tensors/interpreters/precision.py
@@ -449,6 +449,10 @@ class FixedPrecisionTensor(AbstractTensor):
     __matmul__ = matmul
     mm = matmul
 
+    def reciprocal(self):
+        ones = self * 0 + 1
+        return ones / self
+
     # Approximations:
     def inverse(self, iterations=8):
         """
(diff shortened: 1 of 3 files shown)


syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -498,12 +498,8 @@ def sigmoid(self, method="exp"):
...
 
         if method == "exp":
-            # Inverse can only be used on matrices
-            if len(self.shape) == 1:
-                one = self * 0 + 1
-                result = one / (1 + (self * -1).exp())
-            else:
-                result = (1 + (self * -1).exp()).inverse()
+            one = self * 0 + 1
+            result = one / (1 + (self * -1).exp())
```
LaRiffle: So basically you're suggesting .inverse() doesn't work ? :(
(The benefit of .inverse is that it's supposed to be faster)

LaRiffle: But yeah you're right it's failiing apparently sometimes, I was discussing of this two weeks ago on the call with André

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -560,6 +607,7 @@ def _tanh_chebyshev(tensor, maxval: int = 6, terms: int = 32):
           where c_i is the ith Chebyshev series coefficient and P_i is ith polynomial.
         The approximation is truncated to +/-1 outside [-maxval, maxval].
         Args:
+            tensor (tensor): values where the tanh needs to be approximated
```
LaRiffle: typo this should be here :)

test/torch/tensors/test_precision.py
```diff
@@ -449,6 +449,7 @@ def test_torch_sigmoid_approx(workers):
     alice, bob, james = workers["alice"], workers["bob"], workers["james"]
 
     fix_prec_tolerance_by_method = {
+        "chebyshev": {3: 6 / 100, 4: 1 / 100, 5: 1 / 100},
```
LaRiffle: To improve test speed could we remove the case precision=5 everywhere where it occurs? From what I see =4 is already compelling enough :)

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -482,45 +486,88 @@ def exp(self, iterations=8):
...
+    def sign(self):
+        return (self > 0) + (self < 0) * (-1)
+
+    @staticmethod
+    def _sigmoid_exp(tensor):
         """
-        Approximates the sigmoid function
+        Compute the sigmoid using the exp approximation
+        sigmoid(x) = 1 / (1 + exp(-x))
+        sigmoid(x) = (sigmoid(|x|) - 0.5) * sign(x) + 0.5
```
LaRiffle: Detail a bit the method for people which don't know about numerical stability etc

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -449,6 +449,10 @@ def matmul(self, *args, **kwargs):
     __matmul__ = matmul
     mm = matmul
 
+    def reciprocal(self):
+        ones = self * 0 + 1
+        return ones / self
```
LaRiffle: Let's keep this like this, until I come back to trying my approx inverse trick

syft/frameworks/torch/tensors/interpreters/precision.py
gmuraru: Hmm..this is for ```_tanh_chebyshev``` (forgot to put it in the previous PR)

PR #3220, opened 2020-03-18
changed "python" to system specific python version

MrMonk3y:
"python" is no longer hardcoded. the system version of python is used instead


karlhigley:
@MrMonk3y We found [a simpler way to do this](https://github.com/OpenMined/PySyft/pull/3225/files#diff-ca6bbfb74fe6c78d8a570b6f8d4256aeR21) in #3225.

PR #3221, opened 2020-03-18
included websockets-example-MNIST-parallel again

MrMonk3y:
should be tested due to some errors with **kwargs and woker.evaluate()


karlhigley:
Superseded by #3222.

PR #3222, opened 2020-03-18
included websockets-example-MNIST-parallel test again

MrMonk3y:
websockets-example-MNIST-parallel has some problems with **kwargs and worker.evaluate() which should be tested


test/notebooks/test_notebooks.py
```diff
@@ -70,7 +70,6 @@
 exclusion_list_folders = [
     "examples/tutorials/websocket",
     "examples/tutorials/advanced/monitor_network_traffic",
-    "examples/tutorials/advanced/websockets_mnist_parallel",
```
gmuraru: Q: Shouldn't be better to comment it? and leave a TODO?

karlhigley: Oh yeah, this is a confusing change but it’s correct. This removes it from the test exclusions so that it runs again (which is why the tests are failing on this PR.) Once the tests pass, it’s TODONE. 😄

PR #3223, opened 2020-03-18
Use `PlaceholderIds` in `Actions` and `Plans` (instead of full `Placeholders`)

Jasopaum:
Introduce a class that could be used as id instead of the current string | int.

The rationale behind that was to disambiguate id and string | int in messages and actions.
For instance, we could have only placeholder ids (as ObjectIds) in Plan's actions instead of the whole placeholders and we wouldn't take them for string or integers.


syft/execution/placeholder.py
```diff
@@ -97,14 +102,7 @@ def detail(worker: AbstractWorker, tensor_tuple: tuple) -> "PlaceHolder":
...
 
-        if not hasattr(worker, "_tmp_placeholders"):
-            worker._tmp_placeholders = {}
-
-        if tensor_id not in worker._tmp_placeholders:
-            tensor = PlaceHolder(owner=worker, id=tensor_id, tags=tags, description=description)
-            worker._tmp_placeholders[tensor_id] = tensor
-
-        return worker._tmp_placeholders[tensor_id]
+        return PlaceHolder(owner=worker, id=tensor_id, tags=tags, description=description)
```
LaRiffle: How do you make sure you don't create 2 PlaceHolder with the same tags?

syft/execution/plan.py
```diff
@@ -246,17 +247,19 @@ def add_placeholder(self, tensor, node_type=None):
...
 
-    def replace_with_placeholders(self, obj, **kw):
+    def replace_with_placeholder_ids(self, obj, **kw):
         """
-        Replace in an object all FrameworkTensors with Placeholders
+        Replace in an object all FrameworkTensors with Placeholder ids
         """
         if isinstance(obj, (tuple, list)):
-            r = [self.replace_with_placeholders(o, **kw) for o in obj]
+            r = [self.replace_with_placeholder_ids(o, **kw) for o in obj]
```
LaRiffle: I'm not sure I understand the strategy: you're now sotring ObjectIds instead of tags right? What are the benefits?

karlhigley: You remember the `self.worker._tmp_placeholders` variable that held placeholders during deserialization in order to avoid creating multiple identical `Placeholders`, because duplicate `Placeholders` breaks instantiation with actual tensors? This is a strategy for getting rid of that, by storing `Placeholders` at the `Plan` level and using that store to avoid duplication.

karlhigley: Using ids here makes it possible to deserialize without creating duplicate `Placeholders` and then index into the `Placeholder` store on the `Plan` to make sure that all the places that reference the same id use the same `Placeholder` object.

syft/execution/placeholder.py
Jasopaum: I don't, but I don't think it was done before neither

karlhigley: The check to avoid duplicate placeholders moves after the deserialization. Since `Actions` no longer directly contain `Placeholders` (which move to a field on `Plan`), we don't have to worry about deserializing the same `Placeholder` multiple times as long as the property on `Plan` doesn't contain duplicates.

syft/execution/plan.py
```diff
@@ -214,7 +215,7 @@ def add_placeholder(self, tensor, node_type=None):
             tensor: the tensor to replace with a placeholder
             node_type: Should be "input" or "output", used to tag like this: #<type>-*
         """
-        if tensor.id not in self.placeholders.keys():
+        if tensor.id not in self.placeholders:
```
karlhigley: I think we still want `self.placeholders` indexed by integers and pointing to `Placeholder` objects.

syft/execution/plan.py
```diff
@@ -684,6 +704,13 @@ def detail(worker: AbstractWorker, plan_tuple: tuple) -> "Plan":
         plan.tags = sy.serde.msgpack.serde._detail(worker, tags)
         plan.description = sy.serde.msgpack.serde._detail(worker, description)
 
+        # Replace non-instanciated placeholders from plan.placeholders by instanciated placeholders
+        # from state.state_placeholders
+        # TODO this definitely isn't the right strategy. Maybe state shouldn't contain
+        # instanciated placeholders but values directly?
```
karlhigley: I'm kinda confused by this part. Not sure I have my head fully wrapped around why State contains Placeholders. 🤔

karlhigley: (Then we'd be able to fetch the right `Placeholder` using the value contained in an `ObjectId`.)

syft/execution/plan.py
karlhigley: (Then we'd be able to fetch the right Placeholder using the value contained in an ObjectId.)

Jasopaum: Yep, I didn't change anything here. I just removed the `.keys()` because it's not needed: when you check if `x in dict`, Python will look at the keys of `dict`. But I can put it back if it's clearer

syft/execution/plan.py
Jasopaum: I asked Théo who told me that it was to avoid tensor-placeholder operations.
If state contains placeholders, then we can do operation between the plan's placeholders and the ones in its state.
But I agree, Placeholders shouldn't have anything to do with States as they're static

syft/execution/plan.py
karlhigley: Ah, makes sense.

syft/execution/plan.py
```diff
@@ -274,6 +275,20 @@ def replace_with_placeholders(self, obj, arg_ids, result_ids, **kw):
             else:
                 return None
 
+    def replace_ids_with_placeholders(self, obj):
+        """
+        Replace in an object all ids with Placeholders
```
karlhigley: This only applies to `ObjectIds`/`PlaceholderIds` now, right?

syft/execution/plan.py
karlhigley: I think I'm okay with this strategy for now. Based on Slack conversations, sounds like we should leave `State` containing instantiated `Placeholders` for the time being.

syft/execution/plan.py
```diff
@@ -775,6 +800,13 @@ def unbufferize(worker: AbstractWorker, protobuf_plan: PlanPB) -> "Plan":
         if protobuf_plan.description:
             plan.description = protobuf_plan.description
 
+        # Replace non-instanciated placeholders from plan.placeholders by instanciated placeholders
+        # from state.state_placeholders
+        # TODO this definitely isn't the right strategy. Maybe state shouldn't contain
+        # instanciated placeholders but values directly?
```
karlhigley: I think the approach is okay, but wonder if this makes sense to extract as a method since it happens in multiple places.

syft/execution/state.py
```diff
@@ -64,7 +64,9 @@ def read(self):
                 # for all the placeholders in this sub plan, we report a copy of them
                 # in the parent plan and notify their origin using the #inner tag
                 for placeholder in self.state_placeholders:
+                    ex_id = placeholder.id
                     placeholder = placeholder.copy()
+                    placeholder.id = ex_id
```
karlhigley: Does `copy` change the ID of the original `Placeholder`? Weird!

syft/serde/protobuf/proto.py
```diff
@@ -56,6 +57,7 @@
     torch.Size: SizePB,
     # Syft types
     AdditiveSharingTensor: AdditiveSharingTensorPB,
+    ObjectId: IdPB,  # TODO harmonize naming between syft and proto
```
karlhigley: One way to do this without breaking everything would be to create an `ObjectId`/`PlaceholderId` schema on the Protobuf side that contains an `Id` field.

syft/execution/computation.py
```diff
@@ -130,15 +133,17 @@ def bufferize(worker: AbstractWorker, action: "ComputationAction") -> "Computati
...
             else:
                 return_ids = action.return_ids
 
             for return_id in return_ids:
-                if type(return_id) == PlaceHolder:
-                    protobuf_op.return_placeholders.extend(
-                        [sy.serde.protobuf.serde._bufferize(worker, return_id)]
+                if isinstance(return_id, ObjectId):
+                    # TODO hack to know when we have an ObjectId, we store it
+                    # return_placeholders and not in return_ids
```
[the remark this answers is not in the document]
Jasopaum: It was to know that I need to deserialize what's inside as an ObjectId and not as a string | int as this how classic ids are deserialized.
But as I only use ObjectIds for Plans and Placeholders, right now, I think I can remove that and do the extra logic in Plan deserialization

syft/execution/placeholder.py
```diff
@@ -20,6 +21,10 @@ def __init__(self, owner=None, id=None, tags: set = None, description: str = Non
         """
         super().__init__(id=id, owner=owner, tags=tags, description=description)
 
+        # TODO currently we only use ObjectId on placeholder but if we use them, we
+        # should use them everywhere
```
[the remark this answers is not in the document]
Jasopaum: I was hoping for a quick extension to all objects because I don't want to have Placeholder as a different object... But I can change the naming for now and change it back when we extend it to other objects.
What do you think?

syft/execution/plan.py
Jasopaum: Yep! All objects in Plan actions should be placeholders (except constants) and they are all encoded with an ObjectId

syft/execution/state.py
Jasopaum: It does, maybe because we don't want to have 2 objects with the same id on the same worker

syft/serde/protobuf/proto.py
Jasopaum: How bad would it be to rename protobuf's `Id` to `ObjectId`/`PlaceholderId`?

syft/execution/computation.py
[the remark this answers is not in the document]
karlhigley: Oh, that makes sense. We can leave this for a later PR to sort out.

syft/execution/computation.py
```diff
@@ -188,14 +193,7 @@ def unbufferize(
...
         if return_placeholders:
-            if len(return_placeholders) == 1:
-                action = ComputationAction(
-                    command, target, tuple(args), kwargs, return_placeholders[0]
-                )
-            else:
-                action = ComputationAction(
-                    command, target, tuple(args), kwargs, return_placeholders
-                )
+            action = ComputationAction(command, target, tuple(args), kwargs, return_placeholders)
```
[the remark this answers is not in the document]
karlhigley: Nevermind!

syft/execution/placeholder.py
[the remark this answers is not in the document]
karlhigley: I think it's still beneficial to know what type the id goes with somehow. Using a separate class allows us to put that information into type annotations, which is kinda nice. On the other hand, a proliferation of tiny id classes might not be great. 🤷‍♂ 

Could there be a base `Id` class that defines the fields/behaviors and then a bunch of sub-classes that only exist to capture/enforce id types?

syft/execution/state.py
karlhigley: Ah, right.

syft/serde/protobuf/proto.py
karlhigley: Kinda bad? It's not really intended for either of those specific purposes.

syft/execution/computation.py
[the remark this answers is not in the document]
Jasopaum: Hum, finally, it was indeed a trick to have somewhere to put ObjectIds and somewhere else to put classic ids

syft/serde/protobuf/proto.py
Jasopaum: Even if ObjectId was used everywhere?

karlhigley: If we were going to do that, we'd probably want to rename the Python class to `Id`, because I'm not sure everything that uses an `Id` is actually using it as the id for an object.

karlhigley: (Still seems like we might want a way to immediately tell what type an id goes with from the class though.)

syft/execution/computation.py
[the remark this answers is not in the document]
Jasopaum: How would you sort it out?

syft/execution/placeholder.py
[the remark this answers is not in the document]
Jasopaum: Hum, I don't think `Ids` will have lots of functionalities so I'm not a fan of having a Id class for each PySyft object...

syft/serde/protobuf/proto.py
Jasopaum: I don't really get why we would need to know which type the goes to 🤔

karlhigley: Two reasons I can think of:
- It's handy for serialization/deserialization.
- It prevents people from misusing the ids from one class to load objects of another class (to prevent weird collisions of both accidental and intentional natures.)

syft/execution/placeholder.py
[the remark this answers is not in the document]
karlhigley: I think we could kick this can down the road by calling it `PlaceholderId` for now. Since a one-to-one correspondence is required between Python classes and Protobuf schemas, that would require adding a schema for `PlaceholderId` though.

It's not perfect and we may just refactor it out later, but it keeps the scope of this PR limited and make it easier to think through the potential implications.

syft/serde/protobuf/proto.py
Jasopaum: - Do you think serialization/deserialization could differ between different id classes?
- Hum, I agree on this point

karlhigley: Depends on whether they're represented by separate classes, I suppose. If they are, then seems like the serialization could be different.

syft/execution/computation.py
```diff
@@ -130,15 +134,17 @@ def bufferize(worker: AbstractWorker, action: "ComputationAction") -> "Computati
...
             else:
                 return_ids = action.return_ids
 
             for return_id in return_ids:
-                if type(return_id) == PlaceHolder:
-                    protobuf_op.return_placeholders.extend(
-                        [sy.serde.protobuf.serde._bufferize(worker, return_id)]
+                if isinstance(return_id, PlaceholderId):
+                    # NOTE to know when we have a PlaceholderId, we store it
+                    # in return_placeholder_ids and not in return_ids
```
karlhigley: Is this still true? Line 197 below looks like it's using the `return_ids` field.

syft/execution/placeholder_id.py
```diff
@@ -0,0 +1,42 @@
+import syft as sy
+from syft.workers.abstract import AbstractWorker
+
+from syft_proto.execution.v1.placeholder_id_pb2 import PlaceholderId as PlaceholderIdPB
+
+
+class PlaceholderId:
+    """ ObjectIds are used to uniquely identify PySyft objects.
+    """
```
karlhigley:
```suggestion
    """
    PlaceholderIds are used to identify which Placeholder tensors should be used
    as the inputs and outputs of Actions.
    """
```

syft/execution/plan.py
```diff
@@ -70,7 +70,7 @@ def method2plan(*args, **kwargs):
     )
 
 
-class Plan(AbstractObject, ObjectStorage):
+class Plan(AbstractObject):
```
karlhigley: Holy cow, didn't realize this PR would remove `ObjectStorage` as a parent of `Plan`. 👍

syft/serde/protobuf/proto.py
```diff
@@ -34,7 +35,7 @@
 from syft_proto.generic.pointers.v1.pointer_tensor_pb2 import PointerTensor as PointerTensorPB
 from syft_proto.messaging.v1.message_pb2 import ObjectMessage as ObjectMessagePB
 from syft_proto.messaging.v1.message_pb2 import TensorCommandMessage as CommandMessagePB
-from syft_proto.types.syft.v1.id_pb2 import Id as IdPB
```
karlhigley: I suspect we may still need to import `Id` here too.

test/serde/serde_helpers.py
```diff
@@ -644,6 +645,25 @@ def compare(detailed, original):
     ]
 
 
+# syft.execution.placeholder_id.PlaceholderId
+def make_objectid(**kwargs):
```
karlhigley:
```suggestion
def make_placeholderid(**kwargs):
```

test/serde/protobuf/test_protobuf_serde_full.py
```diff
@@ -33,9 +33,10 @@
 samples[syft.execution.computation.ComputationAction] = make_computation_action
 samples[syft.execution.communication.CommunicationAction] = make_communication_action
 samples[syft.execution.plan.Plan] = make_plan
-samples[syft.generic.pointers.pointer_tensor.PointerTensor] = make_pointertensor
 samples[syft.execution.protocol.Protocol] = make_protocol
 samples[syft.execution.state.State] = make_state
+samples[syft.execution.placeholder_id.PlaceholderId] = make_objectid
```
karlhigley:
```suggestion
samples[syft.execution.placeholder_id.PlaceholderId] = make_placeholderid
```

test/serde/msgpack/test_msgpack_serde_full.py
```diff
@@ -54,17 +54,18 @@
     syft.frameworks.torch.tensors.interpreters.crt_precision.CRTPrecisionTensor
 ] = make_crtprecisiontensor
 samples[syft.frameworks.torch.tensors.decorators.logging.LoggingTensor] = make_loggingtensor
-samples[syft.generic.pointers.multi_pointer.MultiPointerTensor] = make_multipointertensor
 samples[syft.execution.plan.Plan] = make_plan
 samples[syft.execution.state.State] = make_state
 samples[syft.execution.computation.ComputationAction] = make_computation_action
 samples[syft.execution.communication.CommunicationAction] = make_communication_action
 samples[syft.execution.protocol.Protocol] = make_protocol
+samples[syft.execution.placeholder_id.PlaceholderId] = make_objectid
```
karlhigley:
```suggestion
samples[syft.execution.placeholder_id.PlaceholderId] = make_placeholderid
```

syft/execution/computation.py
Jasopaum: Yep, but in the protobuf schema, I use the `return_placeholder_ids` field for placeholderId and the `return_ids` field for classic ids.
Maybe I don't need to do so?

syft/execution/plan.py
Jasopaum: This could have been removed before but it was still here somehow. Anyway, it's better now :)

syft/serde/protobuf/proto.py
Jasopaum: It's not used anywhere... Why would we need to import it?

karlhigley: Whoa, didn't realize this was the only place where `IdPB` was used so far. It will definitely be used more in the future, but we can put it back when we need it.

Jasopaum: It was imported but not even used...
But we definitely need to use that!

syft/execution/computation.py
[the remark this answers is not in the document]
karlhigley: I'm not sure! Seems like a good reason to postpone dealing with it. 😅

PR #3227, opened 2020-03-19
Plans support for typechecking, lists and dicts

### LINKED ISSUE #3184: Improve Plans flexibility

tudorcebere:
# Description

This PR fixes #3184. 

New features:

- Plans now have support for type-checking at runtime based on the type provided at build time.
- Plans now support usage of lists, tuples and dicts as inputs.
- Plans now have a number of arguments check, build time and call time should have the same number of arguments.

# Type of change:
-  Bug fix
- New feature

# How Has This Been Tested?

Wrote a few tests that verify that:

- type-checking works, it should raise a RunTimeWarning if the types differ.
- argument number check works, build time and call time should not differ, if they differ, a RunTimeError should occur.
- lists of tensors work as input.
- dicts work as input. 
- a nested structure containing lists and dicts works as input.

# TODO:
- add support for variables when building plans. Not sure if this behaviour is wanted, I will open an issue with this.
- add support for lists, dicts and tuples as output.
- make a better scheme on dicts as inputs, as they are unordered.

### DIFF
--- a/syft/execution/plan.py
+++ b/syft/execution/plan.py
@@ -20,7 +20,6 @@ from syft.workers.abstract import AbstractWorker
 from syft.execution.placeholder import PlaceHolder
 
 from syft_proto.execution.v1.plan_pb2 import Plan as PlanPB
-from syft_proto.execution.v1.computation_action_pb2 import ComputationAction as ComputationActionPB
 
 
 class func2plan(object):
@@ -233,17 +232,17 @@ class Plan(AbstractObject, ObjectStorage):
                         f"Please use instead torch.tensor(..., dtype=torch.int32) for example."
                     )
 
-                placeholder.tags.add(f"#input-{self._tmp_args_ids.index(tensor.id)}")
-                if tensor.id in self._tmp_result_ids:
-                    placeholder.tags.add(f"#output-{self._tmp_result_ids.index(tensor.id)}")
+                placeholder.tags.add(f"#input-{arg_ids.index(tensor.id)}")
+                if tensor.id in result_ids:
+                    placeholder.tags.add(f"#output-{result_ids.index(tensor.id)}")
 
                 placeholder.tags.add(f"#type_FrameworkTensor")
             elif node_type == "output":
-                if tensor.id in self._tmp_result_ids:
-                    placeholder.tags.add(f"#output-{self._tmp_result_ids.index(tensor.id)}")
+                if tensor.id in result_ids:
+                    placeholder.tags.add(f"#output-{result_ids.index(tensor.id)}")
 
-                if tensor.id in self._tmp_args_ids:
-                    placeholder.tags.add(f"#input-{self._tmp_args_ids.index(tensor.id)}")
+                if tensor.id in arg_ids:
+                    placeholder.tags.add(f"#input-{result_ids.index(tensor.id)}")
             else:
                 raise ValueError("node_type should be 'input' or 'output'.")
 
@@ -251,7 +250,7 @@ class Plan(AbstractObject, ObjectStorage):
 
         return self.placeholders[tensor.id]
 
-    def add_placeholder_fcall(self, obj, node_type=None):
+    def add_placeholder_fcall(self, obj, arg_ids, node_type=None):
         """
         Create a placeholder for the arguments, even if they are not tensors. These placeholders are used for
         type check at the runtime of the plan. Every placeholder that is created or is already existing will
@@ -275,15 +274,15 @@ class Plan(AbstractObject, ObjectStorage):
             self.placeholders[obj_id] = placeholder
 
             if node_type == "fcall":
-                if obj_id in self._tmp_args_ids:
-                    placeholder.tags.add(f"#fcall-{self._tmp_args_ids.index(obj_id)}")
+                if obj_id in arg_ids:
+                    placeholder.tags.add(f"#fcall-{arg_ids.index(obj_id)}")
                     type_tag = "type_" + type(obj).__name__
                     placeholder.tags.add(f"#{type_tag}")
             else:
                 raise ValueError("node_type should be 'fcall'")
             self.var_count += 1
         else:
-            self.placeholders[obj_id].tags.add(f"#fcall-{self._tmp_args_ids.index(obj_id)}")
+            self.placeholders[obj_id].tags.add(f"#fcall-{arg_ids.index(obj_id)}")
 
     def replace_with_placeholders(self, obj, arg_ids, result_ids, **kw):
         """
@@ -349,9 +348,9 @@ class Plan(AbstractObject, ObjectStorage):
             args: Input arguments to run the plan
         """
 
-        def generate_input_ids(plan, args, fcall_argvs=False):
+        def generate_input_ids(args, arg_ids, fcall_argvs=False):
             """
-            Add all input tensors to _tmp_args_ids, even if they are in nested structures.
+            Add all input tensors to arg_ids, even if they are in nested structures.
 
             Note: Only supports dict, list and tuples, might be better to check if the object is iterable for
             better support.
@@ -359,26 +358,25 @@ class Plan(AbstractObject, ObjectStorage):
             Arguments:
                 plan: the current plan.
                 args: arguments of the plan at build time.
-                fcall_argvs: we want the original set of inputs of the build to be added to _tmp_args_ids, even if
+                fcall_argvs: we want the original set of inputs of the build to be added to arg_ids, even if
                 they are not tensors, useful for the type check.
             """
             for arg in args:
                 if isinstance(arg, FrameworkTensor):
-                    plan._tmp_args_ids.append(arg.id)
+                    arg_ids.append(arg.id)
                     continue
 
                 if isinstance(arg, dict):
-                    generate_input_ids(plan, list(arg.values()))
+                    generate_input_ids(list(arg.values()), arg_ids)
                 elif isinstance(arg, list):
-                    generate_input_ids(plan, arg)
+                    generate_input_ids(arg, arg_ids)
 
                 if fcall_argvs:
-                    plan._tmp_args_ids.append(id(arg))
+                    arg_ids.append(id(arg))
 
         self.owner.init_plan = self
-        self._tmp_args_ids = []
-        generate_input_ids(self, args, fcall_argvs=True)
-
+        arg_ids = []
+        generate_input_ids(args, arg_ids, fcall_argvs=True)
         with sy.hook.trace.enabled():
             # We usually have include_state==True for functions converted to plan
             # using @func2plan and we need therefore to add the state manually
@@ -389,7 +387,6 @@ class Plan(AbstractObject, ObjectStorage):
 
         results = (results,) if not isinstance(results, tuple) else results
 
-        arg_ids = [t.id for t in args if isinstance(t, FrameworkTensor)]
         result_ids = [t.id for t in results if isinstance(t, FrameworkTensor)]
 
         for arg in args:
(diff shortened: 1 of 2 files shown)


syft/execution/plan.py
```diff
@@ -384,6 +450,48 @@ def __setattr__(self, name, value):
...
+            kwargs: original kwargs received on __call__.
+        """
+
+        def extract_type(obj: PlaceHolder) -> str:
+            """
+            Function to extract type on the
+            """
+            for tag in obj.tags:
+                if "type_" in tag:
+                    return tag[6:]
```
LaRiffle: Should be 5 no ?

syft/execution/plan.py
```diff
@@ -324,12 +387,16 @@ def build(self, *args):
...
 
-        arg_ids = [t.id for t in args if isinstance(t, FrameworkTensor)]
         result_ids = [t.id for t in results if isinstance(t, FrameworkTensor)]
 
         for arg in args:
             self.replace_with_placeholders(arg, arg_ids, result_ids, node_type="input")
 
+        for arg in args:
+            # adding the input tag to the tensors, even if nested, adding fcall tag to every argument
+            self.replace_with_placeholders(arg, arg_ids, result_ids, node_type="input")
```
LaRiffle: This is duplictae with l 393 I think:
`self.replace_with_placeholders(arg, arg_ids, result_ids, node_type="input")`

syft/execution/plan.py
```diff
@@ -248,6 +250,40 @@ def add_placeholder(self, tensor, arg_ids, result_ids, node_type=None):
...
+        The placeholders created will use the same unique incremental index.
+
+        Args:
+            obj: The object from the arguments of the build.
+            node_type: Only accepts "fcall".
+        """
+        if isinstance(obj, FrameworkTensor):
+            obj_id = obj.id
+        else:
+            obj_id = id(obj)
```
LaRiffle: Here you can use our IdProvider: `sy.ID_PROVIDER.pop()`

syft/execution/plan.py
```diff
@@ -248,6 +250,40 @@ def add_placeholder(self, tensor, arg_ids, result_ids, node_type=None):
 
         return self.placeholders[tensor.id]
 
+    def add_placeholder_fcall(self, obj, arg_ids, node_type=None):
+        """
```
LaRiffle: I'm not sure I understand the role of "fcall", what is it here for?

syft/execution/plan.py
```diff
@@ -384,6 +450,48 @@ def __setattr__(self, name, value):
...
+            raise RuntimeError(
+                f"Plan {self.name} requires {len(inputs)} arguments, received {len(args)}."
+            )
+
+        for index, arg in enumerate(args):
+            placeholder_type = extract_type(inputs[index])
+            argument_type = type(arg).__name__
+
+            tensor_type_error = (
+                isinstance(arg, FrameworkTensor) == placeholder_type == "FrameworkTensor"
```
LaRiffle: isinstance returns a bool, how could it be == to a "FrameworkTensor" str?

syft/execution/plan.py
```diff
@@ -396,41 +504,67 @@ def __call__(self, *args, **kwargs):
...
+            """
+
+            for arg in args:
+                if isinstance(arg, FrameworkTensor):
+                    placeholder = input_placeholders.pop(0)
+                    placeholder.instantiate(arg)
+                elif isinstance(arg, (list, tuple, set)):
+                    instantiate_input(input_placeholders, arg)
+                elif isinstance(arg, dict):
+                    instantiate_input(input_placeholders, list(arg.values()))
```
LaRiffle: for dict, is there a risk that the value() don't  come in the expected order? (dict is not ordered)

syft/execution/plan.py
```diff
@@ -396,41 +504,67 @@ def __call__(self, *args, **kwargs):
...
+        # Instantiate all the input placeholders in the correct order
+        input_placeholders = sorted(self.find_placeholders("#input"), key=tag_sort("input"))
 
-            # This ensures that we return the output placeholder in the correct order
-            output_placeholders = sorted(self.find_placeholders("#output"), key=tag_sort("output"))
-            response = [p.child for p in output_placeholders]
+        if self.forward is not None:
+            if self.include_state:
+                args = (*args, self.state)
+            return self.forward(*args)
```
LaRiffle: I have a slight preference here for the if/else structure to better understand the branching logic

test/execution/test_plan.py
```diff
@@ -1125,4 +1125,158 @@ def plan_test_2(x, y):
...
+
+    dummy_input_list = [th.tensor([1, -2]), th.tensor([1, 2]), 2, False]
+
+    wrong_type_inputs = [
+        [pointer_to_data_1, pointer_to_data_2, 2, 2],
+        [pointer_to_data_1, pointer_to_data_2, 2.0, False],
+    ]
+
+    for test_number, wrong_type_input in enumerate(wrong_type_inputs):
+        plan_test.build(*dummy_input_list)
```
LaRiffle: Can you build a plan twice? (This is a question, I've never tried it!)

test/execution/test_plan.py
```diff
@@ -1125,4 +1125,158 @@ def plan_test_2(x, y):
...
+        pointer_plan = plan_test.send(device_1)
+        with pytest.warns(RuntimeWarning) as e:
+            pointer_plan(*wrong_type_input)
```
LaRiffle: Shouldn't you also send your arguments to the same location before giving them to pointer_plan?

Instead of sending plan_test, the trick is to set plan_test.forward = None to make sure it will use the built version :)

LaRiffle: oh sorry I had'nt seen the : ` device_1 = sy.VirtualWorker(hook, id="test_dev_1", data=(x11, x12))`  that's nice!

But maybe use my trick to avoid extra potential bugs when using pointerplans

syft/execution/plan.py
tudorcebere: The actual form of the type is #type_<int>, respecting the convention from the other tags.

syft/execution/plan.py
tudorcebere: I will write better documentation in the code in the next commit. The fcall tag is used to denote the placeholders that were originally in the input. For example if I build a plan with the following arguments:

- th.tensor([1,2])
- [th.tensor([4,5]), th.tensor([6, 7])]
- True

I want to add Placeholders with the #fcall tag (denoting that those placeholders were in the input of the function) and with their types (from build time): FrameworkTensor, list, bool.  The original implementation would not flag the fact that those Placeholders were given as input at build time, they would leave only the #input tag (could not easily deduce what placeholders were used in input function) and the nontensor inputs were not receiving a placeholder, making typing impossible for those specific inputs (nested structures/variables in the future :) ).

syft/execution/plan.py
```diff
@@ -384,6 +450,48 @@ def __setattr__(self, name, value):
...
                 self.__setattr__(f"{name}_{tensor_name}", tensor)
 
+    def validate_input(self, args, kwargs):
+        """
+        Function used in __call__ to validate the input based on the input received at build time.
+
+        Args:
+            args: original args received on __call__.
+            kwargs: original kwargs received on __call__.
+        """
```
Jasopaum: If I understand well, you're tagging the placeholder with a custom string to tell what is the "build time" type of this placeholder and then check that the runtime type is the same as what's written in the tag.
Wouldn't it be neater to add this information in the placeholder (as an attribute for instance) instead?

syft/execution/plan.py
Jasopaum: Not sure I followed everything.
Is the function you're talking about the plan?
Do we have now both #input and #fcall tags? Why having both?
Why did you call it "fcall"?

tudorcebere: Yes. We want to make plans closer to how actual functions work (without them being actual functions). Let's say we want to build a plan with an underlying function that receives two parameters.

x:  (and the build input) th.tensor([1,2])
y: (and the build input) [th.tensor([4,5]), th.tensor([6, 7])]

The generated placeholders:

- placeholder for th.tensor([1, 2]) will have the following tags: input, fcall, type.
- placeholder for [th.tensor([4, 5]), th.tensor([6, 7])] will have the following tags: fcall, type. (this is not an input tensor)
- placeholder for th.tensor([4, 5]) will have the following tags: input
- placeholder for th.tensor([6, 7]) will have the following tags: input

The fcall tag tells me that this placeholder is a placeholder for a build time argument (even if not FrameworkTensor). This is useful when doing the type check/arguments verification. In the old implementation, the list from the arguments would have not received a Placeholder, making type checking impossible.

The naming might not be the greatest, I'll document this better and think for a better name (suggestions would be great)?

syft/execution/plan.py
tudorcebere: My bad, this should work like a xor between isinstance(arg, FrameworkTensor) and  placeholder_type == "FrameworkTensor". Forgot a pair of parentheses, will fix in the next commit.

syft/execution/plan.py
tudorcebere: Nice observation, I'll add this to the TODO list and I will think about this, currently this is might be bug prone.

syft/execution/plan.py
tudorcebere: It might be, you are right, I am not sure. I taught that this might not be wanted (to add a few new attributes), as there would be just a few placeholders that would actually use this.

tudorcebere: Another problem is that I still need a mechanism to flag nontensors inputs at build time, like lists for example, I considered that this would add some unwanted complexity to the Placeholder class.

Jasopaum: Maybe we can discuss about this with other people but I think it would be better to have this in the Placeholder class. This way, we could also have some kind of `validate_type` on instantiation and avoid the comparison of substrings (which I really don't like).

Jasopaum: @OpenMined/syft-core-team?

Jasopaum: > Another problem is that I still need a mechanism to flag nontensors inputs at build time, like lists for example, I considered that this would add some unwanted complexity to the Placeholder class.

I'll need to read what you do with non-tensor inputs in more detail but this should go in the Placeholder logic, you're right

test/execution/test_plan.py
tudorcebere: I have to test this, I assumed it works. Could it be a useful thing?

syft/execution/plan.py
tudorcebere: Yes. We want to make plans closer to how actual functions work (without them being actual functions). Let's say we want to build a plan with an underlying function that receives two parameters.

x: (and the build input) th.tensor([1,2])
y: (and the build input) [th.tensor([4,5]), th.tensor([6, 7])]

The generated placeholders:

placeholder for th.tensor([1, 2]) will have the following tags: input, fcall, type.
placeholder for [th.tensor([4, 5]), th.tensor([6, 7])] will have the following tags: fcall, type_list. (this is not an input tensor)
placeholder for th.tensor([4, 5]) will have the following tags: input
placeholder for th.tensor([6, 7]) will have the following tags: input
The fcall tag tells me that this placeholder is a placeholder for a build time argument (even if not FrameworkTensor). This is useful when doing the type check/arguments verification. In the old implementation, the list from the arguments would have not received a Placeholder, making type checking impossible.

The naming might not be the greatest, I'll document this better and think for a better name (suggestions would be great).

tudorcebere: Nontensor inputs would require a placeholders as well (for the typecheck), but if we create two attributes in Placeholder (something like known_type: str and build_argument: bool), the problem would be solved without tags.

karlhigley: Yeah, let's preferentially use new attributes over string comparisons. 👍

tudorcebere:
I will close this PR and open a new one: #3296 with the serialized input approach.

Issue #3212, opened 2020-03-17
Enhance encryption API


LaRiffle:
Two features here:

- Make `crypto_provider` kwargs mandatory in `.share()`and check that it's not a client_worker
- Modify `.encrypt()` to add a kwarg and be used not only for Paillier encryption but only MPC (default should be MPC,  and it should call for mpc .fix_prec(...).share(...)  under the hood)

nin-ed:
I can try the first part.

AshwinRam2000:
I can take this up

PR #3234, opened 2020-03-21
Add a `protocol` kwarg to `encrypt()` method to select between MPC and Paillier HE

Syzygianinfern0:
Fixes [#3212](https://github.com/OpenMined/PySyft/issues/3212)

> Modify .encrypt() to add a kwarg and be used not only for Paillier encryption but only MPC (default should be MPC, and it should call for mpc .fix_prec(...).share(...) under the hood)

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/native.py
+++ b/syft/frameworks/torch/tensors/interpreters/native.py
@@ -992,14 +992,30 @@ class TorchTensor(AbstractTensor):
         else:
             return self.child.torch_type()
 
-    def encrypt(self, *args, **kwargs):
+    def encrypt(self, encryption_method='mpc', **kwargs):
         """This method will encrypt each value in the tensor using Multi Party
-        Computation (default) or Paillier homomorphic encryption.
-        :param kwargs:
-        :return:
+        Computation (default) or Paillier homomorphic encryption
         """
-        x_shared = self.fix_prec().share(*args, **kwargs)
-        return x_shared
+        if encryption_method.lower() == 'mpc':
+            args_fix_prec = kwargs.pop('args_fix_prec')
+            workers = kwargs.pop('workers')
+            crypto_provider = kwargs.pop('crypto_provider')
+
+            if args_fix_prec:
+                x_shared = self.fix_prec(**args_fix_prec).share(*workers, crypto_provider=crypto_provider, **kwargs)
+            else:
+                x_shared = self.fix_prec().share(*workers, crypto_provider=crypto_provider, **kwargs)
+            return x_shared
+        elif encryption_method.lower() == 'paillier':
+            x = self.copy()
+            x_encrypted = PaillierTensor().on(x)
+            x_encrypted.child.encrypt_(kwargs['public_key'])
+            return x_encrypted
+        else:
+            raise NotImplementedError(
+                "Currently the .encrypt() method only supports Paillier Homomorphic "
+                "Encryption and Secure Multi-Party Computation"
+            )
 
     def decrypt(self, private_key):
         """This method will decrypt each value in the tensor, returning a normal
--- a/syft/frameworks/torch/tensors/interpreters/native.py
+++ b/syft/frameworks/torch/tensors/interpreters/native.py
@@ -992,20 +992,20 @@ class TorchTensor(AbstractTensor):
         else:
             return self.child.torch_type()
 
-    def encrypt(self, encryption_method="mpc", **kwargs):
+    def encrypt(self, protocol="mpc", **kwargs):
         """
         This method will encrypt each value in the tensor using Multi Party
         Computation (default) or Paillier Homomorphic Encryption
 
         Args:
-            encryption_method (str): Currently supports 'mpc' for Multi Party
+            protocol (str): Currently supports 'mpc' for Multi Party
                 Computation and 'paillier' for Paillier Homomorphic Encryption
             **kwargs:
                 With Respect to MPC accepts:
                     workers (list): Parties involved in the sharing of the Tensor
                     crypto_provider (syft.VirtualWorker): Worker responsible for the
                         generation of the random numbers for encryption
-                    args_fix_prec (dict): To be parsed as a kwarg for the .fix_prec() method
+                    Keyword Args: To be parsed as kwargs for the .fix_prec() method
 
                 With Respect to Paillier accepts:
                     public_key (phe.paillier.PaillierPublicKey): Can be obtained using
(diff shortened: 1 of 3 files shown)


imskr:
Please always use other than your master branch for creating PR

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -992,19 +992,57 @@ def torch_type(self):
...
-            *public_key a public key created using
-                syft.frameworks.torch.he.paillier.keygen()
+            encryption_method (str): Currently supports 'mpc' for Multi Party
+                Computation and 'paillier' for Paillier Homomorphic Encryption
+            **kwargs:
+                With Respect to MPC accepts:
+                    workers (list): Parties involved in the sharing of the Tensor
+                    crypto_provider (syft.VirtualWorker): Worker responsible for the
+                        generation of the random numbers for encryption
+                    args_fix_prec (dict): To be parsed as a kwarg for the .fix_prec() method
```
LaRiffle: Could you also allow fix_precision kwargs to be given alongside the share ones (workers & crypto_provider)? Would be simpler for the end user :)

LaRiffle: (I think we would provide kwargs and not args for fix_precision)

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -992,19 +992,57 @@ def torch_type(self):
...
+                        ```public_key, private_key = sy.frameworks.torch.he.paillier.keygen()```
+        Returns:
+            An encrypted version of the Tensor following the protocol specified
+
+        Raises:
+            NotImplementedError: If protocols other than the ones mentioned above are queried
+
         """
+        if encryption_method.lower() == "mpc":
+            args_fix_prec = kwargs.pop("args_fix_prec")
```
LaRiffle: Use .get() when the parameter is not compulsory (like in fix prec)

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -992,19 +992,57 @@ def torch_type(self):
...
+            workers = kwargs.pop("workers")
+            crypto_provider = kwargs.pop("crypto_provider")
+
+            if args_fix_prec:
+                x_shared = self.fix_prec(**args_fix_prec).share(
+                    *workers, crypto_provider=crypto_provider, **kwargs
```
LaRiffle: with my suggestion, then you will had to remove kwargs here I guess

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -992,19 +992,57 @@ def torch_type(self):
...
+            else:  # If no args for .fix_prec()
+                x_shared = self.fix_prec().share(
+                    *workers, crypto_provider=crypto_provider, **kwargs
+                )
+            return x_shared
+
+        elif encryption_method.lower() == "paillier":
+            x = self.copy()
+            x_encrypted = PaillierTensor().on(x)  # Instantiate the class
+            x_encrypted.child.encrypt_(kwargs["public_key"])  # Perform blacHomomorphic Encryption
```
LaRiffle: Extract public_key like you did for mpc (it's nicer :) ) 
and typo: blacHomomorphic

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -992,19 +992,57 @@ def torch_type(self):
         else:
             return self.child.torch_type()
 
-    def encrypt(self, public_key):
-        """This method will encrypt each value in the tensor using Paillier
-        homomorphic encryption.
+    def encrypt(self, encryption_method="mpc", **kwargs):
```
LaRiffle: I have a slight preference for `protocol` instead of `encryption_method` (mostly because its shorter and because we often speak of crypto protocols)

test/torch/tensors/test_native.py
```diff
@@ -205,3 +205,24 @@ def forward(self, x):
 
     ## Forward on the remote model
     pred = model_net(tensor_remote)
+
+
+def test_encrypt():
+    hook = syft.TorchHook(torch)
```
LaRiffle: Check other examples you can improt workers directly:

ex: in test_additive_sharing.py
```
def test_eq(workers):
    me, alice, bob, crypto_provider = (
        workers["me"],
        workers["alice"],
        workers["bob"],
        workers["james"],
    )
```

test/torch/tensors/test_native.py
```diff
@@ -205,3 +205,24 @@ def forward(self, x):
...
+
+    bob = syft.VirtualWorker(hook, id="bob")
+    alice = syft.VirtualWorker(hook, id="alice")
+    james = syft.VirtualWorker(hook, id="james")
+
+    x = torch.randint(10, (1, 5), dtype=torch.float32)
+    x_encrypted = x.encrypt(workers=[bob, alice], crypto_provider=james, args_fix_prec={"base": 10})
+    assert torch.all(torch.eq(x_encrypted.get().float_prec(), x))
```
LaRiffle: Oh we've forgotten to add a generic decrypt method!!
Would you be willing to do also this one? :) 
for MPC .decrypt() would call `.get().float_prec()` 
and for paillier, it would'nt change

Syzygianinfern0: Sure. Do I submit it in another PR?
I've incorporated the suggestions mentioned.

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -992,19 +992,59 @@ def torch_type(self):
...
+            workers = kwargs.pop("workers")
+            crypto_provider = kwargs.pop("crypto_provider")
+            kwargs_fix_prec = kwargs  # Rest of kwargs for fix_prec method
+
+            if kwargs_fix_prec:
+                x_shared = self.fix_prec(**kwargs_fix_prec).share(
+                    *workers, crypto_provider=crypto_provider
+                )
+            else:  # If no kwargs for .fix_prec()
+                x_shared = self.fix_prec().share(*workers, crypto_provider=crypto_provider)
```
Jasopaum: I don't think this case is different from the one above, is it?
Doesn't `self.fix_prec(**kwargs_fix_prec)` work with an empty dict?

Syzygianinfern0: Yes. You are correct. 
That style was necessary for an older version of code where the kwargs were "None" rather than an empty dict.

test/torch/tensors/test_native.py
LaRiffle: In this PR would be more coherent if you can do it !

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -1042,16 +1042,44 @@ def encrypt(self, protocol="mpc", **kwargs):
...
+        Returns:
+            An decrypted version of the Tensor following the protocol specified
+
+        Raises:
+            NotImplementedError: If protocols other than the ones mentioned above are queried
+
+        """
+
+        if protocol.lower() == "mpc":
+            x_encrypted = self.copy()
```
LaRiffle: Do you need  copy() here?

Syzygianinfern0: It served a purpose of not making it an inplace operation

LaRiffle: Oh I see 
Ok I got your point, let's keep it your way

PR #3237, opened 2020-03-22
Add coverage range

imskr:
Allow range between 90 and 100

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,5 +1,5 @@
 coverage:
-    range: 90..100
+    range: 95..100
     round: down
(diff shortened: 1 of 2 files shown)


codecov.yml
```diff
@@ -1,3 +1,8 @@
+coverage:
+    range: 90..100
```
karlhigley: The current lower bound on coverage is 95%: https://github.com/OpenMined/PySyft/blob/master/.github/workflows/tests.yml#L69

imskr: Done

Issue #3233, opened 2020-03-21
NLLLoss is not supported in SMPC


LaRiffle:
We'd like support for the negative log likelihood loss in SMPC.

The Negative Log Likelihood Loss (or Cross Entropy Loss)  is the most common setting for classification problems with C classes. It increases as the predicted probability diverges from the actual label.

Details here: https://pytorch.org/docs/stable/nn.html#nllloss

codeboy5:
I would like to take up this issue

LaRiffle:
Done :)

codeboy5:
Hi, @LaRiffle are we looking for Cross Entry Loss (https://pytorch.org/docs/stable/nn.html#crossentropyloss) or NLL Loss (https://pytorch.org/docs/stable/nn.html#nllloss)? 
A slight difference between the two as Cross-Entropy combines softmax and NLL. I can implement both of these if you would like that?

LaRiffle:
Which one would you think is the most classic in classification? :)

codeboy5:
Cross-Entropy :)

LaRiffle:
Let's run for this one then 🚀

Syzygianinfern0:
I would like to work on this. 
@codeboy5 you can ping me over Slack for Colab as @Syzygianinfern0

karlhigley:
The easiest way to collaborate is probably to push WIP code and work from a shared branch.

PR #3242, opened 2020-03-22
[WIP] Negative Log-Likelihood and Cross Entropy Losses support for SMPC

Syzygianinfern0:
Fixes [3233](https://github.com/OpenMined/PySyft/issues/3233)
>The Negative Log Likelihood Loss (or Cross Entropy Loss) is the most common setting for classification problems with C classes. It increases as the predicted probability diverges from the actual label.


LaRiffle:
Don't forget to add tests :)

PR #3254, opened 2020-03-26
Init Crypten Plan + Rebased

gmuraru:
**Attention** Those changes are not meant to go into master, I will change the PR to target a different branch.

Rebased with master Crypten using Plans (also integrated with SyftCrypTensor).
 * Added the initial implementation (context setup with ```run_multiparty_computation``` and test file) from Ayoub
 * ```SyftCrypTensor``` will be a shell for the operations that will be done on the MPCTensor -- Integrated changes from Ajay
 * Traced the ```add``` and ```get_plain_text``` methods

How the tested function looks:
```
     @run_multiworkers([alice, bob], master_addr="127.0.0.1")
     @sy.func2plan()
     def plan_func():
         alice_tensor = crypten.load("crypten_data", 1)
         bob_tensor = crypten.load("crypten_data", 2)
 
         crypt = alice_tensor + bob_tensor
         result = crypt.get_plain_text()
         return result
```

How the plan looks
```
<Plan plan_func id:96985944591 owner:me built>
def plan_func():
    <syft.execution.placeholder_id.PlaceholderId object at 0x7f555de1f750> = crypten.load(crypten_data, 1)
    <syft.execution.placeholder_id.PlaceholderId object at 0x7f555de1f890> = crypten.load(crypten_data, 2)
    <syft.execution.placeholder_id.PlaceholderId object at 0x7f555de1f9d0> = _1.add(<syft.execution.placeholder_id.PlaceholderId object at 0x7f555de1fbd0>)
    <syft.execution.placeholder_id.PlaceholderId object at 0x7f555de1fb10> = _3.get_plain_text()
    return _4
```

### DIFF
--- a/.github/workflows/tutorials.yml
+++ b/.github/workflows/tutorials.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.6, 3.7]
+        python-version: [3.7]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/tutorials.yml
+++ b/.github/workflows/tutorials.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.7]
+        python-version: [3.6, 3.7]
 
     steps:
     - uses: actions/checkout@v2
--- a/pip-dep/requirements.txt
+++ b/pip-dep/requirements.txt
@@ -14,3 +14,4 @@ torch~=1.4.0
 tornado==4.5.3
 websocket_client~=0.57.0
 websockets~=8.1.0
+git+https://github.com/facebookresearch/CrypTen.git@68e0364c66df95ddbb98422fb641382c3f58734c#egg=crypten
--- a/pip-dep/requirements.txt
+++ b/pip-dep/requirements.txt
@@ -14,4 +14,4 @@ torch~=1.4.0
 tornado==4.5.3
 websocket_client~=0.57.0
 websockets~=8.1.0
-git+https://github.com/facebookresearch/CrypTen.git@e39a7aaf65436706321fe4e3fc055308c78b6b92#egg=crypten
+git+https://github.com/facebookresearch/CrypTen.git@4ad37cb97827b3364d433251d46c75d980bc3cf6#egg=crypten
--- a/pip-dep/requirements.txt
+++ b/pip-dep/requirements.txt
@@ -14,4 +14,4 @@ torch~=1.4.0
 tornado==4.5.3
 websocket_client~=0.57.0
 websockets~=8.1.0
-git+https://github.com/facebookresearch/CrypTen.git@3069264128999f91d21eac811f852dfc10064d27#egg=crypten
+git+https://github.com/gmuraru/CrypTen.git@2aa0c2f14b472a8df6a85be1cfc4630479a10b2f#egg=crypten
--- a/pip-dep/requirements.txt
+++ b/pip-dep/requirements.txt
@@ -14,4 +14,4 @@ torch~=1.4.0
 tornado==4.5.3
 websocket_client~=0.57.0
 websockets~=8.1.0
-git+https://github.com/facebookresearch/CrypTen.git@68e0364c66df95ddbb98422fb641382c3f58734c#egg=crypten
+git+https://github.com/facebookresearch/CrypTen.git@e39a7aaf65436706321fe4e3fc055308c78b6b92#egg=crypten
--- a/pip-dep/requirements.txt
+++ b/pip-dep/requirements.txt
@@ -14,4 +14,4 @@ torch~=1.4.0
 tornado==4.5.3
 websocket_client~=0.57.0
 websockets~=8.1.0
-git+https://github.com/facebookresearch/CrypTen.git@4ad37cb97827b3364d433251d46c75d980bc3cf6#egg=crypten
+git+https://github.com/facebookresearch/CrypTen.git@3069264128999f91d21eac811f852dfc10064d27#egg=crypten
--- a/syft/__init__.py
+++ b/syft/__init__.py
@@ -25,15 +25,23 @@ logger = logging.getLogger(__name__)
 # Tensorflow / Keras dependencies
 # Import Hooks
 
+__all__ = []
 if dependency_check.tfe_available:
     from syft.frameworks.keras import KerasHook
     from syft.workers.tfe import TFECluster
     from syft.workers.tfe import TFEWorker
 
-    __all__ = ["KerasHook", "TFECluster", "TFEWorker"]
+    __all__.extend(["KerasHook", "TFECluster", "TFEWorker"])
 else:
     logger.info("TF Encrypted Keras not available.")
-    __all__ = []
+
+if dependency_check.crypten_available:
+    from syft.frameworks.torch.tensors.crypten.syft_crypten import SyftCrypTensor
+
+    __all__.extend(["SyftCrypTensor"])
+else:
+    logger.info("CrypTen not available.")
+
 
 # Pytorch dependencies
 # Import Hook
@@ -113,6 +121,7 @@ __all__.extend(
         "AutogradTensor",
         "FixedPrecisionTensor",
         "LargePrecisionTensor",
+        "SyftCrypTensor",
         "PointerTensor",
         "MultiPointerTensor",
         "PrivateGridNetwork",
--- a/syft/__init__.py
+++ b/syft/__init__.py
@@ -35,14 +35,6 @@ if dependency_check.tfe_available:
 else:
     logger.info("TF Encrypted Keras not available.")
 
-if dependency_check.crypten_available:
-    from syft.frameworks.torch.tensors.crypten.syft_crypten import SyftCrypTensor
-
-    __all__.extend(["SyftCrypTensor"])
-else:
-    logger.info("CrypTen not available.")
-
-
 # Pytorch dependencies
 # Import Hook
 from syft.frameworks.torch.hook.hook import TorchHook
--- a/syft/__init__.py
+++ b/syft/__init__.py
@@ -113,7 +113,6 @@ __all__.extend(
         "AutogradTensor",
         "FixedPrecisionTensor",
         "LargePrecisionTensor",
-        "SyftCrypTensor",
         "PointerTensor",
         "MultiPointerTensor",
         "PrivateGridNetwork",
--- a/syft/__init__.py
+++ b/syft/__init__.py
@@ -44,7 +44,7 @@ from syft.grid.private_grid import PrivateGridNetwork
 from syft.grid.public_grid import PublicGridNetwork
 
 # Import sandbox
-from syft.sandbox import create_sandbox, hook
+from syft.sandbox import create_sandbox, make_hook
 
 # Import federate learning objects
 from syft.frameworks.torch.fl import FederatedDataset, FederatedDataLoader, BaseDataset
@@ -118,7 +118,7 @@ __all__.extend(
         "PrivateGridNetwork",
         "PublicGridNetwork",
         "create_sandbox",
-        "hook",
+        "make_hook",
         "combine_pointers",
         "FederatedDataset",
         "FederatedDataLoader",
--- a/syft/execution/placeholder.py
+++ b/syft/execution/placeholder.py
@@ -27,6 +27,17 @@ class PlaceHolder(AbstractTensor):
         self.expected_shape = tuple(shape) if shape is not None else None
         self.child = None
 
+    def __getattribute__(self, name):
+        try:
+            # Try to find the attribute in the current object
+            # we need some attributes like: instantiate, id, tags
+            response = object.__getattribute__(self, name)
+        except AttributeError:
+            child = object.__getattribute__(self, "child")
+            response = getattr(child, name)
+
+        return response
+
     def instantiate(self, tensor):
         """
         Add a tensor as a child attribute. All operations on the placeholder will be also
--- a/syft/execution/plan.py
+++ b/syft/execution/plan.py
@@ -225,7 +225,7 @@ class Plan(AbstractObject):
         """
         object.__setattr__(self, name, value)
 
-        if isinstance(value, FrameworkTensor):
+        if isinstance(value, FrameworkTensor) or isinstance(value, AbstractTensor):
             self.role.register_state_tensor(value)
         elif isinstance(value, FrameworkLayerModule):
             for tensor_name, tensor in value.named_tensors():
--- a/syft/execution/plan.py
+++ b/syft/execution/plan.py
@@ -8,6 +8,7 @@ import io
 import torch
 
 import syft as sy
+from syft import dependency_check
 from syft.execution.computation import ComputationAction
 from syft.execution.placeholder import PlaceHolder
 from syft.execution.role import Role
@@ -23,6 +24,8 @@ from syft.workers.abstract import AbstractWorker
 from syft_proto.execution.v1.plan_pb2 import Plan as PlanPB
 from syft_proto.execution.v1.computation_action_pb2 import ComputationAction as ComputationActionPB
 
+if dependency_check.crypten_available:
+    import crypten
 
 class func2plan(object):
     """Decorator which converts a function to a plan.
--- a/syft/execution/plan.py
+++ b/syft/execution/plan.py
@@ -16,7 +16,6 @@ from syft.execution.state import State
 from syft.execution.translation.abstract import AbstractPlanTranslator
 from syft.execution.translation.default import PlanTranslatorDefault
 from syft.generic.frameworks.types import FrameworkTensor
-from syft.generic.tensor import AbstractTensor
 from syft.generic.frameworks.types import FrameworkLayerModule
 from syft.generic.object import AbstractObject
 from syft.generic.pointers.pointer_plan import PointerPlan
@@ -225,7 +224,7 @@ class Plan(AbstractObject):
         """
         object.__setattr__(self, name, value)
 
-        if isinstance(value, FrameworkTensor) or isinstance(value, AbstractTensor):
+        if isinstance(value, FrameworkTensor):
             self.role.register_state_tensor(value)
         elif isinstance(value, FrameworkLayerModule):
             for tensor_name, tensor in value.named_tensors():
@@ -247,8 +246,6 @@ class Plan(AbstractObject):
             if self.include_state:
                 args = (*args, self.state)
             return self.forward(*args)
-        elif not self.is_built:
-            return self.build(args)
         else:
             return self.role.execute(args)
 
--- a/syft/execution/plan.py
+++ b/syft/execution/plan.py
@@ -25,9 +25,6 @@ from syft.workers.abstract import AbstractWorker
 from syft_proto.execution.v1.plan_pb2 import Plan as PlanPB
 from syft_proto.execution.v1.computation_action_pb2 import ComputationAction as ComputationActionPB
 
-if dependency_check.crypten_available:
-    import crypten
-
 
 class func2plan(object):
     """Decorator which converts a function to a plan.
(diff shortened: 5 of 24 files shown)


syft/frameworks/crypten/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+
+import crypten.communicator as comm
+import crypten
+
+
+def load(tag: str, src: int):
+    if src == comm.get().get_rank():
+        # Means the data is on one of our local workers
+
+        worker = syft.local_worker.get_worker_from_rank(src)
```
gmuraru: @youben11 I changed the load to look after the worker given a rank (the PR I merged into the CrypTen branch for plans did not have those changes and it fails when it looks for the tagged data on the ```syft.local_worker```)

pip-dep/requirements.txt
```diff
@@ -14,3 +14,4 @@ torch~=1.4.0
 tornado==4.5.3
 websocket_client~=0.57.0
 websockets~=8.1.0
+git+https://github.com/facebookresearch/CrypTen.git@e39a7aaf65436706321fe4e3fc055308c78b6b92#egg=crypten
```
youben11: We should keep with commit '68e0364c66df95ddbb98422fb641382c3f58734c' (or any other that don't break with pysyft) for the moment, as the last version introduces some changes that requires torch nightly as well as other functionalities we doesn't support yet.

syft/__init__.py
```diff
@@ -25,15 +25,23 @@
...
-    __all__ = ["KerasHook", "TFECluster", "TFEWorker"]
+    __all__.extend(["KerasHook", "TFECluster", "TFEWorker"])
 else:
     logger.info("TF Encrypted Keras not available.")
-    __all__ = []
+
+if dependency_check.crypten_available:
+    from syft.frameworks.torch.tensors.crypten.syft_crypten import SyftCrypTensor
+
+    __all__.extend(["SyftCrypTensor"])
```
youben11: I'm not sure about this, but if SyftCrypTensor is a tensor type that its only purpose is for building plans with crypten functionalities, then should we make it available in the API?

gmuraru: Removed it

pip-dep/requirements.txt
gmuraru: Will try to keep track with the latest version of CrypTen (also removed the Python 3.6 Tests and TutorialTests from this PR) and if there are problems with the Torch, will go back to a more stable version

syft/execution/plan.py
```diff
@@ -321,8 +328,12 @@ def build(self, *args):
 
         results = (results,) if not isinstance(results, tuple) else results
 
-        arg_ids = [t.id for t in args if isinstance(t, FrameworkTensor)]
-        result_ids = [t.id for t in results if isinstance(t, FrameworkTensor)]
+        arg_ids = [
+            t.id for t in args if isinstance(t, FrameworkTensor) or isinstance(t, AbstractTensor)
+        ]
+        result_ids = [
```
gmuraru: This is needed because we return a ```SyftCrypTensor``` which is an ```AbstractTensor```

syft/execution/plan.py
```diff
@@ -422,7 +433,12 @@ def __call__(self, *args, **kwargs):
                 if _self is None:
                     response = eval(cmd)(*args, **kwargs)  # nosec
                 else:
-                    response = getattr(_self, cmd)(*args, **kwargs)
+                    try:
+                        response = getattr(_self, cmd)(*args, **kwargs)
```
gmuraru: Q: What is the better method to do it?
The problem is that when running the plan and we reach the ```get_plain_text``` instruction, ```_self``` is a ```PlaceHolder``` and ```.child``` is an ```MPCTensor```. The placeholder does not have the ```get_plain_text``` method and in the current implementation I try to resolve it by checking the child

syft/execution/plan.py
karlhigley: Why is `SyftCrypTensor` not a `FrameworkTensor`? Is Crypten not a framework?

syft/execution/plan.py
karlhigley: If `Placeholder` isn't properly forwarding to the `child`, we should probably fix that in `Placeholder`.

syft/frameworks/crypten/hook/hook.py
```diff
@@ -0,0 +1,22 @@
...
+def get_hooked_crypten_func(func_api_name, func):
+    cmd_name = f"crypten.{func_api_name}"
+
+    @tracer(func_name=cmd_name)
+    @wraps(func)
+    def overloaded_func(*args, **kwargs):
+        try:
+            response = func(*args, **kwargs)
+        except RuntimeError:
+            response = SyftCrypTensor(tensor=th.zeros([]))
```
gmuraru: Q: Tried with a simple ```th.zeros([])``` (without explicitly specifying the shape). It seems that currently, the shape does not matter.

syft/execution/plan.py
```diff
@@ -20,6 +22,9 @@
 from syft_proto.execution.v1.plan_pb2 import Plan as PlanPB
 from syft_proto.execution.v1.computation_action_pb2 import ComputationAction as ComputationActionPB
 
+if dependency_check.crypten_available:
+    import crypten
```
karlhigley: Plans shouldn't know anything about specific frameworks, so we should avoid adding Crypten-specific code if we can help it. (@OpenMined/syft-core-team has an open issue to remove the PyTorch-specific code.)

syft/messaging/message.py
```diff
@@ -676,3 +676,57 @@ def detail(worker: AbstractWorker, msg_tuple: tuple) -> "WorkerCommandMessage":
             sy.serde.msgpack.serde._detail(worker, command_name),
             sy.serde.msgpack.serde._detail(worker, message),
         )
+
+
+class CryptenInit(Message):
```
karlhigley: This seems like the kind of thing PySyft has left to be added on top by other libraries (like Grid.) I'm not opposed to having this functionality in PySyft, but the inconsistency strikes me as pretty weird and a sign that there may be a design issue here.

syft/workers/base.py
```diff
@@ -28,13 +28,16 @@
...
 from syft.messaging.message import IsNoneMessage
+from syft.messaging.message import CryptenInit
 from syft.messaging.message import Message
 from syft.messaging.message import ObjectMessage
 from syft.messaging.message import ObjectRequestMessage
 from syft.messaging.message import PlanCommandMessage
 from syft.messaging.message import SearchMessage
 from syft.workers.abstract import AbstractWorker
 
+from syft.frameworks.crypten import run_party
```
karlhigley: `BaseWorker` intentionally doesn't know anything about frameworks, so would like to find a way to avoid letting that knowledge leak in. One possibility would be to create a `CryptenWorker` that overrides or adds to the message dispatch table.

.github/workflows/tutorials.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.6, 3.7]
+        python-version: [3.7]
```
karlhigley: This implies we'd be removing support for Python 3.6 entirely. @iamtrask, thoughts on that?

If we want to keep 3.6 support for PySyft (but not Crypten), possible we could split the workflows into one for each Python version and exclude the Crypten tests in the 3.6 workflow.

syft/execution/placeholder.py
```diff
@@ -34,7 +34,7 @@ def instantiate(self, tensor):
         """
         if isinstance(tensor, PlaceHolder):
             self.child = tensor.child
-        elif tensor.is_wrapper:
+        elif hasattr(tensor, "is_wrapper") and tensor.is_wrapper:
```
gmuraru: This is needed because after we built a plan the ```tensor``` (received as a parameter) is a ```MPCTensor``` (does not have the attribute ```is_wrapper```.
Another solution can be seen in a previous commit [here](https://github.com/OpenMined/PySyft/pull/3254/commits/d884650282b6cd45c8c80562cb431d9a81310988#diff-3fd9f3371ec1373832b99032d1212bad) (I kinda like more this approach)

karlhigley: Either way is probably fine. Using `hasattr` protects us against future tensor types that don't have the attribute though.

.github/workflows/tutorials.yml
gmuraru: Sorry, this should target a separate branch -- the changes are like this because the latest CrypTen master has support for python >= 3.7

syft/workers/base.py
gmuraru: Good idea :)

karlhigley:
BTW, some of the changes I suggested above are probably things the @OpenMined/syft-core-team should tackle. I'm not trying to push all the PySyft refactoring onto the Crypten team, I just want to make sure we have a good foundation to build on that will support what Crypten needs. Where the foundation is lacking, I want to make sure that we're talking about the gaps and planning for how to resolve them.

Some things can probably be addressed in this PR (e.g. `CryptenWorker`), some things could be deferred for later (e.g. `Placeholder` method forwarding), and some things might be worthy of their own issues (e.g. establishing groups of communicating workers in a consistent way.) The last category especially is where I think @OpenMined/syft-core-team could help.

syft/execution/placeholder.py
```diff
@@ -25,6 +25,17 @@ def __init__(self, owner=None, id=None, tags: set = None, description: str = Non
             self.id = PlaceholderId(self.id)
         self.child = None
 
+    def __getattribute__(self, name):
```
gmuraru: @karlhigley moved the logic here.
We also need tome attributes from Placeholder (like the ```id```, ```tags```, ```instantiate``` method, etc.).
and for some of them, we forward them to the child.

Another solution for this would be to use a list of attributes that we know are called on the placeholder.
This list should be kept with the attributes like: ```id```, ```tags``` ... (all the parameters in the constructor) + the methods we define in the Placeholder.

What would look like then:
```
def __getattribute__(self, name):
    if name in ["id", "tags", ....]:
       return object.__getattribute__(self, name)
   else:
       child = object.__getattribute__(self, "child")
       return getattr(child, name)
```

karlhigley: I think you could maybe use `hasattr` to check if Placeholder defines `name` and delegate to the child otherwise.

.github/workflows/tutorials.yml
karlhigley: I'd really ❤️  for it to target the main branch. We want Crypten support, it's just a matter of figuring out how to get it integrated without upending the rest of the apple cart. We're squeezed on one side by Tensorflow (<= 3.7) and on the other side by Crypten. We really could narrow PySyft to 3.7 support only, just want to make sure that isn't going to break something else.

syft/execution/placeholder.py
karlhigley: The suggestion above doesn't work, because `hasattr` calls `__getattribute__`. Oops!

syft/execution/plan.py
LaRiffle: Interesting point. Maybe it is not because currenly SyftCrypTensor is just a helper to build Plans for Crypten, it doesn't serve other purpose than building plans. But if it proves to be more useful, than it should be definitely a FrameworkTensor

syft/frameworks/torch/tensors/crypten/syft_crypten.py
```diff
@@ -0,0 +1,52 @@
...
+    @property
+    def data(self):
+        return self
+
+    @data.setter
+    def data(self, new_data):
+        self.tensor = new_data.child
+        return self
+
+    @tracer(method_name="add")
```
LaRiffle: so do we need this one, or is the add call directly forwarded to the child?

syft/messaging/message.py
LaRiffle: Sorry I didn't get the meaning of "This seems like the kind of thing PySyft has left to be added on top by other libraries"
What would you think would be the best way to proceed here?

syft/workers/base.py
LaRiffle: Alternatively, how bad would that be to put the import in the `def run_crypten_party` ?
I've already seen things like this but I don't know how evil it is

.github/workflows/tutorials.yml
iamtrask: Yes - I agree that we should be targeting the main branch. Having multiple separate version of PySyft is really a last resort nuclear option.

youben11: We have currently two blocking issues to merge this into master, the first is this python 3.6 support, the second is that crypten is now using torch nightly, and it will actually breaks badly if ran with torch==1.4.0

syft/workers/base.py
gmuraru: I think the problem is that we are exposing some methods that are not necessarily needed by the ```VirtualWorker``` (@karlhigley this is the reason?)

.github/workflows/tutorials.yml
gmuraru: What do you think should be the resolution here?
I am thinking that will be a good idea to test with an older version of Crypten (that had support for pytorch 1.4) and target the Syft master? The downside of this is that we are not up-to-date with the Crypten changes.

syft/frameworks/torch/tensors/crypten/syft_crypten.py
gmuraru: It is forwarded to the child (in our case we do not have a child).

But, if I change in the ```__init__``` method to have this:
```
self.child = tensor
```
Then when building the plan, the traces look like this:
```
-> for log in sy.hook.trace.logs:
(Pdb) p sy.hook.trace.logs
[(('crypten.load', None, ('crypten_data', 1), {}), SyftCrypTensor>tensor(0.)), (('crypten.load', None, ('crypten_data', 2), {}), SyftCrypTensor>tensor(0.)), (('__add__', tensor(0.), (tensor(0.),), {}), tensor(0.)), (('get_plain_text', SyftCrypTensor>tensor(0.), (), {}), SyftCrypTensor>None)]
(Pdb)
```
And we get an error in the plan with:
```
 ValueError: The following tensor was used but is not known in this plan: 
E                   0.0
E                   Possible reasons for this can be:
E                   - This tensor is external to the plan and should be provided using the state. See more about plan.state to fix this.```

This is because (I think) the values for the add operations are not wrapped in SyftCrypTensor. The solution (there might be another) was to trace the ```add``` operation in the ```SyftCryptensor``` class.

File: pip-dep/requirements.txt [hunk 14]
```diff
@@ -14,3 +14,4 @@ torch~=1.4.0
 tornado==4.5.3
 websocket_client~=0.57.0
 websockets~=8.1.0
+git+https://github.com/facebookresearch/CrypTen.git@4ad37cb97827b3364d433251d46c75d980bc3cf6#egg=crypten
```
gmuraru, line 17: Q: Should we move this to a separate requirement file?

.github/workflows/tutorials.yml
gmuraru: Changed the target to ```master```

syft/messaging/message.py
```diff
@@ -52,6 +52,37 @@ def __repr__(self):
...
+
+        Returns:
+            CryptenInit message.
+
+        Examples:
+            message = detail(sy.local_worker, msg_tuple)
+        """
+        return CryptenInit(sy.serde.msgpack.serde._detail(worker, msg_tuple[0]))
+
+
```
youben11, line 85: Seems like this is all duplicated 
```suggestion
```

test/serde/serde_helpers.py
```diff
@@ -1417,6 +1417,37 @@ def compare(detailed, original):
...
+                CODE[syft.messaging.message.CryptenInit],
+                (
+                    (CODE[tuple], (0, 2, (CODE[str], (b"127.0.0.1",)), 8080)),
+                ),  # (Any) simplified content
+            ),
+            "cmp_detailed": compare,
+        },
+    ]
+
+
```
youben11, line 1450: This is a deuplicate as well
```suggestion
```

pip-dep/requirements.txt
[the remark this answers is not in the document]
gmuraru: Will do

.github/workflows/tutorials.yml
LaRiffle: FWIW we've retargeted to a crypten branch after lots of discussion becuse we need torch nightly
We'll make sure this Openmined:crypten branch is updated regularly

syft/frameworks/torch/tensors/crypten/syft_crypten.py
```diff
@@ -0,0 +1,50 @@
+import torch
```
LaRiffle, line 1: Should we remove it from this PR if we don't use it?

syft/generic/frameworks/hook/trace.py
```diff
@@ -60,6 +60,9 @@ def trace_wrapper(*args, **kwargs):
                     recording sub actions
             """
 
+            if not hasattr(syft, "hook") or syft.hook == None:
```
LaRiffle, line 63: When can this happen?

syft/messaging/message.py
LaRiffle: Agreed

syft/generic/frameworks/hook/trace.py
gmuraru: The trace for the error was this:
```

    @pytest.fixture(scope="session", autouse=True)
    def hook():
>       hook = TorchHook(torch)

test/conftest.py:130: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
syft/frameworks/torch/hook/hook.py:145: in __init__
    self._hook_native_tensor(crypten.mpc.MPCTensor, TorchTensor)
syft/frameworks/torch/hook/hook.py:277: in _hook_native_tensor
    self._hook_native_methods(tensor_type)
syft/generic/frameworks/hook/hook.py:102: in _hook_native_methods
    native_method = getattr(tensor_type, attr)
../../../miniconda3/envs/pysyft/lib/python3.7/site-packages/crypten/cryptensor.py:39: in __getattribute__
    dummy = cls(None)
../../../miniconda3/envs/pysyft/lib/python3.7/site-packages/crypten/mpc/mpc.py:73: in __init__
    super().__init__(requires_grad=requires_grad)
../../../miniconda3/envs/pysyft/lib/python3.7/site-packages/crypten/cryptensor.py:132: in __init__
    self._reset_gradients()
```

What (I think is happening) -- when we hook we use the ```getattr```  -- but that method is changed in the CrypTensorMetaclass (metaclass for the MPCTensor). The problem is when we hook a method and that method is called in the ```getattr``` (in getattr we call an already hooked method and we do not have the syft.hook initialized -- syft.hook is initialized at the end of the TorchHook init method)

syft/frameworks/torch/tensors/crypten/syft_crypten.py
gmuraru: Yes, Done

gmuraru:
> I left very few comments but I guess the main part left is that CrypTenWorker.
> Regarding this, I do have a slight concern about creating workers dedicated to specific frameworks. In particular because it means you're making a choice when starting a set up on which interaction you'll allow, and the data you have and use won't be available for workers which are not from this framework.
> I'd rather envision adding to workers some specialities: by default you would get a worker which is fine for pytorch but not let's say for crypten. But at any point you could say `alice.add_crypten_support()` or smthg like this and your worker could also support specific operations for this framework, without loosing its compatibility with pytorch workers. I don't know if this makes sense?

It makes a lot of sense @LaRiffle. Currently I want to have this PR merged and after I can add support for this. If that sounds ok? Or should I push the changes in this PR? (I am thinking that this PR becomes bigger and bigger and it might be hard to track all the changes)

syft/generic/frameworks/hook/trace.py
LaRiffle: Amazing, thanks for the through explanation!

PR #3256, opened 2020-03-26
Fix the placeholder serialization.

### LINKED ISSUE #3214: Fix parameter serialization
In some situations, parameters are not serialized properly. I suspect this is due to our implementation of parameter.data

Here is one example:
```python
class Net(sy.Plan):
    def __init__(self):
        super(Net, self).__init__()
        self.fc1 = nn.Linear(1, 1)

    def forward(self, x):
        return self.fc1(x)

plan = Net()
plan.build(th.tensor([1.2]))

x = th.tensor([-1.0])
expected = plan(x)

plan.fix_precision().share(alice, bob, crypto_provider=charlie)
print(plan.state.tensors())
ptr_plan = plan.send(james)

# Fetch plan
fetched_plan = plan.owner.fetch_plan(ptr_plan.id_at_location, james)
print('***')
print(fetched_plan.state.tensors())
```
Output
```
[Parameter containing:
(Wrapper)>FixedPrecisionTensor>[AdditiveSharingTensor]
	-> [PointerTensor | me:94226517866 -> alice:74685210613]
	-> [PointerTensor | me:30028513485 -> bob:91228892047]
	*crypto provider: charlie*, Parameter containing:
(Wrapper)>FixedPrecisionTensor>[AdditiveSharingTensor]
	-> [PointerTensor | me:16955185561 -> alice:5015164314]
	-> [PointerTensor | me:77573712688 -> bob:21883177159]
	*crypto provider: charlie*]
***
[FixedPrecisionTensor>[AdditiveSharingTensor]
	-> [PointerTensor | me:94226517866 -> alice:74685210613]
	-> [PointerTensor | me:30028513485 -> bob:91228892047]
	*crypto provider: charlie*, FixedPrecisionTensor>[AdditiveSharingTensor]
	-> [PointerTensor | me:16955185561 -> alice:5015164314]
	-> [PointerTensor | me:77573712688 -> bob:21883177159]
	*crypto provider: charlie*]
```

rotalex:
Fixes #3214. 
The problem was that when reconstructing the network placeholders, the is_wrapper parameter from tensor was ignored.
The apparent issue with parameter serialization was due to the fact that when fetching a plan the wrapper parameter was incorrectly reconstructed as a plain parameter. This is actually done inside the placeholder deserialization function, where the "is_wrapper" field is stripped from the parameter tensors inside the instantiate method, hence the missing "Wrapper" when printing the deserialized network's parameters.

### DIFF
--- a/syft/execution/placeholder.py
+++ b/syft/execution/placeholder.py
@@ -29,8 +29,6 @@ class PlaceHolder(AbstractTensor):
         """
         Add a tensor as a child attribute. All operations on the placeholder will be also
         executed on this child tensor.
-
-        We remove wrappers or Placeholders if is there are any.
         """
         if isinstance(tensor, PlaceHolder):
             self.child = tensor.child
--- a/syft/execution/placeholder.py
+++ b/syft/execution/placeholder.py
@@ -34,8 +34,6 @@ class PlaceHolder(AbstractTensor):
         """
         if isinstance(tensor, PlaceHolder):
             self.child = tensor.child
-        elif tensor.is_wrapper:
-            self.instantiate(tensor.child)
         else:
             self.child = tensor
         return self
--- a/syft/execution/placeholder.py
+++ b/syft/execution/placeholder.py
@@ -29,6 +29,8 @@ class PlaceHolder(AbstractTensor):
         """
         Add a tensor as a child attribute. All operations on the placeholder will be also
         executed on this child tensor.
+
+        We remove Placeholders if is there are any.
         """
         if isinstance(tensor, PlaceHolder):
             self.child = tensor.child
--- a/test/execution/test_plan.py
+++ b/test/execution/test_plan.py
@@ -7,6 +7,7 @@ import torch.nn.functional as F
 import torch.optim as optim
 
 import syft as sy
+from itertools import starmap
 from syft.generic.pointers.pointer_tensor import PointerTensor
 from syft.generic.frameworks.types import FrameworkTensor
 from syft.execution.plan import Plan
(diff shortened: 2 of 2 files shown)


test/execution/test_plan.py
```diff
@@ -640,6 +640,7 @@ def forward(self, x):
     # Compare with local plan
     assert th.all(decrypted - expected.detach() < 1e-2)
     # assert fetched_plan.state.state_placeholders != plan.state.state_placeholders #TODO
```
gmuraru: Q: Are the state placeholder still not working?

karlhigley: What's the problem with state placeholders?

karlhigley: (I see that #3214 has something to do with `Parameters`, but not sure what about `Placeholders` is problematic.)

syft/execution/placeholder.py
```diff
@@ -34,8 +34,6 @@ def instantiate(self, tensor):
         """
         if isinstance(tensor, PlaceHolder):
             self.child = tensor.child
-        elif tensor.is_wrapper:
-            self.instantiate(tensor.child)
```
karlhigley: Removing this seems questionable, because we do want to be able to instantiate a `Placeholder` with a wrapper tensor.

test/execution/test_plan.py
```diff
@@ -640,6 +640,7 @@ def forward(self, x):
...
+    assert str(fetched_plan.state.tensors()) == str(plan.state.tensors())
```
karlhigley: Do we have to compare string representations here? Could we compare the state tensors directly?

test/execution/test_plan.py
```diff
@@ -640,7 +641,13 @@ def forward(self, x):
     # Compare with local plan
     assert th.all(decrypted - expected.detach() < 1e-2)
     # assert fetched_plan.state.state_placeholders != plan.state.state_placeholders #TODO
-    assert str(fetched_plan.state.tensors()) == str(plan.state.tensors())
+
+    assert all(
+        starmap(
```
karlhigley: TIL about `starmap`. 😃

karlhigley:
Thanks for bearing with me while I wrapped my head around this PR. 👍

Issue #2994, opened 2020-02-02
Allow PySyft to package plan operations either as a list of individual commands or as TorchScript


cereallarceny:
TorchScript allows for the execution of a PyTorch model outside of Python using the LibTorch C++ library and is compiled by PyTorch’s native JIT compiler. PySyft compiles all plan operations into a list called the “list of procedures” (it also compiles any tensors you include with the plan as the plan’s “state”). Currently, these operations are listed in order, one by one. For instance, you may have the following plan in Python:

```py
def plan_double_abs(x):
    x = x + x
    x = torch.abs(x)
    return x
```

Which then returns the `__add__` operation, followed by the `abs` operation. Each operation also comes with various tensors or pointer tensors (which are literally pointers in memory to a tensor that doesn’t exist until you create it) that instruct you what to name the result of each operation. Syft.js utilizes this to then convert each operation’s commandname into a valid TensorFlow.js command (more on this later).

However, PyTorch for Android and iOS do not have the ability to execute an ordered list of operations because they only have access to LibTorch in C++. LibTorch does not expose operations such as `add` and `abs`, and instead is only capable of executing TorchScript.

For this project, we need to instruct PySyft to be able to simplify plans either into TorchScript (for Android and iOS) or into a list of procedures and state tensors (for Javascript).

vvmnnnkv:
It seems initial thinking was that Plan with list of operations will be traced with torch.jit.trace.
The question is, do we need same Plan protobuf type to (be able to) hold both versions of Plan or it's fine to have ops list plan to be syft_proto.execution.v1.Plan and torchscript version to be syft_proto.types.torch.v1.ScriptModule?

cereallarceny:
I would imagine we can do this conversion on the fly in PyGrid and avoid
storing it twice on protobuf.  If it doesn't take long to run the plan
through the JIT compiler then I say that we do that.  Protobuf can only
have the one true value, and then when we host a plan on PyGrid then it
runs the compiler, splitting the plan into two versions and storing each
separately.  Go ahead and do this Vova if you want.

karlhigley:
I'd probably make the JIT compiled version a field of type `syft_proto.types.torch.v1.ScriptModule` on the `Plan` schema, because I'm not sure that the Torchscript will contain 100% of the information required to make Plans work. (It will definitely contain all the information required to run the computations in the Plan, but handling inputs/outputs/ids/etc isn't necessarily part of that.)

cereallarceny:
[...]

PR #3263, opened 2020-03-27
Add Plan translation to torchscript

vvmnnnkv:
Based on this discussion: https://github.com/OpenMined/PySyft/issues/2994#issuecomment-595333791

Current idea is having Plan translation done on PyGrid side.
When the "list of ops" Plan is hosted, PyGrid will translate it to multiple variants (e.g. torchscript, tfjs) and then deliver variant requested by worker client.

Right now it works with syft-proto from this branch: https://github.com/OpenMined/syft-proto/pull/80


vvmnnnkv:
Updated example [notebooks ](https://github.com/OpenMined/PySyft/blob/vvm/torchscript-plan/examples/experimental/FL%20Training%20Plan/Host%20Plan.ipynb) that work against updated PyGrid: https://github.com/OpenMined/PyGrid/pull/532

syft/execution/placeholder.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, owner=None, id=None, tags: set = None, description: str = Non
         if not isinstance(self.id, PlaceholderId):
             self.id = PlaceholderId(self.id)
 
-        self.expected_shape = shape
+        self.expected_shape = tuple(shape) if shape is not None else None
```
karlhigley, line 27: Starting to think we should maybe have a generic `Shape` class, even if it just inherits from `tuple`.

karlhigley: (Not a change for this PR necessarily, just thinking out loud.)

syft/execution/plan.py
```diff
@@ -306,11 +307,19 @@ def get_args_shape(self):
...
         return [ph.expected_shape for ph in self.role.input_placeholders()]
 
-    def translate_with(
-        self, plan_translator: "sy.execution.translation.abstract.AbstractPlanTranslator"
-    ):
+    def add_translation(self, plan_translator: "AbstractPlanTranslator"):
         return plan_translator(self).translate()
 
+    def remove_translation(self, plan_translator: Union[str, AbstractPlanTranslator] = "default"):
+        if plan_translator == "default":
```
karlhigley, line 314: Could we make this a real translator so the type is just `AbstractPlanTranslator`?

syft/execution/role.py
```diff
@@ -122,6 +122,20 @@ def _instantiate_inputs(self, args):
         )
         PlaceHolder.instantiate_placeholders(input_placeholders, args)
 
+        # Last extra argument is a state?
```
karlhigley, line 125: Was that a pre-existing thing or something new from this PR?

syft/execution/translation/abstract.py
```diff
@@ -5,4 +5,7 @@ def __init__(self, plan):
         self.plan = plan
```
karlhigley, line 5: Since it's an abstract base class, this should inherit from `ABC` (`from abc import ABC`.)

pip-dep/requirements.txt
```diff
@@ -7,7 +7,8 @@ phe~=1.4.0
 Pillow~=6.2.2
 requests~=2.22.0
 scipy~=1.4.1
-syft-proto~=0.2.9.a1
+#syft-proto~=0.2.9.a1
+git+git://github.com/OpenMined/syft-proto@vvm/placeholder-shape#egg=syft_proto
```
karlhigley, line 11: Is there another PR to go with this? Don't see one on `syft-proto`.

syft/execution/role.py
vvmnnnkv: This is new. This allows PlanTranslatorTorchscript to pass Plan's state tensors as input arguments to a Plan, so they don't have to be recorded in torchscript as nn.Parameters.
Not elegant, but this should be refactored away when ComputationPhase is introduced.

pip-dep/requirements.txt
vvmnnnkv: Oops. Created: https://github.com/OpenMined/syft-proto/pull/87

syft/execution/translation/torchscript.py
```diff
@@ -0,0 +1,36 @@
...
+
+        args_shape = plan.get_args_shape()
+        args = PlaceHolder.create_placeholders(args_shape)
+
+        # Temporarily remove reference to original function
+        tmp_forward = plan.forward
+        plan.forward = None
+
+        # To avoid storing Plan state tensors in torchscript, they will be send as parameters
+        plan_params = plan.parameters()
```
Jasopaum, line 23: nit: this variable is not used.
Also, I didn't get your comment just above. Is it related to something torchscript-specific?

syft/execution/role.py
```diff
@@ -116,6 +122,20 @@ def _instantiate_inputs(self, args):
...
+        if (
+            self.state is not None
```
Jasopaum, line 127: nit: self.state should never be None (in the constructor, it is set as a new Role if none is provided)

syft/execution/translation/torchscript.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, plan):
+        super().__init__(plan)
+
+    def translate(self):
+        plan = self.plan
+
+        args_shape = plan.get_args_shape()
+        args = PlaceHolder.create_placeholders(args_shape)
+
+        # Temporarily remove reference to original function
```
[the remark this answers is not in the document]
vvmnnnkv: To force `Plan.__call__` to use recorded operations instead of original function:
https://github.com/OpenMined/PySyft/blob/vvm/torchscript-plan/syft/execution/plan.py#L236

syft/execution/translation/torchscript.py
vvmnnnkv: > Is it related to something torchscript-specific?

Yes, jit.trace() fails if tensors with require_grad=True are used in operations but created outside of traced scope, like this:
```
t = torch.randn(3,3, requires_grad=True)

def my_func(x):
   x = x + t * 2
   return t

torch.jit.trace(my_func, (torch.ones(3, 3),)) #  <== Fails because t is created somewhere outside of my_func
```

We either need to register such tensors via nn.Module (this will store tensor inside torchscript, which we want to avoid here) or pass them as an argument into traced function:
```
t = torch.randn(3,3, requires_grad=True)

def my_func(x, t):
   x = x + t * 2
   return t

torch.jit.trace(my_func, (torch.ones(3, 3), t)) #  <== works
```

vvmnnnkv: Fixed `plan_params` 👍

syft/execution/translation/torchscript.py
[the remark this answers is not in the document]
Jasopaum: Yep, but does it mean that `jit.trace` doesn't work if we execute the Plan with forward()?
I mean, the behavior should be the same. Or maybe this _should_ is the problem

syft/execution/translation/torchscript.py
Jasopaum: Hum, and it wouldn't work with the `state.read()`, I guess

Issue #2744, opened 2019-11-17
Single run_websocket_server.py file


gmuraru:
**Is your feature request related to a problem? Please describe.**
Currently we have a ```run_websocket_server.py``` file in the root of the project, but also in different tutorials. I think we should have only one such file and the notebooks to import the needed functionality from it.

**Describe the solution you'd like**
Have only one ```run_websocket_server.py``` file in the top dir of the project.
The notebooks should import this module.

**Describe alternatives you've considered**
Move the file to a separate directory like ```start_websocket```

**Additional context**
This will reduce the complexity/size of the repository, reduce the points of failures.

bhadreshpsavani:
May i work on this issue?

LaRiffle:
Sure! I just assigned you :)

bhadreshpsavani:
Today I analyzed the issue and here are my observations,

1. I have compared all the  `run_websocket_server.py` in a different directory, they don't contain the same code. 

2. I have also analyzed codebase and come to know that we also have one another file named `start_websocket_servers.py` file which is using `run_websocket_server.py` in the same directories in the tutorials.

I think that suggested changes will not be valid unless we make generalized `start_websocket_servers.py`  and  `run_websocket_server.py` files to solve this issue

@LaRiffle  shall we do it or keep it as it is right now?

gmuraru:
Sorry for the late response. My opinion is that we should (in a way) unify some of the files.
One of the idea is to have a ```start_websocket_servers``` that could be parameterized and depending on one paramter start specific workers.
Eg:
   ```start_websocket_server --notebook mnist-parallel```` and this should start the set of workers for the ```MNIST-Parallel tutorial```.

What do you think @LaRiffle ?

JameerBabu:
Is this issue free to work ? If so, please assign me @LaRiffle

vineeth14:
Hey, I'd love to work on this issue

JameerBabu:
@gmuraru It was assigned to me and I started working on it.

JameerBabu:
@vineeth14 .. I am working on it .. please take some other issue.

karlhigley:
Could whoever has code (even if it's partial or not working yet) push a WIP PR and link it to this issue? It's hard for the maintainers to tell whether people are still working on something unless we can see code being written. 😄

JameerBabu:
[...]

PR #3265, opened 2020-03-28
Merging tests run_websocket_server.py file 

vineetjai:
Solves #2744


gmuraru:
There is also a file name ```start_websocket_servers.py``` in the ```websockets_mnist```. Can't the file be merged into one?

test/run_websocket_server.py
```diff
@@ -116,19 +27,38 @@ def start_websocket_server_worker(
...
-        help="if set, websocket server worker will be started in verbose mode",
+        help="""if set, websocket server worker will be started in verbose mode""",
     )
-
+    parser.add_argument(
+        "--notebook",
+        type=str,
+        default="normal",
+        help="""can run websocket server for websockets examples of mnist/mnist-parallel or
+    pen_testing/steal_data_over_sockets. Type 'mnist' for starting server
```
gmuraru: Q: Shouldn't there be 1/2 more level of indentation?

vineetjai: corrected.

vineetjai:
> There is also a file name `start_websocket_servers.py` in the `websockets_mnist`. Can't the file be merged into one?

I had already merged `run_websocket_servers.py` from the `websockets_mnist`.  While `start_websocket_servers.py` calls `run_websocket_servers.py`. They can't be merged into one file because of different arguments they pass through `run_websocket_servers.py `.

run_websocket_server.py
```diff
@@ -8,7 +8,7 @@
 import numpy as np
 from torchvision import datasets
 from torchvision import transforms
-
+from syft.frameworks.torch.federated import utils
```
gmuraru, line 11: Q: Shouldn't here be fl?

examples/tutorials/advanced/websockets-example-MNIST/start_websocket_servers.py
```diff
@@ -0,0 +1,23 @@
+import subprocess
+import sys
+from pathlib import Path
+
+python = Path(sys.executable).name
+
+FILE_PATH = Path(__file__).resolve().parents[5].joinpath("run_websocket_server.py")
```
gmuraru, line 7: I think here should be one level lower (4 instead of 5)

vineetjai: corrected.

run_websocket_server.py
vineetjai: I was running on an older version of syft-proto. It has to be fl according to new version.

examples/tutorials/advanced/websockets-example-MNIST/start_websocket_servers.py
```diff
@@ -0,0 +1,23 @@
+import subprocess
+import sys
```
gmuraru, line 2: There is also another file in ```webosckets-mnist```? From a first look, they seem identical? Can't one be removed

run_websocket_server.py
vineetjai: resolved.

examples/tutorials/advanced/websockets-example-MNIST/start_websocket_servers.py
vineetjai: Even we merge this file and make a new file in the main repo. then there has to be a file in `websockets-example-MNIST` and `websockets_mnist` which call this `start_websocket_servers.py` in the main repo which calls `run_websocket_server.py`. So I think there will be a lot of arguments that will involve which will not make codebase shorter but also make it complex.

vineetjai: So I think we have to leave it.

gmuraru: Why not use directly the file from ```websockets_mnist``` (like call only that file?)

vineetjai: So you are saying for `websockets-example-MNIST/start_websocket_servers.py` we can use  `websockets_mnist/start_websocket_servers.py` ?

examples/tutorials/advanced/websockets-example-MNIST/start_websocket_servers.py
```diff
@@ -0,0 +1,15 @@
+import subprocess
```
gmuraru, line 1: Q: Why not simply remove this and in all the places where this was used ad the other one -- what I understand is that we currently have 2 python files and this one calls the other one. My idea is to delete this one and change all the occurrences where we called this file to the other one

vineetjai: I think we can delete this file because its folder doesn't contain a notebook in which it is used. Now, this file is of no use. So should I delete it?

vineetjai:
> This does reduce code duplication, but still has multiple `run_websocket_server.py` files. Could we use a single `run_websocket_server.py` file for everything?

There are only these `run_websocket_server.py` files
```
./test/run_websocket_server.py
./test/scripts/run_websocket_server.py
./run_websocket_server.py
```
I had already include mergeable parts of `/test/run_websocket_server.py` and `./test/scripts/run_websocket_server.py` into `./run_websocket_server.py`. Also, `/test/run_websocket_server.py` and `./test/scripts/run_websocket_server.py` are calling `/run_websocket_server.py` with different arguments. So I think these three can't be merged into a large single file. If you have any suggestions let me know.

PR #3266, opened 2020-03-28
Send Dataset_pointer

abogaziah:
- Added functionality to get a dataset from a remote federated_dataset object without breaking it.
if the user calls `fed_dataset["alice"].get()` for example that would leave a pointer that points to noting and it will break the whole object 
- also updated the dataset_pointer representation to include (id, owner -> id at location, location)
- Solved issue #3267


syft/frameworks/torch/fl/dataset.py
```diff
@@ -253,6 +253,11 @@ def workers(self):
 
         return list(self.datasets.keys())
 
+    def get_dataset(self, worker):
```
Jasopaum: It's not the part of the code I know the best but do we want to have a more granular way of getting datasets?
For instance, what if a worker wants to get only one of its dataset in the FederatedDataset?

syft/frameworks/torch/fl/dataset.py
```diff
@@ -253,6 +253,11 @@ def workers(self):
...
+        dataset = self[worker].get()
+        del self.datasets[worker]
```
Jasopaum: Do we really want that? What is the usecase for `get_dataset()`?

abogaziah: because if the user used federated_dataset.datasets[worker].get() that would leave it with a pointer that points to nothing. that will require the pointer to be deleted from the datasets dict if we want the federated_dataset object to still work. the user won't be able to do it if he doesn't have a deep understanding of the implementation because it will cause errors in other parts of his code like dataset_loaders for example. this is a way to keep the object always clean and ready to use

syft/frameworks/torch/fl/dataset.py
abogaziah: it's not mine but I believe this implementation doesn't allow more than one dataset per worker since it uses a dict to store them, if it has {bob:ptr1, alice:ptr2} and you add alice:ptr3 it will replace ptr2 not add it

abogaziah:
I solved issue #3267 in the latest commits

PR #3271, opened 2020-03-29
Remove implicit inplace fix_precision for PointerTensor

### LINKED ISSUE #3261: fix_precision() is inplace when applied to a pointer tensor.
**Description of the bug**
The method fix_precision() applied on pointers is 'inplace'. This should not be the case

In the following code:

```
a = torch.Tensor([2., 3.]).send(bob)
a.fix_precision()
a = a.get()
print(type(a))
```

The variable  `a` after calling `get()` is a fixed precision tensor which is a bug, because the original variable `a` defined as `torch.Tensor([2., 3.])` is not a fixed precision tensor.

However, this bug is not existing in case when  `a` is not a pointer:

```
a = torch.Tensor([2., 3.])
a.fix_precision()
print(type(a))
```

`a` here is not a fixed precision tensor. which is the desired behavior.

**Desktop:**
 - OS: Archlinux
 - Version 0.3.2

sukhadj:
Fixes #3261


syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -812,12 +812,14 @@ def fix_prec(self, *args, storage="auto", field_type="int100", no_wrap: bool = F
         if not kwargs.get("owner"):
             kwargs["owner"] = self.owner
 
+        # we create a copy of pointer which prevents inplace fix precision
         if self.is_wrapper:
-            self.child = self.child.fix_prec(*args, **kwargs)
+            self_copy = self.copy()
```
gmuraru: Q: Is there a scenario where the ```PointerTensor``` is not a wrapper?
```
import torch
import syft

hook = syft.TorchHook(torch)
alice = syft.VirtualWorker(id="alice", hook=hook)

x = torch.Tensor([1,2,3])
ptr = x.send(alice)
ptr.child.fix_precision()
```

sukhadj: Code outside this if block takes care of it (which is not modified).
`(Wrapper)>FixedPrecisionTensor>tensor([1000, 2000, 3000])` is the output, which is correct!

sukhadj:
Is there any reason that we have two different methods for in-place and non in-place precision?
Shouldn't we just have one method with inplace as argument? :thinking:

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -812,12 +812,14 @@ def fix_prec(self, *args, storage="auto", field_type="int100", no_wrap: bool = F
...
+            self_copy.child = self.child.fix_prec(*args, **kwargs)
```
LaRiffle: Actually I'm not 100% comfortable with copying the  tensor.
What about this?
```python
if self.is_wrapper:
    child = self.child.fix_prec(*args, **kwargs)
    if no_wrap:
        return child
    else:
        return child.wrap()
```
If this works, then I guess we should go in that direction:)

sukhadj: It works and this looks better approach!!

Issue #3236, opened 2020-03-21
Extend nn.Module() with Support for Federated Learning of non-parameter information


iamtrask:
**Is your feature request related to a problem? Please describe.**

https://github.com/OpenMined/PySyft/issues/2498

**Describe the solution you'd like**

Solution: extend nn.Module() with a .parameters_and_statistics() iterator which will look for this kind of information so that we can make sure that federated learning properly moves everything from machien to machine.

**Describe alternatives you've considered**
We could also expose .statistics() different from .parameters() so that federaetd learning optimizers could treat them differnetly

**Additional context**
It's important to consider that FL averaging might not work for all statistics like this.

SagarMoghe:
Can I take up this problem?

AshwinRam2000:
I can take this up.

karlhigley:
Assigned you both. Hope that whoever starts working on it first can post a WIP PR with their early commits so you can collaborate on it.

SagarMoghe:
Thank you

ratmcu:
@karlhigley Can I work on this issue as well. I have already worked on a use case similar. So we can collaboratively solve this.

SagarMoghe:
@iamtrask @karlhigley , I am new to Syft so can you please guide me where in the code base should I start to look for relevant code to the related issue. Thank you

sachin-101:
@iamtrask From the comments in #2498 it seems the BatchNorm bug was due to different versions. So you can close this issue if it is solved.

karlhigley:
I think the proposed feature is still useful.

SagarMoghe:
@iamtrask @karlhigley  I did some research and I went around and snooped into nn.module code and noticed that line 87 of [nn.module](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/module.py) has the variable that contains information I need. I am not sure where exactly in the pysyft codebase should I implement the function to get the value of this variable. I am new to pysyft and dont know much about the code structure. Please help me.
PS: does making a function .parameters_and_statistics() which returns the self._buffers containing all the non parameterised attributes of the neuralNet and returning them to the user solves the problem of the issue?
Thank you

ratmcu:
@SagarMoghe I already added a function and working on testing it,
I am also adding features to get all the non parameter elements of iterated by an arbitrary function that could be added to the hooked pytorch nn module to be considered at the send().
I would like if we can work on testing the new feature, rather than repeating the same.

[...]

PR #3273, opened 2020-03-30
iterator for all the tensor elements other than parameters of nn.module

ratmcu:
new internal iterator could be extended with tensor element buffers (parameters, buffers ) in the nn.Module. this update holds all the internal buffers inside the nn.Module.


ratmcu:
@iamtrask do you think at the moment this modification would be suffice for the movement of non parametric tensor elements as demanded by https://github.com/OpenMined/PySyft/issues/3236 ? . I was hoping until we elegantly add a hook for the buffers underlying the nn.Module and Optimizers(which is not using a buffer definition at the moment from pytorch, which I hope we can add a feature in pytorch.)  we can keep updating the list of iterators from the nn.Module  if a new iterator of elements comes up.

iamtrask:
Still failing codecov - but if you can get that to pass (above 95%) we're good :)

syft/frameworks/torch/hook/hook.py
```diff
@@ -601,8 +617,12 @@ def module_send_(nn_self, *dest, force_send=False, **kwargs):
             if module_is_missing_grad(nn_self):
                 create_grad_objects(nn_self)
 
-            for p in nn_self.parameters():
-                p.send_(*dest, **kwargs)
+            # for p in nn_self.parameters():
+            #     p.send_(*dest, **kwargs)
```
karlhigley, line 621: Looks like the code below replaces these lines, so probably good to remove them.

karlhigley:
(@imskr This is another example where the main code coverage tests pass but Codecov still fails.)

karlhigley:
Test failure is from an unrelated known flaky test.

PR #3279, opened 2020-03-30
Added get_nn_buffer to get access to non parameterised variables of m…

SagarMoghe:
Added get_nn_buffer to get access to non parameterised variables of model.


SagarMoghe:
Solution for issue #3236

iamtrask:
how does this compare to https://github.com/OpenMined/PySyft/pull/3273/files

SagarMoghe:
@iamtrask , I will update the method name asap. I guess #3273 is requesting selected parameters and passing them, where as I am requesting the entire buffer that is maintained by pytorch nn.module. I feel it would be safe to use as future pytorch algorithms are most likely  to keep using the buffer and will only break its functionality when pytorch deprecates the named_buffer. Any future algorithm using some other parameter other than running mean or std will automatically get sent as it would be a part of the buffer and wont have to be explicitly added to the iterator list.

iamtrask:
(my approval is pending Theo's recommendation)

ratmcu:
@iamtrask in https://github.com/OpenMined/PySyft/pull/3273 I was driven by the decision of only depending on the nn.Module for iteratable parameters, Seems pytorch follows this design pattern in adding extra elements. So in future we'd only have to update the list of functions from nn.Module in the hook.

SagarMoghe:
Made some changes into my previous push after discussing with @karlhigley . Made a new buffer class in generic folder to handle all the buffers in future. Currently it only gathers data from torch and must be later integrated with hooks to send data to worker nodes.

SagarMoghe:
As of now the buffer is only collected in a variable and I would will collaborate with someone experienced in hooks to integrate my class to hooks and send it to worker nodes. Testing can be performed once data is sent to worker node. In case of any issue in later iteration, we can simply delete the merged file without affecting the integrity of code as the class added is added from an external python file i.e buffers.py

SagarMoghe:
I will create a test where I make a model and compare the buffer values directly from pytorch and from my buffer in pysyft. Does this sound right?

karlhigley:
Sounded like you wanted to be able to send the buffers along with the model, so the test should probably `send` the model+buffers to another worker, `get` them back, and then check that received buffers are the same as the sent buffers.

PR #3284, opened 2020-03-31
Generalize `Role` to any type of `Action` (computation or communication)

karlhigley:


syft/execution/role.py
```diff
@@ -10,7 +10,7 @@
 import torch
```
karlhigley: Even though `Role` no longer directly references `torch`, this import can't be removed because `Actions` sometimes call methods like `torch.manual_seed()`, which require `torch` to be available in the scope where `eval` runs. Crypten runs into the same issue so we need a better solution for this, but I'll leave that for a future PR.

karlhigley: Along related lines, sure would love to get rid of that `eval`.

test/execution/test_plan.py
```diff
@@ -160,7 +160,7 @@ def serde_plan(x):
     serde_plan_detailed = serde._detail(bob, serde_plan_simplified)
 
     t = th.tensor([1.0])
-    expected = serde_plan_detailed(t)
+    expected = serde_plan(t)
```
Jasopaum, line 163: Wow, good catch! How did that happen?

syft/execution/plan.py
```diff
@@ -182,13 +182,13 @@ def build(self, *args):
...
-        self.role.register_computation_inputs(args)
+        self.role.register_inputs(args)
 
         # Register outputs in role
-        self.role.register_computation_outputs(results)
+        self.role.register_outputs(results)
 
         for log in sy.hook.trace.logs:
-            self.role.register_computation_action(log)
+            self.role.register_action(log, ComputationAction)
```
Jasopaum, line 191: As the type of action will be automatically inferred, don't you think we could do it in the Role?

test/execution/test_plan.py
karlhigley: I think it's been that way since the test was written. Probably just a copy/paste oops.

syft/execution/plan.py
karlhigley: We could! I figured this was in the spirit of putting the restriction that `Plans` can only contain `ComputationActions` in `Plan` instead of `Role`. I also kinda like that `Roles` don't know anything about `Action` types.

karlhigley: We're going to have refactor this further one way or another, for sure. I was kinda leaning toward passing in an `Action` object (instead of the `log`.)

Jasopaum: Hum, but then, the Plan would have to build the `Action`s?
Also, what happens if we try to build a ComputationAction with communication stuff inside? 🤔

karlhigley: We were already assuming that all actions created in a `Role` would be `ComputationActions`, so this is no worse in that sense. It’s actually more flexible because now `Protocols` could add `CommunicationActions` to a `Role` without changing the `Role` code.

`Plans` should never contain `CommunicationActions`, so I don’t see a problem there. We do need to add logic somewhere to determine which `Actions` are computations vs communications and create the right ones, but we need it for `Protocols` and we’re not there yet. 🤷‍♀️

Jasopaum: No, I agree with what you did! I was reacting to the "passing in an Action object".
And my second thought didn't have anything to do here, sorry, it was just a random question ^^

Jasopaum: And I'm starting to (slowly) build protocols so if you have any ideas about determining which actions are Computations and which one are Computations, I'm listening 😄

karlhigley: Hmm, well, my first thought is that communication actions are all defined by PySyft and not by the various ML frameworks. Could we use that somehow? I mean, we can always just build a list of the communication actions and assume everything else is a computation action, but there might be a more elegant way to do it. 🤔

Issue #3259, opened 2020-03-27
Indexing pointer tensors with a list


artix41:
When running the following code
```python
hook = sy.TorchHook(torch)
bob = sy.VirtualWorker(hook, id="bob")

a = torch.tensor([10,20,30])
print(a[[0,2]])

b = torch.tensor([10,20,30]).send(bob)
print(b[[0,2]])
```
the first tensor `a` get displayed correctly as 
```python
tensor([10, 30])
```
but the second tensor raises the following error:
```python
IndexError: too many indices for tensor of dimension 1
```

I have Python 3.7.6, PySyft 0.2.4 and PyTorch 1.4.0.

I'm new to PySyft, so maybe it is an expected behavior for some reasons, but I don't really see why (since remote tensors are supposed to have the same functionalities as normal tensors as far as I understand).

rotalex:
I can take a look.

PR #3285, opened 2020-03-31
Fix list tensor index

rotalex:
Fix list as an index for syft tensors. Issue: #3259 also related:#2628

Problem solution: the logic from #2628 was mostly correct but it was placed in the wrong function. The return_list parameter should have been added to build_unwrap_args_with_rules which actually deals with function parameters.
Added a small test to check if tensors can be indexed with lists of indexes. 

Questions: Should we keep the changes from #2628 PR?


Syzygianinfern0:
@rotalex I am working on a similar problem. Please contact me on Slack as @Syzygianinfern0

syft/generic/frameworks/hook/hook_args.py
```diff
@@ -320,10 +321,12 @@ def build_unwrap_args_with_rules(args, rules, return_tuple=False):
     lambdas = [
         typed_identity(a)  # return the same obj with an identity fct with a type check if needed
         if not r  # if the rule is a number == 0.
+        else build_unwrap_args_with_rules(a, r, True, True)
```
gmuraru, line 324: Q: Why are we forcing here to return a tuple and a list?

gmuraru:
LGTM!

syft/generic/frameworks/hook/hook_args.py
karlhigley, line 324: Since the `return_list` flag causes the function to coerce the return type to `list`, could we do without the flag and make this line something like:
```suggestion
        else list(build_unwrap_args_with_rules(a, r, True))
```
If so, the lines below could also be removed.

syft/generic/frameworks/hook/hook_args.py
```diff
@@ -347,6 +350,9 @@ def build_unwrap_args_with_rules(args, rules, return_tuple=False):
     except KeyError:
         f = many_fold
 
+    if return_list:
+        return lambda x: list(f(lambdas, x))
+
```
karlhigley, line 355:
```suggestion
```

PR #3269, opened 2020-03-29
Fix overloading for torch.nn and Functional modules

Syzygianinfern0:
This PR fixes the following
- By #3253 the automatic overloading of torch modules was lost.
- By setting up the decorators and an efficient \_\_init\_\_.py, the functionality is restored.
- Also includes the overloading of AvgPooling and Conv2d from torch.nn which were earlier missing
- Move tests from test_precision related to tor APIs to test_nn.py and test_functional.py

PR #3286, opened 2020-04-01
Overload

Syzygianinfern0:
Follow Up of #3269 

>  This file is normally inside the nn/ directory. In the precision.py file when we import nn, the nn/\_\_init\_\_.py gets called. Currently all other classes (ie Conv2d, AvgPool2d, etc) behave well when imported in the nn/\_\_init\_\_.py. 
But due to the RNN Classes requiring the import of FixedPrecisionTensor (line 14) and precision.py requiring the import of nn module, there is a problem.

Changes:
- [x] Move rnn.py back to nn/ (non-breaking)
<details>
  <summary> How? </summary>

# Here is How
([from docs](https://docs.python.org/2.0/ref/import.html))
> The first form of import statement (`import foo.bar`) binds the module name in the local namespace to the module object, and then goes on to import the next identifier, if any. 

Hence, by **changing the import from a bound to an unbound-to-module import**, we can workaround the circular import. 
More information [here](https://stackoverflow.com/questions/22187279/python-circular-importing)
</details>

- [ ] Update tests to use the overloaded module


test/torch/nn/test_nn.py
```diff
@@ -4,7 +4,6 @@
 import torch.nn as nn
 import torch.nn.functional as F
 
-import syft.frameworks.torch.rnn as rnn
 import syft.frameworks.torch.nn as nn2
```
Jasopaum, line 7: Could you rename this to something else than `nn2`? `syft_nn` for instance, or just `nn` and `torch_nn` for the one above

syft/frameworks/torch/nn/rnn.py
```diff
@@ -47,10 +42,12 @@ def init_hidden(self, input):
         This method initializes a hidden state when no hidden state is provided
         in the forward method. It creates a hidden state with zero values.
         """
+        import syft.frameworks.torch.tensors.interpreters.precision as precision
```
Jasopaum, line 45: Doesn't it work with the import at the beginning of the file?

Syzygianinfern0: No. This is the only configuration (afaik).
Otherwise a major refraction/making it independant of that requirement is required.

test/torch/nn/test_nn.py
Syzygianinfern0: Done 😊

karlhigley:
Test failure is from an unrelated and known flaky test.

PR #3295, opened 2020-04-02
Remove `eval` & `import torch` from `Role` using framework package store

karlhigley:
This creates a global registry of the available framework packages,
which can be used when executing `Actions` in `Plans`. This removes the
last remaining `eval` in the code base, and removes the need for `Role`
to import framework packages referenced in `Plans` (like `torch` or
`crypten`.)

### DIFF
--- a/syft/execution/role.py
+++ b/syft/execution/role.py
@@ -132,17 +132,7 @@ class Role(AbstractObject):
         return_placeholder = self._fetch_placeholders_from_ids(return_placeholder)
 
         if _self is None:
-            cmd_path = cmd.split(".")
-
-            package_name = cmd_path[0]
-            subpackage_names = cmd_path[1:-1]
-            method_name = cmd_path[-1]
-
-            package = framework_packages[package_name]
-            for subpackage_name in subpackage_names:
-                package = getattr(package, subpackage_name)
-            method = getattr(package, method_name)
-
+            method = self._fetch_package_method(cmd)
             response = method(*args, **kwargs)
         else:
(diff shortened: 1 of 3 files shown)


syft/execution/role.py
```diff
@@ -5,8 +5,7 @@
 
 import copy
 
-# TODO torch shouldn't be used here
-import torch
+from syft.generic.frameworks.types import framework_packages
```
Jasopaum, line 8: A bit weird to import this from a module called `types` isn't it?

syft/execution/role.py
```diff
@@ -133,7 +132,18 @@ def _execute_action(self, action):
...
+            cmd_path = cmd.split(".")
+
+            package_name = cmd_path[0]
+            subpackage_names = cmd_path[1:-1]
+            method_name = cmd_path[-1]
+
+            package = framework_packages[package_name]
+            for subpackage_name in subpackage_names:
+                package = getattr(package, subpackage_name)
+            method = getattr(package, method_name)
```
Jasopaum, line 144: Do you think we could put this in a `execute_command` method to make it easier to read?

syft/execution/role.py
karlhigley: Yeah, it is. That's where the dependency checks for available imports happen though. We could import it in the `__init__.py`, I suppose.

syft/execution/role.py
karlhigley: Yeah, an `Extract Method` refactor is a good idea here. 👍

PR #3296, opened 2020-04-02
Plans support for typechecking, lists and dicts using serialized input.

### LINKED ISSUE #3184: Improve Plans flexibility

tudorcebere:
# Description

This PR fixes #3184. 

New features:
- Plans now have a new attribute: serialized_input, that will serialized the input at build time and compare it with the serialized input at call time.
- Plans now have support for type-checking at runtime based on the type provided at build time.
- Plans now support usage of lists, tuples and dicts as inputs.
- Plans now have a number of arguments check, build time and call time should have the same number of arguments.

# Type of change:
- New feature

# How Has This Been Tested?

Wrote a few tests that verify that:
- type-checking works, it should raise a RunTimeWarning if the types differ.
- argument number check works, build time and call time should not differ. If they differ, a TypeError should occur. I have chosen TypeError to be consistent with python functions.
- dicts from build time have the same keys as dicts from call time (nested as well). If they differ, a KeyError should occur.
- lists, dicts and tuples of tensors work as input (and any nested combination).

# TODO:
~~Adapt the changes to the new Role class.~~
~~Add syft-proto support.~~
~~Add protobuffer support.~~
~~Remove the usage of pickle (only a PoC). ~~
~~Check interaction with torchscript. ~~


syft/execution/role.py
```diff
@@ -117,23 +135,31 @@ def execute(self, args_):
...
+            """
+
+            if isinstance(obj, FrameworkTensor):
+                placeholder = input_placeholders.pop(0)
+                placeholder.instantiate(obj)
+            elif isinstance(obj, (list, tuple, set)):
+                for elem in obj:
+                    instantiate_nested_input(input_placeholders, elem)
+            elif isinstance(obj, dict):
+                _ = [
```
Jasopaum, line 154: I don't think you need the `_ = `. And also, maybe a classic for loop would be more readable and easier to understand than a list comprehension here

syft/execution/type_wrapper.py
```diff
@@ -0,0 +1,261 @@
...
+from warnings import warn
+from typing import Union
+from syft_proto.execution.v1.type_wrapper_pb2 import NestedTypeWrapper as NestedTypeWrapperPB
+from syft_proto.execution.v1.type_wrapper_pb2 import ClassType as ClassTypePB
+
+class NestedTypeWrapper:
+    """
+        Class for input type serialization and type checking for nested structures.
+    """
+    def __init__(self, nested_type=None):
```
Jasopaum, line 12: I think the `if` would be better in the `serialize_nested_type` method
Also, I'm not sure the pattern of instantiating a class and doing what you want to do at the same time is the most intuitive, I'll try to think about it but I guess it's ok

syft/execution/type_wrapper.py
```diff
@@ -0,0 +1,261 @@
...
+        if nested_type:
+            self.serialized_nested_type = NestedTypeWrapper.serialize_nested_type(nested_type)
+        else:
+            self.serialized_nested_type = None
+
+    @staticmethod
+    def serialize_nested_type(input_arg: any) -> Union[list, tuple, dict, type]:
+        """
+            Function to serialize input of a function/Plan, including nested types.
```
Jasopaum, line 21: nit: it's a method here

syft/execution/type_wrapper.py
```diff
@@ -0,0 +1,261 @@
...
+                call: the called argument length.
+                path: the nested path to reach that obj.
+
+            Returns:
+                 None
+        """
+        raise TypeError(f"{obj_type} {obj_name} {path} has length {call}, while being build with length {build}.")
+
+    @staticmethod
+    def raise_wrong_no_arguments_err(obj_type: str, obj_name: str, build: int, call: int) -> None:
```
Jasopaum, line 79: I think no_arguments is a bit ambiguous, maybe nb_arguments or number_arguments would be better

syft/execution/type_wrapper.py
```diff
@@ -0,0 +1,261 @@
...
+
+            Params:
+                obj_type: the type of the object returned by calling .__name__ on type(obj).
+                obj_name: the name/id of the object
+                args: the arguments to be compared with the reference/build one.
+
+            Returns:
+                None
+        """
+        def check_type_nested_structure(obj_type: str, obj_name: str, build: any, call: any, path: str) -> None:
```
Jasopaum, line 130: I don't think you need an "outer" recursion here, check_type_nested_structure and input_check do the same thing

syft/execution/type_wrapper.py
```diff
@@ -0,0 +1,261 @@
...
+            iterable_supported_list = (list, tuple, dict)
+
+            if type(call) not in iterable_supported_list:
+                if not isinstance(call, build):
+                    NestedTypeWrapper.raise_typecheck_warn(obj_type, obj_name, build.__name__, type(call).__name__, path)
+                return
+
+            if type(build) != type(call):
```
Jasopaum, line 138: Is build a type or a value? You use `isinstance(call, build)` and type(build)` which is weird
Also, I think we can find a better naming for build and call

syft/execution/plan.py
```diff
@@ -163,7 +159,10 @@ def parameters(self):
             return []
 
     def build(self, *args):
-        """Builds the plan.
+        """
+        Outdated!!
```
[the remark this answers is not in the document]
karlhigley: I think the docs haven't kept pace with the `Plan` refactoring.

syft/execution/plan.py
```diff
@@ -113,6 +107,7 @@ def __init__(
         id: Union[str, int] = None,
         owner: "sy.workers.BaseWorker" = None,
         tags: List[str] = None,
+        serialized_input: list = None,
```
karlhigley, line 110: Could we call this field something like `input_types`? I've been finding `serialized_input` a bit confusing.

syft/execution/plan.py
```diff
@@ -229,8 +229,9 @@ def __setattr__(self, name, value):
             for tensor_name, tensor in value.named_tensors():
                 self.__setattr__(f"{name}_{tensor_name}", tensor)
 
-    def __call__(self, *args):
+    def __call__(self, *args, **kwargs):
```
karlhigley, line 232: I don't think this method uses `kwargs`.

syft/execution/plan.py
```diff
@@ -229,8 +229,9 @@ def __setattr__(self, name, value):
...
         """
+        Outdated!!
```
karlhigley, line 234: Let's capture an issue to update the doc strings in `Plan` and associated classes. Then we can save that work for a later PR.

syft/execution/role.py
```diff
@@ -69,15 +68,34 @@ def register_inputs(self, args_):
         """ Takes input arguments for this role and generate placeholders.
         """
         # TODO Should we be able to rebuild?
-        self.input_placeholder_ids = tuple(self._build_placeholders(arg).value for arg in args_)
+        def register_nested_inputs(obj) -> None:
+            if isinstance(obj, (list, tuple)):
+                _ = [register_nested_inputs(elem) for elem in obj]
+            elif isinstance(obj, dict):
+                _ = [register_nested_inputs(v) for _, v in sorted(obj.items())]
+            elif isinstance(obj, FrameworkTensor):
```
karlhigley, line 76: This kind of nested collection traversal is a common pattern between this method and `register_nested_outputs`, and we use a similar pattern elsewhere in the code base. I wonder if we can extract a common traversal utility that accepts a function or a `type->function` dictionary to specify what to do with the leaf elements?

syft/execution/translation/torchscript.py
```diff
@@ -17,13 +17,29 @@ def translate(self):
 
         # Temporarily remove reference to original function
         tmp_forward = plan.forward
-        plan.forward = None
 
-        # To avoid storing Plan state tensors in torchscript, they will be send as parameters
+        # To avoid storing Plan state tensors inside the torchscript,
```
karlhigley, line 21: Seems a bit unrelated to the feature being added in this PR. Is there a separate PR for this fix?

syft/execution/type_wrapper.py
```diff
@@ -0,0 +1,261 @@
+import syft as sy
+from syft.workers.abstract import AbstractWorker
+from warnings import warn
+from typing import Union
+from syft_proto.execution.v1.type_wrapper_pb2 import NestedTypeWrapper as NestedTypeWrapperPB
+from syft_proto.execution.v1.type_wrapper_pb2 import ClassType as ClassTypePB
+
+class NestedTypeWrapper:
```
karlhigley, line 8: Is the purpose of this class to capture the types of input args? If so, there's probably a more descriptive name for it. Maybe something like `InputTypeDescriptor`?

syft/execution/type_wrapper.py
```diff
@@ -0,0 +1,261 @@
...
+        Class for input type serialization and type checking for nested structures.
+    """
+    def __init__(self, nested_type=None):
+        if nested_type:
+            self.serialized_nested_type = NestedTypeWrapper.serialize_nested_type(nested_type)
+        else:
+            self.serialized_nested_type = None
+
+    @staticmethod
+    def serialize_nested_type(input_arg: any) -> Union[list, tuple, dict, type]:
```
karlhigley, line 19: I don't think this is exactly serialization (which would be the responsibility of the serdes.) Maybe it's something like `enumerate_nested_types`?

syft/execution/type_wrapper.py
```diff
@@ -0,0 +1,261 @@
...
+                build: the build/reference argument type.
+                call: the called argument type.
+                path: the nested path to reach that obj.
+
+            Returns:
+                 None
+        """
+        warn(
+            f"{obj_type} {obj_name} {path} has type {build}, while being built with type {call}.",
+            RuntimeWarning,
```
karlhigley, line 58: If the types are mismatched, that seems like a case where we can't gracefully continue execution, right? If so, these should probably be exceptions.

syft/execution/type_wrapper.py
```diff
@@ -0,0 +1,261 @@
...
+            return tuple(result) if isinstance(input_arg, tuple) else result
+
+        if isinstance(input_arg, dict):
+            serialized_dict = {k: NestedTypeWrapper.serialize_nested_type(v) for k, v in input_arg.items()}
+            return serialized_dict
+
+        return type(input_arg)
+
+    @staticmethod
+    def raise_typecheck_warn(obj_type: str, obj_name: str, build: str, call: str, path: str) -> None:
```
karlhigley, line 42: Could these methods take the object as a param and extract the type/name/id/string repr to form the error messages?

syft/serde/protobuf/native_serde.py
```diff
@@ -35,6 +36,28 @@ def _unbufferize_none(worker: AbstractWorker, obj: "Empty") -> "type(None)":
     """
     return None
 
+def _bufferize_type(worker: AbstractWorker, obj) -> ClassTypePB:
+    proto_type = ClassTypePB()
+
+    if isinstance(obj, type):
+        serialized_type = syft.serde.msgpack.serde._simplify(worker, obj)
```
karlhigley, line 43: The Protobuf serde shouldn't depend on the msgpack serde. (We plan to replace msgpack with Protobuf, so the msgpack serde will eventually go away.)

karlhigley:
(Resolving conflicts with `master` might clear up some of the fixes for other issues that crept in here.)

syft/serde/msgpack/native_serde.py
```diff
@@ -364,6 +363,20 @@ def _detail_ndarray(
...
+    module_path = my_type.__module__
+    full_path_type = module_path + "." + my_type.__name__
+    return full_path_type.encode("utf-8")
+
+
+def _detail_type(worker: AbstractWorker, type_represantation: bytes) -> type:
+    string_representation = type_represantation.decode("utf-8")
+    result = pydoc.locate(string_representation)
+    if result is None:
+        return object
```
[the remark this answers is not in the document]
tudorcebere: A generic object for the isinstance operation, currently I am doing this because I am using pydoc to recreate the original type (this assumes that both ends use the same file layout). If pydoc failes, this means that the specific type was not found and I want to drop typechecking on that element. This will go away in time, after the protobuf support becomes big enough and we can rely on the protobuf type (this is actually your idea :+1: ). Might add a warning and maybe a future TODO mark.

syft/execution/type_wrapper.py
tudorcebere: It depends, if they are nested objects (dict, list, tuple), they will be objects (instances of those nested types), if not, build will be a type, call will be an obj (an instance of that type, if the input is correct).

syft/execution/role.py
tudorcebere: Made one for Role, I could make it generic and usable in all classes, but might be awkard to use/to understand, I believe the one for Role is enough.

syft/execution/role.py
```diff
@@ -54,22 +50,47 @@ def input_placeholders(self):
...
         # TODO Should we be able to rebuild?
-        self.input_placeholder_ids = tuple(
-            self._store_placeholders(arg).value for arg in args_ if isinstance(arg, PlaceHolder)
+
+        traversal_function = lambda x: self.input_placeholder_ids.append(
+            self._store_placeholders(x).value
         )
 
+        self.input_placeholder_ids = []
+        Role.nested_object_traversal(args_, traversal_function, PlaceHolder)
```
Jasopaum, line 78: Can't we use the return value of this function to fill `self.input_placeholder_ids`?

syft/execution/role.py
```diff
@@ -54,22 +50,47 @@ def input_placeholders(self):
...
-        self.output_placeholder_ids = tuple(
-            self._store_placeholders(result).value for result in results
+
+        traversal_function = lambda x: self.output_placeholder_ids.append(
+            self._store_placeholders(x).value
         )
 
+        results = (results,) if not isinstance(results, tuple) else results
+        self.output_placeholder_ids = []
+        Role.nested_object_traversal(results, traversal_function, PlaceHolder)
```
Jasopaum, line 91: Same comment as above

syft/execution/role.py
```diff
@@ -111,10 +132,16 @@ def execute(self, args_):
     def _instantiate_inputs(self, args_):
         """ Takes input arguments for this role and generate placeholders.
         """
-        input_placeholders = tuple(
+
+        def traversal_function(obj):
+            placeholder = input_placeholders.pop(0)
+            placeholder.instantiate(obj)
```
Jasopaum, line 138: I'm not sure I have a better solution right now but this doesn't seem very clean

syft/execution/role.py
```diff
@@ -159,31 +186,21 @@ def _store_placeholders(self, obj):
...
-        elif isinstance(obj, PlaceHolder):
+
+        def traversal_function(obj):
             if obj.id.value not in self.placeholders:
                 self.placeholders[obj.id.value] = obj
             return obj.id
-        else:
-            return obj
+
+        return Role.nested_object_traversal(obj, traversal_function, PlaceHolder)
```
Jasopaum, line 195: Wait, isn't it weird to have the object traversal in both `_store_placeholders` and the methods that use it?

syft/execution/role.py
tudorcebere: It was the cleanest way to use a local reference, I could use `functools.partial` to be more explicit about it, but it won't be any fancier I think.

syft/execution/role.py
tudorcebere: `nested_object_traversal` returns the actuall "tree" scheme, while I want the flattened version of the tree, this would require a second pass through it, I don't think it's worth it.

syft/execution/role.py
tudorcebere: Performance wise, it's almost the same, but it's awkard, didn't see it, i'll make `register_inputs` and `register_outputs` not use _store_placeholders anymore (I think _store_placeholders could be removed with a `traversal_function` thing, but i'll leave it like this for readability if you are ok with this.

syft/execution/type_wrapper.py
tudorcebere: The outer recursion is to handle the original input of the function, while the inner one is checking the nested types. I could make it as one, but I think it will be uglier and I should handle the path kinda pretty awkard.

test/execution/test_plan.py
```diff
@@ -904,3 +904,28 @@ def plan_type_warn(dic):
...
+
+    # Result of torch autograd
+    torch_grads = autograd_test(X)
+
+    # Result of traced backprop
+    autograd_test.forward = None
+    plan_grads = autograd_test(X)
+
+    # (debug out)
+    print("Traced Plan:\n", autograd_test.code)
```
Jasopaum: Maybe remove that

PR #3303, opened 2020-04-03
Tracing from inside the placeholders

Jasopaum:
The idea is here is to perform the tracing during plan building from the inside of the placeholders.
This way, the placeholders could directly write actions into their parent plan.

Problems:
- Placeholders having a Role and Roles having Placeholders can lead to infinite recursion on serde if we don't take care.

TODO:
- Modify protobuf serde.


syft/execution/role.py
```diff
@@ -295,15 +293,24 @@ def detail(worker: AbstractWorker, role_tuple: "Role") -> tuple:
...
             placeholders[ph.id.value] = ph
 
-        return Role(
+        role = Role(
             id=id_,
             owner=worker,
             actions=actions,
             state=state,
-            placeholders=placeholders,
+            # placeholders=placeholders,
```
karlhigley, line 301: It might still be okay to pass the `placeholders` here. Should be okay to modify the `placeholders` after they're set on the `Role`, yeah?

syft/execution/role.py
```diff
@@ -295,15 +293,24 @@ def detail(worker: AbstractWorker, role_tuple: "Role") -> tuple:
...
             input_placeholder_ids=input_placeholder_ids,
             output_placeholder_ids=output_placeholder_ids,
         )
+        for ph in placeholders.values():
+            ph.role = role
+        for ph in state.state_placeholders:
+            ph.role = role
+
+        role.placeholders = placeholders
+        role.state = state
```
karlhigley, line 311:
```suggestion
```
If they're passed in to the `__init__`, seems like these are redundant. 🤔

syft/execution/placeholder.py
```diff
@@ -7,7 +7,15 @@
 
 
 class PlaceHolder(AbstractTensor):
-    def __init__(self, owner=None, id=None, tags: set = None, description: str = None):
+    def __init__(
+        self,
+        role=None,
+        tracing=False,
```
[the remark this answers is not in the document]
Jasopaum: I'm not sure because some placeholders are created when a Plan is being built.

syft/execution/placeholder.py
```diff
@@ -24,6 +32,47 @@ def __init__(self, owner=None, id=None, tags: set = None, description: str = Non
         if not isinstance(self.id, PlaceholderId):
             self.id = PlaceholderId(self.id)
         self.child = None
+        self.role = role
+        self.tracing = tracing
+
+    def get_class_attributes(self):
```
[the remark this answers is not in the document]
Jasopaum: This method is used in the hooking process so I don't think I can change its name.
Basically, it's used when wrapping the response of a method (with a Placeholder in this case) and to know which attributes should be the same in the response as in the `self`

syft/execution/plan.py
```diff
@@ -172,25 +172,27 @@ def build(self, *args):
         """
         self.owner.init_plan = self
 
+        # Enable tracing
+        for ph in self.role.placeholders.values():
```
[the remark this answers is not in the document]
Jasopaum: Sure, I like this!

syft/generic/frameworks/hook/hook.py
```diff
@@ -236,10 +236,20 @@ def _hook_syft_placeholder_methods(self, tensor_type: type, syft_type: type):
         comparators to the hooking
         """
 
+        def create_fn(base_method, name):
+            def new_method(self, *args, **kwargs):
```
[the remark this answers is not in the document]
Jasopaum: Oops, yes, I didn't want to think about a name when testing and forgot to change it after. Thanks!

syft/execution/placeholder.py
```diff
@@ -24,6 +32,47 @@ def __init__(self, owner=None, id=None, tags: set = None, description: str = Non
...
+        """
+        Specify all the attributes need to build a wrapper correctly when returning a response.
+        """
+        return {"role": self.role, "tracing": self.tracing, "owner": self.owner}
+
+    @classmethod
+    def handle_func_command(cls, command):
```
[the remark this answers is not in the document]
Jasopaum: Hum... Not sure I follow you there 🤔

syft/execution/role.py
Jasopaum: My problem here is that placeholders have a role and roles have placeholders so on detail, I could have infinite recursion if I don't take care.
My solution (until we find a better one) was to not detail the role of a placeholder on `Placeholder.detail()` but detail the placeholders of a Role and set their role to the current role being detailed.

syft/execution/placeholder.py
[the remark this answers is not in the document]
karlhigley: Ah, good point.

syft/execution/state.py
```diff
@@ -71,13 +71,7 @@ def read(self):
...
                     parent_plan.role.placeholders[placeholder.child.id] = placeholder
 
-        tensors = []
-        for placeholder in self.state_placeholders:
-            # State elements from sub plan should not be reported when read() is used
-            if "#inner" not in placeholder.tags:
-                tensor = placeholder.child
-                tensors.append(tensor)
-        return tensors
+        return [ph for ph in self.state_placeholders if "#inner" not in ph.tags]
```
karlhigley, line 74: Wonder if `#inner` could become a (derived?) property instead of a tag. 🤔

Jasopaum: Sure! I didn't want to change that right now but I'd be glad to do it if everybody agrees 👍

karlhigley: Yeah, not at all required for this PR, but figured it followed in the spirit of the other tag removal you did.

syft/execution/placeholder.py
```diff
@@ -21,11 +29,72 @@ def __init__(self, owner=None, id=None, tags: set = None, description: str = Non
...
+        # Send it to the appropriate class and get the response
+        response = new_type.handle_func_command(new_command)
+
+        # Find first placeholder in args
+        ph_arg = None
+        for arg in args:
+            # TODO throw an error or handle the case where no placeholder provided in args
+            if isinstance(arg, PlaceHolder):
+                ph_arg = arg
+
```
karlhigley, line 78: I suppose this could implemented with a check after the loop:
```suggestion
        # Find first placeholder in args
        ph_arg = None
        for arg in args:
            if isinstance(arg, PlaceHolder):
                ph_arg = arg

        if ph_arg is None:
            raise SomeException
```

syft/execution/plan.py
```diff
@@ -25,6 +29,61 @@
 from syft_proto.execution.v1.computation_action_pb2 import ComputationAction as ComputationActionPB
 
 
+class FrameworkWrapper:
```
karlhigley, line 32: Optional: Might be worth pulling the tracing stuff out to a separate file.

syft/execution/placeholder.py
Jasopaum: Hum, after reflexion, is it possible that no PlaceHolder is there? Being in PlaceHolder.handle_func_command, maybe we don't need any check

karlhigley: Makes sense to me. Let's leave it off.

syft/execution/plan.py
```diff
@@ -25,6 +29,61 @@
...
+        return trace_wrapper
+
+
+@contextmanager
+def trace(package, role):
+    try:
+        wrapped = FrameworkWrapper(package, role)
+        yield wrapped
+    except:
+        raise
```
karlhigley, line 84: Maybe this should just re-raise the original exception?

karlhigley: Oh, right, that's what this does. 🤦

karlhigley:
🚀 🎸 🥇

PR #3314, opened 2020-04-06
Don't call arguments or locals `args` or `kwargs` (unless they're `*args` or `**kwargs`)

karlhigley:
It's a small thing, but it makes debugging hard when the actual `*args` and `**kwargs` are shadowing or shadowed by arguments and locals.


karlhigley:
They have special meaning in `pdb`.

Jasopaum:
Oh right!
I'll approve, even if I'm not sure why one of the tests doesn't pass...

karlhigley:
It looks like an unrelated failure from a flaky test. I merged a change that runs the tests in a different random order every time, which is turning up a lot of those. I'll re-run and see if we get lucky on a retry.

PR #3321, opened 2020-04-07
Adding CKKSTensor: CKKS homomorphic encryption scheme

youben11:
Introduces a new tensor type for evaluating addition, subtraction and multiplication of tensors in a component-wise fashion of encrypted tensors using the CKKS homomorphic encryption scheme.


pip-dep/requirements.txt
```diff
@@ -9,6 +9,7 @@ requests~=2.22.0
 scipy~=1.4.1
 syft-proto~=0.2.9.a1
 tblib~=1.6.0
+git+https://github.com/OpenMined/TenSEAL.git#egg=tenseal
```
karlhigley, line 12: If we're going to add a dependency, we should probably do it via a versioned package.

youben11:
> I think we should probably make `tenseal` an optional dependency with a dependency check to see if it's available. Otherwise LGTM, but I'll defer to others on the new tensor type. +1

Yeah sure, I actually forgot about that, thank you !

pip-dep/requirements.txt
youben11: We have some packaging issues on the TenSEAL side, that we can't deal with for the moment, is it impossible to do it this way?

karlhigley: Well, I'm not saying we absolutely can't, but I'd generally discourage it since it can make PySyft brittle when changes happen upstream. What are the upstream packaging issues? TenSEAL is an OpenMined library, so maybe we should iron those out and get a versioned release published.

youben11: TenSEAL has a binary extension that need to be packaged, and we should automate the build and packaging for different platforms, and we have some other priorities ATM. I'm working on TenSEAL and will be around if any changes comes in. We should just consider it a temporary thing until we get it published.

iamtrask: Hmm - i think this temporary thing is a big enough detraction from packaging standards to warrant holding it out of master.  Happy to push things in a branch for now though until it's ready.

youben11:
I'm currently not sure either to make TenSEAL an optional dependency or not :thinking:

pip-dep/requirements.txt
karlhigley: Alternately: make it an optional dependency, and pin the dependency to a specific commit. The issue with brittleness comes from optimistically assuming that future versions will continue to be compatible (like PySyft used to do.)

youben11: Just pushed a package that should work on most linux distro, next is to automate the build for other platforms.

pip-dep/requirements.txt
```diff
@@ -9,6 +9,7 @@ requests~=2.22.0
 scipy~=1.4.1
 syft-proto~=0.2.9.a1
 tblib~=1.6.0
+tenseal
```
karlhigley, line 12: This dependency should specify a version:
```suggestion
tenseal~=0.1.0.dev0
```

youben11: done

youben11:
Ping @LaRiffle @iamtrask

syft/frameworks/torch/tensors/interpreters/ckks.py
```diff
@@ -0,0 +1,146 @@
...
+        homomorphic encryption.
+
+        Args:
+            context: a TenSEALContext created using 
+                syft.frameworks.torch.he.tenseal.context()
+            scale: the scale to be used to encode tensor values.
+        """
+
+        output = CKKSTensor().on(self.child)
+        output.child.encrypt_(context)
```
Syzygianinfern0, line 37: Shouldn't `scale` be passed as a arg here too?

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -1036,10 +1039,23 @@ def encrypt(self, protocol="mpc", **kwargs):
 
             return x_encrypted
 
+        elif protocol.lower() == "ckks":
+            context = kwargs.get("public_key")
+            scale = kwargs.get("scale")
+            if scale is None:
+                scale = DEFAULT_CKKS_SCALE
```
Syzygianinfern0, line 1046: Can be simplified to `val = kwargs.get("val", "default value")` to be consistent with the current code style

syft/frameworks/torch/tensors/interpreters/ckks.py
```diff
@@ -0,0 +1,146 @@
...
+        torch tensor.
+
+        Args:
+            secret_key: optional if the tensor was created in 
+                the current context. Used to decrypt the tensor.
+        """
+        if secret_key is None:
+            x_decrypted = self.child.decrypt()
+        else:
+            x_decrypted = self.child.decrypt(secret_key)
```
Syzygianinfern0, line 65: This calls TenSeal's CPP magic I suppose. Is that right? 
Beautiful 😍

syft/frameworks/torch/tensors/interpreters/native.py
youben11: True !

syft/frameworks/torch/tensors/interpreters/ckks.py
youben11: yeah the child is a c++ object

syft/frameworks/torch/tensors/interpreters/ckks.py
youben11: Well spotted !

syft/frameworks/tenseal/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+    poly_modulus_degree=DEFAULT_CKKS_N, coeff_mod_bit_sizes=DEFAULT_CKKS_COEFF_MOD
+):
+    """Returns a public context (containing public keys and attributes)
+    and the secret key.
+    """
+    c = context(SCHEME_TYPE.CKKS, poly_modulus_degree, 0, coeff_mod_bit_sizes)
+    secret_key = c.secret_key()
+    c.make_context_public()
+
+    return c, secret_key
```
LaRiffle, line 20: Do we want to define tenseal as a framework? it looks to me more as a tensor dependency, though I'm not sure how it should be modeled.

youben11: So TenSEAL is more about general use, we don't want to put all the pysyft glue there, but as we use it with pysyft, we have to adapt some function to the framework. So I thought that a place where key generation and default parameters would be is in `pysyft.frameworks.tenseal` but I'm also not sure.

LaRiffle: @karlhigley @Jasopaum your feedback might be useful

Syzygianinfern0:
```
ERROR: Could not find a version that satisfies the requirement tenseal==0.1.2.dev0 (from -r pip-dep/requirements.txt (line 12)) (from versions: 0.1.1a0)
ERROR: No matching distribution found for tenseal==0.1.2.dev0 (from -r pip-dep/requirements.txt (line 12))
```
Latest test. Maybe we would like to update that if we are having tenseal as a dep.

iamtrask:
Ready to merge one conflicts are resolved! 🙂

eric-vader:
I am interested in the changes, @youben11
Any plans to still merge the branch since its approved?

youben11:
Hi @eric-vader ! Unfortunately, this changes won't be merged soon due to a lot of technical challenges we faced. You can still look into [TenSEAL](https://github.com/OpenMined/TenSEAL) if you need to perform ML with HE. TenSEAL currently doesn't have a general purpose tensor type, which is expected on the PySyft side, so we preferred to make the integration once it's ready.

PR #3323, opened 2020-04-08
Call native func with original name if not overloaded

youben11:
When a pure torch tensors gets called with a torch function, the behavior is to call the function name prefixed with 'native_', but this wouldn't work if the function isn't overloaded, so we may check first if the `native_func_name` exists and call `func_name` directly if it doesn't.


gmuraru:
You have to add a test such that codecov will pass

karlhigley:
There are already coverage tests included in the main test suite and Codecov is finicky, so I'd ignore it for now. @imskr, any idea how we can tweak Codecov to stop failing on tiny coverage changes?

karlhigley:
(Adding a test is still a good idea, of course.)

imskr:
@karlhigley we need to add some parameters to ignore tiny changes.

youben11:
Well, other tests are failing here.

gmuraru:
Currently, it fails in test ```test_instantiate_tfe_layer``.
On this
```
E       AssertionError: 
E       Not equal to tolerance rtol=0.001, atol=0
E       
E       Mismatched elements: 4 / 20 (20%)
E       Max absolute difference: 0.00031287
E       Max relative difference: 0.00110612
E        x: array([[-0.163999, -1.074379, -1.394909, -0.963268,  0.888889],
E              [-0.163999, -1.074379, -1.394909, -0.963268,  0.888889],
E              [-0.163999, -1.074379, -1.394909, -0.963268,  0.888889],
E              [-0.163999, -1.074379, -1.394909, -0.963268,  0.888889]])
E        y: array([[-0.164181, -1.074468, -1.395135, -0.9634  ,  0.889202],
E              [-0.164181, -1.074468, -1.395135, -0.9634  ,  0.889202],
E              [-0.164181, -1.074468, -1.395135, -0.9634  ,  0.889202],...
```

This is probably due to the random order of the tests. Will try to re-run the tests and check if it passes - if this continues to happen on the PRs (this particular failing we should open an issue)

PR #3324, opened 2020-04-08
Translation to German Tutorial "Part 13a"

JMBehnken:
Translation of the thirteenth tutorial "Part 13a" to german by native speaker.

The notebooks depend on each other (e. g. saved model and started servers).

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,6 +1,6 @@
 coverage:
     range: 95..100
-    round: down
+    round: up
     precision: 2
 
 comment:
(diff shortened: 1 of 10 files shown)


JMBehnken:
I do not understand why the tests fail. By running the notebooks locally everything works out fine and my code is exactly the same as in the english original. I have checked this multiple times.

Can you point me to some resources to learn why this filas or how to avoid it?

karlhigley:
Looks like flaky tests to me. I updated the automated testing workflows to skip the core library tests when PRs only change tutorials/translations, which should address that.

karlhigley:
Merging the latest `master` into these branches should pull in the fix I mentioned, but I'm unable to fetch the code and do that myself since the branch names contain double quotes, which git thinks is an invalid remote branch to fetch from.

PR #3327, opened 2020-04-08
Gradient clipping - add tolerance

gmuraru:
This PR is opened as a result of making the tests more robust.

There was a problem with one [test](https://github.com/OpenMined/PySyft/pull/3317) and this PR tries to solve the problem


syft/frameworks/torch/hook/hook.py
```diff
@@ -810,12 +810,7 @@ def param_is_pointer_tensor(param):
                     total_norm = 0
                 for p in parameters:
                     param_norm = p.grad.data.norm(norm_type)
-                    # Remote PySyft tensor
-                    if param_is_pointer_tensor(p):
-                        total_norm += param_norm ** norm_type
-                    # Local PySyft tensor
-                    else:
-                        total_norm += param_norm.item() ** norm_type
+                    total_norm += param_norm ** norm_type
```
gmuraru, line 813: Change such that both return tensors

karlhigley:
Tutorials failures are unrelated, merging.

PR #3334, opened 2020-04-10
Whitelist `serde` from Random-Order Testing

Syzygianinfern0:
Addresses issue in `test_serde_simplification` which causes tests to fail.

The `serde` isn't supposed to be run in random order as the serialization has to be complete. But random testing includes asynchronous execution so _maybe_ `ser` isnt complete before `de`


karlhigley:
This would disable the serde tests, which probably isn't what we want.

PR #3352, opened 2020-04-14
Communication actions plus plus (Adds `name` field)

Prtfw:


syft/execution/communication.py
```diff
@@ -17,6 +18,7 @@ class CommunicationAction(Action):
     def __init__(
         self,
         obj_id: Union[str, int],
+        name: str,  # @TODO: maybe refactor into enum later?
```
karlhigley, line 21: Yeah, an `enum` might be a good idea. On `ComputationAction`, there's such a wide variety of methods (from PyTorch and Tensorflow) that can be captured that it didn't seem worth it. Here, since we're only dealing with PySyft methods, it might make more sense.

syft/execution/communication.py
```diff
@@ -28,6 +30,12 @@ def __init__(
         super().__init__()
 
         self.obj_id = obj_id
+        if name in ["move", "remote_send", "remote_get", "get", "share", "share_"]:
+            self.name = name
+        else:
+            raise ValueError(
+                f"name ${name} for CommunicationAction is not in the list of supported actions"
```
karlhigley, line 37: 👍

Prtfw:
@karlhigley 
thanks for the help with syft-proto!... now I can use your "template" for the new field args (assuming we do want to add it)

note: I messed up in my github-fu and accidentally overwrite your commits but i was able to manually find the diff and re-apply them... 

I don't expect this to be working right now (at the very lease because of the new field args for communication actions) but feel free to give it a eyeball incase I am going in the wrong direction. 🙏

syft/generic/pointers/pointer_tensor.py
```diff
@@ -418,9 +432,16 @@ def value(self, *args, **kwargs):
         Returns:
             A pointer to a Tensor
         """
-        command = ("value", self, args, kwargs)
+        ptr_id = self.id_at_location
+        source = self.owner.id
+        _kwargs = {}
+        _args = ()
+        message = TensorCommandMessage.communication(
```
karlhigley, line 439: I'm not sure if this one should be a computation or a communication. 🤔 

The more I think about it the more I think you have it right, but I don't really get what `value` is for since we also have `get`. 🤷

syft/execution/communication.py
```diff
@@ -28,21 +28,31 @@ def __init__(
         super().__init__()
 
         self.obj_id = obj_id
+        if name in ["move", "remote_send", "remote_get", "get", "share", "value", "share_"]:
```
Jasopaum, line 31: What is value? Does it really represent a communication?

syft/execution/communication.py
```diff
@@ -17,8 +15,10 @@ class CommunicationAction(Action):
     def __init__(
         self,
         obj_id: Union[str, int],
+        name: str,  # @TODO: maybe refactor into enum later?
         source: Union[str, int],
         destinations: List[Union[str, int]],
+        args_: Tuple,  # The arguments to the method call, needed for .share()
         kwargs_: dict,
     ):
         """Initialize an communication action
```
Jasopaum, line 24: I think I should have done it earlier sorry, but could you add a few lines to explain what represents each attribute for a CommunicationAction?

syft/generic/pointers/pointer_tensor.py
```diff
@@ -404,12 +413,17 @@ def share(self, *args, **kwargs):
             A pointer to an AdditiveSharingTensor
         """
 
-        # Send the command
-        command = ("share", self, args, kwargs)
+        ptr_id = self.id_at_location
+        source = self.owner.id
+        _kwargs = {}
+        message = TensorCommandMessage.communication(
+            ptr_id, "share", source, [self.location.id], args, _kwargs
```
Jasopaum, line 420: Maybe I'm wrong but I think args for share are really the location where you want to send shares... So maybe this should go into `location` instead of `[self.location.id]` which is the location of the tensor to share (so basically the source of the communication action we want to perform)

syft/execution/communication.py
karlhigley: I wondered the same thing, but it isn't a computation. I think it might be equivalent to `get` but without removing the remote instance? Not sure.

syft/generic/pointers/pointer_tensor.py
karlhigley: Yeah, it probably should, but for this PR I'd be satisfied with just adding names and fixing the things that the extra field breaks. (We can keep refactoring from there of course, but trying to keep the scope of each PR limited.)

syft/messaging/message.py
```diff
@@ -111,11 +111,11 @@ def computation(name, target, args_, kwargs_, return_ids):
         return TensorCommandMessage(action)
 
     @staticmethod
-    def communication(obj_id, source, destinations, kwargs_):
+    def communication(obj_id, name, source, destinations, args_, kwargs_):
```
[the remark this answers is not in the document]
Prtfw: ahh my bad... will fix!

syft/execution/communication.py
Prtfw: took a closer look, it appears value is currently broken and not being used very much at all... (we could potentially remove it)

syft/generic/pointers/pointer_tensor.py
Prtfw: @Jasopaum you are right, the args are just a list of where to SMPC "share" the tensor. this also means we can remove `args` I think, which is great news... I have updated the code to reflect that.

I am using send_msg instead of send_command so I think location is self.owner... 
(because the doc string for `send_msg => args: location = A BaseWorker instance that lets you provide the destination to send the message.`)
plz let me know if this understanding is incorrect 

also destination is where a communication action is being sent so I think that is the ids of the workers we can pull out of `args`... also let me know if this is incorrect

syft/generic/pointers/pointer_tensor.py
Prtfw: (got ride of it : )

syft/execution/communication.py
```diff
@@ -28,6 +27,14 @@ def __init__(
         super().__init__()
 
         self.obj_id = obj_id
+        if name in ["move", "remote_send", "mid_get", "remote_get", "get", "share", "share_"]:
```
karlhigley, line 30: We can probably use this as the list of supported methods to determine whether a particular action is a computation or a communication in the tracing. Could the list move to a (hard-coded) static method?

Prtfw: can do!

PR #3353, opened 2020-04-14
Explict Specification of Encryption Protocol in PaillierTensor Notebook

### LINKED ISSUE #3329: Call decrypt() without specifying the protocol
When we call decrypt on a tensor, we don't want to always provide the protocol if it can be known otherwise.

I think that the current protocol that encrypted the tensor can be checked by looking at the child's type.

### LINKED ISSUE #3333: Need changes in Experimental notebook for PaillierTensor
**To Reproduce**
Run PySyft/examples/experimental/PaillierTensor.ipynb and on running 3rd block it will throw an error.

**Expected behavior**
Should encrypt tensor without throwing the error 😁

**Additional context**
change encrypt and decrypt like:
x = th.Tensor([1,2,3]).encrypt(protocol="paillier", public_key=pub)
out = out.decrypt(protocol="paillier", private_key=pri)

Syzygianinfern0:
Fixes #3333 by updating method to follow current API


Syzygianinfern0:
> PS: it doesn't fix #3329

Yes. It will be made after the #3321 is merged, as mentioned in your issue. That's why it's currently as a `Draft PR`. **Not ready yet.** 
I've fixed #3333 just as it will be related to it.

PR #3354, opened 2020-04-14
[WIP] Make `wrap` methods `return self` (as a trial balloon)

karlhigley:


karlhigley:
Closing this since @gmuraru and @iamtrask are charting a path forward on this refactoring epic.

PR #3356, opened 2020-04-14
Fix flakiness in `test_local_remote_gradient_clipping`

karlhigley:
This test was implicitly relying on `torch.Tensor's` approximate `eq`
when comparing a tensor with a float. Instead, let's compare floats
directly, which makes this test always fail.


karlhigley:
Merging despite unrelated flaky test failure. The game of whack-a-mole continues!

Issue #3344, opened 2020-04-12
`test_instantiate_tfe_layer` is flaky


karlhigley:
**Describe the bug**
This test semi-regularly fails 

**To Reproduce**
Run the test (or full suite) until it fails.

**Expected behavior**
The test should reliably pass.

**Screenshots**
```
        # compares results and raises error if not equal up to 0.001
>       np.testing.assert_allclose(actual, expected, rtol=0.001)
E       AssertionError: 
E       Not equal to tolerance rtol=0.001, atol=0
E       
E       Mismatched elements: 4 / 20 (20%)
E       Max absolute difference: 0.00031287
E       Max relative difference: 0.00110612
E        x: array([[-0.163999, -1.074379, -1.394909, -0.963268,  0.888889],
E              [-0.163999, -1.074379, -1.394909, -0.963268,  0.888889],
E              [-0.163999, -1.074379, -1.394909, -0.963268,  0.888889],
E              [-0.163999, -1.074379, -1.394909, -0.963268,  0.888889]])
E        y: array([[-0.164181, -1.074468, -1.395135, -0.9634  ,  0.889202],
E              [-0.164181, -1.074468, -1.395135, -0.9634  ,  0.889202],
E              [-0.164181, -1.074468, -1.395135, -0.9634  ,  0.889202],...
```

**Additional context**
This started being an issue when the order of the test suite was randomized (specifically to shake out flaky tests like this.)

jvmncs:
There are two potential issues here:
1. Randomized functions in tests should be seeded, since integration tests should be deterministic. By randomizing the order of tests, the seed that these tolerance values are tuned against might be getting reset.
2. The rtol value here is working against the test definition, which is what makes this test dependent on seed.

To explain 2 a bit more:The loss of precision from TFE's fixed point representation truncating floating points is relatively worse for values closer to 0, and sampling kernel_initializer from a standard normal means we have a high likelihood of sampling numbers near 0. Relative error between the expected and actual value of the kernel will thus increase for numbers near 0 (whereas absolute error should be better).

If you want to randomize testing order, I'd double check that you're setting seeds after sampling the testcase order (so that each test can still be deterministic). I'd additionally either change the kernel_initializer to a distribution that doesn't have an expected value at 0 (normal -> uniform), or switch from rtol to atol (this is less satisfying but should at least make the test less susceptible to precision errors).

karlhigley:
[...]

PR #3358, opened 2020-04-14
Adjust comparison thresholds in `test_instantiate_tfe_layer`

karlhigley:
This test has a known issue with sampling from a normal distribution
that might land close to zero, making small absolute errors result in
large relative errors. This bumps the relative tolerance and adds an
absolute tolerance to prevent the test from accepting large absolute
errors on large values that result in small relative errors.

More on this in [Issue #3344](https://github.com/OpenMined/PySyft/issues/3344#issuecomment-612652546).


karlhigley:
Merging despite unrelated serialization test failure to reduce sources of test flakiness in `master`.

PR #3360, opened 2020-04-15
Translate Plan to torchscript on build time

vvmnnnkv:
Next step for #3263
Make translation on Plan build time.
(2nd attempt :))

### DIFF
--- a/syft/workers/base.py
+++ b/syft/workers/base.py
@@ -1181,11 +1181,12 @@ class BaseWorker(AbstractWorker, ObjectStorage):
         if workers is not None:
             if not isinstance(workers, list):
                 workers = [workers]
-            workers.append(self)
         else:
-            # self is referenced in self.__known_workers
-            # NOTE: for some reason self._known_workers may contain a Plan
-            workers = [w for _, w in self._known_workers.items() if isinstance(w, AbstractWorker)]
+            workers = [w for w in self._known_workers.values() if isinstance(w, AbstractWorker)]
+
+        # self is not referenced in self._known_workers when auto_add=False
+        if self not in workers:
+            workers.append(self)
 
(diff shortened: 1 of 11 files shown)


test/serde/msgpack/test_msgpack_serde_full.py
```diff
@@ -105,6 +102,8 @@ def test_serde_coverage():
 @pytest.mark.parametrize("cls", samples)
 def test_serde_roundtrip(cls, workers, hook, start_remote_worker):
     """Checks that values passed through serialization-deserialization stay same"""
+    serde_worker = syft.VirtualWorker(id=f"serde-worker-{cls.__name__}", hook=hook, auto_add=False)
+    workers["serde_worker"] = serde_worker
```
karlhigley, line 106: This is a really good idea! 🎉

syft/workers/base.py
```diff
@@ -1181,11 +1181,12 @@ def serializer(self, workers=None) -> codes.TENSOR_SERIALIZATION:
...
-            workers.append(self)
         else:
-            # self is referenced in self.__known_workers
-            # NOTE: for some reason self._known_workers may contain a Plan
-            workers = [w for _, w in self._known_workers.items() if isinstance(w, AbstractWorker)]
+            workers = [w for w in self._known_workers.values() if isinstance(w, AbstractWorker)]
+
+        # self is not referenced in self._known_workers when auto_add=False
+        if self not in workers:
+            workers.append(self)
```
karlhigley, line 1189: Looks like we converged on a solution here! I think the version in #3370 is a little simpler, but the effect is the same.

karlhigley:
Failures are from known flaky tests, merging.

PR #3361, opened 2020-04-15
Handle torchscript state arg outside of Role

vvmnnnkv:
In https://github.com/OpenMined/PySyft/pull/3263, plan with state was traced to torchscript with extra arg that passes state inside the plan. The last arg was handled in Role.

However, adding this extra argument interferes with #3296 by @tudorcebere

This PR moves handling of the last arg from Role into wrapper func in torchscript translator class, so the Role itself is not affected and gets correct number of arguments.


karlhigley:
Flaky tests strike again. This looks good though!

PR #3362, opened 2020-04-15
[WIP] Complete real world tutorial, use-case for SMPC-FL, PATE-DP, Grid 

NiWaRe:
# Use-case: Image Classifier for prof. Photographers and 
The PATE framework for privacy-preserving self-supervised learning is a very important example, which has many real-world usecases and therefore is also a very well suited vor showcasing the PySyft, PyGrid capabilities. 
I noticed that the framework is implemented and there exist tutorials of using PATE on simple queries but I didn’t find a tutorial putting the complete workflow together. 
After discussing with @LaRiffle it seems to be a good idea to do just that based on a real-world example which is reasonably similar to use-cases as the [Stanford skin-cancer classifier](https://cs.stanford.edu/people/esteva/nature/) 
*This use-cases therefore uses the Cifar10 dataset to train an Image-Classifier based on the scenario of multiple professional photographers wanting to automatically annotate their photographs and/or a model-provider wanting to create/sell a state-of-the-art Object-Classifier.*

# The original approach: (Not possible)
The original idea was to showcase the PATE framework with Encrypted FL and DP based on a Style-Transfer-Example, as this could have potentially visualized the effects of the teacher-aggregation and the laplace mechanisms in an interesting manner. After thinking about aggregating the Gram-Matrices (for the style representation following [this approach](https://www.cv-foundation.org/openaccess/content_cvpr_2016/papers/Gatys_Image_Style_Transfer_CVPR_2016_paper.pdf) ) I didn't find a simple way to combine that with assumed discrete output of the model which is being created in the standard PATE setting. For the style transfer the Gram-Matrices could have been potentially averaged and then used in the style loss to convert any image into a general kind of drawn image (not any particular style, which would have been considered as private in this setting)
If somebody has an idea, he/she can potentially use the idea well for visualizing the combination in the FL process or the noise in the DP-mechanisms. 

I will be working on the chosen tutorial during the next days, corrections, suggestions, etc. are of course always welcome.


NiWaRe:
### Update: First draft of the "Simple Version" (see notes in notebook)
The following sub-components have been tested (quick test)
* **PATE-Setup:** Encrypted distribution of data on to "hosting-groups" (2 workers: the teacher and the student) The role of the model-owner is assumed for the local machine. (To be a neutral third party extra functionality is needed, see *Shortcomings* and introduction in notebook)
* **Encrypted-FL Student Training:** Tested with a VERY simple model (after first epoch 10% -> which is about right as it would equal guessing) Still takes VERY LONG time (no CUDA with SMPC possible)
* **Encrypted Label Prediction of each teacher:** Straight forward (also very slow)
* **Noisy Voting, DP-secured labeling of student, public dataset:** Passive Sec. Approach, trusted aggregator necessary (some kind of `.remote_share()` would need to be possible for that to work)
* **Student Model Training:** Very much faster. Performance not representative because no complete training was done and the complete notebook wasn't run in one go entirely (still running :D )

I got a couple of things to work and improve. The goal is to also use the use-case a experimental ground for production-near tests. <br>
*Any feedback (especially concerning the **general concept** and **how to speed up things** ), ideas, help is welcome.*

NiWaRe:
Much to do lately, working again on it starting next week.

PR #3364, opened 2020-04-15
Add method to add/remove operations from base worker

### LINKED ISSUE #3363: Add support for Crypten in worker
A feature request. Currently, for CrypTen we add all the methods that are "specific" for this framework into the ```BaseWorker``` and we need to have a solution to register/unregister those operations.

Have separate methods to add/remove support for CrypTen

Have dedicated workers, but see the discussion from [here](https://github.com/OpenMined/PySyft/pull/3254)

**Additional context**

gmuraru:
Fixes #3363 

Some initial changes to add the mechanism to add/remove methods from a base worker such that it will be able to support the implemented CrypTen operations.

Currently, there is added only CrypTen support, but I choose to implement it in this way (is only an initial draft) such that we could easily add more frameworks in the future.

WIP: Need to add tests to make sure the add/remove mechanism work as intended.

### DIFF
--- a/syft/frameworks/crypten/context.py
+++ b/syft/frameworks/crypten/context.py
@@ -208,8 +208,6 @@ def run_multiworkers(
             if was_initialized:
                 crypten.init()
 
-            sy.local_worker.remove_crypten_support()
-
             return return_values
 
         return wrapper
--- a/syft/frameworks/crypten/context.py
+++ b/syft/frameworks/crypten/context.py
@@ -25,10 +25,6 @@ def _launch(func, rank, world_size, master_addr, master_port, queue, func_args, 
     for key, val in communicator_args.items():
         os.environ[key] = str(val)
 
-    if rank == 1:
-        import pdb
-
-        pdb.set_trace()
     crypten.init()
     return_value = func(*func_args, **func_kwargs)
     crypten.uninit()
--- a/syft/frameworks/crypten/context.py
+++ b/syft/frameworks/crypten/context.py
@@ -152,6 +152,7 @@ def run_multiworkers(
                 zip(range(1, len(workers) + 1), [worker.id for worker in workers])
             )
 
+            sy.local_worker.add_crypten_support()
             sy.local_worker._set_rank_to_worker_id(rank_to_worker_id)
 
             # Start local party
@@ -206,6 +207,8 @@ def run_multiworkers(
                 thread.join()
             if was_initialized:
                 crypten.init()
+
+            sy.local_worker.remove_crypten_support()
 
             return return_values
 
--- a/syft/frameworks/crypten/context.py
+++ b/syft/frameworks/crypten/context.py
@@ -25,6 +25,10 @@ def _launch(func, rank, world_size, master_addr, master_port, queue, func_args, 
     for key, val in communicator_args.items():
         os.environ[key] = str(val)
 
+    if rank == 1:
+        import pdb
+
+        pdb.set_trace()
     crypten.init()
     return_value = func(*func_args, **func_kwargs)
     crypten.uninit()
--- a/syft/frameworks/crypten/worker_support.py
+++ b/syft/frameworks/crypten/worker_support.py
@@ -93,7 +93,7 @@ def run_crypten_party_jail(worker: BaseWorker, message: CryptenInitJail):
     return ObjectMessage(return_value)
 
 
-def add_support_to_workers(worker):
+def add_support_to_worker(worker):
     worker._message_router[CryptenInitPlan] = types.MethodType(run_crypten_party_plan, worker)
     worker._message_router[CryptenInitJail] = types.MethodType(run_crypten_party_jail, worker)
 
--- a/syft/frameworks/crypten/worker_support.py
+++ b/syft/frameworks/crypten/worker_support.py
@@ -87,7 +87,7 @@ def add_support_to_workers(worker):
     worker._message_router[CryptenInitJail] = types.MethodType(run_crypten_party_jail, worker)
 
     for method in methods_to_add:
-        method_name = method.__name
+        method_name = method.__name__
         if not hasattr(worker, method_name):
             setattr(worker, method_name, types.MethodType(method, worker))
 
--- a/syft/frameworks/crypten/worker_support.py
+++ b/syft/frameworks/crypten/worker_support.py
@@ -1,6 +1,7 @@
 from typing import Dict
 import types
 
+import syft as sy
 from syft.workers.base import BaseWorker
 
 from syft.messaging.message import ObjectMessage
@@ -65,6 +66,8 @@ def run_crypten_party_jail(worker: BaseWorker, message: CryptenInitJail):
     from syft.frameworks.crypten import utils
 
     worker.rank_to_worker_id, world_size, master_addr, master_port = message.crypten_context
+    sy.local_worker.rank_to_worker_id = worker.rank_to_worker_id
+
     ser_func = message.jail_runner
     onnx_model = message.model
     crypten_model = None if onnx_model is None else utils.onnx_to_crypten(onnx_model)
--- a/syft/frameworks/crypten/worker_support.py
+++ b/syft/frameworks/crypten/worker_support.py
@@ -33,6 +33,7 @@ def run_crypten_party_plan(worker: BaseWorker, message: tuple) -> ObjectMessage:
     """
 
     worker.rank_to_worker_id, world_size, master_addr, master_port = message.crypten_context
+    sy.local_worker.rank_to_worker_id = worker.rank_to_worker_id
 
     plans = worker.search("crypten_plan")
     assert len(plans) == 1
--- a/syft/frameworks/crypten/worker_support.py
+++ b/syft/frameworks/crypten/worker_support.py
@@ -1,0 +1,99 @@
+from typing import Dict
+import types
+
+from syft.workers.base import BaseWorker
+
+from syft.messaging.message import ObjectMessage
+from syft.messaging.message import CryptenInitPlan
+from syft.messaging.message import CryptenInitJail
+
+from syft.frameworks.crypten import run_party
+
+
+def get_worker_from_rank(worker: BaseWorker, rank: int) -> BaseWorker:
+    assert hasattr(worker, "rank_to_worker_id"), "First need to call run_crypten_party"
+    return worker._get_worker_based_on_id(worker.rank_to_worker_id[rank])
+
+
+def _set_rank_to_worker_id(worker: BaseWorker, rank_to_worker_id: Dict[int, int]) -> None:
+    worker.rank_to_worker_id = rank_to_worker_id
+
+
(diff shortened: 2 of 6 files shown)


syft/workers/base.py
```diff
@@ -1160,6 +1160,18 @@ def _get_msg(self, index):
 
         return sy.serde.deserialize(self.msg_history[index], worker=self)
 
+    def add_crypten_support(self):
+        """Add CrypTen specific methods"""
+        from syft.workers.worker_support import add_support
```
LaRiffle, line 1165: Probably here the import could be global

syft/workers/worker_support.py
```diff
@@ -0,0 +1,18 @@
...
+from syft.frameworks.crypten.worker_support import add_support_to_worker, remove_support_from_worker
+
+supported_frameworks = {}
+
+if dependency_check.crypten_available:
+    supported_frameworks["crypten"] = [add_support_to_worker, remove_support_from_worker]
+
+
+def add_support(worker: BaseWorker, framework: str) -> None:
+    add_support_to_worker(worker)
```
LaRiffle, line 14: Use your `supported_frameworks` dict :)

syft/frameworks/crypten/worker_support.py
```diff
@@ -0,0 +1,66 @@
...
+    assert rank != None
+
+    return_value = run_party(plan, rank, world_size, master_addr, master_port, (), {})
+    return ObjectMessage(return_value)
+
+
+def add_support_to_worker(worker):
+    worker._message_router[CryptenInit] = types.MethodType(run_crypten_party, worker)
+
+    for method in methods_to_add:
```
LaRiffle, line 53: maybe don't re-add it if it's already there

gmuraru: Done

gmuraru:
> As a suggestion: we may need a function to check if a worker supports a specific framework

Should we keep in each worker a set of supported frameworks? (and have a method with -- ```is_supported(framework_name)``` where we simply check if we support that framework (like we have the methods attached to the worker)

test/crypten/test_context.py
```diff
@@ -1,8 +1,32 @@
...
+
+def test_add_crypten_support(workers):
+    alice = workers["alice"]
+
+    for method in methods_to_add:
+        assert not hasattr(
+            alice, method.__name__
+        ), f"Worker should not have method {method.__name__}"
+
+    alice.add_crypten_support()
```
iamtrask, line 19: What exactly is involved in 'Adding crypten support?" Why would it not be supported by default?

gmuraru: There was a discussion [here] (https://github.com/OpenMined/PySyft/pull/3254).
Also, added screenshots - I hope I did not miss any information
(http://swarm.cs.pub.ro/~gmuraru/OpenMined/pic1.png)
(http://swarm.cs.pub.ro/~gmuraru/OpenMined/pic2.png)

**TL;DR:** We discussed that maybe the framework information might not be contained in the ```BaseWorker``` and we should have a mechanism to register those "specific framework methods". If this is not wanted, I can switch to something else :D

syft/workers/base.py
gmuraru: I left it into the method, because making it global it would create an import error.

```worker_support.py``` needs BaseWorker which is not defined :\

File: pip-dep/requirements.txt [hunk 5]
```diff
@@ -17,3 +17,4 @@ torch~=1.4.0
 tornado==4.5.3
 websocket_client~=0.57.0
 websockets~=8.1.0
+git+https://github.com/gmuraru/CrypTen.git@2aa0c2f14b472a8df6a85be1cfc4630479a10b2f#egg=crypten
```
gmuraru, line 20: This should be removed after merging the [Fix plan](https://github.com/OpenMined/PySyft/pull/3422) changes

syft/frameworks/crypten/context.py
```diff
@@ -159,6 +160,8 @@ def wrapper(*args, **kwargs):
             if was_initialized:
                 crypten.init()
 
+            sy.local_worker.remove_crypten_support()
```
LaRiffle, line 163: I wonder if we should remove crypten support?

syft/frameworks/crypten/worker_support.py
```diff
@@ -0,0 +1,68 @@
...
+        Args:
+            message (CryptenInit): should contain the rank_to_worker_id, world_size, master_addr and master_port.
+
+        Returns:
+            An ObjectMessage containing the return value of the crypten function computed.
+        """
+
+    worker.rank_to_worker_id, world_size, master_addr, master_port = message.crypten_context
+
+    plans = worker.search("crypten_plan")
```
LaRiffle, line 33: .search has changed and now no longer returns pointers right?

LaRiffle: it's request_search which does

syft/frameworks/crypten/worker_support.py
```diff
@@ -0,0 +1,68 @@
...
+            break
+
+    assert rank != None
+
+    return_value = run_party(plan, rank, world_size, master_addr, master_port, (), {})
+    return ObjectMessage(return_value)
+
+
+def add_support_to_worker(worker):
+    worker._message_router[CryptenInit] = types.MethodType(run_crypten_party, worker)
```
LaRiffle, line 51: What is the role of `types.MethodType` ?

LaRiffle: Ah ok :) https://stackoverflow.com/questions/37455426/advantages-of-using-methodtype-in-python

syft/workers/worker_support.py
```diff
@@ -0,0 +1,18 @@
...
+from syft.frameworks.crypten.worker_support import add_support_to_worker, remove_support_from_worker
+
+supported_frameworks = {}
+
+if dependency_check.crypten_available:
+    supported_frameworks["crypten"] = [add_support_to_worker, remove_support_from_worker]
+
+
+def add_support(worker: BaseWorker, framework: str) -> None:
+    supported_frameworks[framework][0](worker)
```
LaRiffle, line 14: I'm not sure this scales: maybe you could replace the list [add_support_to_worker, remove_support_from_worker] by a dict with "standard" keys, so that if you need to has more on them you don't mess up with indices

syft/workers/worker_support.py
gmuraru: Done

syft/workers/worker_support.py
gmuraru: Done

syft/frameworks/crypten/worker_support.py
gmuraru: Hmmm...```request_search``` is used if you want to take a plan from the worker that built the plan, right?

In this scenario, the plan is already sent after it was built (to each worker) and the worker will simply search for the plan in the objects they have.

pip-dep/requirements.txt
```diff
@@ -10,7 +10,7 @@ Pillow~=6.2.2
 requests~=2.22.0
 RestrictedPython~=5.0
 scipy~=1.4.1
-syft-proto~=0.3.0
+syft-proto~=0.3.1
```
youben11, line 13: We can go for the 0.4.1 which got released after this one

syft/frameworks/crypten/context.py
```diff
@@ -87,7 +87,7 @@ def _send_party_info(worker, rank, msg, return_values, model=None):
     return_values[rank] = utils.unpack_values(response.object, model)
 
 
-def run_multiworkers(workers: list, master_addr: str, master_port: int = 15463):
```
youben11, line 90: Q: I didn't get why you always change ports George :p

syft/workers/worker_support.py
```diff
@@ -0,0 +1,22 @@
...
+from syft.frameworks.crypten.worker_support import add_support_to_worker, remove_support_from_worker
+
+
+supported_frameworks = {}
+
+if dependency_check.crypten_available:
+    supported_frameworks["crypten"] = {
+        "add_support": add_support_to_worker,
+        "remove_support": remove_support_from_worker,
+    }
```
youben11, line 14: What about adding the framework name in these functions?

```suggestion
from syft.frameworks.crypten.worker_support import add_support_to_worker as add_crypten_support, remove_support_from_worker as remove_crypten_support


supported_frameworks = {}

if dependency_check.crypten_available:
    supported_frameworks["crypten"] = {
        "add_support": add_crypten_support,
        "remove_support": remove_crypten_support,
    }
```

syft/frameworks/crypten/context.py
gmuraru: Ahh...Because while testing I get the error "port already in use" (if the tests fail...I can change it back)

pip-dep/requirements.txt
gmuraru: Hmm, first I want to merge this [branch](https://github.com/OpenMined/PySyft/pull/3453) and then this should be fixed :D

syft/frameworks/crypten/context.py
gmuraru: So you suggest the following:
If we want to run with a plan/jail we add the support to workers and this support should not be removed, until explicitly the user requests that

syft/frameworks/crypten/context.py
gmuraru: Done

youben11: Ah, that's because you have processes still running in the background. I have a script in my workspace that I call if processes keeps running, it's a single command script, might be helpful:

```bash
#!/bin/bash

kill -9 `ps |grep python | awk '{$1=$1};1' |cut -f1 -d ' '`
```

gmuraru:
This one can be merged -- the coverage is failing -- we need to add more tests :(

youben11:
> This one can be merged -- the coverage is failing -- we need to add more tests :(

That's my fault, I should add full tests for the Jail functionality as soon as I get it to work with training a model.

syft/frameworks/crypten/context.py
gmuraru: ### TODO: Don't remove this.

pip-dep/requirements.txt
```diff
@@ -10,7 +10,7 @@ Pillow~=6.2.2
 requests~=2.22.0
 RestrictedPython~=5.0
 scipy~=1.4.1
-syft-proto~=0.3.0
+syft-proto~=0.4.1
```
gmuraru, line 13: Will change this back since the current PR does not have some changes required for tests to be able to pass -- will bump the version in a future PR

syft/frameworks/crypten/worker_support.py
LaRiffle: All good!

gmuraru:
> As a suggestion: we may need a function to check if a worker supports a specific framework

Ahh...will add this :)

PR #3366, opened 2020-04-15
Restrict `pytest-randomly` version to avoid `3.3.0`

karlhigley:
`pytest-randomly 3.3.0` adds support for `pytest-xdist` but does so in a
way that requires `pytest-xdist` to be installed, which we don't need.
It's probably an accidental change, but until it's resolved we'll just
avoid that version (which breaks our builds.)

### DIFF
--- a/pip-dep/requirements_dev.txt
+++ b/pip-dep/requirements_dev.txt
@@ -1,3 +1,4 @@
+bandit
 black>=10.3b0
 bumpversion>=0.5.3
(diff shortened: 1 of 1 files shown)


Syzygianinfern0:
Can confirm! 
Tested locally 🎉

karlhigley:
There are still failures from flaky tests here, but this addresses the `<INTERNALERROR>` messages that were plaguing all the open PRs.

PR #3371, opened 2020-04-17
feat: add plan translation to tfjs

Nolski:
This is similar to @vvmnnnkv 's translation code except will translate plans to Tensorflow.js. This also sets up ComputationActions to be translated to TensorflowPython as well.


syft/common/util.py
```diff
@@ -3,6 +3,13 @@
 
 import torch
 import numpy as np
+from enum import Enum
+
+
+class TranslationTargets(Enum):
+    PYTORCH = 'torch'
+    TENSORFLOW = 'tf'
+    TENSORFLOW_JS = 'tfjs'
```
karlhigley, line 12: Since translation is limited to `Plans` (at least for now), let's put this somewhere in the `execution/translation` package. It could go in the `__init__.py` for now.

syft/execution/computation.py
```diff
@@ -1,17 +1,22 @@
...
+from pythreepio.utils import Command
+from pythreepio.errors import TranslationMissing
 
 
 class ComputationAction(Action):
     """Describes mathematical operations performed on tensors"""
 
-    def __init__(self, name, target, args_, kwargs_, return_ids):
+    def __init__(self, name, target, args_, kwargs_, return_ids, 
+                 base_framework=TranslationTargets.PYTORCH.value):
```
karlhigley, line 19: Since we'll be translating full `Plans` and probably not inter-mingling actions targeted at different frameworks (yet anyway),  I imagined this kind of logic living in the `PlanTranslator` class. Want to keep the `Action` classes as close to pure data structures as possible.

syft/execution/translation/tfjs.py
```diff
@@ -0,0 +1,21 @@
...
+    """Performs translation from 'list of ops' Plan into torchscript Plan"""
+
+    def __init__(self, plan):
+        self.threepio = Threepio('torch', 'tfjs', None)
+        super().__init__(plan)
+
+    def translate(self):
+        plan = self.plan.copy()
+        for action in plan.role.actions:
+            action.base_framework = TranslationTargets.TENSORFLOW_JS
```
karlhigley, line 20: Seems like this method should do the work of translating from one framework to another directly. Took me a while to grok that the translation was happening on action creation and in the setter.

syft/execution/computation.py
```diff
@@ -36,6 +36,15 @@ def __init__(self, name, target, args_, kwargs_, return_ids):
         self.kwargs = kwargs_
         self.return_ids = return_ids
 
+    def copy(self):
```
Nolski, line 39: I know I'm not supposed to touch `Action`s but having a copy method makes it a bit easier to translate list of actions. I hope this is a reasonable compromise :)

syft/execution/role.py
```diff
@@ -34,6 +35,7 @@ def __init__(
         output_placeholder_ids: Tuple[int, str] = None,
         # General kwargs
         id: Union[str, int] = None,
+        operations: Dict[str, List[Action]] = None
```
Nolski, line 38: Operations here is a `dict` containing the `list` of the corresponding `Action`s translated into other frameworks. I believe this requires us to update syft-proto

syft/execution/translation/threepio.py
```diff
@@ -0,0 +1,56 @@
+from torch import jit
+from pythreepio.threepio import Threepio
+from pythreepio.utils import Command
+from pythreepio.errors import TranslationMissing
+from syft.execution.translation import TranslationTarget
+from syft.execution.translation.abstract import AbstractPlanTranslator
+
+
+class PlanTranslatorThreepio(AbstractPlanTranslator):
```
Nolski, line 9: Base class which will allow us to easily add support for other frameworks using threepio later on

syft/execution/computation.py
karlhigley: Yeah, this is a good addition.

syft/execution/plan.py
```diff
@@ -95,6 +97,8 @@ class Plan(AbstractObject):
         owner: plan owner
         tags: plan tags
         description: plan description
+        base_framework: The underlying framework (pytoch, tensorflow) which the plan is to be executed in
+        frameworks: A list of frameworks which the plan will also support
```
karlhigley, line 101: Seems like `frameworks` might be left over from an intermediate form of this PR or something?

syft/execution/role.py
```diff
@@ -34,6 +35,7 @@ def __init__(
         output_placeholder_ids: Tuple[int, str] = None,
         # General kwargs
         id: Union[str, int] = None,
+        operations: Dict[str, List[Action]] = None,
```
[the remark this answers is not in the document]
karlhigley: If we combine them into one field like:
```python
actions: Dict[str, List[Action]]
```
then maybe the `base_framework` attribute could be used as the key to quickly grab the relevant `Actions`?

syft/execution/translation/threepio.py
```diff
@@ -0,0 +1,56 @@
...
+        threepio = Threepio(self.plan.base_framework, to_framework, None)
+        function_name = action.name.split(".")[-1]
+        if action.target is None:
+            # Translate normally if action isn't a method of a tensor
+            args = action.args
+            cmd = threepio.translate(Command(function_name, action.args, action.kwargs))
+        else:
+            # Otherwise reformat into proper translation
+            args = [action.target, *action.args]
+            cmd = threepio.translate(Command(function_name, args, action.kwargs))
```
tudorcebere, line 25: Could this be refactored into something like:
```
if action.target is None:
    args = ...
else:
    args = ...
cmd = ...
```
or even

```
args = actions.args if ...
cmd = ...
```

test/execution/test_translation.py
```diff
@@ -40,6 +42,55 @@ def foo(x, state):
...
+
+        def forward(self, x):
+            x = self.fc1(x)
+            x = F.relu(x)
+            x = self.fc2(x)
+            return x
+
+    plan = Net()
+    plan.build(th.zeros(10, 2))
+    orig_plan = plan.copy()
```
tudorcebere, line 91: :rocket: :+1:

syft/execution/plan.py
```diff
@@ -95,6 +97,8 @@ class Plan(AbstractObject):
         owner: plan owner
         tags: plan tags
         description: plan description
+        base_framework: The underlying framework (pytoch, tensorflow) which the plan is to be executed in
```
tudorcebere, line 100: Typo on pytorch

syft/execution/translation/threepio.py
```diff
@@ -0,0 +1,57 @@
...
+        new_action.target = None
+        return new_action
+
+    def translate_framework(self, to_framework: str) -> Role:
+        """Translates current plan's Role to specified framework"""
+        role = self.plan.role.copy()
+        plan = self.plan.copy()
+        # Check to see if plan has been translated to this framework yet
+        if plan.roles.get(to_framework, None) is not None:
+            plan.default_frameourcork = to_framework
```
vvmnnnkv, line 37: Is it a real Plan property?

Nolski: :man_facepalming:

test/execution/test_translation.py
```diff
@@ -40,6 +42,64 @@ def foo(x, state):
...
+    assert serde_plan.role.actions[0].name == "tf.mul"
+    assert len(serde_plan.role.actions[0].args) == 2
+
+    # check that translation is not lost after serde
+    serde_plan_full = deserialize(serialize(plan_js))
+    assert serde_plan_full.role.actions[0].name == "tf.mul"
+    assert len(serde_plan_full.role.actions[0].args) == 2
+
+
+@pytest.mark.skip(reason="Missing translation for torch.nn.functional.linear")
```
Nolski, line 82: Are we not allowed to have skipped tests? It would be great to keep this in here until threepio support comes about (which is soon)

karlhigley: Hmm, I agree, we should be able to skip tests. Not sure why this counts toward coverage. Quick fix would probably be to keep the test here but comment it out, and we can uncomment it if we figure out a way to make skipped tests not count toward coverage requirements.

PR #3372, opened 2020-04-17
Skip the privacy attack notebooks during notebook tests

karlhigley:
The black box model inversion notebook is downloading EMNIST and using
up all the space in the automated test environments.


karlhigley:
So many flaky tests!

PR #3373, opened 2020-04-18
Revert "Translate Plan to torchscript"(#3361) and "Handle torchscript arg"(#3360)

Syzygianinfern0:
The failures in tests related to `Plans` of `nn.Module` where `nn.Parameter` not getting hooked successfully is attributed to changes in these commits. Somehow the tests related to the issue passed in the tests of this PR 🤔🤔🤔

#3360 
#3361 
## Tested with this Code
```python
import torch as th
import syft as sy

hook = sy.TorchHook(th)
sharan = sy.VirtualWorker(hook, id="sharan")
crypto_provider = sy.VirtualWorker(hook, id="james")

torch = th
syft = sy

class Net(sy.Plan):
    def __init__(self):
        super(Net, self).__init__()
        self.fc1 = th.nn.Linear(1, 1)

    def forward(self, x):
        return self.fc1(x)


plan = Net()
plan.build(th.tensor([1.2]))
original_weight = plan.fc1.weight.clone()

plan.fix_precision()
plan.fc1.weight.float_prec_()
# ^^^ 💀 Raises AttributeError: 'Parameter' object has no attribute 'child' 💀

assert (plan.fc1.weight - original_weight) < 10e-2
```
<br>

[An example of the tests failing](https://github.com/OpenMined/PySyft/runs/595917181)

I've made sure this test is **not flaky** by running in a loop each with a different seed (just to be on a safe side as anyway seeds won't affect the presence of attributes of an object 😅😅)

### DIFF
--- a/examples/experimental/PaillierTensor.ipynb
+++ b/examples/experimental/PaillierTensor.ipynb
@@ -2,25 +2,13 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
-   "metadata": {},
-   "outputs": [
-    {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
-      "Falling back to insecure randomness since the required custom op could not be found for the installed version of TensorFlow. Fix this by compiling custom ops. Missing file was '/Users/atrask/Laboratory/openmined/PySyft/venv/lib/python3.6/site-packages/tf_encrypted/operations/secure_random/secure_random_module_tf_1.15.0.so'\n"
-     ]
-    },
-    {
-     "name": "stdout",
-     "output_type": "stream",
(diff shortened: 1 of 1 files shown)


Syzygianinfern0:
@karlhigley Please have a look

karlhigley:
Hmm, a bit difficult to understand how they passed before and fail nor if they aren’t flaky. A conflict with something in `master`?

Syzygianinfern0:
> Hmm, a bit difficult to understand how they passed before and fail now if they aren’t flaky. A conflict with something in `master`?

Yes. I felt the same too. Maybe @vvmnnnkv was not on a recent version of master.

PR #3375, opened 2020-04-19
Rework `PlanTranslatorTorchscript` so it doesn't strip `Parameters`

karlhigley:
## Description

It turns out `jit.trace()` strips `Parameters` out of the tensor chain,
which causes issues with `fix_precision`. Don't fully understand why
`jit.trace()` does that, but operating on a copy of the original `Plan`
avoid mutating the original `State` tensors, which fixes the issue.

## Type of change

Please mark options that are relevant.

- [ ] Added/Modified tutorials
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

* [ ] I have added tests for my changes
* [ ] I did follow the [contribution guidelines](https://github.com/OpenMined/PySyft/blob/master/CONTRIBUTING.md)
* [ ] I have commented my code following [Google style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).

PR #3376, opened 2020-04-19
Rework `PlanTranslatorTorchscript` so it doesn't strip `Parameters`

Syzygianinfern0:
Resolve Conficts, merge with master
>It turns out `jit.trace()` strips `Parameters` out of the tensor chain,
which causes issues with `fix_precision`. Don't fully understand why
`jit.trace()` does that, but operating on a copy of the original Plan
avoid mutating the original State tensors, which fixes the issue.
@karlhigley #3375


karlhigley:
`test_fit` is known to be flaky. 😢

PR #3383, opened 2020-04-21
Fix `Plan` tracing that involves `backward()`/autograd

karlhigley:
## Description

For training to work on web/mobile, we need `Plan` tracing to be able to handle training, which involves calling `backward()` on a model inside `Plan`. It's not working currently, so this PR adds a failing test as a seed for developing a fix.

## Type of change

## Checklist:

* [X] I have added tests for my changes


karlhigley:
A `requires_grad` parameter could be good. 👍

I'm not entirely sure what changes when building a `Plan` with `requires_grad=True` either, but the PyTorch JIT stops complaining.

vvmnnnkv:
Got simple backward trace working!

I'm not sure we want to trace autograd for all input values with requires_grad. Maybe this should be done by invoking `Plan.build()` explicitly with AutogradTensor arguments and not assume Plan will add them automatically?

syft/execution/computation.py
```diff
@@ -201,3 +202,32 @@ def unbufferize(
             action = ComputationAction(command, target, tuple(args_), kwargs_, tuple(return_ids))
 
         return action
+
```
karlhigley, line 205: 👍 

This seems like the kind of thing that could live in `Action`, so that it can be shared across `Action` types.

syft/execution/placeholder.py
```diff
@@ -146,8 +146,19 @@ def copy(self):
...
+            tags=self.tags,
+            owner=self.owner,
+            shape=self.expected_shape,
+        )
         placeholder.child = self.child
+
+        if self.tracing:
+            command = ("copy", self, tuple(), {}), placeholder
+            self.role.register_action(command, syft.execution.computation.ComputationAction)
+
```
karlhigley, line 161: Hadn't considered that we might want to do `copy`ing inside `Plans`!

syft/execution/plan.py
```diff
@@ -201,9 +203,16 @@ def build(self, *args):
             for arg in args
         )
 
+        # Wrap arguments that require gradients with AutogradTensor
+        # Note that AutogradTensor comes before Placeholder in the chain
+        # so that all operations that happen inside AutogradTensor are recorded by Placeholder
```
karlhigley, line 208: Not sure I follow this comment. What wouldn't work if the chain was `Placeholder > AutogradTensor > ... other tensors in chain ...`?

syft/execution/plan.py
```diff
@@ -622,6 +636,65 @@ def unbufferize(worker: AbstractWorker, protobuf_plan: PlanPB) -> "Plan":
...
+        input_names = {ph.id: f"_arg{i+1}" for i, ph in enumerate(inputs)}
+        output_names = {ph.id: f"_out{i+1}" for i, ph in enumerate(outputs)}
+        state_names = {ph.id: f"_state{i+1}" for i, ph in enumerate(states)}
+        var_names = {}
+
+        out = ""
+        out += f"def {self.name}("
+        out += stringify(inputs)
+        out += "):\n"
+        for action in self.role.actions:
```
karlhigley, line 680: This part seems partially redundant with the new method on `ComputationAction`. Can this delegate to the `__str__` methods on the component classes?

Also, can it be a `__str__` method itself?

test/execution/test_plan.py
```diff
@@ -708,3 +709,38 @@ def plan_test_2(x, y):
...
+    # Result of torchscript'ed traced backprop
+    ts_plan_grads = autograd_test.torchscript(X)
+
+    # (debug out)
+    print("Traced Plan:\n", autograd_test.code)
+    print("Torchscript Plan:\n", autograd_test.torchscript.code)
+
+    # Test all results are equal
+    assert torch_grads.eq(plan_grads).all()
+    assert torch_grads.eq(ts_plan_grads).all()
```
karlhigley, line 746: This seems like two tests rolled into one. The Torchscript translation part should probably go to `test_translation.py`.

karlhigley:
> I'm not sure we want to trace autograd for all input values with requires_grad.

When would we not want that?

>  Maybe this should be done by invoking Plan.build() explicitly with AutogradTensor arguments and not assume Plan will add them automatically?

I like the idea of being able to enable/disable that functionality, but I think it would ideally be explicit, like an `autograd=True` kwarg or something.

syft/execution/placeholder.py
vvmnnnkv: I recall you mentioned that you don't want to capture `send` action in `placeholder.send`, but rather in same place as other commands are captured.
Perhaps we should try doing same for copy?

syft/execution/plan.py
vvmnnnkv: If placeholder comes before autograd in chain, trace will record `backward` method call on autograd tensor, but not anything that happens inside backward. So trace will just have `backward` command, which won't work in torchscript.
We'll probably want to do same thing with other custom tensor types, so all ops are captured right before the torch tensor.

syft/execution/placeholder.py
karlhigley: I think it's fine for now.

syft/execution/plan.py
karlhigley: Hmm, okay. Let's leave that for a later PR.

syft/execution/computation.py
vvmnnnkv: I checked CommunicationAction, it seems to have different structure.
If we'll have such method in Action, it will need to distinguish between communication/computation.
It's probably better for communication to have own method?

syft/execution/plan.py
vvmnnnkv: Some time earlier code representation was removed from `__str__` so I figured it's better to make it separate method.

vvmnnnkv: Removed duplication by using communication action's method to create string representation.
Though, it's not very elegant because single communication action doesn't track placeholder's names, so we need to pass variable names.

vvmnnnkv:
> I like the idea of being able to enable/disable that functionality, but I think it would ideally be explicit, like an autograd=True kwarg or something.

Added `trace_autograd` kwarg `func2plan` and `build` methods.

syft/execution/computation.py
karlhigley: The structure is different, but they should at least partially converge. This method doesn't seem to be very structure dependent though, other than the list of attributes to `stringify`. Seems like there could be a common, general method backing them both, maybe with args to specify which fields to `stringify` or something.

Could happen here or be deferred for a later PR. 🤷

syft/execution/plan.py
```diff
@@ -159,7 +162,7 @@ def parameters(self):
         else:
             return []
 
-    def build(self, *args):
+    def build(self, *args, trace_autograd=False):
```
Jasopaum, line 165: Hum, I'm not sure having a trace_autograd attribute and a trace_autograd arg in build is the best solution. I think that you did that because of __call__ but maybe both could use the attribute? Or __call__ could also take it as an argument even if it seems odd

syft/execution/plan.py
```diff
@@ -607,6 +632,27 @@ def unbufferize(worker: AbstractWorker, protobuf_plan: PlanPB) -> "Plan":
 
         return plan
 
+    @property
+    def code(self) -> str:
```
Jasopaum, line 636: Why is this called code and not __str__?

syft/execution/placeholder.py
```diff
@@ -157,6 +168,36 @@ def create_from(tensor, owner, role=None, tracing=False):
         """
         return PlaceHolder(owner=owner, role=role, tracing=tracing).instantiate(tensor)
 
+    @staticmethod
+    def insert(tensor, after, owner, role=None, tracing=False):
```
Jasopaum, line 172: Just a side comment:
With these, are Placeholders now Tracers? ^^

syft/execution/plan.py
vvmnnnkv: There was something like that in `__str__` before and it was removed. I guess it might be too large and inconvenient for str representation. `.code` is inspired by pytorch's torchscript module :)

syft/execution/placeholder.py
karlhigley: This part seems weird to me too. Starting to think `Placeholder` has too many responsibilities, and maybe we should split placeholding and tracing into separate classes.

syft/execution/plan.py
vvmnnnkv: `trace_autograd` is Plan build option, so I added it as kward in `build` method.
And also it can be specified with decorator. So you can specify it when building the Plan both ways.

syft/execution/placeholder.py
vvmnnnkv: Hmm, why Tracers? `.insert` is similar to `.create_from`, with extra arg to create Placeholder deeper in chain.

karlhigley: Because `Placeholders` weren’t designed to go deeper in the chain, they were supposed to hold the place of a tensor chain. Putting them deeper in the chain makes tracing work, but changes how the class is used in a way that makes the name inaccurate.

karlhigley: (To be clear, I’m still okay with merging this and resolving the naming issue later, but I do see an issue. It might even be the one @Jasopaum is referring to.) :smile:

Jasopaum: I agree with what was said here 🙂

PR #3386, opened 2020-04-21
Adding a Protobuf/Msgpack Serializable Interface

tudorcebere:
## Description

Currently, if you want to enrol a type into a protobuf schema, you have to edit syft/serde/protobuf files, which is not very natural and might not scale in time. This solution generates those structures at runtime by checking which class is implementing the interface.

~placed the interface in syft/interfaces, placing it in syft/serde will create some awkward cyclic 
dependency import problem.~ Solved this one by removing AbstractWorker from syft/serde/serde.py and making the msgpack import local.

~currently, I'm exploiting the fact that in __subclasses__ we find only the classes that directly inherits the interface. By doing this, I can be sure on the fact that the class won't rely on the implementation from its parent class and implement it's own. (so we won't have to deal with the inheritance chain at unbufferize). This idea could be bad, because of having to inherit every time ProtobufInterface. Any feedback would be great.~ Thanks @gmuraru + @karlhigley  for the cool tips! :)

* merged the two interfaces into one.

* removed global code might rely on runtime variables, wrote init functions for global variables in serde/probuf/serde.py, serde/protobuf/proto.py and serde/msgpack/serde.py. 

## Type of change

## Checklist:


syft/interfaces/protobuf_interface.py
```diff
@@ -0,0 +1,26 @@
...
+        return bufferize_implemented and unbufferize_implemented and get_protobuf_schema_implemented
+
+    original_subclasses = {s for s in cls.__subclasses__() if check_implementation(s)}
+    sub_sets = {
+        s for c in cls.__subclasses__() for s in get_all_subclasses(c) if check_implementation(s)
+    }
+    return original_subclasses.union(sub_sets)
+
+
+class ProtobufInterface:
```
karlhigley, line 15: 👍 

Can we apply this same class to msgpack serialization with `simplify`/`detail` and call it something like `Serializable`?

tudorcebere: Sure! I could add a wrapper for libraries as well to have a similar behavior,

karlhigley: Libraries?

tudorcebere: Instead of keeping a list with those objects from libraries (torch, numpy) we can use in msgpack/protobuf, it might be useful to have a wrapper with the same behaviour as this interface? Might be overkill.

tudorcebere:
PaillierTensor implements simplify/detail, but it's not available in proto.json. Should we add him there?

syft/__init__.py
```diff
@@ -39,22 +39,19 @@
...
-from syft.grid.public_grid import PublicGridNetwork
-
-# Import sandbox
-from syft.sandbox import create_sandbox, make_hook
-
-# Import federate learning objects
-from syft.frameworks.torch.fl import FederatedDataset, FederatedDataLoader, BaseDataset
-from syft.federated.train_config import TrainConfig
+# communication interfaces
+import syft.interfaces
```
karlhigley, line 43: These should probably live in the `serde` package with the other serialization code.

syft/__init__.py
```diff
@@ -84,6 +81,17 @@
...
+# Import grids
+from syft.grid.private_grid import PrivateGridNetwork
+from syft.grid.public_grid import PublicGridNetwork
+
+# Import sandbox
+from syft.sandbox import create_sandbox, make_hook
+
+# Import federate learning objects
+from syft.frameworks.torch.fl import FederatedDataset, FederatedDataLoader, BaseDataset
+from syft.federated.train_config import TrainConfig
```
karlhigley, line 93: We have an outstanding TODO to vaporize `TrainConfig`, which is definitely out of scope for this PR, but I'd avoid importing it here just so we don't give the impression that people should be using it.

syft/execution/action.py
```diff
@@ -1,7 +1,9 @@
 from abc import ABC
+from syft.interfaces.protobuf_interface import ProtobufInterface
+from syft.interfaces.msgpack_interface import MsgpackInterface
 
 
-class Action(ABC):
+class Action(ABC, ProtobufInterface, MsgpackInterface):
```
karlhigley, line 6: For the sake of brevity, could we combine these interfaces and call the resulting class `SyftSerializable` or something like that?

syft/serde/msgpack/serde.py
```diff
@@ -39,49 +39,15 @@
...
-from syft.execution.protocol import Protocol
-from syft.messaging.message import TensorCommandMessage
-from syft.messaging.message import ObjectMessage
-from syft.messaging.message import ObjectRequestMessage
-from syft.messaging.message import IsNoneMessage
-from syft.messaging.message import GetShapeMessage
-from syft.messaging.message import ForceObjectDeleteMessage
-from syft.messaging.message import SearchMessage
-from syft.messaging.message import PlanCommandMessage
-from syft.messaging.message import WorkerCommandMessage
```
karlhigley, line 71: 🚀 🎸

syft/serde/protobuf/proto.py
```diff
@@ -57,21 +57,11 @@
...
-    ObjectMessage: ObjectMessagePB,
-    TensorCommandMessage: CommandMessagePB,
-    CommunicationAction: CommunicationActionPB,
-    ComputationAction: ComputationActionPB,
-    PlaceHolder: PlaceholderPB,
-    Plan: PlanPB,
-    PointerTensor: PointerTensorPB,
-    Protocol: ProtocolPB,
-    Role: RolePB,
-    State: StatePB,
```
karlhigley, line 72: 🚀 🎸

syft/serde/protobuf/serde.py
```diff
@@ -43,20 +44,7 @@
...
-    PlaceholderId,
-    PlaceHolder,
-    Plan,
-    PointerTensor,
-    Protocol,
-    Role,
-    State,
-    TensorCommandMessage,
-]
+OBJ_PROTOBUF_TRANSLATORS = list(get_protobuf_subclasses(ProtobufInterface))
```
karlhigley, line 47: 👍

test/serde/msgpack/test_msgpack_serde_full.py
```diff
@@ -88,6 +88,7 @@
 
 samples[syft.exceptions.GetNotPermittedError] = make_getnotpermittederror
 samples[syft.exceptions.ResponseSignatureError] = make_responsesignatureerror
+samples[syft.frameworks.torch.tensors.interpreters.paillier.PaillierTensor] = make_paillier
```
karlhigley, line 91: Good catch!

syft/__init__.py
tudorcebere: I could to that, but it will generate the horror named cyclic dependency, I could hack it with a try and catch, I saw it in a few places around the codebase. (or we want to get rid of such hackst?).

syft/__init__.py
tudorcebere: Will do, the only problem is that because of global code usage in serde.py, the usage of that interface requires that all of the classes that uses the interfaces are loaded before the global code execution. I'll think of some workaround on it or maybe we want to remove the global code at the expense of some performance? (or maybe at least do some caching, like, we wait for actual execution and afterwards we store the result globally)

syft/execution/action.py
tudorcebere: I have no idea why I haven't done this in the first place. Nice one.

test/serde/serde_helpers.py
```diff
@@ -1955,3 +1955,26 @@ def compare(detailed, original):
             "cmp_detailed": compare,
         }
     ]
+
+
+def make_paillier(**kwargs):
```
karlhigley, line 1960: If it won't block this PR, we could create an issue and kick this part over to the Crypto team.

syft/serde/serde.py
```diff
@@ -7,17 +7,13 @@
...
-
 ## SECTION:  High Level Public Functions (these are the ones you use)
 def serialize(
     obj: object,
-    worker: AbstractWorker = None,
+    worker=None,
     simplified: bool = False,
     force_full_simplification: bool = False,
-    strategy: Callable[[object, AbstractWorker], bin] = msgpack.serialize,
+    strategy: Callable[[object], bin] = None,
```
karlhigley, line 16: It looks like this changes the call signature of the serialization strategies, but I'm not sure I saw a change to the strategies themselves. Do these still match up?

tudorcebere: Yes, I wanted to detach the AbstractWorker import from the file to prevent the circular import, I'll think for a way to bring this back.

syft/serde/msgpack/serde.py
```diff
@@ -169,6 +119,9 @@ def _force_full_simplify(worker: AbstractWorker, obj: object) -> object:
     Returns:
         The simplified object.
     """
+
+    init_global_vars_msgpack()
```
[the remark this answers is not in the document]
tudorcebere: This makes us rely on order of initialization and could create really awkard and hard to find bugs, for example, let's say that when we are loading the first time syft, syft/serde/msgpack/serde.py should be the last one to be loaded (so that all the classes should be loaded to actually get themselves enroled in the interface). I enjoy this lazy init because: 1. forces everybody to use the public API, not the global variables, 2. 100% safe, we are generating those global variables when syft is fully loaded and the interface has all it's subclasses enrolled. The other way would be to mess around in `__init__.py` files, I think it may be better to leave those alone.

syft/serde/serde.py
karlhigley: Makes sense re: avoiding a circular import. I just want to make sure these type annotations are correct(-ish). Maybe it could be `Callable[[object, object]], bin` for now?

syft/serde/syft_serializable.py
```diff
@@ -0,0 +1,94 @@
...
+    sub_sets = {
+        s
+        for c in cls.__subclasses__()
+        for s in get_msgpack_subclasses(c)
+        if check_implementation(s)
+    }
+    return original_subclasses.union(sub_sets)
+
+
+class SyftSerializable:
```
Jasopaum, line 59: Do you think this should be an abstract class and its methods abstract methods?

syft/serde/serde.py
```diff
@@ -40,14 +36,15 @@ def serialize(
     Returns:
         binary: the serialized form of the object.
     """
+    if strategy is None:
+        from syft.serde.msgpack import serialize
```
Jasopaum, line 40: Is this here to avoid some kind of circular imports?

syft/serde/msgpack/serde.py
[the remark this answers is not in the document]
Jasopaum: I'd also prefer if it was done in an __init__.py file but if you think it's too complex, I'm ok with leaving it for now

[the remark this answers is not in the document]
karlhigley: These are all good points! What if the global was an object with a getter that did the lazy initialization inside? Then we wouldn't have to call the `init_global_vars` function in so many places, because it would be built in to the getter.

syft/serde/syft_serializable.py
karlhigley: Oh, good catch. Yeah, this should never be instantiated directly. What do abstract methods look like in comparison to raising `NotImplementedError`?

Jasopaum: I think you just have to declare the method with an `abstractmethod` decorator and do nothing in it.
A difference is that an error will be raised as soon as you create an instance of a class inheriting from the abstract one if it doesn't implement all the abstract methods.
At least, that's what I remember from abstract classes in Python ^^

karlhigley: That sounds good. 👍

karlhigley: (The three of us write some pretty decent Python together.) 😆

test/serde/serde_helpers.py
tudorcebere: I'll create an issue after the merge.

syft/serde/syft_serializable.py
tudorcebere: This will generate some headache on method resolution order (MRO), for example:
SyftSerializable(ABC)
AbstractWorker(ABC, SyftSerializable)

We should drop the ABC from AbstractWorker by doing this (and all the ABCs around the codebase that implements SyftSerializable). Might be a cool idea, not sure if we want it, not sure how it will behave when somethings like this happens, I need to make a few experiments: (here we can't drop Interface2 from the inheritance chain of Interface3)

Interface1(ABC)
Interface2(Interface1)
Interface3(Interface2, SyftSerializable).

tudorcebere: @karlhigley  any thoughts about this?

syft/serde/serde.py
tudorcebere: Yes, as a temporary solution, we'll leave it here and figure out a solution to break those circular changes in a more elegant way.

syft/serde/msgpack/serde.py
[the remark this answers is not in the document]
tudorcebere: My problem is here:
```
from syft.serde.syft_serializable import SyftSerializable, get_protobuf_subclasses

print(len(get_protobuf_subclasses(SyftSerializable))) #12

class Example(SyftSerializable):
    @staticmethod
    def simplify(worker, obj):
        pass

    @staticmethod
    def detail(worker, obj):
        pass

    @staticmethod
    def bufferize(worker, obj):
        pass

    @staticmethod
    def unbufferize(worker, obj):
        pass

    @staticmethod
    def get_protobuf_schema():
        pass

print(len(get_protobuf_subclasses(SyftSerializable))) #13
```
If those globals were generated and a new class is added, the globals will not reflect this, but I figured a way to do this properly I think without having inits everywhere/checks/performance impact. (currently, i'm solving this only if the classes are a part of syft, but in time that might not be the case).

syft/serde/syft_serializable.py
karlhigley: Let's leave it as is for now. Doesn't seem worth it to incur those headaches.

karlhigley: What happens if we switch the order of the parent classes?
```
AbstractWorker(SyftSerializable, ABC)
```
Seems like MRO should work out in that case?

PR #3392, opened 2020-04-23
Drop `custom` support in SecureNN, Flexibility over dtype in ASTs

### LINKED ISSUE #3389: AST dtype arg should only accept long and int
Currently the dtype arg for initialising AdditiveSharingTensor accepts None, "long", "int", "custom". But as an attempt for standardisation we don't want user to be able to call with arg "custom" it is there only for internal operations(like secureNN).

Raise a ValueError with a response along the lines of "Invalid value for dtype arg, use long or int"

**Additional context**

Syzygianinfern0:
## Description
Fixes #3389 

- [x] Assert SecureNN dtype is not `custom`
- [x] Flexibility over dtype in ASTs

## Type of change

## Checklist:


syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -50,25 +50,24 @@ def __init__(
                 which this tensor should be searchable for
             description: an optional string describing the purpose of the
                 tensor
+        Raises:
+            ValueError: Invalid dtype argument (`long` and `int` only)
+            ValueError: Shares is not a dict (optionally wrapped)
```
Syzygianinfern0, line 55: Line #97

LaRiffle, line 55: Can we also support int32, int64 and the associated torch dtypes? https://pytorch.org/docs/stable/tensor_attributes.html#torch.torch.dtype

LaRiffle: To be more flexible for new users

karlhigley:
One possibility to allow internal access while disabling external access might be to use a "private" method, like:
```python
class Example:
    def __init__(self, dtype="int"):
        if !(dtype == "int" or dtype == "long"):
            raise
        self.dtype = dtype
	
    @staticmethod
    def _init_custom(*args, **kwargs):
        del kwargs['dtype']
        example = Example(*args, **kwargs)
        example.dtype = "custom"
        return example
```

syft/frameworks/torch/mpc/securenn.py
```diff
@@ -141,6 +144,16 @@ def select_share(alpha_sh, x_sh, y_sh):
...
+    assert (
+        alpha_sh.dtype != "custom"
+    ), "`custom` dtype shares are unsupported in SecureNN, use dtype = `long` or `int` instead"
+    assert (
+        x_sh.dtype != "custom"
+    ), "`custom` dtype shares are unsupported in SecureNN, use dtype = `long` or `int` instead"
+    assert (
+        y_sh.dtype != "custom"
+    ), "`custom` dtype shares are unsupported in SecureNN, use dtype = `long` or `int` instead"
+
```
LaRiffle, line 156:
```suggestion
    assert (
        alpha_sh.dtype == x_sh.dtype == y_sh.dtype != "custom"
    ), "`custom` dtype shares are unsupported in SecureNN, use dtype = `long` or `int` instead"

```

syft/frameworks/torch/mpc/securenn.py
```diff
@@ -499,6 +521,12 @@ def division(x_sh, y_sh, bit_len_max=None):
     Returns:
         element-wise integer division of x_sh by y_sh
     """
+    assert (
+        x_sh.dtype != "custom"
+    ), "`custom` dtype shares are unsupported in SecureNN, use dtype = `long` or `int` instead"
+    assert (
+        y_sh.dtype != "custom"
+    ), "`custom` dtype shares are unsupported in SecureNN, use dtype = `long` or `int` instead"
```
LaRiffle, line 529:
```suggestion
    assert (
        x_sh.dtype == y_sh.dtype != "custom"
    ), "`custom` dtype shares are unsupported in SecureNN, use dtype = `long` or `int` instead"
```

LaRiffle:
> you can let user enter these new ones for initialisation but inside __init__ based on conditional set self.dtype as long or int only so that it works well with rest of the code without any changes.

Maybe the most reasonable for the moment

Syzygianinfern0:
Dropping `int16` and `int8` support (for the moment). 
For instance, creating a `int16` would set its field to be 2**16 but dtype to be `int`
> set self.dtype as long or int only 

This creates a problem during `serde` operations as **only the dtype is encoded and sent** and is hence when deserialized (from dtype of `int`) it gives a field of 2**32 :disappointed:

Yugandhartripathi:
Looks good and ready for merge!

PR #3393, opened 2020-04-23
Test placeholder forward

gmuraru:
Add a missing test for checking if the placeholder (from the plans) is correctly forwarding an attribute if it is not found in its class.
Taken from [here](https://github.com/OpenMined/PySyft/pull/3377)

### DIFF
--- a/syft/frameworks/crypten/context.py
+++ b/syft/frameworks/crypten/context.py
@@ -10,7 +10,9 @@ import crypten
 from crypten.communicator import DistributedCommunicator
 
 
-def _launch(func, rank, world_size, master_addr, master_port, queue, func_args, func_kwargs):
+def _launch(
+    func, rank, world_size, master_addr, master_port, queue, func_args, func_kwargs
+):  # pragma: no cover
     communicator_args = {
         "RANK": rank,
         "WORLD_SIZE": world_size,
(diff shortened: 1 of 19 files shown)


LaRiffle:
I was quite confuse I first thought you where instantiating the class and not an instance of it :D

syft/frameworks/crypten/context.py
```diff
@@ -10,7 +10,7 @@
 from crypten.communicator import DistributedCommunicator
 
 
-def _launch(func, rank, world_size, master_addr, master_port, queue, func_args, func_kwargs):
+def _launch(func, rank, world_size, master_addr, master_port, queue, func_args, func_kwargs): # pragma: no cover
```
gmuraru, line 13: Needed because this is run on the remote worker and it seems is not seen by the coverage package :\

test/execution/test_placeholder.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+import torch as th
+
+from syft.execution.placeholder import PlaceHolder
+
+
+def test_placeholder_forwarding():
```
karlhigley, line 7: I'm probably missing something, but it doesn't seem like this test requires the other Crypten-related changes in this PR?

gmuraru: Currently, I want to make sure I got the coverage rate > 95% (and see what else it fails on the tests). After that, I want to rebase the CrypTen branch with what it is on the master (and those changes should disappear).

iamtrask:
Status on this?

gmuraru:
This can be closed since we added other changes. Will reopen it if needed

Issue #2573, opened 2019-08-28
Implement maxpool2d for AdditiveSharedTensor


mari-linhares:
**Is your feature request related to a problem? Please describe.**

This is crucial for being able to run CNN models in an encrypted way.

**Describe the solution you'd like**

We already have `maxpool` implemented, I believe `maxpool2d` will consist basically of a wrapper that splits a matrix into multiple tensors and run `maxpool` on them.

This operation could be parallelized but this is out of the scope of this issue.

**Acceptance criteria**

- [x] Implement maxpool2d (#2578)
- [x] Test maxpool2d (#2578)
- [ ] Make it fast (aka be able to run mnist CNN with it)
- [ ] Implement maxpool2d_deriv
- [ ] Test maxpool2d_deriv

PR #3399, opened 2020-04-23
Avg & Maxpooling with SMPC

abogaziah:
## Description
resolves issue #2573 
overloading torch.nn.F.max_pool2d and avg_pool2d in order to use it in architectures like ResNet

## Type of change

## Checklist:

### DIFF
--- a/syft/frameworks/torch/nn/__init__.py
+++ b/syft/frameworks/torch/nn/__init__.py
@@ -1,5 +1,7 @@
 from syft.frameworks.torch.nn.conv import Conv2d
-from syft.frameworks.torch.nn.functional import conv2d, maxpool2d, avgpool2d
+from syft.frameworks.torch.nn.functional import conv2d
+from syft.frameworks.torch.nn.functional import maxpool2d
+from syft.frameworks.torch.nn.functional import avgpool2d
 from syft.frameworks.torch.nn.functional import dropout
 from syft.frameworks.torch.nn.functional import linear
 from syft.frameworks.torch.nn.pool import AvgPool2d
--- a/syft/frameworks/torch/nn/__init__.py
+++ b/syft/frameworks/torch/nn/__init__.py
@@ -1,5 +1,5 @@
 from syft.frameworks.torch.nn.conv import Conv2d
-from syft.frameworks.torch.nn.functional import conv2d
+from syft.frameworks.torch.nn.functional import conv2d, maxpool2d, avgpool2d
 from syft.frameworks.torch.nn.functional import dropout
 from syft.frameworks.torch.nn.functional import linear
 from syft.frameworks.torch.nn.pool import AvgPool2d
@@ -26,6 +26,8 @@ def nn(module):
         module.conv2d = conv2d
         module.dropout = dropout
         module.linear = linear
+        module.max_pool2d = maxpool2d
+        module.avg_pool2d = avgpool2d
 
     module.functional = functional
 
--- a/syft/frameworks/torch/nn/functional.py
+++ b/syft/frameworks/torch/nn/functional.py
@@ -170,9 +170,11 @@ def _pool(tensor, kernel_size: int = 2, stride: int = 2, mode="max"):
     if mode is "max":
         for channel in range(a_w.shape[0]):
             result.append(a_w[channel].max())
-    else:
+    elif mode is "mean":
         for channel in range(a_w.shape[0]):
             result.append(torch.mean(a_w[channel]))
+    else:
+        raise Exception("unknown pooling mode")
 
     result = torch.stack(result).reshape(output_shape)
     return result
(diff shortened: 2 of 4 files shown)


syft/frameworks/torch/nn/__init__.py
```diff
@@ -1,5 +1,5 @@
 from syft.frameworks.torch.nn.conv import Conv2d
-from syft.frameworks.torch.nn.functional import conv2d
+from syft.frameworks.torch.nn.functional import conv2d, maxpool2d, avgpool2d
```
Syzygianinfern0, line 2:
```suggestion
from syft.frameworks.torch.nn.functional import conv2d
from syft.frameworks.torch.nn.functional import maxpool2d
from syft.frameworks.torch.nn.functional import avgpool2d
```
Just to be consistent with `__init__.py` files eveywhere

syft/frameworks/torch/nn/functional.py
```diff
@@ -150,3 +150,56 @@ def conv2d(input, weight, bias=None, stride=1, padding=0, dilation=1, groups=1):
...
+        tensor,
+        size=output_shape + kernel_size,
+        stride=(stride * a_strides[0], stride * a_strides[1]) + a_strides,
+    )
+    a_w = a_w.reshape(-1, *kernel_size)
+    result = []
+    if mode is "max":
+        for channel in range(a_w.shape[0]):
+            result.append(a_w[channel].max())
+    else:
```
Syzygianinfern0, line 173: It be good if the `else` captures just the `mean` arg with other modes leading to raising Exceptions

syft/frameworks/torch/nn/functional.py
```diff
@@ -150,3 +150,56 @@ def conv2d(input, weight, bias=None, stride=1, padding=0, dilation=1, groups=1):
...
+        for channel in range(a_w.shape[0]):
+            result.append(torch.mean(a_w[channel]))
+
+    result = torch.stack(result).reshape(output_shape)
+    return result
+
+
+def pool2d(tensor, kernel_size: int = 2, stride: int = 2, mode="max"):
+    assert len(tensor.shape) == 4
```
Syzygianinfern0, line 182: `torch.nn.functional.max_pool2d` currently supports both 3D and 4D tensors. So might be good to add compatibility for 3D tensors too.

syft/frameworks/torch/nn/functional.py
```diff
@@ -150,3 +150,56 @@ def conv2d(input, weight, bias=None, stride=1, padding=0, dilation=1, groups=1):
...
+        batches,
+        channels,
+        (tensor.shape[2] - kernel_size) // stride + 1,
+        (tensor.shape[3] - kernel_size) // stride + 1,
+    )
+    result = []
+    for batch in range(batches):
+        for channel in range(channels):
+            result.append(_pool(tensor[batch][channel], kernel_size, stride, mode))
+    result = torch.stack(result).reshape(out_shape)
```
Syzygianinfern0, line 196: 👍👍

syft/frameworks/torch/nn/functional.py
abogaziah: that's an initial version to get it to work then we'll iterate on it

syft/frameworks/torch/nn/__init__.py
abogaziah: why do you think that's necessary? is there a use case that benefits from that?

syft/frameworks/torch/nn/functional.py
abogaziah: well spotted! I actually thought about that but I made it that way to avoid writing a test case for the expectation 😂 will do it next commit

syft/frameworks/torch/nn/__init__.py
Syzygianinfern0: Doesn't change the functionality in anyway. It's just conventional to do so 😄

AlanAboudib:
@Syzygianinfern0 did you check for any potential GC leaks?

syft/frameworks/torch/nn/functional.py
```diff
@@ -150,3 +150,58 @@ def conv2d(input, weight, bias=None, stride=1, padding=0, dilation=1, groups=1):
...
+    a_w = a_w.reshape(-1, *kernel_size)
+    result = []
+    if mode is "max":
+        for channel in range(a_w.shape[0]):
+            result.append(a_w[channel].max())
+    elif mode is "mean":
+        for channel in range(a_w.shape[0]):
+            result.append(torch.mean(a_w[channel]))
+    else:
+        raise Exception("unknown pooling mode")
```
LaRiffle, line 177: You can use a more precise type of Error, ValueError might be appropriate

syft/frameworks/torch/nn/functional.py
Syzygianinfern0: Can we have this done before we merge? I guess it might be work if we have something like 
```python
if len(tensor.shape) == 3:
    return squeeze(pool2d(unsqueeze(tensor))
```
or something like that??

Syzygianinfern0:
> @Syzygianinfern0 did you check for any potential GC leaks?

Major leaks currently. But testing on codebase of #3403 they disappear :partying_face:... Related issues I guess :thinking:

PR #3404, opened 2020-04-24
Remove `init_plan` attribute from `BaseWorker`

### LINKED ISSUE #3402: Decouple `Plan` tracing from `BaseWorker`
Tracing nested `Plans` still depends on stashing state on `BaseWorker` via the `init_plan` attribute. We'd like to decouple `Plan` tracing from the hooks and `BaseWorker`, so that it's self-contained and can trace multiple `Roles` in parallel for `Protocols`.

 TBD how to do this while keeping nested `Plan` tracing working. Maybe each `Plan` could have an attribute that points to its parent or root-level `Plan`, where the root-level `Plan` points to itself?

- Dropping nested `Plan` functionality. Not even sure who uses it for what.

**Additional context**
#3303 removed (most of?) the coupling between `Plan` tracing and the hooks. This change was out of scope for that PR, but is very much related.

karlhigley:
This PR removes the `init_plan` attribute from `BaseWorker`, which is currently used to keep track of the root `Plan` when tracing nested `Plans`. In order to accomplish that, it also removes nested `Plan` functionality.

Fixes #3402


karlhigley:
Refactoring dependency graph (thus far):

Screen Shot 2020-04-24 at 4 12 09 PM

syft/execution/placeholder.py
```diff
@@ -36,6 +36,7 @@ def __init__(
         self.child = None
         self.role = role
         self.tracing = tracing
+        self.inner = False
```
Jasopaum, line 39: If we keep this, could we find a more informative naming? Something like is_inner_state_placeholder?

karlhigley: Good thought! Ended up getting rid of it though.

karlhigley:
I would like to get rid of `Placeholder.owner`, but in order to do so we might have to make it not an `AbstractObject`, which I suspect would up-end our hook-based tracing. Leaving it for now, since all tensor objects have `owners` at the moment.

syft/execution/role.py
```diff
@@ -35,29 +35,19 @@ def __init__(
         state_tensors=None,
```
Jasopaum, line 35: `state_tensors` is not used anymore in the constructor, is it?

syft/execution/state.py
```diff
@@ -53,25 +52,10 @@ def read(self):
         If run while a plan is building, declare all the state tensors to the plan
         currently building.
         """
-        # TODO clean this function
```
Jasopaum, line 56: 🎉

karlhigley:
Merging despite the flakiness of `test_torch_sigmoid_approx`.

PR #3407, opened 2020-04-26
Update Crypten Head

gmuraru:
## Description
Update CrypTen to point to the master branch from Facebook Research.

### DIFF
--- a/syft/frameworks/crypten/__init__.py
+++ b/syft/frameworks/crypten/__init__.py
@@ -19,7 +19,7 @@ def load(tag: str, src: int, **kwargs):
         result = crypten.load_from_party(preloaded=result, src=src, **kwargs)
 
     else:
-        result = crypten.load_from_party(f=load, src=src, **kwargs)
+        result = crypten.load_from_party(preloaded=-1, src=src, **kwargs)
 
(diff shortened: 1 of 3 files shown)


syft/frameworks/crypten/__init__.py
```diff
@@ -16,10 +16,10 @@ def load(tag: str, src: int, **kwargs):
         assert len(results) == 1
 
         result = results[0]
-        result = crypten.native_load(f=None, preloaded=result, src=src, **kwargs)
+        result = crypten.load_from_party(preloaded=result, src=src, **kwargs)
 
     else:
-        result = crypten.native_load(f=None, src=src, **kwargs)
+        result = crypten.load_from_party(f=load, src=src, **kwargs)
```
gmuraru, line 22: This is only needed because ```f``` or ```preloaded``` argument must not be ```None```

youben11, line 22: I think it would make more sense to pass `preloaded=-1` to do the same as above but specifying that it's not useful (-1)

gmuraru: Done :)

PR #3442, opened 2020-05-03
Fix `move` PT to its own location

### LINKED ISSUE #3435: Move() on same machine
Dear all,

I was not sure whether to classify this as a "bug" or "new feature", please move it in case this is wrong.
Is the behavior of the move() method intended to throw an error, when trying to move an object to a virtual worker that already possesses the object?
In case this happens, a objectnotfounderror is raised: 

_ObjectNotFoundError: Object "56137057453" not found on worker!!! You just tried to interact with an object ID:56137057453 on <VirtualWorker id:manager_alice #objects:12> which does not exist!!! Use .send() and .get() on all your tensors to make sure they're on the same machines. If you think this tensor does exist, check the ._objects dictionary on the worker and see for yourself!!! The most common reason this error happens is because someone calls .get() on the object's pointer without realizing it (which deletes the remote object and sends it to the pointer). Check your code to make sure you haven't already called .get() on this pointer!!!_

Of course one can handle this in the control logic by checking the object ownership before calling the move method, but I think it would be nice if the move() method just ignores the request in case one tries to move objects that a worker already possesses.
When looping over multiple workers this would simplify the control logic.

Syzygianinfern0:
## Description

> the move() method intended to throw an error, when trying to move an object to a virtual worker that already possesses the object

Fixes #3435 

## Type of change

## Checklist:

### DIFF
--- a/syft/generic/pointers/pointer_tensor.py
+++ b/syft/generic/pointers/pointer_tensor.py
@@ -274,6 +274,9 @@ class PointerTensor(ObjectPointer, AbstractTensor):
         if self.owner.id == destination.id:
             return self.get()
 
+        if self.location.id == destination.id:
+            return self
+
         ptr = self.remote_send(destination, requires_grad=requires_grad)
 
         # We make the pointer point at the remote value. As the id doesn't change,
--- a/test/torch/pointers/test_pointer_tensor.py
+++ b/test/torch/pointers/test_pointer_tensor.py
@@ -329,6 +329,12 @@ def test_move(workers):
     z = y.move(me)
     assert (z == t).all()
 
+    # Move object to same location
+    alice.clear_objects()
+    t = torch.tensor([1.0, 2, 3, 4, 5]).send(bob)
+    t = t.move(bob)
+    assert torch.all(torch.eq(t.get(), torch.tensor([1.0, 2, 3, 4, 5])))
+
 
 def test_combine_pointers(workers):
     """


syft/generic/pointers/pointer_tensor.py
```diff
@@ -274,6 +274,9 @@ def move(self, destination: AbstractWorker, requires_grad: bool = False):
         if self.owner.id == destination.id:
             return self.get()
 
+        if self.location.id == destination.id:
```
Jasopaum, line 277: I think this should be handle at a lower level, maybe on the reception of a Communication Message.
Otherwise, we may need this kind of check at several places (for different methods doing doing stuff with pointers). What do you think?

karlhigley: There's a similar check in the de-serialization, but I don't think it's sufficient to resolve this issue. Not sure where the right place for this is. I suppose we could add a decorator that wraps the function with this check and the one above? Not sure how else to avoid duplication.

Jasopaum: I don't think it should be done in serde, it's too late at this point.
I think it should be done on the worker side, when a worker receives the order to send a tensor (basically all the communication action will resume to that at the end, right?) it should check the source and destination of the communication and do things accordingly to that.
But it's just how I see it, maybe there's better?

karlhigley: I think this is even before a message gets sent; it's checking whether the worker has been asked to move a tensor to itself and bypassing sending a message to itself in that case. I can't think of a better place to do this yet, but I'm guessing we'll figure it out as we rework the communication methods for `Protocols`. 🤷

PR #3446, opened 2020-05-04
Allowed users only as Tuple

gmuraru:
## Description
There was a discrepancy ```allowed_users``` between what the documentation (comments) was saying and how it was used:
```allowed_users=("User")``` will not create a tuple, will be created only if you add the ```,``` after the first element.

## Checklist:

**Question**: Should we also allow List?
**LE**: Changed it such that ```allowed_users``` is a list


karlhigley:
I kinda wonder why `allowed_users` is a tuple instead of a list. There's nothing inherently fixed-length about `allowed_users` AFAICT.

gmuraru:
> I kinda wonder why `allowed_users` is a tuple instead of a list. There's nothing inherently fixed-length about `allowed_users` AFAICT.

I can change it to a list

gmuraru:
> > I kinda wonder why `allowed_users` is a tuple instead of a list. There's nothing inherently fixed-length about `allowed_users` AFAICT.
> 
> I can change it to a list. The only difference is that tuples are immutable.

Changed it back to a tuple because I got errors in the ```hook``` saying that list is not hashable :\

PR #3452, opened 2020-05-05
Send models with @run_multiworkers

youben11:
## Description

After introducing Jails and the mechanisms for sending Crypten models across workers, we needed a way to send those models while initializing a Crypten computation using our `@run_multiworkers` decorator. 

With this PR, a function decorated with `@run_multiworkers([ALICE, BOB], master_addr="127.0.0.1", model=model, dummy_input=dummy_input)` where `model` is a `torch.nn.Module` and dummy input is a `torch.Tensor` that can be evaluated using the `model`, every party running on different workers will have access to the `model` inside their functions, which then should be an equivalent Crypten model.

## Type of change

## Checklist:

### DIFF
--- a/syft/frameworks/crypten/context.py
+++ b/syft/frameworks/crypten/context.py
@@ -197,7 +197,7 @@ def run_multiworkers(
                 threads.append(thread)
 
             # Wait for local party and sender threads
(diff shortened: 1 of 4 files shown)


gmuraru:
Some ```Lint``` errors

PR #3457, opened 2020-05-06
Ability to seed the `IdProvider`, Verbosity Toggling from `TorchHook`

Syzygianinfern0:
## Seed the IdProvider
During most of my debugging sessions (especially the GC ones), the need to trace the course of specific tensors was very important. And the way to do that was through their IDs as that is basically the identity of any syft object. 

These changes give an ability to seed the `IdProvider` in the following ways
 ```python
sy.ID_PROVIDER.seed()
sy.manual_seed()
hook = sy.TorchHook(th, seed=0)
```

## Set Verbosity from Hook
Also, the ability to toggle the verbosity of all `VirtualWorkers` through the hook is more comfortable rather toggling each worker individually. It can be done this way now.
 ```python
hook = sy.TorchHook(th, verbose=True)
```

## Toggle Verbosity Using Hook
Toggle the verbosity at will! This is super useful while debugging a specific line
eg. 
```python
a = torch.ones(1, 5)
b = torch.ones(1, 5)

hook.set_verbose(True)
a = a.encrypt(workers=[alice, bob], crypto_provider=crypto_provider)
hook.set_verbose(False)

b = b.encrypt(workers=[alice, bob], crypto_provider=crypto_provider)
print(f"Output: {(a*b).decrypt()}")
```
Output:
```
worker <VirtualWorker id:me #objects:0> sending (ObjectMessage tensor([[...]])) to <VirtualWorker id:alice #objects:6>
worker <VirtualWorker id:alice #objects:6> received ObjectMessage
worker <VirtualWorker id:me #objects:0> sending (ObjectMessage tensor([[...]])) to <VirtualWorker id:bob #objects:6>
worker <VirtualWorker id:bob #objects:6> received ObjectMessage

Output: tensor([[1., 1., 1., 1., 1.]])
```
Something like this can be included in the beginner tutorials about the VirutalWorkers to better understand what happens under the hood. :tada: 

## Type of change

## Checklist:


syft/__init__.py
```diff
@@ -131,3 +131,7 @@ def pool():
     from syft.generic.id_provider import IdProvider
 
     ID_PROVIDER = IdProvider()
+
+
+def manual_seed(seed=0):
```
Syzygianinfern0, line 136: @karlhigley is this a good place for this? I'm stumped as to where this has to go :thinking:

syft/workers/base.py
```diff
@@ -112,7 +112,14 @@ def __init__(
         self.id = id
         self.is_client_worker = is_client_worker
         self.log_msgs = log_msgs
-        self.verbose = verbose
+        if hasattr(hook, "verbose") and hook.verbose is True:
+            self.verbose = True
+        else:
+            self.verbose = verbose
```
LaRiffle, line 118: I guess the verbose option provided in the worker init should be prioritary on the hook.verbose

syft/workers/base.py
```diff
@@ -112,7 +112,14 @@ def __init__(
...
+
+        if id != "me" and isinstance(hook, sy.TorchHook) and hasattr(hook, "_virtual_workers"):
+            hook._virtual_workers.append(self)
```
LaRiffle, line 121: Wanna make _virtual_workers a `set()` to avoid `if id != "me" ` ?

syft/workers/base.py
Syzygianinfern0: Done!

syft/workers/base.py
Syzygianinfern0: Done!

syft/__init__.py
karlhigley: It looks like it's only used once, so maybe it could be in-lined in the torch hook?

syft/__init__.py
```diff
@@ -131,3 +131,7 @@ def pool():
...
+    syft.ID_PROVIDER.seed(seed)
```
karlhigley, line 137:
```suggestion
```

syft/frameworks/torch/hook/hook.py
```diff
@@ -97,6 +103,9 @@ def __init__(
         # Save the provided torch module as an attribute of the hook
         self.torch = torch
         self.framework = self.torch
+        if seed is not None:
+            syft.manual_seed(seed)
```
karlhigley, line 107:
```suggestion
            syft.ID_PROVIDER.seed(seed)
```

syft/workers/base.py
```diff
@@ -112,7 +112,14 @@ def __init__(
         self.id = id
         self.is_client_worker = is_client_worker
         self.log_msgs = log_msgs
-        self.verbose = verbose
+        if verbose is None:
+            self.verbose = hook.verbose if hasattr(hook, "verbose") else False
+        else:
+            self.verbose = verbose
+
+        if isinstance(hook, sy.TorchHook) and hasattr(hook, "_virtual_workers"):
```
karlhigley, line 120: It seems like `hasattr(hook, "_virtual_workers")` is always `True`, given that the hook sets `self._virtual_workers = {self.local_worker}`?

syft/workers/base.py
```diff
@@ -112,7 +112,14 @@ def __init__(
...
+            hook._virtual_workers.add(self)
```
karlhigley, line 121: This looks like it adds all workers to `hook._virtual_workers`. Either the name `_virtual_workers` or something about the logic doesn't quite make sense here. 🤔

syft/messaging/message.py
```diff
@@ -300,7 +300,7 @@ def __init__(self, obj_id, user, reason):
 
     def __str__(self):
         """Return a human readable version of this message"""
-        return f"({type(self).__name__} {(self.obj_id, self.user, self.reason)})"
+        return f"({type(self).__name__} {(self.obj_id if hasattr(self, 'obj_id') else '', self.user, self.reason)})"
```
karlhigley, line 303: This bypasses the error, but I think the underlying ssue is that the field on the message is called `object_id` (instead of `obj_id`.)

syft/frameworks/torch/hook/hook.py
```diff
@@ -122,10 +131,14 @@ def __init__(
             # be agnostic to the means by which workers communicate (such as
             # peer-to-peer, sockets, through local ports, or all within the
             # same process)
-            self.local_worker = VirtualWorker(hook=self, is_client_worker=is_client, id="me")
+            self.local_worker = VirtualWorker(
```
karlhigley, line 134: Wondering if we should rename `self.local_worker` to just `self.worker`, because that would make it possible to rename `self._virtual_workers` below to `self.local_workers`, which seems potentially more accurate.

syft/workers/base.py
Syzygianinfern0: That would make it fail the first time while instantiating the "me" / local_worker from the torch hook.
Like [here](https://github.com/Syzygianinfern0/PySyft/blob/e9ecd4594129f0f307ff7fefaae136cbab8769ad/syft/frameworks/torch/hook/hook.py#L134)

syft/workers/base.py
Syzygianinfern0: We keep track of the "virtual" workers (alice, bob, etc.) here. Hence the naming convention.

syft/frameworks/torch/hook/hook.py
Syzygianinfern0: I feel that the current naming scheme seems to be more appropriate though. (@LaRiffle can we have your thoughts on this)

local_worker ->  Worker owned by the local machine
virtual_worker -> Machine present elsewhere but "virtually" present by ```alice = sy.VirtualWorker(hook, id="alice")```

syft/workers/base.py
karlhigley: This is in BaseWorker though, which means all workers get added to this dictionary.

syft/frameworks/torch/hook/hook.py
karlhigley: All VirtualWorkers are local but not all VirtualWorkers are virtual. Something seems off about that. Not saying you should fix it in this PR, just saying this PR is a bit confusing due to underlying naming issues.

PR #3465, opened 2020-05-07
Update `Role.register_action()` to allow registering `CommunicationActions`

### LINKED ISSUE #3424: Register `CommunicationActions` in `Roles` (for `Protocols`)
`Role.register_action()` should be able to register both `ComputationActions` and `CommunicationActions`. Some of the assumptions about which fields are present only make sense for `ComputationActions`, which causes errors when trying to register `CommunicationActions`.

Make the change easy:
* Add tests for `Role.register_action()` with `ComputationActions` to guard the existing functionality against regressions.
* Move fields shared by both action types into the `Action` base class. These fields can be provided as standard `args` to `register_action()`.
* Provide fields that are unique to `ComputationActions` as `kwargs` instead of `args`

Make the easy change:
* Add tests for `Role.register_action()` with `CommunicationActions` to drive the new functionality
* Provide fields that are unique to `CommunicationActions` as `kwargs`
* Use the `action_type` param to decide which `kwargs` are expected to be present
* Refactor any private methods used by `register_action()` (like `_store_placeholders`) to work with both `Action` types

We could have separate methods to register the two types of `Action`. That would push the decision about which `Action` type to register out into the tracing code in the hooks, tracing wrappers, and `Placeholders` though. Might be better to make that decision in one centralized place to avoid duplication and discrepancies.

**Additional context**
This is an early step toward tracing `CommunicationActions` into `Protocols`.

karlhigley:
## Description

This mostly standardizes the fields with `ComputationAction`, which makes it possible to register both action types interchangeably in `Roles`.

Depends on OpenMined/syft-proto#100

Resolves #3424

## Type of change

## Checklist:

* [X] I have added tests for my changes


syft/workers/base.py
```diff
@@ -549,23 +551,22 @@ def execute_computation_action(self, action: ComputationAction) -> PointerTensor
                 raise ResponseSignatureError(new_ids)
 
     def execute_communication_action(self, action: CommunicationAction) -> PointerTensor:
```
Jasopaum, line 553: Seems like you don't use the `return_ids` and `return_values` attributes. Do you think we will use them at some point?

karlhigley: I think we probably should, but it's not required to make the existing tests pass. As we start tracing more `CommunicationActions`, I think we'll have to fix this, particularly for things like `get`.

karlhigley:
At this point, the serialization seems to be working, but there's an issue with an in-place `detach_` in one of the Udacity tests, which I don't fully understand yet. That's the last piece remaining to make this PR potentially mergeable.

syft/execution/action.py
```diff
@@ -10,4 +20,258 @@ class Action(ABC):
...
 
-    pass
+    def __init__(
+        self, name: str, target, args_: tuple, kwargs_: dict, return_ids: tuple, return_value=False
+    ):
+        """Initialize an action
+
+        Args:
+            name (String): The name of the method to be invoked (e.g. "__add__")
+            target (Tensor): The object to invoke the method on
```
Jasopaum, line 30: Is it necessarily a tensor here or can it be a Syft object as well?

syft/execution/action.py
```diff
@@ -10,4 +20,258 @@ class Action(ABC):
...
+            args_ (Tuple): The arguments to the method call
+            kwargs_ (Dictionary): The keyword arguments to the method call
+            return_ids (Tuple): primarily for our async infrastructure (Plan, Protocol, etc.), the id of
+                action results are set by the client. This allows the client to be able to predict where
+                the results will be ahead of time. Importantly, this allows the client to pre-initalize the
+                pointers to the future data, regardless of whether the action has yet executed. It also
+                reduces the size of the response from the action (which is very often empty).
+            return_value (boolean): return the result or not. If true, the result is directly returned,
```
Jasopaum, line 38: I know this has nothing to do with this PR but could `return_value ` be inferred from the method name?

syft/execution/action.py
```diff
@@ -10,4 +20,258 @@ class Action(ABC):
...
+            sy.serde.msgpack.serde._detail(worker, target),
+            sy.serde.msgpack.serde._detail(worker, args_),
+            sy.serde.msgpack.serde._detail(worker, kwargs_),
+            sy.serde.msgpack.serde._detail(worker, return_ids),
+            sy.serde.msgpack.serde._detail(worker, return_value),
+        )
+
+    @staticmethod
+    @abstractmethod
+    def bufferize(worker: AbstractWorker, action: "Action", protobuf_action):
```
Jasopaum, line 174: Python doesn't check for the method signature when you implement an abstract method? Well, this language is... surprising 😅

syft/execution/action.py
karlhigley: Maybe? Not sure, but might be worth a look. Create an issue?

syft/execution/action.py
karlhigley: I'm not sure. In concept, it certainly could be (and should be) a Syft object, but it may only work for tensors right now.

Jasopaum: Ok, we keep it as is for now then 👍

syft/execution/action.py
Jasopaum: Anyway, it's a detail

syft/serde/protobuf/serde.py
```diff
@@ -401,22 +401,27 @@ def bufferize_arg(worker: AbstractWorker, arg: object) -> ArgPB:
...
+
     return protobuf_arg
 
 
 def unbufferize_args(worker: AbstractWorker, protobuf_args: list) -> list:
     args_ = []
     for protobuf_arg in protobuf_args:
-        args_.append(unbufferize_arg(worker, protobuf_arg))
-    return args_
+        unbufferized = unbufferize_arg(worker, protobuf_arg)
```
Jasopaum, line 411: I would have done it with a list comprehension here, but it's a detail

karlhigley: You're right, that's cleaner. Fixing.

PR #3479, opened 2020-05-09
[WIP] Fix Identity Operations

### LINKED ISSUE #3439: flatten() removes the object!
**Describe the bug**
when .flatten() is called on AST or a remote torch tensor it removes the original tensor from the worker
**To Reproduce**
A= torch.rand(4)
a=A.fix_prec().share(bob, alice, crypto_provider=james) #or A.send(bob)
b=a.flatten()
a.get()

note: this happens only with 1D tensor, 2D works fine

Syzygianinfern0:
> I think the issue is that in torch the .flatten operation on 1d array is actually an inplace operation. You can do the above example with .view(-1) and it will work fine.
I have fixed the issue locally by modifying the is_inplace_method in torch_attributes.py

This can be triggered by any operation that is an identity operation as while hooking and registering the processed tensor, (due to some GC reasons) the earlier version of the tensor is deleted (de-registered).

For an identity operation, this deletes the original tensor that we are wokring with hence giving rise to such a probem

Fixes #3439 

- [x] Fix the issues
- [x] Fix stuff I break due to this
- [ ] Add tests

## Type of change

## Checklist:


iamtrask:
Looks like an additive sharing test is failing

PR #3484, opened 2020-05-09
Make an `ObjectStore` something a `Worker` has (instead of is)

karlhigley:
## Description

This aims to decouple "having object storage" from "being a worker," which should help make it possible for other objects (e.g. `Protocols`) to have their own "scoped" object storage, so that intermediate results of executing multi-party computations
* are private
* don't pollute the main object storage.

## Type of change

## Checklist:


syft/exceptions.py
```diff
@@ -327,9 +327,8 @@ def __init__(self, obj_id, worker):
         )
         message += (
             "Use .send() and .get() on all your tensors to make sure they're "
-            "on the same machines. "
-            "If you think this tensor does exist, check the ._objects dictionary "
-            "on the worker and see for yourself!!! "
+            "on the same machines. If you think this tensor does exist, check "
+            "the object_store dictionary on the worker and see for yourself!!! "
```
Jasopaum, line 331: nit: these 3 `!` look a bit agressive ^^

syft/generic/object_storage.py
```diff
@@ -2,11 +2,11 @@
...
+from syft.exceptions import ObjectNotFoundError
 from syft.generic.frameworks.types import FrameworkTensor
 from syft.generic.frameworks.types import FrameworkTensorType
 from syft.generic.tensor import AbstractTensor
-
-from syft.exceptions import ObjectNotFoundError
+from syft.workers.abstract import AbstractWorker
 
 
 class ObjectStorage:
```
Jasopaum, line 12: Is it an `ObjectStorage` or an `ObjectStore`?

syft/generic/object.py
```diff
@@ -58,20 +58,16 @@ def describe(self, description: str) -> "AbstractObject":
...
-            self.tags.add(tag)
-            if self.owner is not None:
-                # NOTE: this is a fix to correct faulty registration that can sometimes happen
-                if self.id not in self.owner._objects:
-                    self.owner.register_obj(self)
-                # note: this is a defaultdict(set)
-                self.owner._tag_to_object_ids[tag].add(self.id)
-            else:
-                raise RuntimeError("Can't tag a tensor which doesn't have an owner")
+        if self.owner is not None:
```
Jasopaum, line 61: nit but I think it would be slightly clearer to write
```
if self.owner is None:
    raise

<do stuff>
```

syft/generic/object.py
```diff
@@ -58,20 +58,16 @@ def describe(self, description: str) -> "AbstractObject":
...
+            self.tags = self.tags or set()
```
Jasopaum, line 62: According to the constructor, this should never be None

syft/generic/object_storage.py
```diff
@@ -96,18 +98,18 @@ def set_obj(self, obj: Union[FrameworkTensorType, AbstractTensor]) -> None:
                 else:
                     self._tag_to_object_ids[tag].add(obj.id)
 
-    def rm_obj(self, remote_key: Union[str, int], force=False):
+    def rm_obj(self, obj_id: Union[str, int], force=False):
```
Jasopaum, line 101: 🙏

syft/workers/base.py
```diff
@@ -226,6 +228,24 @@ def _recv_msg(self, message: bin):
...
+
+        Args:
+            return_self: flag, whether to return self as return value
+
+        Returns:
+            self, if return_self if True, else None
+
+        """
+        self.object_store.clear_objects()
+        return self if return_self else None
```
Jasopaum, line 247: I'm not sure we'll ever want to return a `Worker` when calling a function called `clear_objects`, right?
Maybe we could return `self.object_store`?

syft/generic/object_storage.py
karlhigley: Yeah, I should have renamed this. Will update.

syft/workers/base.py
karlhigley: Yeah, this is not great, but it's required to maintain compatibility with the Udacity course. I'll add a comment to that effect.

PR #3499, opened 2020-05-10
Enable state for roles in func2protocol decorator

Jasopaum:
## Description

Add a states arg in func2protocol decorator.
We may want to think about how this State will be used.

## Type of change

## Checklist:


test/execution/test_protocol.py
```diff
@@ -231,6 +231,28 @@ def protocol(roles):
...
 
+def test_stateful_protocol(workers):
+    shapes = {"alice": ((1,),), "bob": ((1,),)}
+    states = {"alice": (th.tensor([1]), th.tensor([3])), "bob": (th.tensor([5]),)}
+
+    @sy.func2protocol(args_shape=shapes, states=states)
+    def protocol(roles):
+        # fetch tensors from stores
+        # TODO will we fetch tensors that come from states this way?
+        # With tags?
```
karlhigley, line 242: 🤔 

For `Plans`, they're fetched with `state.read()`, right? I remember you saying you weren't a fan of that, so maybe this is a good time to think about how to do it better.

Jasopaum: Yep, they're fetched with `read`. As we are going towards something very similar for other tensors fetched from the worker store, I think we'll need to talk about differences between them/if we can totally merge these 2 concepts (from an API point of view at least).

karlhigley: Maybe in the Syft Core sync tomorrow?

karlhigley:
I suppose we could call the method that reads from `State` objects something like `load_state()`. 🤔

Jasopaum:
> I suppose we could call the method that reads from `State` objects something like `load_state()`. 🤔

You mean replacing `state.read()` or `role.load()`?

karlhigley:
> You mean replacing `state.read()` or `role.load()`?

I was thinking we'd end up with `role.load()` loading from an `ObjectStore` and `role.load_state()` loading from `State`. Not sure if that makes any sense, but it might?

Jasopaum:
> > You mean replacing `state.read()` or `role.load()`?
> 
> I was thinking we'd end up with `role.load()` loading from an `ObjectStore` and `role.load_state()` loading from `State`. Not sure if that makes any sense, but it might?

It might! But I still don't know how these 2 are related...
The wires behind the `role.load()` aren't connected yet. I can implement a first `load_state` though, searching by tags, what do you think?

karlhigley:
Sounds okay to me! Do we even need tags yet though? 🤷

Jasopaum:
Oh, you want to write `a, b, c = role.load_state()`?
Yeah, it's definitely possible!

karlhigley:
I mean, I'm not sure that _want_ to, but it would be fine for now as an incremental step. 😄

PR #3521, opened 2020-05-13
 Create a `VirtualWorker` for each `Role` and use it in the tests 

### LINKED ISSUE #3490: Create temporary `VirtualWorkers` for tracing in each `Role` of a `Protocol`
`Protocols` currently require `Workers` to exist before tracing, which hampers the future assignment of multiple `Workers` to a `Role`.

When creating `Roles` in `@sy.func2protocol`, create a `VirtualWorker` for each `Role` and assign it to an attribute on that `Role`.

* ???

**Additional context**
Depends on #3488.

karlhigley:
## Description

Creates a `VirtualWorker` for each `Role` to use during `Protocol` tracing.

Resolves #3490

## Checklist:

* [X] I have commented my code following [Google style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
* [X] I did follow the [contribution guidelines](https://github.com/OpenMined/PySyft/blob/master/CONTRIBUTING.md)

### DIFF
--- a/syft/execution/protocol.py
+++ b/syft/execution/protocol.py
@@ -32,12 +32,13 @@ class func2protocol(object):
     This class should be used only as a decorator.
     """
 
-    def __init__(self, args_shape=None):
+    def __init__(self, roles: list = [], args_shape: dict = {}):
         self.args_shape = args_shape
+        self.role_names = roles
 
     def __call__(self, protocol_function):
         # create the roles present in decorator
-        roles = {role_id: Role() for role_id in self.args_shape.keys()}
+        roles = {role_id: Role() for role_id in self.role_names}
 
         protocol = Protocol(
             name=protocol_function.__name__,
--- a/syft/execution/protocol.py
+++ b/syft/execution/protocol.py
@@ -48,19 +48,16 @@ class func2protocol(object):
             owner=sy.local_worker,
         )
 
-        # Build the protocol automatically
-        # TODO We can always build automatically, can't we? Except if workers doesn't have
-        # tensors yet in store. Do we handle that?
-        if self.args_shape:
-            try:
-                protocol.build()
-            except TypeError as e:
-                raise ValueError(
-                    "Automatic build using @func2protocol failed!\nCheck that:\n"
-                    " - you have provided the correct number of shapes in args_shape\n"
-                    " - you have no simple numbers like int or float as args. If you do "
-                    "so, please consider using a tensor instead."
-                )
+        try:
+            protocol.build()
+        except TypeError as e:
+            raise ValueError(
+                "Automatic build using @func2protocol failed!\nCheck that:\n"
+                " - you have provided the correct number of shapes in args_shape\n"
+                " - you have no simple numbers like int or float as args. If you do "
+                "so, please consider using a tensor instead."
+            )
+
         return protocol
 
 
--- a/syft/execution/protocol.py
+++ b/syft/execution/protocol.py
@@ -19,6 +19,7 @@ from syft.generic.frameworks.types import FrameworkTensor
 from syft.generic.frameworks.types import FrameworkLayerModule
 from syft.generic.object import AbstractObject
 from syft.workers.abstract import AbstractWorker
+from syft.workers.virtual import VirtualWorker
 
 from syft_proto.execution.v1.protocol_pb2 import Protocol as ProtocolPB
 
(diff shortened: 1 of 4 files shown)


test/execution/test_protocol.py
```diff
@@ -2,16 +2,33 @@
...
+    @sy.func2protocol(roles=["alice", "bob"], args_shape={"alice": ((1,),)})
     def protocol(roles):
-        tensor = roles["alice"].fetch(th.tensor([1]))
+        alice = roles["alice"]
+        bob = roles["bob"]
 
-        tensor.send(bob)
+        tensor = alice.fetch(th.tensor([1]))
+
+        tensor.send(bob.worker)
```
karlhigley, line 26: I don't love this syntax and I'm not sure it will work when the `Roles` are executed, but it does make the tracing tests pass. 🤷

syft/execution/protocol.py
```diff
@@ -32,12 +33,16 @@ class func2protocol(object):
...
-    def __init__(self, args_shape=None):
+    def __init__(self, roles: list = [], args_shape: dict = {}):
         self.args_shape = args_shape
+        self.role_names = roles
 
     def __call__(self, protocol_function):
         # create the roles present in decorator
-        roles = {role_id: Role() for role_id in self.args_shape.keys()}
+        roles = {
+            role_id: Role(worker=VirtualWorker(id=role_id, hook=sy.local_worker.hook))
```
Jasopaum, line 43: Isn't `sy.local_worker.hook` the default, here?

syft/execution/protocol.py
```diff
@@ -139,7 +141,8 @@ def build(self):
         self.toggle_tracing(True)
         self.is_building = True
 
-        results = self.forward(self.roles)
+        sorted_roles = [self.roles[name] for name in sorted(self.roles.keys())]
```
Jasopaum, line 144: If I understood correctly, the users will have to provide args in alphabetical order of the corresponding role id?

test/execution/test_protocol.py
Jasopaum: Let's polish that later?

syft/execution/protocol.py
karlhigley: Yeah, that might not be ideal. I suppose we could just make them match with the list of role names provided above.

syft/execution/protocol.py
karlhigley: Oh, maybe so!

syft/execution/protocol.py
Jasopaum: Do you think we can give `**self.roles` and then the user have to name the args as he named the roles?
Still not ideal but a bit better maybe. What do you think?

syft/execution/protocol.py
karlhigley: Hmm, I don't think it is, but maybe it should be?

Jasopaum: Oh, sorry, I thought it was!
It could, or we could also have a "local_hook" instead (or with) the "local_worker", not sure I see everything but I find this `sy.local_worker.hook` a bit odd

syft/execution/protocol.py
```diff
@@ -141,8 +141,7 @@ def build(self):
         self.toggle_tracing(True)
         self.is_building = True
 
-        sorted_roles = [self.roles[name] for name in sorted(self.roles.keys())]
-        results = self.forward(*sorted_roles)
+        results = self.forward(*self.roles.values())
```
Jasopaum, line 144: Won't this assign values to args in the order of the dict (random)?

syft/execution/protocol.py
karlhigley: Completely agree on it being odd! Workers and hooks are coupled in some weird ways we'd probably have to untangle.

syft/execution/protocol.py
karlhigley: In Python 3.7, the order is guaranteed to match insertion order: https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects

It's also true in Python 3.6, but it was provisional at that point: https://docs.python.org/3/whatsnew/3.6.html#new-dict-implementation

Jasopaum: Oh, really?
This goes against all I know about dicts 😆

Jasopaum: Oh, maybe they store metadata around the real data structure to keep this order when creating an iterator?

karlhigley: Looks like they adopted a data structure that maintains the order of entries and uses a sparse index: https://mail.python.org/pipermail/python-dev/2012-December/123028.html

Issue #3464, opened 2020-05-07
New PySyft workers are created with FSS `Plans` in their object storage


karlhigley:
**Describe the bug**
When a new worker is created, it comes with some `Plans` already loaded into the object storage:
```
{67822474598: <Plan Plan id:67822474598 owner:bob Tags: #fss_eq_plan_1 built>,
 42739784794: <Plan Plan id:42739784794 owner:bob Tags: #fss_eq_plan_2 built>,
 98376427733: <Plan Plan id:98376427733 owner:bob Tags: #fss_comp_plan_1 built>,
 48876976143: <Plan Plan id:48876976143 owner:bob Tags: #fss_comp_plan_2 built>,
 22552634480: <Plan Plan id:22552634480 owner:bob Tags: #xor_add_1 built>,
 96804360037: <Plan Plan id:96804360037 owner:bob Tags: #xor_add_2 built>}
```

**To Reproduce**
```
hook = TorchHook(torch)
bob = syft.VirtualWorker(id="bob", hook=hook, is_client_worker=False)
bob._objects
```

**Expected behavior**
New workers should start with empty object stores.

**Additional context**
Probably a result of the recent Function Secret Sharing PR.

jmaunon:
I confirm this bug as well

PR #3522, opened 2020-05-13
Trace generation actions like `torch.rand()` into `Roles` in `Protocols`

### LINKED ISSUE #3492: Make it possible for `Protocols` to generate tensors on specific `Roles`/`Workers`
When tracing and executing `Protocols`, it's not yet possible to locally generate new tensors (e.g. with `torch.rand()`) and have them registered on the correct `Role`/`Workers`.

Give each `Role` a `torch` attribute that contains a `FrameworkWrapper` around the actual `torch` module for tracing into that `Role`.

* I suppose we could instead add an additional `kwarg` for which `Role` to trace into onto all `torch` methods and strip it out in the `FrameworkWrapper` before forwarding to `torch`. 🤔 

**Additional context**
This may depend on #3490 (not sure.)

karlhigley:
## Description

Makes it possible to trace tensor generation methods called on framework modules (like `torch`) into `Protocol Roles`.

Resolves #3492

NOTE TO REVIEWERS: This PR is based on #3521, and only the last commit is new.

## Checklist:
* [X] My changes are covered by tests.
* [X] I have run [the pre-commit hooks](https://github.com/OpenMined/PySyft/blob/master/CONTRIBUTING.md#setting-up-pre-commit-hook) to format and check my code for style issues.
* [X] I have commented my code following [Google style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).

### DIFF
--- a/syft/execution/role.py
+++ b/syft/execution/role.py
@@ -36,6 +36,7 @@ class Role(SyftSerializable):
     ):
         self.id = id or sy.ID_PROVIDER.pop()
         self.worker = worker or sy.local_worker
+
         self.actions = actions or []
 
         # All placeholders
--- a/syft/execution/role.py
+++ b/syft/execution/role.py
@@ -12,10 +12,10 @@ from syft.execution.placeholder import PlaceHolder
 from syft.execution.placeholder_id import PlaceholderId
 from syft.execution.state import State
 from syft.generic.frameworks.types import FrameworkTensor
+from syft.serde.syft_serializable import SyftSerializable
 from syft.workers.abstract import AbstractWorker
 
 from syft_proto.execution.v1.role_pb2 import Role as RolePB
-from syft.serde.syft_serializable import SyftSerializable
 
 
 class Role(SyftSerializable):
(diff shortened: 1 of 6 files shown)


syft/workers/base.py
```diff
@@ -182,8 +182,6 @@ def __init__(
 
         # storage object for crypto primitives
         self.crypto_store = PrimitiveStorage(owner=self)
-        # declare the plans used for crypto computations
-        sy.frameworks.torch.mpc.fss.initialize_crypto_plans(self)
```
Jasopaum, line 186: I don't think you really meant to remove that

syft/execution/role.py
```diff
@@ -47,6 +49,8 @@ def __init__(
         self.state = state or State()
         self.tracing = False
 
+        self.torch = FrameworkWrapper(package=self.worker.torch, role=self, owner=self.worker)
```
Jasopaum, line 52: I think we could even have a for loop to add all the frameworks we support here

syft/workers/base.py
karlhigley: I really did, because it makes the tests fail and shouldn't have been there in the first place.

karlhigley: I opened an issue for it, but there hasn't been a fix: https://github.com/OpenMined/PySyft/issues/3464

syft/execution/role.py
karlhigley: Oh yeah! That would be cool.

syft/workers/base.py
karlhigley: I can move the fix to a separate PR, though. 🤷

karlhigley: Removed and submitted as #3525

PR #3530, opened 2020-05-14
Serde refactor, lazy dataloaders, better serde interface

tudorcebere:
## Description

Serde module has a lot of global variables which won't allow dynamic usage of external libraries and are not looking very well either. replaced those with lazy dataloaders. This is a steps towards external libraries, but more work has to be done towards it. Currently, we can dynamically add new types in serde, but we can't use it properly due to the available messages, more to be added soon (after removing msgpack).

## Checklist:


AlanAboudib:
For your info @tudorcebere , here is how I currently register SyferText objects to serde.

```def register_to_serde(class_type: type):
    """Adds a class `class_type` to the `serde` module of PySyft.

    This is important to enable SyferText types to be sent to remote workers.

    Args:
        class_type (type): The class to register to PySyfts' serde module.
            This enables serde to serialize and deserialize objects of that class.

    Returns:
        (int): The proto ID it is registered with
    """

    # Get the maximum integer index of detailers and add 1 to it
    # to create a new index that does not exist yet
    proto_id = max(list(serde.detailers.keys())) + 1

    # Add the simplifier
    serde.detailers[proto_id] = class_type.detail

    # Add the simplifier
    serde.simplifiers[class_type] = (proto_id, class_type.simplify)

    return proto_id


```

karlhigley:
@AlanAboudib Worth mentioning that msgpack serialization will eventually be replaced by Protobuf, so something to be aware of that's on the horizon and getting closer.

AlanAboudib:
@karlhigley that's good to know. However, if PySyft implements its own version of my function, I suppose I would just use it out of the box

karlhigley:
@AlanAboudib One of the key differences between msgpack and Protobuf is that Protobuf requires message schemas, so while we may implement some version of that functionality that allows external libraries to register classes for serialization, using it will require some work on the library end that we can't automate or abstract away.

AlanAboudib:
@karlhigley hmm, sounds we have some work to do on our end if we would like to  support pysyft > 0.2.5 :D

syft/serde/protobuf/serde.py
```diff
@@ -2,71 +2,246 @@
...
-            MAP_PYTHON_TO_PROTOBUF_CLASSES[proto_class] = proto_class.get_protobuf_schema()
-        (
-            bufferizers,
-            forced_full_bufferizers,
-            unbufferizers,
-        ) = _generate_bufferizers_and_unbufferizers()
+
+    # Check to see if there is a bufferizer
+    # for this type. If there is, return the bufferized object.
+    # breakpoint()
```
karlhigley, line 200: Stray debugging comment, I think?

Issue #3494, opened 2020-05-10
Trace `fetch` and `load` actions into `Protocols`


karlhigley:
**Is your feature request related to a problem? Please describe.**
It's not yet possible to `fetch` objects from or `store` object to the local `Worker's` object store inside a `Protocol`, which makes it impossible to provide inputs or extract outputs.

**Describe the solution you'd like**
Flesh out the existing `fetch` method and add a parallel `store` method. Make sure they can be traced as `Actions` inside `Protocols`. Consider adding a new `Action` type to capture these (e.g. `WorkerAction`?)

**Describe alternatives you've considered**
* Load inputs and store outputs outside the `Protocol Roles`. This would require more glue code around `Protocols` in the worker libraries and would provide less flexibility.

**Additional context**
A stub for the `fetch` method was added in #3488.

PR #3538, opened 2020-05-17
protocol store and load + tracing

Prtfw:
## Description

Thank you for your contribution to the PySyft repository.
Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

If your pull request closes a GitHub issue, then set its number below.

Resolves https://github.com/OpenMined/PySyft/issues/3494

## Checklist:
* [ ] I have run [the pre-commit hooks](https://github.com/OpenMined/PySyft/blob/master/CONTRIBUTING.md#setting-up-pre-commit-hook) to format and check my code for style issues.


Prtfw:
@karlhigley 
Hey bauss... so this PR is kind of shaping up... was wondering if we 100% need a new action type workerAction? 

I think the pro is that it's conceptually cleaner... but the potential con is more code to maintain... or maybe I am missing some key reason / other pros entirely... wanted to get your / Jason's thoughts 

Thanks 🙏

syft/generic/object_storage.py
```diff
@@ -71,15 +71,22 @@ def get_obj(self, obj_id: Union[str, int]) -> object:
         Returns:
             Object with id equals to `obj_id`.
         """
-
+        obj = None
         try:
-            obj = self._objects[obj_id]
+            obj = self._objects.get(obj_id, None)
+            if obj is None:
```
Jasopaum, line 77: I don't think we should handle placeholders in a different fashion than tensors. Do you know why we need something special for placeholders?

Jasopaum: Oh, is it because of `PlaceholderId`?

syft/generic/object_storage.py
```diff
@@ -71,15 +71,22 @@ def get_obj(self, obj_id: Union[str, int]) -> object:
...
+                # probably a placeholder
+                found = (self._objects[x] for x in self._objects if x.value == obj_id)
+                try:
+                    obj = next(found)
+                except StopIteration as e:
+                    raise ObjectNotFoundError(obj_id, self)
+
         except KeyError as e:
```
Jasopaum, line 85: I don't think we need the try except anymore as you use `.get()`

syft/execution/role.py
```diff
@@ -171,14 +171,39 @@ def execute(self):
...
+
     def load(self, tensor):
         """ Load tensors used in a protocol from worker's local store
         """
-        # TODO mock for now, load will use worker's store in a future work
-        if self.tracing:
-            return PlaceHolder.create_from(tensor, role=self, tracing=True)
-        else:
+        # load will use worker's store
+        if type(tensor) == str or type(tensor) == int:
```
Jasopaum, line 194: Can you use `isinstance` instead?
`if isinstance(tensor, (str, int))`

syft/execution/role.py
```diff
@@ -171,14 +171,39 @@ def execute(self):
 
         return tuple(p.child for p in output_placeholders)
 
+    def store(self, tensor):
+        """ store tensors used in a protocol to worker's local store
+        """
+        if "<class 'torch.Tensor'>" == str(type(tensor)):
```
Jasopaum, line 177: I think we'll also want to store other things than tensors. Parameters or stuff like that for instance.

syft/execution/role.py
```diff
@@ -171,14 +171,39 @@ def execute(self):
 
         return tuple(p.child for p in output_placeholders)
 
+    def store(self, tensor):
```
Jasopaum, line 174: I'm still not sure about this will be used but would we also need something to update the value of an object?
The "usecase" would be: at the beginning of the protocol, I use `load` and at the end I update the initial value with the one we have computed during the protocol.
Maybe that needs discussion though

karlhigley: Imagine `Role` has its own `ObjectStore`, which works like a separate "namespace" that holds inputs, intermediate results, and outputs. Two possibilities for how `load/store` could work:
* `load/store` transfer or copy tensors from a `Worker's` `ObjectStore` to a `Role's` `ObjectStore` for inputs and transfer/copy the outputs back
* `load` "reads" inputs from a `Role's` `ObjectStore` (expecting them to be pre-populated) and "writes" outputs back to the `ObjectStore` (expecting them to be extracted afterward.

Jasopaum: Oh, that's clearer, thanks!
So in my opinion, we do need an update feature, maybe in another method?

syft/generic/object_storage.py
Prtfw: yep... but I fixed up the code a bit to make it hopefully cleaner.

syft/execution/role.py
karlhigley: `update` would be like an in-place operation where the new value overwrites the old one?

Prtfw: add a storage_id kwargs so you can "over-write" and update. let me know how it looks.

syft/execution/worker_actions.py
```diff
@@ -0,0 +1,115 @@
...
+from syft_proto.execution.v1.worker_action_pb2 import (
+    WorkerAction as WorkerActionPB,
+    # HACK replace this with the correct syft proto objects and update proto version in reqs
+    # and update documentations
+)
+
+WORKER_ACTION_METHODS = ["store", "load"]
+
+
+class WorkerAction(Action):
```
Jasopaum, line 13: I'm bad at naming stuff but I feel like this is a bit too generic, right? I mean, all actions are performed by a worker

syft/generic/object_storage.py
```diff
@@ -70,15 +70,21 @@ def get_obj(self, obj_id: Union[str, int]) -> object:
...
+            obj = self._objects.get(obj_id, None)
+            if obj is None:
+                # check for placeholder_ids
+                found = (self._objects[x] for x in self._objects if x.value == obj_id)
+                try:
+                    obj = next(found)
+                except StopIteration as e:
+                    # the object was not found using key look up or looking for placeholder_ids
+                    raise ObjectNotFoundError(obj_id, self)
+        except Exception as e:
```
Jasopaum, line 85: I don't think this try/except does much, maybe we can simply remove it?

syft/generic/object_storage.py
```diff
@@ -70,15 +70,21 @@ def get_obj(self, obj_id: Union[str, int]) -> object:
...
         try:
-            obj = self._objects[obj_id]
-        except KeyError as e:
-            if obj_id not in self._objects:
-                raise ObjectNotFoundError(obj_id, self)
-            else:
-                raise e
-
+            # if object is a tesor
+            obj = self._objects.get(obj_id, None)
```
Jasopaum, line 76: You don't want to use self.object_store.get_obj because it raises an exception?

syft/generic/object_storage.py
```diff
@@ -70,15 +70,21 @@ def get_obj(self, obj_id: Union[str, int]) -> object:
...
+        obj = None
         try:
-            obj = self._objects[obj_id]
-        except KeyError as e:
-            if obj_id not in self._objects:
-                raise ObjectNotFoundError(obj_id, self)
-            else:
-                raise e
-
+            # if object is a tesor
```
Jasopaum, line 75: typo: tesor -> tensor

syft/generic/object_storage.py
```diff
@@ -70,15 +70,21 @@ def get_obj(self, obj_id: Union[str, int]) -> object:
...
+            obj = self._objects.get(obj_id, None)
+            if obj is None:
+                # check for placeholder_ids
+                found = (self._objects[x] for x in self._objects if x.value == obj_id)
```
Jasopaum, line 79: Maybe the placeholders should be stored in `_objects` with id.value as a key? Would it make things easier?

test/execution/test_protocol.py
```diff
@@ -178,6 +178,29 @@ def protocol(alice, bob):
...
+        tensor2 = bob.load(stored_id)
+        t1plus = tensor1 + 1
+        t2plus = tensor2 + 1
+        return t1plus, t2plus
+
+    assert len(protocol.roles) == 2
+    assert "alice" in protocol.roles
+    assert "bob" in protocol.roles
+
+    # TODO add asserts for trace / traced_actions
```
Jasopaum, line 198: It's done, right? :)

test/serde/serde_helpers.py
```diff
@@ -1419,6 +1419,12 @@ def compare(detailed, original):
     ]
 
 
+# syft.execution.worker_actions.WorkerAction
+def make_worker_action(**kwargs):
+    # @TODO: add test
```
Jasopaum, line 1424: Don't forget this one

syft/generic/object_storage.py
Prtfw: yep... currently I get 
E   RecursionError: maximum recursion depth exceeded

syft/execution/worker_actions.py
Prtfw: Would worker_storage_actions or worker_protocol_actions or role_actions or worker_role_actions be better?
I just used WorkerAction to get going because that's what Karl called it initially. Open for suggestions though.

karlhigley: Yeah, bad naming on my part. Accurate, but not specific. 😓 

Based on your suggestions, what about `StorageActions`?

syft/generic/object_storage.py
karlhigley: What's the stack trace for that error? I probably messed something up when I extract `ObjectStore`.

syft/generic/object_storage.py
karlhigley: If it would help, we could split `_objects` by type or something. 🤔

syft/messaging/message.py
```diff
@@ -121,6 +123,14 @@ def communication(name, target, args_, kwargs_, return_ids):
         action = CommunicationAction(name, target, args_, kwargs_, return_ids)
         return TensorCommandMessage(action)
 
+    @staticmethod
+    def workerAction(name, target, args_, kwargs_, return_ids):
```
karlhigley, line 127: You must be writing other languages right now too! 😄 
```suggestion
    def worker_action(name, target, args_, kwargs_, return_ids):
```

test/execution/test_protocol.py
```diff
@@ -178,6 +178,29 @@ def protocol(alice, bob):
     assert "remote_get" in (action.name for action in traced_actions)
 
 
+def test_trace_store_load():
+    @sy.func2protocol(roles=["alice", "bob"], args_shape={"alice": ((1,),), "bob": ((1,),)})
+    def protocol(alice, bob):
+        t = th.Tensor([101])
+        stored = bob.store(t)
```
karlhigley, line 185: Still feel like I should mention that I was thinking of `load` as something that happens at the beginning and `store` as something that happens at the end. There's no reason it _has_ to be that way I suppose, but I'd love to see at least one test that does that. I think it will become increasingly important to test as @Jasopaum makes progress on `Executions`.

syft/generic/object_storage.py
Prtfw: @karlhigley  I was reproduce the stack trace now that tests are working again,
here you go!
image

test/execution/test_protocol.py
Prtfw: added to the same test that loads then stores.

syft/generic/object_storage.py
karlhigley: Based on the line numbers, I think the issue must be coming from [around here](https://github.com/OpenMined/PySyft/pull/3538/files#diff-f1b6d98d314c69d43920489c6a7e469cL74-R75), but I don't see code that looks like the snippets in the stack trace. Not 100% sure how it ends up calling `get_obj()` again inside `get_obj()`. Might be worth experimenting with the `try/except` version that uses `self._objects[obj_id]` instead of `self._objects.get(obj_id, None)` to see if that helps.

PR #3585, opened 2020-05-21
Adding torch.dtype support in serde.protobuf

tudorcebere:
## Description

Support for torch.dtype in serde.protobuf.

## Checklist:


syft/serde/protobuf/torch_serde.py
```diff
@@ -5,6 +5,7 @@
 import io
 
 import torch
+import pydoc
```
karlhigley, line 8: Haven't seen this used in the Syft code base before. 🤔

tudorcebere: Actually, I've used it in the typechecking PR when serializing types. I am sending the path of the type as a string and afterwards locate it to get the actual type in torch. Similar usage would be in ` serde.protobuf.native_serde.py`.

tudorcebere: By doing this we can drop `protobuf.torch.serde.TORCH_DTYPE_STR` in the future.

PR #3561, opened 2020-05-20
Split `FrameworkHook` into multiple parts

karlhigley:
## Description

This splits `FrameworkHook` into `TensorHook`, `PointerHook`, and `StringHook` in order to make it easier to figure out which parts can be replaced with the new style of tensor hooking in #3559.

## Checklist:
* [X] My changes are covered by tests.
* [X] I have run [the pre-commit hooks](https://github.com/OpenMined/PySyft/blob/master/CONTRIBUTING.md#setting-up-pre-commit-hook) to format and check my code for style issues.
* [X] I have commented my code following [Google style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).

(See the [the contribution guidelines](https://github.com/OpenMined/PySyft/blob/master/CONTRIBUTING.md) for additional tips.)

PR #3590, opened 2020-05-21
Revert "Split `FrameworkHook` into multiple parts"

karlhigley:
Reverts OpenMined/PySyft#3561


karlhigley:
It apparently may cause issues due to changing the order of imports and/or hook application. Waiting to see how the tests turn out.

PR #3591, opened 2020-05-21
Avoids multiple peers registration in the same grid environment.

### LINKED ISSUE #3576: Calling syft.grid.register() twice should throw an informative error
If you call syft.grid.register() twice in the same python runtime it should raise an error describing that you can't do this - that they shoudl restart the python runtime and try again.

jefersonf:
## Description

I just added a `_registered_peer` variable to hold the very first grid peer registration. So, any register made later, within the same runtime, the same instance is returned.

Resolves #3576 

## Checklist:


syft/grid/__init__.py
```diff
@@ -3,20 +3,27 @@
...
+        if not kwargs:
+            args = args = {"max_size": None, "timeout": 444, "url": DEFAULT_NETWORK_URL}
+        else:
+            args = kwargs
+
+        peer_id = str(uuid.uuid4())
+        _registered_peer = Network(peer_id, **args)
+        _registered_peer.start()
+    else:
+        raise UserWarning(f"You're already a registered peer {_registered_peer}")
```
jefersonf, line 27: I was wondering if there is a need to throw an error in the user's face if he tries to register multiple times. I think just returning the registered peer for the first time would be okay.  What do you think?

IonesioJunior: Yep, I think we can return the registered peer instance created by the first call. But we probably need to print a warning (not an exception or error message) to clarify to the user that they already have a registered peer.

PR #3595, opened 2020-05-22
Implement msgpack serde for RoleAssignments

Jasopaum:
## Description

We cannot implement protobuf serde because workers can't yet be bufferized.

## Checklist:


syft/execution/role_assignments.py
```diff
@@ -10,13 +16,20 @@ class RoleAssignments:
...
                 the Protocol to which the RoleAssignments is associated.
         """
-        self.assignments = {role_id: [] for role_id in role_ids}
+        if assignments:
+            self.assignments = assignments
+        elif role_ids:
+            self.assignments = {role_id: [] for role_id in role_ids}
+        else:
+            raise ValueError(
+                "You need to provide role_ids or assignments in RoleAssignments' constructor"
```
tudorcebere, line 31: I think this is a typo, should be RoleAssignament's.

syft/execution/role_assignments.py
```diff
@@ -10,13 +16,20 @@ class RoleAssignments:
     which other parties are participating and communicate with them if needed.
     """
 
-    def __init__(self, role_ids):
+    def __init__(self, role_ids: list = [], assignments: dict = {}):
```
tudorcebere, line 19: I think there might be a hidden bug when settings obj refs as default arguments (might not occur here, but I would personally avoid it just for future usage).

```
class Test:
    def __init__(self, test={}):
        self.test = test       

    def add_elem(self, k, v):
        self.test[k] = v

    def print(self):
        print(self.test)

t1 = Test()
t1.add_elem("1", 1)
t1.print()
# {"1": 1}

t2 = Test()
t2.add_elem("2", 2)
t2.print()
# {"1": 1, "2": 2}         
```

syft/execution/role_assignments.py
Jasopaum: The class name is plural, so there is no typo, I think (English is not my native language so I may be wrong though ^^)

syft/execution/role_assignments.py
Jasopaum: Oh yep, I already had this kind of bug! Nice catch 👍

syft/execution/role_assignments.py
karlhigley: Bizarrely, this is the correct punctuation in English. It's a shit language and I make no excuses.

PR #3596, opened 2020-05-23
Host private datasets locally

### LINKED ISSUE #3570: I want to be able to send private tensors to my own grid worker

me = sy.grid.register("Andrew")

b = th.tensor([1,2,3,4]).tag("#hello").private_tensor(allowed_users=("Ionesio","Andrew"))

me.host_dataset(b)

This raises an error when it should allow me to host data on my own node.

IonesioJunior:
## Description
Add the ability to set up and host private datasets locally, making them searchable to other peers and managing its access level.

Resolves #3570

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/native.py
+++ b/syft/frameworks/torch/tensors/interpreters/native.py
@@ -802,7 +802,7 @@ class TorchTensor(AbstractTensor):
 
         if self.is_wrapper:
             self.child = (
-                syft.PrivateTensor(*args, **kwargs)
+                syft.PrivateTensor(tags=self.tags, *args, **kwargs)
                 .on(self.child, wrap=False)
                 .register_credentials(tuple(allowed_users))
             )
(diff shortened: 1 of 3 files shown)


IonesioJunior:
**Note:**
Some methods used to convert tensor types ( such as `private_tensor` or `fix_precision`) makes the new tensors unable to be searched by its tags (these tags aren't forwarded during the process).

Since the new tensor does not have the tags set up, it's not added to the list of searchable  tensors, making them invisible to other workers.https://github.com/OpenMined/PySyft/blob/eb0b87da9747e921643366491f5f8e0aa2cbc4ce/syft/generic/object_storage.py#L93-L99

**Proposed solution:**
I just explicitly forward the tags during the object instantiation process.

iamtrask:
> **Note:**
> Some methods used to convert tensor types ( such as `private_tensor` and `fix_precision`) makes the new tensors unable to be searched by its tags (these tags aren't forwarded during the process).
> 
> Since the new tensor does not have the tags set up, it's not added to the list of searchable tensors, making them invisible to other workers.
> 
> https://github.com/OpenMined/PySyft/blob/eb0b87da9747e921643366491f5f8e0aa2cbc4ce/syft/generic/object_storage.py#L93-L99
> 
> **Proposed solution:**
> I just explicitly forward the tags during the object instantiation process.

This seems like a good workaround for now. We should capture this as an issue.

PR #3599, opened 2020-05-23
Remove protocol parameter in decrypt

### LINKED ISSUE #3329: Call decrypt() without specifying the protocol

youben11:
## Description

The PR remove the need to a `protocol` parameter in `decrypt` and warn the user to not use it, without raising an error.

Resolves #3329 

## Checklist:


syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -1049,19 +1047,23 @@ def decrypt(self, protocol="mpc", **kwargs):
...
         Raises:
             NotImplementedError: If protocols other than the ones mentioned above are queried
 
         """
 
-        if protocol.lower() == "mpc":
+        protocol = kwargs.get("protocol", None)
+        if protocol:
+            warnings.warn("protocol should no longer be used in decrypt")
+
```
LaRiffle, line 1060: Beware of the case where we have an AutogradTensor

youben11: I thought that passing tests meant it's working fine, but yeah there was no tests for decrypting an AutogradTensor. Fixed that and added test for decrypting with/without grad

PR #3607, opened 2020-05-25
Instantiate Protocols' placeholders on object reception

Jasopaum:
## Description

The objective is to instantiate Placeholders for a role that receives an object during a protocol execution.

## Checklist:
* [ ] I have run [the pre-commit hooks](https://github.com/OpenMined/PySyft/blob/master/CONTRIBUTING.md#setting-up-pre-commit-hook) to format and check my code for style issues.


syft/workers/base.py
```diff
@@ -380,7 +386,7 @@ def send(
             obj.id_at_origin = obj.id
 
         # Send the object
-        self.send_obj(obj, worker)
+        self.send_obj(obj, worker, placeholder_id)
```
karlhigley, line 389: This seems a bit awkward and reinforces my sense that it might be cleaner to package the `placeholder_id` into the `obj` somehow.

syft/workers/base.py
```diff
@@ -425,19 +431,19 @@ def handle_object_msg(self, obj_msg: ObjectMessage):
...
+        if isinstance(obj, FrameworkTensor) and (
+            obj.requires_grad and obj.origin is not None and obj.id_at_origin is not None
+        ):
+            obj.register_hook(obj.trigger_origin_backward_hook(obj.origin, obj.id_at_origin))
+
+        # Check if we received an object from a protocol run
+        if placeholder_id in self._protocol_placeholders:
+            self._protocol_placeholders[placeholder_id].instantiate(obj)
+        else:
+            # TODO is it really in the else here or do we want to register anyway?
```
karlhigley, line 445: Seems like it depends on whether the `Placeholders` are already registered in the `ObjectStore`? I'm not sure if they are. If so, then it doesn't seem like we'd need to register them again. If not, maybe this is covered by the `load/store` methods from @Prtfw's PR?

test/execution/test_protocol.py
```diff
@@ -38,7 +38,7 @@ def test_trace_communication_actions_send():
     def protocol(alice, bob):
         tensor = alice.torch.tensor([1])
 
-        tensor.send(bob.worker)
+        tensor.send(bob)
```
karlhigley, line 41: 🎉 

What magic makes this work? 😄

test/execution/test_protocol.py
```diff
@@ -305,3 +175,41 @@ def protocol(role1, role2):
...
+        a = x + y
+        b = a.send(role2)
+
+        # On role2
+        res = b + z
+        # d = c.send(role1)
+
+        # # On role1
+        # res = d * 2
+
```
karlhigley, line 198: Are these commented lines working at this point? If not, we can probably just remove them.

karlhigley:
Oh, one more thing: I'm normally in favor of removing code instead of commenting it, but in this case I'd suggest commenting out the other communication methods instead of removing them. @Prtfw just did a bunch of work to add those and I think they work as intended for tracing. Also, I want us to have a concrete reminder to come back and sort them out once some of the deep refactoring concludes. 😄

syft/workers/base.py
```diff
@@ -189,6 +189,11 @@ def __init__(
         # storage object for crypto primitives
         self.crypto_store = PrimitiveStorage(owner=self)
 
+        # dict containing all the placeholders of the role assigned to self
+        # TODO should we order them by protocol id/execution id?
+        # ie {exec_id: {**placehohlders}, exec_2: {**other_phs}}
```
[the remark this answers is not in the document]
Jasopaum: But we don't have exec_id yet 😅 
Maybe we should add `Execution` objects right now and make this dict a dict of current `Executions` on the worker?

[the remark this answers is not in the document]
karlhigley: Oh yeah, I don't think it has to happen in this PR. I was just agreeing with your TODO comment! Let's create a separate issue for adding `Executions`. 👍

test/execution/test_protocol.py
Jasopaum: The magic is that I don't really send anything when I do `Placeholder.send()`. Just create a placeholder with `location` as `Role` and trace the action.

test/execution/test_protocol.py
Jasopaum: Oops, yeah, we can remove them! I thought I could have back and forth between workers directly but this needs a bit more work. We simply execute all the actions for all the `Roles` sequentially on this branch

karlhigley:
(As context for anyone else reading this PR, the other communication methods don't quite work for execution yet because they need love and attention separately from `Protocols`, and there's relevant refactoring happening in #3611 and #3629.)

syft/workers/base.py
karlhigley: I'm with you on interacting consistently, but not on "_the_ object store." I'm still thinking of `Executions` having their own `ObjectStore`, which I imagine being where `Placeholders` are stored. (Ideally, the `Worker's` `ObjectStore` would never contain `Placeholders`.)

test/execution/test_protocol.py
karlhigley: Ah, okay, cool! Could you create an issue for the further work then?

syft/execution/protocol.py
```diff
@@ -203,6 +203,7 @@ def assign(self, role_id, worker):
         """ Assign a worker to the specified role.
         """
         self.role_assignments.assign(role_id, worker)
```
karlhigley, line 205: Seems like this assumes one worker per role?

karlhigley: (It’s actually the line below.)

Jasopaum: Hum, indeed... Do you think we'd need to create a class for an instantiated `Role`? I hope not but I don't see any solution right now

syft/workers/base.py
Jasopaum: I think I see what you mean. I'll try to do that in the next PR 🙂

syft/execution/protocol.py
karlhigley: Conceptually, it should be that the worker has a role, rather than the role having a worker (which is a simple way to make tracing work, but shouldn’t be used for execution.) If we introduce an `Execution` class, those are basically instantiated `Roles`, so maybe that would solve this issue?

Jasopaum: Yep, I started thinking about `Executions` and was about to say that the problem doesn't exist so don't mind what I said, I talked before thinking 😅

PR #3629, opened 2020-05-28
Extract communication methods to `AbstractSendable`

karlhigley:
## Description
This moves cross-worker communication methods off of `torch.Tensor` and into `AbstractSendable`, which `torch.Tensor` and other classes inherit from. This should help us provide consistent communication methods across frameworks and different types of Syft objects, which I hope will make it easier to remove wrapper tensors.

Depends on #3636. 

## Type of Change

- [ ] Documentation (non-breaking change which adds documentation)
- [X] Improvement (non-breaking change that improves the performance or reliability of existing functionality)

## Checklist


karlhigley:
Ah, sorry, I rushed to open this PR in order to share it with @gmuraru and @imskr, but I left out some context and should have marked it as a draft.

Behind the linting error, there's an issue with circular dependencies, because moving the communications methods to `AbstractSendable` means that it depends on `PointerTensor`, which is an `AbstractSendable`. Our current plan for what to try next is to add an `AbstractPointer` class (which might be just be a rename of `ObjectPointer`) and rewrite the general case versions of these methods to use that instead of `PointerTensor`. I think that means that `AbstractPointer` will need not to be an `AbstractSendable` though, so the concrete `Pointer` types will probably have to inherit from both. 🤷

karlhigley:
Even this intermediate step is apparently not something we can aim directly at. There's more fundamental refactoring required, so closing this and will revisit later.

PR #3636, opened 2020-05-29
Add utilities for hooking and chaining methods calls along tensor chains

karlhigley:
## Description

Adds utilities for inserting optional `_before_method()`/`_after_method()` hooks on a method decorated with `@hookable`, which allows custom tensor types to insert any functionality they individually require before or after a generic tensor method like `send`. This should help clean up the copious `if isinstance(self, TensorClass)` code in `AbstractTensor` and beyond.

In order to make sure that all custom functionality for types in the tensor chain gets called, this also creates a `chain_call()` function, which accepts a method name with args/kwargs, and calls that method (if it exists) on every tensor in the chain from outermost to innermost, returning the results in an ordered list.

## Type of Change

- [ ] Documentation (non-breaking change which adds documentation)
- [ ] Improvement (non-breaking change that improves the performance or reliability of existing functionality)

## Checklist
- [X] I have added tests for my changes

### DIFF
--- a/syft/generic/abstract/hookable.py
+++ b/syft/generic/abstract/hookable.py
@@ -1,0 +1,32 @@
+from functools import wraps
+
+
+def chain_call(obj, method_name, *args, **kwargs):
+    results = []
+    current = obj
+    while current is not None:
+        method = getattr(current, method_name, None)
+        if method:
+            results.append(method(*args, **kwargs))
+        current = current.child
+    return results
+
+
+def hookable(hookable_method):
+    """Decorator which checks for corresponding hooks and calls them if they exist
+
+    When this decorator is applied to a method, it checks for the existence of `_before_method()`
+    and `_after_method()` hooks, and calls them before/after the correspdoning method if they do.
+
+    This function should be used only as a decorator.
+    """
+    method_name = hookable_method.__name__
+
+    @wraps(hookable_method)
+    def hooked_method(self, *args, **kwargs):
+        chain_call(self, f"_before_{method_name}", *args, **kwargs)
+        return_val = hookable_method(self, *args, **kwargs)
+        chain_call(self, f"_after_{method_name}", *args, **kwargs)
+        return return_val
+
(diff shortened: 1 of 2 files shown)


syft/generic/abstract/hookable.py
```diff
@@ -0,0 +1,32 @@
+from functools import wraps
+
+
+def chain_call(obj, method_name, *args, **kwargs):
```
iamtrask, line 4: this is a nice convenience function - although I'm not sure that we ever truly go more than one child deep (or two children deep in special circumstances) within any call. I do assume that you have a use for it in mind though.

test/generic/test_hookable.py
```diff
@@ -0,0 +1,45 @@
+from syft.generic.abstract.hookable import chain_call
+from syft.generic.abstract.hookable import hookable
+
+
+def test_chain_call():
```
iamtrask, line 5: Oh! I get it... wow that's clever! Love that!

syft/generic/abstract/hookable.py
```diff
@@ -0,0 +1,32 @@
...
+    current = obj
+    while current is not None:
+        method = getattr(current, method_name, None)
+        if method:
+            results.append(method(*args, **kwargs))
+        current = current.child
+    return results
+
+
+def hookable(hookable_method):
```
iamtrask, line 15: This does overlap with ObjectConstructor a bit - but I'm ok with it given that ObjectConstructor doesn't exist yet - and there's a good chance they'll actually be compatible pieces of functionality with each other (aka ObjectConstructor creates the method names which this decorator detects)

syft/generic/abstract/hookable.py
karlhigley: Yeah, we don’t usually reach very far into the chain yet, but having stared at the code for a while, it’s starting to look like it would get simpler/cleaner if we did. There are also a few ad-hog implementations of this pattern floating around.

PR #3643, opened 2020-05-30
Move permissions from `TorchTensor.send()` to `PrivateTensor._before_send()` hook

karlhigley:
## Description
This extracts permission checking into a `_before_send()` hook on the relevant class (`PrivateTensor`), so that `TorchTensor` doesn't have to know as much about the functionality of custom types. This is the first of many similar refactors to move knowledge of custom tensor types out of `TorchTensor`.

## Checklist
- [X] I have added tests for my changes

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/private.py
+++ b/syft/frameworks/torch/tensors/interpreters/private.py
@@ -3,12 +3,11 @@ import syft
 
 from typing import List, Tuple
 
+from syft.exceptions import SendNotPermittedError
+from syft.generic.abstract.tensor import AbstractTensor
 from syft.generic.frameworks.hook import hook_args
 from syft.generic.frameworks.overload import overloaded
-
-
 from syft.workers.abstract import AbstractWorker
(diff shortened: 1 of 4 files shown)


syft/frameworks/torch/tensors/interpreters/private.py
```diff
@@ -61,6 +60,10 @@ def allow(self, user) -> bool:
         """
         return user in self.allowed_users
 
+    def _before_send(self, *location, user: object = None, **kwargs):
+        if not self.allow(user):
+            raise SendNotPermittedError()
```
karlhigley, line 65: Re-reading this, it looks like we might want to consolidate `SendNotPermittedError` and `GetNotPermittedError` into one permissions error.

karlhigley: After a few reads through, I'm pretty sure this still raises the right exceptions in the right places, and there are tests to check that. Will save any changes to the exceptions for later.

PR #3661, opened 2020-06-02
(WIP) Extract `MessageHandler` from `BaseWorker`

karlhigley:
## Description
Moves worker message handling code into a separate class.

## Checklist


syft/workers/base.py
```diff
@@ -110,6 +103,7 @@ def __init__(
         self.hook = hook
 
         self.object_store = ObjectStore(owner=self)
+        self.message_handler = BaseMessageHandler(self.object_store, self)
```
gmuraru, line 106: Q: can't we pass only the ```self``` and get the object store from it?

karlhigley: Yeah, we could. This `MessageHandler` in particular has a dependency on the worker, but in general they should only depend on the `ObjectStore`. If we find that more of them have dependencies on the worker we can consolidate these params, but I'm still trying to encourage future message handlers to be relatively self-contained.

syft/generic/abstract/message_handler.py
```diff
@@ -0,0 +1,20 @@
...
+
+    @abstractmethod
+    def init_routing_table(self):
+        return {}
+
+    def supports(self, msg):
+        return type(msg) in self.routing_table.keys()
+
+    def handle(self, msg):
+        return self.routing_table[type(msg)](msg)
```
youben11, line 20: Why not raising an error here if type(msg) isn't in the routing table and just catch it in workers to see what to do with it?

karlhigley: I considered that, but then I remembered _The_ ~~10~~ 6 _Commandments Of PySyft Development_:

1. Thou shalt make everything a tensor type.
2. Thou shalt not implement messaging systems that circumvent `BaseWorker`.
3. Thou shalt not send data without using either `.send()`, `.get()`, or `.move()`.
4. Thou shalt not use `exec()`.
5. **Thou shalt not implement logic using try/except unless the logic itself does not call anywhere else in PySyft's codebase (purely external libraries.)**
6. Thou shalt never expose context managers to the user.

For the record, I don't make these rules. These were all @iamtrask. 😂

PR #3663, opened 2020-06-02
Convert all `create_pointer()` methods to instance methods

karlhigley:
## Description
This PR converts all `create_pointer()` methods in PySyft to instance methods, in order to both make them consistent and make them easier to hook for prep/clean-up actions provided by the members of a tensor chain.

## Checklist

### DIFF
--- a/syft/generic/abstract/sendable.py
+++ b/syft/generic/abstract/sendable.py
@@ -53,4 +53,4 @@ class AbstractSendable(AbstractObject, SyftSerializable):
         garbage_collect_data: bool = True,
         **kwargs,
     ) -> "AbstractPointer":
-        pass
+        raise NotImplementedError
--- a/syft/generic/abstract/sendable.py
+++ b/syft/generic/abstract/sendable.py
@@ -1,5 +1,17 @@
+# from abc import abstractmethod
+from typing import TYPE_CHECKING
+from typing import Union
+
+import syft as sy
 from syft.serde.syft_serializable import SyftSerializable
 from syft.generic.abstract.object import AbstractObject
+
+# from syft.generic.abstract.pointer import AbstractPointer
+from syft.workers.abstract import AbstractWorker
+
+# this if statement avoids circular imports between base.py and pointer.py
+if TYPE_CHECKING:
+    from syft.generic.abstract.pointer import AbstractPointer
 
 
 class AbstractSendable(AbstractObject, SyftSerializable):
@@ -7,5 +19,38 @@ class AbstractSendable(AbstractObject, SyftSerializable):
     This layers functionality for sending objects between workers on top of AbstractObject.
     """
 
-    def send(self, destination):
-        return self.owner.send_obj(self, destination)
+    def send(
+        self, destination: Union[AbstractWorker, str], garbage_collect_data=None
+    ) -> "AbstractPointer":
+        ptr_id = sy.ID_PROVIDER.pop()
+
+        destination = self.owner.get_worker(destination)
+
+        pointer = self.create_pointer(
+            location=destination,
+            id_at_location=self.id,
+            owner=self.owner,
+            ptr_id=ptr_id,
+            garbage_collect_data=garbage_collect_data,
+        )
+
+        self.owner.send_obj(self, destination)
+
+        return pointer
+
+    # TODO: Make the `create_pointer` method below an abstract method. We can't do this yet, because
+    # we're relying on the wrapper tensors to provide the functionality of creating pointers, which
+    # means that many of the custom tensor types don't yet have pointer types or ways to create
+    # pointers.
+
+    # @abstractmethod
+    def create_pointer(
+        self,
+        location: AbstractWorker = None,
+        id_at_location: (str or int) = None,
+        owner: AbstractWorker = None,
+        ptr_id: (str or int) = None,
+        garbage_collect_data: bool = True,
+        **kwargs,
+    ) -> "AbstractPointer":
+        pass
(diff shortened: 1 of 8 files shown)


syft/generic/pointers/pointer_tensor.py
```diff
@@ -175,77 +176,6 @@ def get_class_attributes(self):
             "garbage_collect_data": self.garbage_collect_data,
         }
 
-    @staticmethod
```
karlhigley, line 178: Note that this `create_pointer` method is actually for creating pointers to plain tensors (not for creating pointers to `PointerTensors`.) It was only used in one place in PySyft, so it has been combined into the `TorchTensor.create_pointer()` method.

syft/generic/pointers/object_pointer.py
```diff
@@ -41,45 +37,6 @@ class ObjectPointer(AbstractSendable, SyftSerializable):
     in the BaseWorker object in self.location.
     """
 
-    def __init__(
```
karlhigley, line 44: Moved into `AbstractPointer`.

syft/workers/abstract.py
```diff
@@ -36,3 +38,31 @@ def _recv_msg(self, message: bin):
             message: The binary message being received.
         """
         pass
+
```
karlhigley, line 41: Must be defined by worker classes, but this leaves the communication and messaging methods completely unspecified, so that various worker types can fill them in.

syft/generic/abstract/sendable.py
```diff
@@ -1,11 +1,56 @@
...
+    ) -> "AbstractPointer":
+        ptr_id = sy.ID_PROVIDER.pop()
+
+        destination = self.owner.get_worker(destination)
+
+        pointer = self.create_pointer(
+            location=destination,
+            id_at_location=self.id,
+            owner=self.owner,
+            ptr_id=ptr_id,
```
AlanAboudib, line 33: I think `ptr_id` can be ignored here, since it will be assigned a value using `sy.ID_PROVIDER` when `AbstractPointer` is initialized with `id=None`.

syft/generic/abstract/sendable.py
```diff
@@ -1,11 +1,56 @@
...
+    def create_pointer(
+        self,
+        location: AbstractWorker = None,
+        id_at_location: (str or int) = None,
+        owner: AbstractWorker = None,
+        ptr_id: (str or int) = None,
+        garbage_collect_data: bool = True,
+        **kwargs,
+    ) -> "AbstractPointer":
+        pass
```
AlanAboudib, line 56: I think it is a good idea to raise `NotImplementedException` here. This is useful for debugging whenever we try to send an object that does not override this method.

syft/generic/pointers/object_pointer.py
```diff
@@ -41,45 +37,6 @@ class ObjectPointer(AbstractSendable, SyftSerializable):
...
-                .grad.child.child). Defaults to None, which means don't point to any attr,
-                just point to then object corresponding to the id_at_location.
-        """
-        super().__init__(id=id, owner=owner, tags=tags, description=description)
-
-        self.location = location
-        self.id_at_location = id_at_location
-        self.garbage_collect_data = garbage_collect_data
-        self.point_to_attr = point_to_attr
-
```
AlanAboudib, line 82: I think the `wrap` method below should be moved to `PointerTensor` since it is designed for tensors in mind. When we inherit directly from `ObjectPointer`, we do not want it to be a pointer to a tensor.

AlanAboudib: In SyferText. Although it does not raise any error.

syft/generic/abstract/sendable.py
karlhigley: Yeah, that might be a reasonable compromise. Making it an `abstractmethod` raises an error immediately when it detects that a child class hasn't implemented it, which is too strict for an incremental rework. Raising `NotImplemented` here is a good way to split the difference in the mean time. 👍

syft/generic/pointers/object_pointer.py
karlhigley: Hmm, yeah, that's a good point. I've been mostly ignoring wrappers for the moment, because the series of changes I've been working on this week is intended to make it easier to remove wrappers altogether.

syft/generic/abstract/sendable.py
karlhigley: That's true in the concrete case, but can't be depended on in this abstract class. (If the id assignment moved into this class, then I'd be on board with ignoring the pointer ids here.)

syft/generic/pointers/object_pointer.py
AlanAboudib: Oh, if wrappers are going to disappear soon, we might keep it for now, not to waste time on something that is going to disappear anyway in a later PR. It is not breaking or anything for SyferText

syft/generic/abstract/sendable.py
AlanAboudib: I am not sure to understand the point. I meant if we are going always to use 

```
id = sy.ID_PROVIDER.pop() 
```

Inside the above `send()` method then it becomes exactly equivalent to ignoring the attribute altogether (I mean only inside the `send()` method)

karlhigley: `send` isn’t necessarily the only place where `create_pointer` will get called. If `create_pointer` was an implementation detail of `send` or if it was always responsible for assigning ids (absorbing that responsibility from other classes), then we could ignore the param.

AlanAboudib: This makes sense

syft/generic/abstract/sendable.py
AlanAboudib: I think your forgot to add the `NotImplementedException` ?

PR #3676, opened 2020-06-06
CrypTen Message Handler

gmuraru:
## Description
Add a message handler for the CrypTen framework that should route the messages to the correct method.

## Affected Dependencies

## How has this been tested?
-  The CrypTen tests are still passing

## Checklist


syft/frameworks/crypten/context.py
```diff
@@ -127,8 +127,7 @@ def wrapper(*args, **kwargs):
 
             rank_to_worker_id = dict(zip(range(0, len(workers)), [worker.id for worker in workers]))
 
-            sy.local_worker.add_crypten_support()
-            sy.local_worker._set_rank_to_worker_id(rank_to_worker_id)
+            sy.local_worker.rank_to_worker_id = rank_to_worker_id
```
gmuraru, line 130: This is kind of ugly. This ```rank_to_worker_id``` is used in the ```load``` function from ```crypten/__init__py```...Thinking how we could remove this or refactor the ```load``` function

syft/__init__.py
```diff
@@ -34,6 +38,11 @@
 else:
     logger.info("TF Encrypted Keras not available.")
 
+if dependency_check.crypten_available:
```
gmuraru, line 41: Added this here and not in ```syft/generic/frameworks/types.py``` because of circular dependecies.

syft/frameworks/crypten/context.py
youben11: We can add some functions that take a worker as argument

gmuraru:
> Wasn't the initial design to leave every worker choose the message handlers to use, instead of adding all registered message handler in `syft.framework_message_handler`?

The initial design (like we had before) was like that (by using ```add_support``` from the ```BaseWorker```).

Now we automatically add the support if we detect a framework (currently is only for ```CrypTen```).

I tried to follow those steps from the discussion between us and @karlhigley:
```
And when adding a new messageHandler we do:
* Extend AbstractMessageHandler
* in the init_routing_table add our methods (routers) for the specific framework
* append a new instance of that class to our self.message_handlers
```

Or shouldn't we add the handlers automatically?

youben11:
yeah, that's what I meant, like every message_handler have a function for checking if the deps are okay, and every worker implements the message handlers he need (if the deps are okay for that specific message handler).

gmuraru:
> yeah, that's what I meant, like every message_handler have a function for checking if the deps are okay, and every worker implements the message handlers he need (if the deps are okay for that specific message handler).

Isn't that happening in ```syft/__init__.py```? There I check if we have the framework installed and I add that framework to a variable ```framework_message_handlers```

youben11:
yeah, but it is kind of decoupled from the message handler itself. I was thinking more like the message handler checking that in a static function, and workers using some handlers if their deps are okay.

syft/workers/base.py
```diff
@@ -994,3 +994,6 @@ def is_framework_supported(cls, framework: str) -> bool:
         :return: True/False
         """
         return framework.lower() in framework_packages
+
+
+BaseWorker.register_message_handlers()
```
gmuraru, line 999: @youben11 moved the logic to the worker level.
Something like this you meant?

@karlhigley what do you think about this?

syft/workers/base.py
```diff
@@ -986,6 +979,13 @@ def force_detail(worker: AbstractWorker, worker_tuple: tuple) -> tuple:
 
         return result
 
+    @staticmethod
+    def register_message_handlers():
+        if sy.dependency_check.crypten_available:
+            from syft.frameworks.crypten.message_handler import CryptenMessageHandler
+
+            BaseWorker._framework_message_handler["crypten"] = CryptenMessageHandler
+
```
youben11, line 989: No need for a function to do that, that attribute could be initialized at first, and message handlers would be appended only if the dependency check passes, which is something the message hander itself checks for (with a static function `MessageHandler.is_supported()` for example)

gmuraru: Hmmm...or couldn't we add this ```registry mechanism``` directly in the ```CryptenMessageHandler``` class file (if we import that file it means we have the dependencies hence we should add it to the ```_framework_message_handlers```?)

@karlhigley what do you think?

syft/frameworks/crypten/message_handler.py
```diff
@@ -0,0 +1,90 @@
...
+
+        plan = plans[0].get()
+
+        rank = None
+        for r, worker_id in self.worker.rank_to_worker_id.items():
+            if worker_id == self.worker.id:
+                rank = r
+                break
+
+        assert rank is not None
```
LaRiffle, line 58: This code is used twice, could be refactor. it?

PR #3682, opened 2020-06-07
Add mean and var methods for AST

LaRiffle:
## Description
Add mean and var methods for AST

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/additive_shared.py
+++ b/syft/frameworks/torch/tensors/interpreters/additive_shared.py
@@ -155,6 +155,13 @@ class AdditiveSharingTensor(AbstractTensor):
         for share in self.child.values():
             return share.shape
 
+    def numel(self):
+        """
+        Return the number of elements
+        """
+        for share in self.child.values():
+            return share.numel()
+
     @property
     def min_value(self):
         if self._min_value is None:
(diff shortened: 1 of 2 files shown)


syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -720,6 +727,18 @@ def chunk(self, shares, *args, **kwargs):
...
         return results
 
+    @overloaded.method
+    def mean(self, shares, **kwargs):
+        result = {}
+        m = 0
+        for worker, share in shares.items():
+            sum_value = share.sum(**kwargs)
+            if m == 0:
+                m += share.numel() // sum_value.numel()
```
Syzygianinfern0, line 737: We can maybe having the  `m` and `sum_value` initialization to just happen just once, why include it in the loop?

```suggestion
        sum_value = share.sum(**kwargs)
        m = next(iter(shares.values())).numel() // sum_value.numel()
        for worker, share in shares.items():
```

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -687,6 +687,12 @@ def eq(self, _self, other):
 
     __eq__ = eq
 
+    def var(self, **kwargs):
+        # Note that this implem is different from pytorch (which divides by m-1)
```
Syzygianinfern0, line 691: Can you provide some more info on this :smile:

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -687,6 +687,12 @@ def eq(self, _self, other):
...
+        mu = self.mean(**kwargs)
+        unbiased_self = self - mu
+        return (unbiased_self * unbiased_self).mean(**kwargs)
```
Syzygianinfern0, line 694: :+1: 

This seems to be the correct way of "variance"
What does torch do different?

PR #3685, opened 2020-06-09
Removed `wrap()` on non-tensor pointers in `respond_to_search()`

AlanAboudib:
## Description

in `syft/workers/message_handlers:respond_to_search()`, all pointers are wrapped before being returned. This should be the case only on `PointerTensor` objects, not only all pointers.

This causes a fail in SyferText where we need to exchange object pointers that are not tensors.

## Affected Dependencies
None

## How has this been tested?
All existing tests

## Checklist

### DIFF
--- a/syft/workers/message_handler.py
+++ b/syft/workers/message_handler.py
@@ -251,7 +251,7 @@ class BaseMessageHandler(AbstractMessageHandler):
             # Wrap only if the pointer points to a tensor.
             # If it points to a generic object, do not wrap.
             if isinstance(obj, torch.Tensor):
-                ptr.wrap()
+                ptr = ptr.wrap()
 
             results.append(ptr)
 
--- a/syft/workers/message_handler.py
+++ b/syft/workers/message_handler.py
@@ -245,7 +245,7 @@ class BaseMessageHandler(AbstractMessageHandler):
             # decision to decide when to delete the tensor.
             ptr = obj.create_pointer(
                 garbage_collect_data=False, owner=sy.local_worker, tags=obj.tags
-            ).wrap()
+            )
 
             results.append(ptr)
 
--- a/syft/workers/message_handler.py
+++ b/syft/workers/message_handler.py
@@ -249,7 +249,7 @@ class BaseMessageHandler(AbstractMessageHandler):
 
             # Wrap only if the pointer points to a tensor.
             # If it points to a generic object, do not wrap.
-            if isinstance(obj, PointerTensor):
+            if isinstance(ptr, PointerTensor):
                 ptr.wrap()
 
             results.append(ptr)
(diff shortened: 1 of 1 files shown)


syft/workers/message_handler.py
```diff
@@ -245,7 +245,13 @@ def respond_to_search(self, msg: SearchMessage) -> List[PointerTensor]:
             # decision to decide when to delete the tensor.
             ptr = obj.create_pointer(
                 garbage_collect_data=False, owner=sy.local_worker, tags=obj.tags
-            ).wrap()
+            )  # .wrap()
+
+            # Wrap only if the pointer points to a tensor.
+            # If it points to a generic object, do not wrap.
+            if isinstance(obj, PointerTensor):
+                ptr.wrap()
```
gmuraru, line 253: Q: if you remove this if completely are the tests failing?

AlanAboudib: Just pushed that to test

gmuraru: Yeah...it is failling :(, it was ok before :+1:

AlanAboudib: I reverted my changes and it still giving an error :O

shubham3121: @AlanAboudib it seemed to be an intermittent and was not the result of your change.It was happening because of timeout when waiting for an async response. I re-ran the tests and all tests passed. You can make your change and we can debug further if the issue happens again.

AlanAboudib: Thank you @shubham3121, I will push the changes then

gmuraru: I think @shubham3121 referred to the changes with
```
    # Wrap only if the pointer points to a tensor.
            # If it points to a generic object, do not wrap.
            if isinstance(obj, PointerTensor):
                ptr.wrap()
```

AlanAboudib: I was trying your idea of just remove `.wrap()` without adding my code. I just push my changes back now

shubham3121: @AlanAboudib , Ques: You want to wrap Tensor objects (not necessarily PointerTensors objects) and not non Tensor objects, is that the behaviour you're looking for ?? 
If so, then the check should be `isinstance(obj, torch.Tensor)`. Because the currently the test is failing because the obj is an Tensor instance and not a PointerTensor object, as a result we are not wrapping it and hence results in error of doing .shape.

AlanAboudib: Oh, my bad, I actually wanted to type:

```
if isinstance(ptr, PointerTensor):
```
Your solution works too. I will do

Thank you so much

shubham3121: @AlanAboudib `ptr.wrap()` should ` prt = ptr.wrap()` as it is not an inline operation.

PR #3689, opened 2020-06-10
Introduce an experimental mechanism for remote exec on Python workers

LaRiffle:
## Description
This is an experimental tool which really helps reducing communication rounds. It doesn't use Plans and is only compatible with Python workers, which is a limitation for PySyft.

However it is very useful for the moment for crypto protocols, as it highly reduces the number of communication and doesn't require to contain only pytorch operations (for example I can make a call to a custom library or use numpy code)

## How has this been tested?

## Checklist

### DIFF
--- a/syft/generic/frameworks/attributes.py
+++ b/syft/generic/frameworks/attributes.py
@@ -7,11 +7,25 @@ from typing import Any
 
 from syft.generic.frameworks.hook.hook import FrameworkHook
 
+allowed_commands = set()
+
 
 class FrameworkAttributes(ABC):
+    _allowed_commands = None
+
     @abstractmethod
     def __init__(self, framework: ModuleType, hook: FrameworkHook):
         pass
+
+    @property
+    def allowed_commands(self):
+        if self._allowed_commands is None:
+            self._allowed_commands = allowed_commands
+        return self._allowed_commands
+
+    @allowed_commands.setter
+    def allowed_commands(self, new_commands):
+        self._allowed_commands = new_commands
 
     # Forcing subclasses to define a class-level constant; see
     # https://stackoverflow.com/a/53417582 for nuance
--- a/syft/generic/utils.py
+++ b/syft/generic/utils.py
@@ -22,13 +22,15 @@ class memorize(dict):
 
 
 def allow_command(func):
-    func_name = f"{func.__module__}.{func.__name__}"
+    module = func.__module__
+    func_name = f"{module}{'.' if module else ''}{func.__name__}"
     allowed_commands.update({func_name})
     return func
 
 
 def remote(func, location):
-    command_name = f"{func.__module__}.{func.__name__}"
+    module = func.__module__
+    command_name = f"{module}{'.' if module else ''}{func.__name__}"
 
     worker = sy.local_worker
 
--- a/syft/generic/utils.py
+++ b/syft/generic/utils.py
@@ -1,3 +1,7 @@
+from syft.generic.frameworks.attributes import allowed_commands
+import syft as sy
+
+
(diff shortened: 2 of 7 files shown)


syft/generic/frameworks/attributes.py
```diff
@@ -7,12 +7,26 @@
...
+    _allowed_commands = None
+
     @abstractmethod
     def __init__(self, framework: ModuleType, hook: FrameworkHook):
         pass
 
+    @property
+    def allowed_commands(self):
+        if self._allowed_commands is None:
+            self._allowed_commands = allowed_commands
```
gmuraru, line 23: Q: In Python this is doing copy of the alllowed_commands, right?

syft/workers/base.py
```diff
@@ -166,6 +166,8 @@ def __init__(
         # storage object for crypto primitives
         self.crypto_store = PrimitiveStorage(owner=self)
 
+        self.syft = sy
```
gmuraru, line 169: Q: Where is this needed? Can't we access it with ```worker.hook.syft```?

syft/generic/frameworks/attributes.py
LaRiffle: I don't think so! But cant' be sure :/

syft/workers/base.py
LaRiffle: Yeah it's a trick, because when you will register with @ allow_command a function "syft.foo.bar", a worker executing the command will try to execute it as a method so "worker.syft.foo.bar", just like we to for torch, when you do "torch.add" it tries "worker.torch.add" that's why I added it...

syft/generic/utils.py
```diff
@@ -15,3 +19,32 @@ def __call__(self, *args):
...
+
+def remote(func, location):
+    command_name = f"{func.__module__}.{func.__name__}"
+
+    worker = sy.local_worker
+
+    if isinstance(location, str):
+        location = worker.get_worker(location)
+
+    def remote_exec(*args, return_value=False, return_arity=1, multiprocessing=False, **kwargs):
```
youben11, line 38: what if the function returns more than a value? (`return_arity=1`)

LaRiffle: If so, the existing mechanism with the ResponseSignatureError is used!
The idea is to have a clean way to avoid this if you already know you're gonna return `n` values

LaRiffle:
> This looks like a nice feature to have! And so, the functions that can be allowed should reside under the `syft` package, right?

Exactly!!
More to come on FSS to see how this can be intensively used :)

syft/generic/utils.py
```diff
@@ -15,3 +19,32 @@ def __call__(self, *args):
     def __missing__(self, key):
         result = self[key] = self.func(*key)
         return result
+
+
+def allow_command(func):
+    func_name = f"{func.__module__}.{func.__name__}"
```
gmuraru, line 25: Q: Is there any chance that ```func.__module``` might be ```""```

syft/generic/utils.py
youben11: And can't this be parameterized? If I want to execute a command that returns (True, 5) for example, can I use the remote decorator? It would be great if this can be passed as a parameter to the decorator with 1 being the default

test/generic/frameworks/test_attributes.py
```diff
@@ -0,0 +1,24 @@
+import pytest
```
gmuraru, line 1: MAYBE: Add a test that throws ```exceptions.ResponseSignatureError```?

LaRiffle:
Comments addressed!

PR #3708, opened 2020-06-15
New strategy for storing rank to id translation tables

### LINKED ISSUE #3516: Update how rank_to_worker_id is stored in workers
**Describe the bug**

An issue we had was that WebsocketServerWorker was storing the received rank_to_worker_id dictionary into `self` then our `crypten.load_from_party` function was trying to get that from `syft.local_worker`, a quick fix was made to store it in `syft.local_worker` as well as self, this will causes issues if :

- The worker running the server is also doing some crypten computation
- The worker running the server is doing crypten computation locally (using virtual workers)

The former was already present, but the second got introduced by the quick fix.

**Possible solution**

Store it in `self` and fetch it from the specific worker on `load_from_party`.

### LINKED ISSUE #3687: Rank to Worker_ID translation should be linked to a specific CrypTen computation
## Description

Server workers are intended to run multiple queries from client in parallel. Starting multiple CrypTen computation on the same worker at a time should be possible, however, due to the current way of storing the rank-to-worker-id translation, starting a new computation will overwrite the old translation dictionary and lead to undefined behavior.

## How to Reproduce
1. Start a server worker supporting CrypTen messages
2. Start two consecutive and long CrypTen computations with different workers involved

## Expected Behavior
Running parallel CrypTen computation should work normally as if they were run sequentially.

## Additional Context
Also see #3516

youben11:
## Description

Each CrypTen computation is identified using a CID (computation id), it might be different in different workers. Translation tables are indexed using CIDs. Global table and the CID of the running computation are stored under `syft.frameworks.crypten`, we assume that parties are separate processes and that CID is overwritten after process creation, which means that even if parties are running in the same machine, they won't manipulate the global CID concurrently.

`get_worker_from_rank` was also updated and I assumed that `syft.local_worker` is able to find all other workers using their id, please correct me if that's not the case.

Fixes #3516 
Fixes #3687 

## Checklist

### DIFF
--- a/syft/frameworks/crypten/__init__.py
+++ b/syft/frameworks/crypten/__init__.py
@@ -6,7 +6,10 @@ import crypten
 from syft.workers.base import BaseWorker
 
 
-RANK_TO_WORKER_ID = {}
+RANK_TO_WORKER_ID = {
+    # Contains translation dictionaries for every computation.
+    # cid (computation id): {rank_to_worker_id dictionary for a specific computation}
+}
 CID = None
 
 
--- a/syft/frameworks/crypten/__init__.py
+++ b/syft/frameworks/crypten/__init__.py
@@ -30,7 +30,7 @@ def get_worker_from_rank(rank: int, cid: int = None) -> BaseWorker:
     if rank_to_worker_id is None:
         raise RuntimeError(
             "CrypTen computation not initiated properly, computation_id doesn't match any rank to"
-            "worker_id transaltion table"
+            "worker_id translation table"
         )
     return syft.local_worker._get_worker_based_on_id(rank_to_worker_id[rank])
 
--- a/syft/frameworks/crypten/__init__.py
+++ b/syft/frameworks/crypten/__init__.py
@@ -3,10 +3,38 @@ import syft
 import crypten.communicator as comm
 import crypten
 
+from syft.workers.base import BaseWorker
+
+
+RANK_TO_WORKER_ID = {}
+CID = None
+
+
+def get_worker_from_rank(cid: int, rank: int) -> BaseWorker:
+    """Find the worker running CrypTen party with specific rank in a certain computation.
+
+    Args:
+        cid: CrypTen computation id.
+        rank: rank of the CrypTen party.
+
+    Returns:
+        BaseWorker corresponding to cid and rank.
+    """
+    rank_to_worker_id = RANK_TO_WORKER_ID.get(cid, None)
+    if rank_to_worker_id is None:
+        raise RuntimeError(
+            "CrypTen computation not initiated properly, computation_id doesn't match any rank to"
+            "worker_id transaltion table"
+        )
+    return syft.local_worker._get_worker_based_on_id(rank_to_worker_id[rank])
+
 
 def load(tag: str, src: int, **kwargs):
     if src == comm.get().get_rank():
-        worker = syft.local_worker.get_worker_from_rank(src)
+        if CID is None:
+            raise RuntimeError("CrypTen computation id is not set.")
+
+        worker = get_worker_from_rank(CID, src)
         results = worker.search(tag)
 
         # Make sure there is only one result
--- a/syft/frameworks/crypten/__init__.py
+++ b/syft/frameworks/crypten/__init__.py
@@ -10,16 +10,22 @@ RANK_TO_WORKER_ID = {}
 CID = None
 
 
-def get_worker_from_rank(cid: int, rank: int) -> BaseWorker:
+def get_worker_from_rank(rank: int, cid: int = None) -> BaseWorker:
     """Find the worker running CrypTen party with specific rank in a certain computation.
 
     Args:
-        cid: CrypTen computation id.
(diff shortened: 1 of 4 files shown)


syft/frameworks/crypten/__init__.py
```diff
@@ -3,10 +3,38 @@
...
+    Returns:
+        BaseWorker corresponding to cid and rank.
+    """
+    rank_to_worker_id = RANK_TO_WORKER_ID.get(cid, None)
+    if rank_to_worker_id is None:
+        raise RuntimeError(
+            "CrypTen computation not initiated properly, computation_id doesn't match any rank to"
+            "worker_id transaltion table"
+        )
+    return syft.local_worker._get_worker_based_on_id(rank_to_worker_id[rank])
```
youben11, line 29: Are all workers recognized by the local_worker?

gmuraru: From what I know, the local worker is automatically created when we import ```syft``` and it should know about all the workers. (@LaRiffle this is the case, right?)

syft/frameworks/crypten/message_handler.py
```diff
@@ -66,24 +65,39 @@ def run_crypten_party_jail(self, msg: CryptenInitJail):
...
-        for r, worker_id in self.worker.rank_to_worker_id.items():
+        for r, worker_id in rank_to_worker_id.items():
             if worker_id == self.worker.id:
                 rank = r
                 break
         return rank
+
+    @staticmethod
+    def _get_computation_id():
+        cid = randint(0, 2 ** 32)
```
gmuraru, line 100: Q: Can't we use the ```id provider``` from syft?

youben11: if it's okay to consume IDs from their than I'm okay with it, however, we can have overlapping IDs in the current implementation without any problem, none of them depends on the other.

syft/frameworks/crypten/message_handler.py
```diff
@@ -66,24 +65,39 @@ def run_crypten_party_jail(self, msg: CryptenInitJail):
             An ObjectMessage containing the return value of the crypten function computed.
         """
 
-        self.worker.rank_to_worker_id, world_size, master_addr, master_port = msg.crypten_context
+        rank_to_worker_id, world_size, master_addr, master_port = msg.crypten_context
+
+        cid = CryptenMessageHandler._get_computation_id()
+        syft_crypten.RANK_TO_WORKER_ID[cid] = rank_to_worker_id
```
gmuraru, line 71: Q: Does it make more sense to name this ```syft_crypten.COMPUTATION_DATA```?

syft/frameworks/crypten/__init__.py
```diff
@@ -3,10 +3,44 @@
...
+    if cid is None:
+        if CID is None:
+            # Neither CID have been set appropriately nor cid have been passed
+            raise ValueError("cid must be set.")
+        cid = CID
+
+    rank_to_worker_id = RANK_TO_WORKER_ID.get(cid, None)
+    if rank_to_worker_id is None:
+        raise RuntimeError(
+            "CrypTen computation not initiated properly, computation_id doesn't match any rank to"
```
gmuraru, line 32: NIT: Leave one empty space at the end

syft/frameworks/crypten/__init__.py
```diff
@@ -3,10 +3,44 @@
...
+            "worker_id transaltion table"
```
gmuraru, line 33: NIT: ```transaltion``` --> ```translation```

syft/frameworks/crypten/__init__.py
LaRiffle: Yup! For virtual workers this is true. For GridNodes it knows automatically the client nodes

syft/frameworks/crypten/message_handler.py
LaRiffle: Yes I guess we should use the id provider!

LaRiffle: And you can brun man of them if need as you do with the wile loop

syft/frameworks/crypten/__init__.py
```diff
@@ -3,10 +3,44 @@
 import crypten.communicator as comm
 import crypten
 
+from syft.workers.base import BaseWorker
+
+
+RANK_TO_WORKER_ID = {}
```
LaRiffle, line 9: Add a comment to say that its a dict whose key are the CID and values are another dict which maps ranks to worker ids.

Also, the current way it's setup I think computation are synchronous and because you delete the key at the end of each computation, is there a scenario where there would be > 1 keys?

youben11: A GridNode for example can receive different request to run a CrypTen computation at the same time, so the number of keys depends on the number of parallel requests.

youben11: it doesn't affect the current way we are using it because we always run a single computation at a time, but I expect a production server to have this use case.

PR #3719, opened 2020-06-18
Fix grid model hosting for large models

IonesioJunior:
## Description
Due to the need to support uploads for large models **_(>=200MB)_**, the approach used by the `serve_model` method has been modified. We changed the use of standard WebSockets messages to a well-known solution used for uploading large files on HTTP servers.

Fix [#23](https://github.com/OpenMined/PyGridNode/issues/23) 

## Affected Dependencies
- **requests_toolbelt**

## How has this been tested?
```
import syft as sy
import torch as th
from syft.workers.node_client import NodeClient
hook = sy.TorchHook(th)

node = NodeClient("ws://<node_address>")

traced_model = th.jit.load('large_traced_model.jit')

node.serve_model(traced_model,
                 model_id="myCustomModel",
                 allow_download=True,
                 allow_remote_inference=True)
```

## Checklist
- [X] I have commented on my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)

### DIFF
--- a/syft/workers/node_client.py
+++ b/syft/workers/node_client.py
@@ -1,4 +1,6 @@
 import json
+import requests
+from requests_toolbelt import MultipartEncoder, MultipartEncoderMonitor
 
 from typing import Union
 from urllib.parse import urlparse
@@ -239,7 +241,7 @@ class NodeClient(WebsocketClientWorker):
         else:
             res_model = model
 
-        serialized_model = serialize(res_model)
+        serialized_model = serialize(res_model).decode(self.encoding)
 
         message = {
             REQUEST_MSG.TYPE_FIELD: REQUEST_MSG.HOST_MODEL,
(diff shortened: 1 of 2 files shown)


syft/workers/node_client.py
```diff
@@ -248,10 +250,31 @@ def serve_model(
...
+            "model": serialized_model,
         }
-        response = self._forward_json_to_websocket_server_worker(message)
-        return self._return_bool_result(response)
+
+        url = self.address.replace("ws", "http") + "/serve-model/"
+
+        # Multipart encoding
+        form = MultipartEncoder(message)
+        upload_size = form.len
```
shubham3121, line 260: Can `form.len` be `None` ? If so, can be put a check and maybe raise a error of message size being null.

PR #3798, opened 2020-06-30
Fix Plan trace for inplace ops, add Role actions pruning

vvmnnnkv:
## Description
Changes:
1. Inplace operations like `X +=1` were traced as `var1 = X.__iadd__(1)` where `var1` is redundant variable that also hides the fact that the result of operation is `X` itself. Hook method is updated to return self as result for inplace operations.
2. Add Role actions pruning, e.g. if some operations/placeholders are not connected to Plan outputs or inputs (via side-effect), these are removed.

## Affected Dependencies
n/a

## How has this been tested?
Unit tests

## Checklist


syft/execution/role.py
```diff
@@ -312,6 +313,100 @@ def copy(self):
...
+
+        # Track operations on output placeholders from bottom to top
+        output_connected_placeholder_ids = set(self.output_placeholder_ids)
+        connected_actions_idx = set()
+        for action_rev_idx, action in enumerate(reversed(self.actions)):
+            return_ids = (
+                {ph.value for ph in action.return_ids if isinstance(ph, PlaceholderId)}
+                if action.return_ids is not None
+                else set()
+            )
```
LaRiffle, line 347: This is very nested and is not optimal for readability
```suggestion
            return_ids = set()
            if action.return_ids is not None:
                for ph in action.return_ids:
                    if isinstance(ph, PlaceholderId):
                        return_ids.add(ph.value)
```

syft/execution/role.py
```diff
@@ -312,6 +313,100 @@ def copy(self):
...
+                len(return_ids.intersection(output_connected_placeholder_ids)) > 0
+            )
+            # Operation has side-effect on placeholder connected to output
+            affects_connected_ph = (
+                target_id in output_connected_placeholder_ids
+                and self._is_inplace_operation(action.name, action.target)
+            )
+
+            if returns_connected_ph or affects_connected_ph:
+                connected_actions_idx.add(len(self.actions) - 1 - action_rev_idx)
```
LaRiffle, line 360: Could you add a line to explain why we don't use action_rev_idx (as we use action_idx below) ? (I guess, because we did a reverse enumerate)

syft/execution/role.py
```diff
@@ -312,6 +313,100 @@ def copy(self):
...
+
+    def _prune_actions(self):
+        """
+        Removes unnecessary actions and placeholders.
+        """
+
+        # Track operations on output placeholders from bottom to top
+        output_connected_placeholder_ids = set(self.output_placeholder_ids)
+        connected_actions_idx = set()
+        for action_rev_idx, action in enumerate(reversed(self.actions)):
```
LaRiffle, line 342: Can you add a line to explain why you need to reverse?

syft/execution/role.py
```diff
@@ -312,6 +313,100 @@ def copy(self):
...
+            )
+
+            if returns_connected_ph or affects_connected_ph:
+                connected_actions_idx.add(len(self.actions) - 1 - action_rev_idx)
+
+                def append_to_connected_ph(ph_id):
+                    output_connected_placeholder_ids.add(ph_id.value)
+
+                Role.nested_object_traversal(action.args, append_to_connected_ph, PlaceholderId)
+                Role.nested_object_traversal(action.target, append_to_connected_ph, PlaceholderId)
```
LaRiffle, line 366: I'm surprised that it implicitely provides `output_connected_placeholder_ids` but why not!

PR #3807, opened 2020-07-01
Fix issues in handle_func_command in syft tensors

LaRiffle:
## Description
This cleans code in handle_func_command and fix bugs due to variable overwriting for syft tensors, and to overloading functions in native.py

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/native.py
+++ b/syft/frameworks/torch/tensors/interpreters/native.py
@@ -299,11 +299,9 @@ class TorchTensor(AbstractTensor):
         """
         Operates as a router for functions. A function call always starts
         by being handled here and 3 scenarii must be considered:
-
         Real Torch tensor:
             The arguments of the function are real tensors so we should
             run the native torch command
-
         Torch wrapper:
             The arguments are just wrappers at the top of a chain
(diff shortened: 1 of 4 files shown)


gmuraru:
Q: Could you add a test for this?

LaRiffle:
> Q: Could you add a test for this?

I've updated the code once more maybe it's clearer now. It's hard to add tests for this, but it's needed for batchnorm that I wanted to put in a separate PR and which is tested. Sounds good?
(and the main function at stakes here, torch.roll, is tested)

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -356,7 +372,7 @@ def handle_func_command(cls, command):
             # Check that the function has not been overwritten
```
gmuraru, line 372: PICKY Stuff: Here the comment should be removed, right? Since we look for ```native_{cmd}```

LaRiffle: haha no it's good to be picky. here the module `native_torch` defined in native.py overwrites native torch functions... Yeah it's a bit strange, I really had to modify a torch function (aka torch.roll) to behave slightly differently

Issue #3813, opened 2020-07-03
How PySyft deal with big training dataset?


NeuZhangQiang:
## Question
How can PySyft deal with big data?

## Further Information
Pysyft use data.send(worker) to share data. What if the dataset is very large? For example, the size of dataset is 100g. 
The PyTorch use DataLoader and Keras use yield to load one batch of data in one time. Can PySyft implement the similar function?
The FederatedDataLoader seem to a solution. But how can we convert the DataLoader (PyTorch) to FederatedDataLoader?

And is it possible to use fit_generator in Pysyft? Any suggestion is appreciate!

chenwanqq:
Personally I haven't found such a method. many base class in PySyft, such as federatedclient, baseworker, Basedataset, directly use Tensor rather than dataset/dataloader. Unforunately the major FL framework, such as  Pysyft, Tensorflow-Federated and FATE don't support big data loading, which strongly limited the usage in many areas, such as CV. 
哥们我跟你研究的方向很像，有兴趣可以通过邮件联系wanqi0605@qq.com

typhoon1104:
hello，I have the question too.I use dataLoader and dataset.But I have a new question that can not get model.

NeuZhangQiang:
@typhoon1104 You need to have a look at this [tutorial](https://github.com/OpenMined/PySyft/tree/v0.2.5/examples/tutorials/advanced/websockets_mnist_parallel). Please note that only pysyft 0.2.5 support this tutorial, while the pysyft 0.2.6 do not support. And I do not know why pysyft 0.2.6 remove this function.

typhoon1104:
> @typhoon1104 You need to have a look at this [tutorial](https://github.com/OpenMined/PySyft/tree/v0.2.5/examples/tutorials/advanced/websockets_mnist_parallel). Please note that only pysyft 0.2.5 support this tutorial, while the pysyft 0.2.6 do not support. And I do not know why pysyft 0.2.6 remove this function.

I use dataset and dataloader of pytorch 1.5.0, it not in PySyft.

NeuZhangQiang:
@typhoon1104 you can try pytorch 1.4.0.

typhoon1104:
I find it is not a question.
But I find a new question that model_str.get() is slow.
It takes 60 seconds to get a model of resnet34.

PR #3823, opened 2020-07-06
updating the websocket client's timeout

wmlba:
## Description
Increasing the timeout will help in solving a lot of timeout issues especially with tensor.send() when the tensor is large and model.send() when the model is big.
This should solve issue #3813 and some other timeout issues.

### DIFF
--- a/syft/workers/websocket_client.py
+++ b/syft/workers/websocket_client.py
@@ -17,7 +17,7 @@ from syft.workers.base import BaseWorker
 
 logger = logging.getLogger(__name__)
 
-TIMEOUT_INTERVAL = 60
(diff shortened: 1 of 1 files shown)


wmlba:
Hopefully the timeout can be specified as an argument when creating the websocket client later.

gmuraru:
Restarted the tests! LGTM!

gmuraru:
Following @youben11 comment, I think we can close this PR, right @wmlba?

wmlba:
> Following @youben11 comment, I think we can close this PR, right @wmlba?

Yes, please. Thanks!

Issue #3852, opened 2020-07-13
FALCON: Basic Operations


abogaziah:
## Feature Description
To build Falcon protocol we will need to build some basic building blocks:

- [X] Select Shares i.e. given two shares(x,y) and a bit (b mod 2): select x if b== 0 ; else select y
- [X] XOR a secret with public bit
- [X] Evaluating [(-1)^B * x ]^m

## Additional Context

[FALCON papaer](https://arxiv.org/pdf/2004.02229v1.pdf)

IanQS:
Hey @abogaziah ! 

Sorry, but would it be possible to add a checkbox in front of each of those tickets so that we can see which ones are in progress or finished without needing to open each ticket? Sorry, it's been a while since I've used Github to track epics and milestones so I'm a little confused

abogaziah:
Hi @IanQS ! I updated the epic, we have split it into 2 milestones where you can know what's done and what's not:  
https://github.com/OpenMined/PySyft/milestone/15
https://github.com/OpenMined/PySyft/milestone/14
Falcon milestone isn't complete yet

gmuraru:
Solved

PR #3856, opened 2020-07-13
Replicated Sharing Tensor (RST): secret sharing, reconstruction, and linear operations

abogaziah:
## Description
This PR will implement the basic operations of the replicative secret sharing scheme in order to be able to implement falcon protocol
for more details: #3852 

## Checklist

Milestones: 
- [x] Secret sharing
- [x] Reconstruction
- [x] Public and private addition / sub

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/native.py
+++ b/syft/frameworks/torch/tensors/interpreters/native.py
@@ -924,20 +924,26 @@ class TorchTensor(AbstractTensor):
                 **kwargs_,
             )
         else:
-            if self.type() == "torch.FloatTensor":
-                raise TypeError("FloatTensor cannot be additively shared, Use fix_precision.")
+            if protocol == "snn":
+                if self.type() == "torch.FloatTensor":
+                    raise TypeError("FloatTensor cannot be additively shared, Use fix_precision.")
 
-            shared_tensor = (
-                syft.AdditiveSharingTensor(
-                    protocol=protocol,
-                    field=field,
-                    dtype=dtype,
-                    crypto_provider=crypto_provider,
-                    owner=self.owner,
+                shared_tensor = (
+                    syft.AdditiveSharingTensor(
+                        protocol=protocol,
+                        field=field,
+                        dtype=dtype,
+                        crypto_provider=crypto_provider,
+                        owner=self.owner,
+                    )
+                    .on(self.copy(), wrap=False)
+                    .init_shares(*owners)
                 )
-                .on(self.copy(), wrap=False)
-                .init_shares(*owners)
-            )
+            elif protocol == "falcon":
+                shared_tensor = syft.ReplicatedSharingTensor(owner=self.owner).share_secret(
+                    self, owners
+                )
+                return shared_tensor
 
         if requires_grad and not isinstance(shared_tensor, syft.PointerTensor):
             shared_tensor = syft.AutogradTensor().on(shared_tensor, wrap=False)
--- a/syft/frameworks/torch/tensors/interpreters/native.py
+++ b/syft/frameworks/torch/tensors/interpreters/native.py
@@ -909,6 +909,11 @@ class TorchTensor(AbstractTensor):
             requires_grad (bool): Should we add AutogradTensor to allow gradient computation,
                 default is False.
         """
+        if protocol == "falcon":
+            shared_tensor = syft.ReplicatedSharingTensor(owner=self.owner).share_secret(
+                self, owners
+            )
+            return shared_tensor
         if self.has_child():
             chain = self.child
 
@@ -924,26 +929,20 @@ class TorchTensor(AbstractTensor):
                 **kwargs_,
             )
         else:
-            if protocol == "snn":
-                if self.type() == "torch.FloatTensor":
-                    raise TypeError("FloatTensor cannot be additively shared, Use fix_precision.")
+            if self.type() == "torch.FloatTensor":
+                raise TypeError("FloatTensor cannot be additively shared, Use fix_precision.")
 
-                shared_tensor = (
-                    syft.AdditiveSharingTensor(
-                        protocol=protocol,
-                        field=field,
-                        dtype=dtype,
-                        crypto_provider=crypto_provider,
-                        owner=self.owner,
-                    )
-                    .on(self.copy(), wrap=False)
-                    .init_shares(*owners)
+            shared_tensor = (
+                syft.AdditiveSharingTensor(
+                    protocol=protocol,
+                    field=field,
+                    dtype=dtype,
+                    crypto_provider=crypto_provider,
+                    owner=self.owner,
                 )
-            elif protocol == "falcon":
-                shared_tensor = syft.ReplicatedSharingTensor(owner=self.owner).share_secret(
-                    self, owners
-                )
-                return shared_tensor
+                .on(self.copy(), wrap=False)
+                .init_shares(*owners)
+            )
 
         if requires_grad and not isinstance(shared_tensor, syft.PointerTensor):
             shared_tensor = syft.AutogradTensor().on(shared_tensor, wrap=False)
--- a/syft/frameworks/torch/tensors/interpreters/replicated_shared.py
+++ b/syft/frameworks/torch/tensors/interpreters/replicated_shared.py
@@ -79,6 +79,8 @@ class ReplicatedSharingTensor(AbstractTensor):
     def add(self, value):
         return self.__switch_public_private(value, self.public_add, self.private_add)
 
+    __add__ = add
+
     def public_add(self, plain_text):
         return self.public_linear_operation(plain_text, add)
 
@@ -87,6 +89,8 @@ class ReplicatedSharingTensor(AbstractTensor):
 
     def sub(self, value):
         return self.__switch_public_private(value, self.public_sub, self.private_sub)
+
+    __sub__ = sub
 
     def public_sub(self, plain_text):
         return self.public_linear_operation(plain_text, sub)
--- a/syft/frameworks/torch/tensors/interpreters/replicated_shared.py
+++ b/syft/frameworks/torch/tensors/interpreters/replicated_shared.py
@@ -1,8 +1,8 @@
-import torch
-from syft.generic.abstract.tensor import AbstractTensor
 import random
-import syft
 from operator import add, sub
+import torch
+import syft
+from syft.generic.abstract.tensor import AbstractTensor
 
 
 class ReplicatedSharingTensor(AbstractTensor):
@@ -15,7 +15,7 @@ class ReplicatedSharingTensor(AbstractTensor):
 
     def share_secret(self, secret, workers):
         number_of_shares = len(workers)
-        workers = self.__arrange_workers(workers)
+        workers = self.__arrange_workers(list(workers))
         shares = self.generate_shares(secret, number_of_shares)
         shares_map = self.__distribute_shares(workers, shares)
         self.child = shares_map
@@ -25,7 +25,7 @@ class ReplicatedSharingTensor(AbstractTensor):
     def __arrange_workers(workers):
         """ having local worker in index 0 saves one communication round"""
         if syft.hook.local_worker in workers:
-            workers.pop(syft.hook.local_worker)
+            workers.remove(syft.hook.local_worker)
             workers = [syft.hook.local_worker] + workers
         return workers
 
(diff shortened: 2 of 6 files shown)


syft/frameworks/torch/tensors/interpreters/replicative_shared.py
```diff
@@ -0,0 +1,65 @@
...
+            shares.append(torch.tensor(random.randrange(self.ring_size)))
+        shares.append(torch.tensor(secret - sum(shares) % self.ring_size))
+        return shares
+
+    @staticmethod
+    def distribute_shares(workers, shares):
+        shares_locations = {}
+        assert len(workers) == len(shares)
+        for i in range(len(shares)):
+            pointer1 = shares[i].send(workers[i])
```
gmuraru, line 33: Q: Shouldn't here the party[i] hold shares i-1 and i + 1?

abogaziah: No, it should know shares i and i+1. if it knows i-1 and i+1 it can reconstruct the secret.
section 3.1 of the paper states: "(x1, x2)
is held by P1, (x2, x3) by P2, and (x3, x1) by P3"

syft/frameworks/torch/tensors/interpreters/replicative_shared.py
```diff
@@ -0,0 +1,65 @@
...
+        self.ring_size = 2 ** 5
+
+    def share_secret(self, secret, workers):
+        number_of_shares = len(workers)
+        shares = self.generate_shares(secret, number_of_shares)
+        shares_locations = self.distribute_shares(workers, shares)
+        self.child = shares_locations
+        return self
+
+    def generate_shares(self, secret, number_of_shares=3):
```
gmuraru, line 21: MAYBE: make this ```staticmethod``` and test it with more shares?

abogaziah: I don't think this can be a static method because it uses an instance variable (self.ring_size), but we can still test it on more shares

gmuraru: PICKY: This is meant to be called only from inside the ```ReplicatedSharingTensor``` class, right? (in that case maybe we can prefix the method name with ```_``` such that we know it is internal?

gmuraru:
@abogaziah I removed some points from this task and created other issues for each one individually.
They can be seen in the [FALCON Milestone](https://github.com/OpenMined/PySyft/milestone/13)

syft/frameworks/torch/tensors/interpreters/replicative_shared.py
```diff
@@ -0,0 +1,154 @@
...
+        pointers = list(shares_map[players[0]])
+        pointers.append(shares_map[players[1]][1])
+        return pointers
+
+    def __sum_shares(self, shares):
+        return sum(shares) % self.ring_size
+
+    def __map_modular_to_real(self, mod_number):
+        """In a modular ring, a number x is mapped to a negative
+         real number ]0,-∞[ iff x > ring_size/2 """
```
gmuraru, line 74: TYPO: ```[``` should be ```]``` and ```]``` should be ```[```

syft/frameworks/torch/tensors/interpreters/replicative_shared.py
```diff
@@ -0,0 +1,154 @@
...
+import syft
+from operator import add, sub
+
+
+class ReplicatedSharingTensor(AbstractTensor):
+    def __init__(
+        self, shares_map=None, owner=None, id=None, tags=None, description=None,
+    ):
+        super().__init__(id=id, owner=owner, tags=tags, description=description)
+        self.child = shares_map
```
gmuraru, line 13: MAYBE: The ```shares_map``` should be only ```shares```. We can get the ```player``` for a share (or the owner) using the ```location``` attribute for the share

gmuraru: This is how it is done in the ```AdditiveSharingTensor```

syft/frameworks/torch/tensors/interpreters/replicative_shared.py
```diff
@@ -0,0 +1,154 @@
...
+        if not self.verify_matching_players(secret):
+            raise ValueError("Shares must be distributed among same parties")
+        z = {}
+        x, y = self.get_shares_map(), secret.get_shares_map()
+        for player in x.keys():
+            z[player] = (operator(x[player][0], y[player][0]), operator(x[player][1], y[player][1]))
+        return ReplicatedSharingTensor(z)
+
+    def public_linear_operation(self, plain_text, operator):
+        players = self.get_players()
```
gmuraru, line 119: MAYBE: If you use ```shares``` and not ```shares_map``` this should simply be ```[player for player in self.shares.location]```

syft/frameworks/torch/tensors/interpreters/replicative_shared.py
```diff
@@ -0,0 +1,154 @@
...
+        number_of_shares = len(workers)
+        workers = self.__arrange_workers(workers)
+        shares = self.generate_shares(secret, number_of_shares)
+        shares_map = self.__distribute_shares(workers, shares)
+        self.child = shares_map
+        return self
+
+    @staticmethod
+    def __arrange_workers(workers):
+        """ having local worker in index 0 saves one communication round"""
```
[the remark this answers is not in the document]
abogaziah: because the shares are local, you won't need to communicate (which is an expensive operation) with another worker to get them thus saved a communication round

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -909,6 +909,11 @@ def share(
             requires_grad (bool): Should we add AutogradTensor to allow gradient computation,
                 default is False.
         """
+        if protocol == "falcon":
+            shared_tensor = syft.ReplicatedSharingTensor(owner=self.owner).share_secret(
```
gmuraru, line 913: Q: I am thinking if we can keep a ```dict``` for those.
Like:
```dict = {
  "falcon": syft.ReplicatedSharingTensor,
  "snn"    : syft.AdditiveSharingTensor
}
```

syft/frameworks/torch/tensors/interpreters/replicative_shared.py
```diff
@@ -0,0 +1,154 @@
+import torch
```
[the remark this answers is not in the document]
abogaziah: according to clean coding principles, comments are very bad practice, they often are misleading, don't change with code. I made the code self-documented so comments are as least as possible, I tried adding comments but it was much harder to read with comments for me than without it

syft/frameworks/torch/tensors/interpreters/replicative_shared.py
```diff
@@ -0,0 +1,154 @@
...
+from syft.generic.abstract.tensor import AbstractTensor
+import random
+import syft
+from operator import add, sub
+
+
+class ReplicatedSharingTensor(AbstractTensor):
+    def __init__(
+        self, shares_map=None, owner=None, id=None, tags=None, description=None,
```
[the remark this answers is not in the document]
abogaziah: I haven't faced a scenario where I need to specify the dtype in this class so far, we can add it if we need it anytime

syft/frameworks/torch/tensors/interpreters/replicative_shared.py
abogaziah: that's not actually a typo, open bracket means the set doesn't include that number. negative numbers are everything between 0 and - infinity not including them, I can remove it if it's confusing, "negative real number" is kinda sufficient

syft/frameworks/torch/tensors/interpreters/replicative_shared.py
```diff
@@ -0,0 +1,154 @@
...
+    ):
+        super().__init__(id=id, owner=owner, tags=tags, description=description)
+        self.child = shares_map
+        self.ring_size = 2 ** 5
+
+    def share_secret(self, secret, workers):
+        number_of_shares = len(workers)
+        workers = self.__arrange_workers(workers)
```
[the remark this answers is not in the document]
abogaziah: let's add tests and see :"D

syft/frameworks/torch/tensors/interpreters/native.py
abogaziah: what would be the use-case of that dict? I don't get it

syft/frameworks/torch/tensors/interpreters/replicative_shared.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def generate_shares(self, plain_text, number_of_shares=3):
+        shares = []
+        for _ in range(number_of_shares - 1):
+            shares.append(torch.tensor(random.randrange(self.ring_size)))
+        shares.append(torch.tensor((plain_text - sum(shares)) % self.ring_size))
+        return shares
+
+    @staticmethod
+    def __distribute_shares(workers, shares):
```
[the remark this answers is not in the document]
abogaziah: " A single leading underscore means you're not supposed to access it "from the outside" -- two leading underscores (w/o trailing underscores) carry the message even more forcefully"
https://www.geeksforgeeks.org/private-methods-in-python/
https://stackoverflow.com/questions/1547145/defining-private-module-functions-in-python
http://www.faqs.org/docs/diveintopython/fileinfo_private.html

syft/frameworks/torch/tensors/interpreters/replicative_shared.py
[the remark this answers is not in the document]
abogaziah: done! I added tests and it works correctly now

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: Dict are better to dynamically add options if needed, and are faster to access instead of if / else-ing on all optiosn

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -0,0 +1,154 @@
...
+
+    @staticmethod
+    def __arrange_workers(workers):
+        """ having local worker in index 0 saves one communication round"""
+        if syft.hook.local_worker in workers:
+            workers.remove(syft.hook.local_worker)
+            workers = [syft.hook.local_worker] + workers
+        return workers
+
+    def generate_shares(self, plain_text, number_of_shares=3):
```
LaRiffle, line 32: I recommend taking example on https://github.com/OpenMined/PySyft/blob/4518a44700cefcd6efd4b43dbf7bf8ae79c52afb/syft/frameworks/torch/tensors/interpreters/additive_shared.py#L334
i.e. remove the noise you've added to the next share instead of removing the sum of all on the last share

abogaziah: What is the difference? in other words, why would you choose this over that?

gmuraru:
I think it looks great as a first PR that enables us to use RSS.
What I think we should implement more is - have the possibility to do operations using "+", "-", etc (I do not know if that is possible for the moment) - There might be needed some ```@overloaded``` work

abogaziah:
@gmuraru done :)

syft/frameworks/torch/tensors/interpreters/native.py
gmuraru: @abogaziah are you making the changes in this PR or we can create a subsequent ticket to do the changes in another issue (we can also mark it as good first issue :D)

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
gmuraru: ping @LaRiffle

syft/frameworks/torch/tensors/interpreters/native.py
abogaziah: I think it needs to be done later, because sharing doesn't work the same way in both cases for now, (AST is more complex) maybe we should do that after refactoring AST

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: I did, I'm looking at how crypten handles it

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: Sounds good! There is also some refacto going on in the AST file in #3924 :)

PR #3860, opened 2020-07-15
test: Add RNN Plan example

juharris:
## Description
I created a language model in a Plan. I was trying to mix the Creating a Plan tutorial and a word based language model from https://github.com/pytorch/examples/tree/master/word_language_model

This was tricky to get working because:
* I wanted to make it work like in the PyTorch Word Language Model example, using 3D tensors (sequence length x batch size x dims) but PySyft's autograd features did not work well with 3D tensors. Many classes assume that tensors are 2D when doing transpose e.g. `MatmulBackward`, `SumBackward`, `TBackward`.
* So I gave up on using 3D tensors and decided to only use the final result from the RNN (and not each output in the sequence). `GRU` from syft/frameworks/torch/nn/rnn.py won't work in a Plan because it uses `torch.sigmoid(z)` and maybe also because of chunking. Instead I made a custom GRU class that does `z.sigmoid()`.
  * I had other issues with PySyft's `rnn.py` in `_init_hidden` (making assumptions on the type of the input) and `_apply_time_step` (uses `stack` which seems like autograd doesn't support).
* Values were getting set to `None` by low-level PySyft operations.  E.g. `param.grad` was `None` after backprop. Sometimes this is expected like when using `nn.Embedding` but usually this was because of a bug somewhere. It was really hard to track these down. It seems like many operations in PySyft default to setting the result to `None` when something unexpected happen. This made something later error.
  * (corrected in another PR) Another example of this was `output += bias` (PyTorch's functional.py) was changing output to `None` when output and bias were `AutogradTensor`s.

## Affected Dependencies
(only required for translating the Plan to Tensorflow.js) An unreleased version of 3p0 with my translations for `select` and `__rsub__` is required to translate the Plan to Tensorflow.js. The example is lenient and will not fail if these translations are missing.

## How has this been tested?
I ran the file I submitted with pytest.

## Checklist


juharris:
@vvmnnnkv I can't tell why those tests fail. Some of those pass on my machine. Others fail but I think it's because I don't have Tensorflow set up.

juharris:
Thanks @iamtrask ! The main issue is that translating the Plan to Tensorflow.js doesn't work yet but I'm making PRs to Threepio to get it working.

juharris:
Tests should pass now. An unreleased version of 3p0 with my translations for `select` and `__rsub__` is required to translate the Plan to Tensorflow.js. The example is lenient and will not fail if these translations are missing.

PR #3876, opened 2020-07-18
[WIP] configure message routing between nodes

Rima:
## Description
This allows configuration systems to update vms/devices/domains and networks through update json files. Also allows each of these entities to fetch the updates of their surroundings from an orchestrator. The orchestrator to a device, is likely to be the "network". Level of visibility of configs is to be determined by some of the following rules:
- A network may "always" know the routes of domains and devices in it with no visibility on VMs.
- A VM may have no clue what networks its devices belong to.
- A device can subscribe to multiple domains.
- A domain may not know the routes to devices subscribed to it, and so communicate only through pub/sub
- A domain may know the routes to some or all devices subscribed to it in the network the domain lives.
- Message forward is possible amongst all entities in a network.
- Only at the network level it is possible to communicate outside the network.

## Affected Dependencies
None

## How has this been tested?
None yet, this is a work in progress, more of the concepts above should be explored.

## Checklist


src/syft/core/nodes/abstract/worker.py
```diff
@@ -86,3 +93,60 @@ def _register_services(self) -> None:
                     self.msg_router[handler_type_subclass] = service_instance
 
         self.services_registered = True
+
+    @type_hints
+    def _fetch_network(self) -> None:
```
iamtrask, line 98: This is the wrong location for this I imagine. Do we also want to allow VMs, Devices, and Domains all to do this? I think you probably want this functionality on the Domain object which is the only thing which is "network aware" quite intentionally. The reason is because the Domain is responsible for making sure that no private information is leaked to the network.

iamtrask: Putting this on the "Worker" class means that all nodes will have it... meaning every VM, and every Device, and every Domain.

iamtrask: Putting this on the "Worker" class means that all nodes will have it... meaning every VM, and every Device, and every Domain.

iamtrask: Even Network nodes themselves would also have it.

Rima: A network for a VM could be as simple as "who is my device?"
A network for a network could be "what's my public route and any routes that I'm allowed to be aware of"

iamtrask: Ah - ok that's cool. Let's not overload the word "network" though. Let's find another word.

src/syft/core/nodes/abstract/worker.py
```diff
@@ -86,3 +93,60 @@ def _register_services(self) -> None:
...
+        # read from json and cache network information.
+        with open('network.json') as f:
+          self.network = json.load(f)
+
+    def _learn_my_specs(self) -> None:
+        with open('specs.json') as f:
+          self.specs = json.load(f)
+
+    @type_hints
+    def sign_message(self, msg: SyftMessage) -> SyftMessage:
```
iamtrask, line 118: So I think that this functionality should be wrapped in a Service instead of put on the worker class. All functionality which processes messages of a certain type on a worker should be packaged into a Service which manages messages of that type.

I also have some services which manage "all messages which are subclasses of X message type"

iamtrask: Functionally equivalent - but in previous versions of Syft the "Worker" class has gotten insanely large because we just add a billion methods for all the kinds of methods it processes. Services are more about code organization than anything.

src/syft/core/nodes/device/network.json
```diff
@@ -0,0 +1,25 @@
+{
```
iamtrask, line 1: Syft shouldn't have user-defined json files. This should be a class with default parameter values and the ability to (optionally) be created from a static file. This will allow PyGrid to create config objects from files, but PySyft will simply run with good defaults.

src/syft/core/message/syft_message.py
```diff
@@ -1,7 +1,69 @@
...
+            defined.
+          - if should_publish_outcomes is True > outcomes routes or broadcast
+            should be defined.
+        """
+        return True
 
 
 class SyftMessage(AbstractMessage):
-    def __init__(self, route, msg_id=None):
+    def __init__(self, route: Route, comms: CommunicationEnvelop, msg_id: UID = None) -> None:
```
iamtrask, line 44: What do you see as the primary difference between Route and CommunicationEnvelop?

iamtrask: From my perspective, Route should contain the information about where a message should be sent (all in one place). It seems CommunicationEnvelop is also trying to do this, yes?

src/syft/core/nodes/device/network.json
Rima: the good default for surrounding workers and specs will probably just be: I don't know anyone else (I'm an isolated worker) and I have no clue what my specs are, but I'll run anyway.

do you see it differently?

iamtrask: I see it the same way. Functionally we're on the same page... literally just the difference between setting defaults in a code object instead of a .json file.

PyGrid can have its own default .json if it wants. I'm just against it in PySyft.

src/syft/core/message/syft_message.py
Rima: The way I think of a route is it's the path to a single worker, might be potentially overloaded with a communication protocol on how to communicate with this worker - but the worker may be able to communicate on multiple protocols, so all in this case would be defined. In the real world, this could be the external and internal IP to the VM processing a message or publishing a message.

A communication envelop, is in nature very specific to a single message floating around between workers, and its function is to tell workers what to do with this message once they're done processing it, such as send me back the outcomes on http or can you forward this message to 10 other workers to calculate the average age of people you have in your dataset, then once done worker X is going to be your aggregator, or we want one of you to volunteer as an aggregator and send me the message back. All that stuff is facilitated by an envelop because an envelop tells this group of workers how to communicate (protocol) and what to communicate (outcomes, this msg, etc.)

src/syft/core/nodes/device/network.json
Rima: makes sense

src/syft/core/io/route.py
```diff
@@ -3,24 +3,54 @@
...
+class BaseRoute(object):
+    connection_details = {}
+
+    def configure_connection(self, protocol: str, host: str, port: int):
+        """
+        the route should have connection details embedded in it.
+        so that nodes operators can utilize it to route messages.
+        """
+        self.connection_details.update({
+            'host': host,
```
iamtrask, line 15: I think about it slightly differently - but not with high confidence. My assumption is that the route should contain "Send me to X node" but that the worker's job is to maintain its own connection and protocol to that node. 

Alternatively, we could consider these details to be optional - things that the node could choose to ignore (or if this protocol wasn't available, it could try to get to the destination through another means).

Or we could consider these to be gold standard and assume that the node would do the heavy lifting to create this kind of connection if it didn't already exist.

All are interesting options.

src/syft/core/io/route.py
```diff
@@ -3,24 +3,54 @@
...
+            'protocol': protocol,
+            'port': port
+        })
+
+    def register_broadcast_channel(self, channel_name: str):
```
iamtrask, line 20: I wonder if from a Route perspective it might be better to create a separate abstraction called a "Group". That way the worker could use channels to talk to the group (according to some naming convention) but if pub/sub wasn't available it could fallback to individually messaging each member of the group.

src/syft/core/io/route.py
```diff
@@ -3,24 +3,54 @@
...
+        """
+        In the case configured protocol is pub/sub or event driven.
+        Args:
+            channel_name: the name of the channel to broadcast on.
+        """
+        self.connection_details.update({'broadcast_channel': name})
+
+    def generate_token(self):
```
iamtrask, line 28: Interesting - can you explain some more detail here?

src/syft/core/message/remote_node_message.py
```diff
@@ -0,0 +1,20 @@
+from syft_message import SyftMessageWithReply
+from ..io.route import Route
+from ...common.id import UID
+
+
+class RegisterNodeMessage(RemoteNodeMessage):
```
iamtrask, line 6: Very nice!

src/syft/core/message/search_message.py
```diff
@@ -0,0 +1,21 @@
+from .syft_message import SyftMessage, SyftMessageWithReply
+from ..io.route import route, Route
+
+class SearchMessage(SyftMessageWithReply):
```
iamtrask, line 4: A nice stub! Lots of functionality to work out here

src/syft/core/message/syft_message.py
```diff
@@ -1,7 +1,17 @@
...
+from ...common.id import UID
+from ...io import Route
 
 class SyftMessage(AbstractMessage):
-    def __init__(self, route, msg_id=None):
+    def __init__(self, route: Route, msg_id: UID = None) -> None:
         self.route = route
         self.msg_id = msg_id
+
+class SignedMessage(SyftMessage):
```
iamtrask, line 10: Very nice concept - seems like the right place for this kind of class to live as well.

src/syft/core/nodes/abstract/remote_nodes.py
```diff
@@ -0,0 +1,94 @@
...
+    Domain = 2
+    Device = 3
+    VM = 4
+
+class Node(object):
+    def __init__(self, type, id = None, name = None, route = None, tags=[]):
+        self.id = id
+        self.name = name
+        self.type = type
+        self.route = route
```
iamtrask, line 17: Oh interesting - I'm starting to see how you're thinking about "Route" here. I think of "Route" as only being something messages have - but if you're also thinking that Nodes would hold a "Route" object to know how to communicate (which totally makes sense) then I see why you would want to have communication information (like host and port) included.

Very interesting.

src/syft/core/nodes/abstract/remote_nodes.py
```diff
@@ -0,0 +1,94 @@
...
+        self.tags = tags
+        self.is_local = is_local
+
+    def as_dict(self, on_key: str = 'id', on_multi_keys: list = []):
```
iamtrask, line 21: Lovely

src/syft/core/nodes/abstract/remote_nodes.py
```diff
@@ -0,0 +1,94 @@
...
+            'type': self.type,
+            'route': self.route,
+            'tags': self.tags}}
+
+    def _flatten_tags_for_search(self):
+        return tags.join('-')
+
+class RemoteNodes(object):
+    """
+    This object keeps a registery of a node's surrounding.
```
iamtrask, line 46: Brilliant documentation here

src/syft/core/io/route.py
Rima: I haven't thought this elaborately, hence it's a place holder at the moment but we can generate a thing like a JWT token for a requester and attach it to the response message. the token would contain things potentially like they're allowed privacy budget, number of requests, duration of access, etc.

Issue #3845, opened 2020-07-10
clone on a remote nn.Module fails


LaRiffle:
## Description
Take a model, .send() it and call clone on it, it will fail.

## How to Reproduce
```
/usr/local/lib/python3.6/dist-packages/syft/frameworks/torch/tensors/interpreters/native.py in clone(self, *args, **kwargs)
    758 
    759         if self.has_child():
--> 760             cloned_tensor.child = self.child.clone(*args, **kwargs)
    761 
    762         return cloned_tensor

TypeError: clone() got an unexpected keyword argument 'memory_format'
```

## How to solve
Add the memory_format kwarg to the clone method of pointer tensor

ibayramli2001:
Same with `copy()` method as it invokes `clone`.

jefersonf:
Hi @LaRiffle, I'm trying to reproduce, but I don't quite understand how.

```
import torch
import torch.nn as nn
import syft as sy

hook = sy.TorchHook(torch)
bob = sy.VirtualWorker(hook, id="bob")
```

I've tested 3 cases.

#### Case 1

```
model = nn.Linear(2, 1)
model_ptr = model.send(bob)
model_copy_ptr = model_ptr.clone()
```

#### Case 2
```
model = nn.Sequential(nn.Linear(2, 2), nn.Linear(2, 1))
model_ptr = model.send(bob)
model_copy_ptr = model_ptr.clone()
```

#### Case 3
```
class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.fc = nn.Linear(2, 1)
    def forward(self, x):
        return nn.functional.relu(self.fc(x))

model = Net()
model_ptr = model.send(bob)
model_copy_ptr = model_ptr.clone()
```

All the cases fail but with a different error like:
```
...
AttributeError: 'Linear' object has no attribute 'clone'
```

How is this clone supposed to be made?

I also ended up noticing that I can do things like these ones, 

```
model_weights_copy = model_ptr.weight.clone()
model_bias_copy = model_ptr.bias.clone()
```


And eventually something like

PR #3904, opened 2020-07-27
fix for copy

omershlo:
## Description
This is a partial solution for #3845. I was able to reproduce the issue for `copy` using: 
```
import torch
import torch.nn as nn
import syft as sy

hook = sy.TorchHook(torch)

model = nn.Linear(2, 1)
model_ptr = model.send(bob)
model_copy_ptr = model_ptr.copy()
```
The suggested fix solves it. 

However, I was not been able to reproduce the error for calling `clone()` on pointer.  It seems like a different bug. 
## Affected Dependencies
For this fix to work `import torch` is required in pointer_tensor.py

## How has this been tested?
There is not much testing required. I checked the above code works and nothing else in the library is broken. 

## Checklist


omershlo:
@LaRiffle please check. 
The test will not work for clone as specified above - there is another bug with cloning a model pointer. I can address it as well but I think it is better to do it as part of another PR.

test/generic/pointers/test_model_pointer.py
```diff
@@ -0,0 +1,23 @@
...
+
+    pred_local = model(data)
+
+    hook = sy.TorchHook(torch)
+    bob = sy.VirtualWorker(hook, id="bob")
+    data_bob = data.send(bob)
+    model_bob = model.send(bob)
+    model_bob_copy = model_bob.copy()
+    pred_remote = model_bob_copy(data_bob)
+    pred_remote = pred_remote.get()
```
LaRiffle, line 21:
```suggestion
    model_bob.get()
    pred_remote = model_bob_copy(data_bob)
    pred_remote = pred_remote.get()
```

LaRiffle:
Ok let's keep clone for a separate PR :)

test/generic/pointers/test_model_pointer.py
omershlo: The suggested change introduces an error. I investigated and this is what I think is happening: 
1) There is no straightforward way to copy a model in torch. 
2) Upon trying to copy a model in soft under the hood you actually call clone. 
3) The difference is that (from the code): Clone should keep ids unchanged, contrary to copy.
4) Therefore we are actually creating two pointers to the same location. Once we "get" the object, it releases the other pointer as well.
5) Tensor on the other hand, do allow copy(). This code will work as expected:   
```
t = torch.tensor([3, 1.0, 2])
p1 = t.send(alice)
p2 = p1.copy()
p1.get()
p2.get()
```
6) In the example from 5 If we replace the `copy()` with `clone()` the same error will occur since p2 will have the same id as p1.

PR #3933, opened 2020-08-03
Make tutorials faster using FSS

LaRiffle:
## Description
This is a small update of the tutorial on Encrypted ML to improve their runtime using Function Secret Sharing

### DIFF
--- a/syft/frameworks/torch/mpc/fss.py
+++ b/syft/frameworks/torch/mpc/fss.py
@@ -41,7 +41,58 @@ COMP = 1
 
 # number of processes
 N_CORES = 8
-MULTI_LIMIT = 10_000
+MULTI_LIMIT = 50_000
+
+
+def keygen(n_values, op):
+    """
+    Run FSS keygen in parallel to accelerate the offline part of the protocol
+
+    Args:
+        n_values (int): number of primitives to generate
+        op (str): eq or comp <=> DPF or DIF
+    """
+    if op == "eq":
+        return DPF.keygen(n_values=n_values)
+    elif op == "comp":
+        if n_values > MULTI_LIMIT:
+            # print('MULTI EVAL', numel, x_masked.owner)
+            multiprocessing_args = []
+            slice_size = math.ceil(n_values / N_CORES)
+            for j in range(N_CORES):
+                n_instances = min((j + 1) * slice_size, n_values) - j * slice_size
+                process_args = (n_instances,)  # TODO add a seed element for the PRG?
+                multiprocessing_args.append(process_args)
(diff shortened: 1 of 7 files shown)


syft/frameworks/torch/mpc/fss.py
```diff
@@ -41,7 +41,58 @@ def full_name(f):
...
+            # print('MULTI EVAL', numel, x_masked.owner)
+            multiprocessing_args = []
+            slice_size = math.ceil(n_values / N_CORES)
+            for j in range(N_CORES):
+                n_instances = min((j + 1) * slice_size, n_values) - j * slice_size
+                process_args = (n_instances,)  # TODO add a seed element for the PRG?
+                multiprocessing_args.append(process_args)
+            p = multiprocessing.Pool()
+            partitions = p.starmap(DIF.keygen, multiprocessing_args)
+            p.close()
```
gmuraru, line 68: Q: It would be better to call ```join```?
Seen [this](https://stackoverflow.com/questions/38271547/when-should-we-call-multiprocessing-pool-join)

syft/frameworks/torch/mpc/fss.py
```diff
@@ -41,7 +41,58 @@ def full_name(f):
...
+
+    Args:
+        n_values (int): number of primitives to generate
+        op (str): eq or comp <=> DPF or DIF
+    """
+    if op == "eq":
+        return DPF.keygen(n_values=n_values)
+    elif op == "comp":
+        if n_values > MULTI_LIMIT:
+            # print('MULTI EVAL', numel, x_masked.owner)
```
gmuraru, line 59: MAYBE: Remove this

syft/frameworks/torch/mpc/fss.py
```diff
@@ -41,7 +41,58 @@ def full_name(f):
...
+    Run FSS keygen in parallel to accelerate the offline part of the protocol
+
+    Args:
+        n_values (int): number of primitives to generate
+        op (str): eq or comp <=> DPF or DIF
+    """
+    if op == "eq":
+        return DPF.keygen(n_values=n_values)
+    elif op == "comp":
+        if n_values > MULTI_LIMIT:
```
gmuraru, line 58: I am thinking that the code from ```op == "com"``` might be included in a separate function, but we can create a refactor ticket for it :D

syft/frameworks/torch/mpc/fss.py
LaRiffle: Interesting!
I should investigate this and observe how performance evolves, will open an issue!

Issue #3826, opened 2020-07-06
Should we support Python <3.8?


iamtrask:
## Question
Given that we want to use typing to enforce code style across the codebase and some of these typing features only exist in python 3.8, should we support any lesser versions?

## Further Information
At present, our tests don't build with Python 3.7 with some sort of error like 

```
=================================== ERRORS ===================================
__________ ERROR collecting tests/syft/typecheck/typecheck_test.py ___________
ImportError while importing test module '/Users/atrask/Dropbox/Laboratory/openmined/PySyft/PySyft/tests/syft/typecheck/typecheck_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/Users/atrask/opt/anaconda3/lib/python3.7/site-packages/_pytest/python.py:493: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
/Users/atrask/opt/anaconda3/lib/python3.7/site-packages/py/_path/local.py:701: in pyimport
    __import__(modname)
/Users/atrask/opt/anaconda3/lib/python3.7/site-packages/_pytest/assertion/rewrite.py:143: in exec_module
    exec(co, module.__dict__)
tests/syft/typecheck/typecheck_test.py:2: in <module>
    from typing import List, Union, Optional, Tuple, Dict, DefaultDict, TypedDict
E   ImportError: cannot import name 'TypedDict' from 'typing' (/Users/atrask/opt/anaconda3/lib/python3.7/typing.py)
```

which is related to https://github.com/OpenMined/PySyft/issues/3722

gmuraru:
As long as we do not have any framework related dependency (Torch, Tensorflow, CrypTen, etc.) that we need to support, I think we should try to support the latest version.

Using the type annotations will help us maintain a cleaner codebase by enforcing more specific restrictions that exist only in Python 3.8

tudorcebere:
If it's not a blocker to other libraries, I am totally into 3.8. We can those features when syft 0.3 is done if we want backwards compatibility.

madhavajay:
Resolved by https://github.com/OpenMined/PySyft/pull/3935

PR #3935, opened 2020-08-04
Support Python <3.8 #3826

madhavajay:
## Description
This PR fixes some minor flake8 issues re-introduced allowing CI to pass, and adds support for both Python 3.7 and Python 3.6 with minimal changes.

The issue was discussed here:
https://github.com/OpenMined/PySyft/issues/3826

Changes required for each version are separated into their own commits.

## Affected Dependencies
None

## How has this been tested?
This has been tested on MacOS with Python 3.8.5, Python 3.7.8 and Python 3.6.11, and additionally GitHub Actions CI.

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.8]
+        python-version: [3.8, 3.7, 3.6]
 
     steps:
     - uses: actions/checkout@v2
@@ -41,9 +41,7 @@ jobs:
 
     - name: Install lint, formatter, security checker
       run: |
-            pip3 install flake8
-            pip3 install black
-            pip3 install bandit
+            pip3 install flake8 black bandit
 
     - name: Lint and format
       run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,4 +54,5 @@ jobs:
 
     - name: Run tests with coverage
       run: |
+            pip install typing-extensions dataclasses
             python setup.py test
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,3 +26,5 @@ pyscaffold[all]
 sqlitedict
 torch>=1.5
 typeguard
+sqlitedict
+typing-extensions # backport to older python 3
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,3 +28,4 @@ torch>=1.5
 typeguard
 sqlitedict
 typing-extensions # backport to older python 3
+dataclasses # backport to python 3.6
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,6 +17,7 @@
 #
 bandit
 black
+dataclasses # backport to python 3.6
 flake8
 flask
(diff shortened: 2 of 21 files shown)


madhavajay:
I am a little confused why I needed to add this:
```
- name: Run tests with coverage
      run: |
+            pip install typing-extensions dataclasses
            python setup.py test

```

I would have thought that listing these packages in both the testing and install_requires inside setup.cfg would be sufficient?
On my local machine these needed to be installed with pip before the tests would work.

Here it kind of suggests that you have to still install them manually?
https://pyscaffold.org/en/latest/dependencies.html

*If run py.test runner, you will have to install those dependencies manually, or do a editable install of your package with pip install -e .[testing].*

Would be great if someone can clarify whats going on here.

gmuraru:
Why do we need those changes in the new PySyft Version?

madhavajay:
The reasoning for this change is that operating systems like Ubuntu LTS come with quite old python versions, so supporting 3.6 means Ubuntu 18.04+ should be covered out of the box.

I looked at going to python 3.5, but as Black is not supported (although its not actually needed for the library) and f"" strings are not supported either this seems like too much of a disruption at this stage, however its possible this may work without significant changes and provide support for Ubuntu 16.04 LTS and many other platforms stuck on python 3.5.

To achieve this the following changes were made:

## Python 3.7
- Imports referring to final and Final were changed to:
```
from typing_extensions import final, Final
```
See more here:
https://github.com/python/typing/tree/master/typing_extensions

## Python 3.6
- The dataclasses backport library was added without any changes to the code.
- `from __future__ import annotations` were removed:

From what I can see they are not being used as removal has not caused any issues.
The "annotations" future is only to delay parsing of types until after load, to prevent issues with referencing custom types in their own definitions. The workaround is to simply define them as strings "CustomType" when referring with in a type definition, however if this becomes cumbersome to support, python 3.6 could be dropped.

See more here:
https://www.python.org/dev/peps/pep-0563/#dropping-annotations-with-o

I would say both of these changes are extremely minor at this stage, but we could choose to only support Python 3.7.

I am happy to incorporate any feedback in this PR, or close without merging if these changes are unacceptable.

madhavajay:
> Why do we need those changes in the new PySyft Version?

It was an open 0.3.0 issue:
https://github.com/OpenMined/PySyft/issues/3826

It seems that perhaps support for 3.6 and 3.7 isn't actually incompatible with the usage of new Python typing constructs.

iamtrask:
> I am a little confused why I needed to add this:
> 
> ```
> - name: Run tests with coverage
>       run: |
> +            pip install typing-extensions dataclasses
>             python setup.py test
> ```
> 
> I would have thought that listing these packages in both the testing and install_requires inside setup.cfg would be sufficient?
> On my local machine these needed to be installed with pip before the tests would work.
> 
> Here it kind of suggests that you have to still install them manually?
> https://pyscaffold.org/en/latest/dependencies.html
> 
> _If run py.test runner, you will have to install those dependencies manually, or do a editable install of your package with pip install -e .[testing]._
> 
> Would be great if someone can clarify whats going on here.

I can't explain why it works this way but I have come to this understanding as well.

PR #3936, opened 2020-08-04
Updated pre-commit-config

madhavajay:
## Description
As per conversation in Slack, the pre commit hooks will prevent tests / checks in CI being easily broken.
I have tested this locally and it is all working currently as expected.

We should document the setup and encourage all developers to adopt the use of these hooks.
See: https://pre-commit.com/

I have updated my additional scripts/pre_commit.sh to use this.
There are some additional tasks that get executed in "scripts/pre_commit.sh" which we probably don't want in the actual commit hook, including running the bandit checks and the tests. If the tests are part of the commit hook one would never be able to submit a commit to fix a broken test without it being fixed first.

However, I think being able to run this command before committing manually is a good way to be sure everything should pass in CI (except for python and os version differences).

## Affected Dependencies
None

## How has this been tested?
I have tested it locally and it will be run in CI.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v1.3.0
+    rev: v3.2.0
     hooks:
     -   id: trailing-whitespace
     -   id: check-docstring-first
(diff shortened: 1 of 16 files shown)


madhavajay:
@cereallarceny Instructions have been added to the CONTRIBUTING.md file.

PR #3946, opened 2020-08-05
Duet Init

iamtrask:
## Description
This is an initial PR for the Duet demo - wherein one person can run a grid server from a notebook and another person (in another notebook somewhere else on planet Earth) can connect to it.

This initial PR is *very* rough. It's using HTTP instead of WebRTC and it's not yet using Syft's fancy new serialization.


iamtrask:
I can't merge in code with pickle yet because it causes security checks to fail - i think this is probably for the best for now.

Issue #3722, opened 2020-06-18
Automatic type-checking of Python type hints


iamtrask:
## Feature Description
In a previous prototype for a PySyft rebuild found [here](https://github.com/OpenMined/PySyft/blob/syft_4v2/syft/check.py), I wrote a new decorator which enforces type checking of python type hints when unit testing. To close this issue:

1) Move the functionality into syft_0.3.0 branch.
2) Add type hints to all methods in the codebase.
3) Create a "CONTRIBUTING.md" file and add hints as a requirement.

tudorcebere:
I'll look into this one.

iamtrask:
All yours!

iamtrask:
Things to watch out for:
- I never tested whether it affected runtime performance (when it should be turned off). I hope it shouldn't but it might.
- I didn't check whether it impacted testing performance.

iamtrask:
Fixed by https://github.com/OpenMined/PySyft/pull/3804

PR #3952, opened 2020-08-06
Mypy 2

madhavajay:
## Description
Fixed all mypy errors with updated settings.

This PR targets this ticket: https://github.com/OpenMined/PySyft/issues/3722

- Fixed syft_decorator issue with Forward References
- Compatible with Python 3.6, 3.7 and 3.8
- Enabled mypy in .pre-commit-config.yaml
- Turned on mypy settings:
  --disallow-incomplete-defs
  --no-implicit-optional
  --warn-unused-ignores
  --warn-redundant-casts
  --strict-equality

## Affected Dependencies
None

## How has this been tested?
Locally MacOS 3.6, 3.7, 3.8 and CI

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,10 +23,24 @@ repos:
     rev: v1.5.1
     hooks:
     -   id: python-use-type-annotations
-# -   repo: https://github.com/pre-commit/mirrors-mypy
-#     rev: v0.782
-#     hooks:
-#     -   id: mypy
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.782
+    hooks:
+    -   id: mypy
+        exclude: src/syft/proto
+        args: [
+            "--ignore-missing-imports",
+            "--scripts-are-modules",
+            "--disallow-incomplete-defs",
+            "--no-implicit-optional",
+            "--warn-unused-ignores",
+            "--warn-redundant-casts",
+            "--strict-equality",
+            # "--warn-unreachable",
+            # "--disallow-untyped-decorators",
+            # "--disallow-untyped-calls",
+            # "--disallow-untyped-defs"
+        ]
 -   repo: local
     hooks:
     -   id: flynt
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,10 +12,11 @@ import inspect
 import os
 import shutil
 import sys
+from pathlib import Path
+from typing import Dict
 
-__location__ = os.path.join(
-    os.getcwd(), os.path.dirname(inspect.getfile(inspect.currentframe()))
-)
+dir_name = inspect.getfile(inspect.currentframe())  # type: ignore
+__location__ = os.path.join(os.getcwd(), os.path.dirname(dir_name))
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
@@ -137,7 +138,7 @@ copyright = "2020, Andrew Trask"
 # built documents.
 #
 # The short X.Y version.
-version = ""  # Is set by calling `setup.py docs`
+default_version = ""  # Is set by calling `setup.py docs`
 # The full version, including alpha/beta/rc tags.
 release = ""  # Is set by calling `setup.py docs`
 
@@ -203,7 +204,7 @@ try:
 except ImportError:
     pass
 else:
-    release = version
+    release = default_version
 
 # A shorter title for the navigation bar.  Default is the same as html_title.
 # html_short_title = None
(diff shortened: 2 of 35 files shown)


.pre-commit-config.yaml
```diff
@@ -23,10 +23,24 @@ repos:
...
+        exclude: src/syft/proto
+        args: [
+            "--ignore-missing-imports",
+            "--scripts-are-modules",
+            "--disallow-incomplete-defs",
+            "--no-implicit-optional",
+            "--warn-unused-ignores",
+            "--warn-redundant-casts",
+            "--strict-equality",
+            # "--warn-unreachable",
```
gmuraru, line 39: Q: we don't want those?

docs/conf.py
```diff
@@ -12,10 +12,11 @@
...
 import shutil
 import sys
+from pathlib import Path
+from typing import Dict
 
-__location__ = os.path.join(
-    os.getcwd(), os.path.dirname(inspect.getfile(inspect.currentframe()))
-)
+dir_name = inspect.getfile(inspect.currentframe())  # type: ignore
+__location__ = os.path.join(os.getcwd(), os.path.dirname(dir_name))
```
gmuraru, line 19: MAYBE: Use ```Path``` than ```join```?

src/syft/core/common/serde/deserialize.py
```diff
@@ -58,8 +58,9 @@ def _deserialize(
                 "provide a schematic to be filled with the hex data provided."
             )
         schematic = schema_type()
-        schematic.ParseFromString(bytes.fromhex(blob))
-        blob = schematic
+        if type(blob) == str:
```
gmuraru, line 61: NITPICK:
```if type(blob) is str```

File: src/syft/core/common/serde/deserialize.py [hunk 4]
```diff
@@ -71,16 +72,20 @@ def _deserialize(
...
-            raise ValueError(
-                "Schema type shouldn't be None when deserializing from json. Please"
-                "provide a schematic to be filled with the json data provided."
+        # QUESTION: Should this be here?
+        # if schema_type is None:
+        #     raise ValueError(
+        #         "Schema type shouldn't be None when deserializing from json. Please"
+        #         "provide a schematic to be filled with the json data provided."
+        #     )
+        if type(blob) == str:
```
gmuraru, line 81: Also here

src/syft/core/common/serde/serializable.py
```diff
@@ -351,32 +359,34 @@ def check_type(s) -> SerdeTypes:
...
+
+            # check if the serde_type is a native type and add it if yes
+            if serde_type is SerdeTypes.SerdeNativeType:
+                native_types.add(s)
+
+            # check if the serde_type is a wrapper type and add it if yes
+            if serde_type is SerdeTypes.SerdeWrapperType:
+                wrapper_types.add(s)
+
+            # even if the current class is not serializable (might be and abstract
```
gmuraru, line 381: TYPO: ```and -> an```

src/syft/core/io/connection.py
```diff
@@ -9,7 +7,8 @@
 from ...decorators import syft_decorator
 
 
-@final
```
gmuraru, line 12: I think it is ok to remove the ```final```

src/syft/core/io/connection.py
```diff
@@ -29,7 +28,7 @@ def recv_eventual_msg_without_reply(
 
 
 # QUESTION: why is this final when VirtualClientConnection inherits from it?
-@final
```
gmuraru, line 32: Also here

.pre-commit-config.yaml
madhavajay: Are you asking why there are commended out ones?
The ones that are commented out are strict enough to cause a lot more issues which I can fix but its better to do these things in batches.

I left them there so they are easy to toggle during development, and as a reminder that we should probably finish them as well.

I only saw 1 error for --warn-unreachable but it could be that the code is just incomplete.
The untyped decorators / calls will mostly be covered by --disallow-untyped-defs but that means a lot of:
```
- def function(self):
+ def function(self) -> None:
```

I was able to get all the other flags complete in time for this PR to be merged at the start of the work day for everyone on EU time zones which is preferable than having a big noisy PR that stays open across other merges.

docs/conf.py
madhavajay: This is a weird one where the code works but the types don't match and I didn't want to disturb it but just make sure that the mypy checks were safely ignored. I can only assume that this is used to get access to the file that imports this file during run time so i figured without a good way to test it, I would leave it alone. Happy to upgrade to pathlib across the board, although I wonder why this isn't already a warning in mypy.

src/syft/core/common/serde/deserialize.py
madhavajay: Yeah i wondered about that, again it wasn't a warning so the question is.... is the type actually the same pointer in memory or is it just equal to the type. Are there going to be some variation in str types at some point which are == but are not "is / ===", not sure but is is probably more explicit, happy to change this as well.

madhavajay: Interestingly, I have the "--strict-equality" flag already on, and mypy doesn't consider these to be incorrect.
I also don't think that the "type" can be coalesced into anything else, unlike say: 1.0 == 1

Issue #3958, opened 2020-08-06
Fixed Precision Tensor: modular operation is broken


abogaziah:
## Description
modular operation (using the % operator) produces a wrong result 
## How to Reproduce
x= torch.tensor([1,2,3]).fix_prec()
y=(x%3)
y.float_prec()= tensor([0.0010, 0.0020, 0.0000])

## Expected Behavior
x= torch.tensor([1,2,3]).fix_prec()
y=(x%3)
y.float_prec()= tensor([1.0, 2.0, 0.0])

PR #3960, opened 2020-08-08
Overload mod operator for the fixed precision tensor

arturomf94:
As pointed out in https://github.com/OpenMined/PySyft/issues/3958 modular operations over fixed precision tensors are broken. This PR overloads the mod operation over fixed precision tensors in [`syft/frameworks/torch/tensors/interpreters/precision.py`](https://github.com/arturomf94/PySyft/blob/d75a2ba0d4542faf1e5d019c00a9d77868f4d723/syft/frameworks/torch/tensors/interpreters/precision.py#L76) and adds a corresponding test in [`test/torch/tensors/test_precision.py`](https://github.com/arturomf94/PySyft/blob/d75a2ba0d4542faf1e5d019c00a9d77868f4d723/test/torch/tensors/test_precision.py#L40)

Closes: https://github.com/OpenMined/PySyft/issues/3958

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/precision.py
+++ b/syft/frameworks/torch/tensors/interpreters/precision.py
@@ -78,12 +78,12 @@ class FixedPrecisionTensor(AbstractTensor):
         """
         Define the modulo operation over object instances.
         """
-        result = self.copy()
+        result = FixedPrecisionTensor()
         if isinstance(divisor, int):
             scaled_divisor = divisor * (self.base ** self.precision_fractional)
-            result.child %= scaled_divisor
+            result.child = self.child % scaled_divisor
         elif isinstance(divisor, FixedPrecisionTensor):
-            result.child %= divisor.child
+            result.child = self.child % divisor.child
         else:
             raise TypeError("Unsupported type for modulo operation")
         return result
--- a/syft/frameworks/torch/tensors/interpreters/precision.py
+++ b/syft/frameworks/torch/tensors/interpreters/precision.py
@@ -9,7 +9,6 @@ from syft.generic.frameworks.overload import overloaded
 from syft.generic.pointers.multi_pointer import MultiPointerTensor
 from syft.generic.abstract.tensor import AbstractTensor
 from syft.workers.abstract import AbstractWorker
-from typing import Union
 
 from syft_proto.frameworks.torch.tensors.interpreters.v1.precision_pb2 import (
     FixedPrecisionTensor as FixedPrecisionTensorPB,
@@ -73,20 +72,6 @@ class FixedPrecisionTensor(AbstractTensor):
                 raise ValueError(
                     "Unsupported arg value for dtype. Use dtype='long' or dtype='int'."
                 )
-
-    def __mod__(self, divisor: Union[int, torch.Tensor]) -> torch.Tensor:
-        """
-        Define the modulo operation over object instances.
-        """
-        result = FixedPrecisionTensor()
-        if isinstance(divisor, int):
-            scaled_divisor = divisor * (self.base ** self.precision_fractional)
-            result.child = self.child % scaled_divisor
-        elif isinstance(divisor, FixedPrecisionTensor):
-            result.child = self.child % divisor.child
-        else:
-            raise TypeError("Unsupported type for modulo operation")
-        return result
 
     def get_class_attributes(self):
         """
@@ -173,6 +158,21 @@ class FixedPrecisionTensor(AbstractTensor):
             pos_nums = self.child / truncation
             self.child = neg_nums * gate + pos_nums * (1 - gate)
             return self
+
+    @overloaded.method
+    def mod(self, _self, divisor):
+        """
+        Define the modulo operation over object instances.
+        """
+        if isinstance(divisor, (int, float)):
+            scaled_divisor = int(divisor * self.base ** self.precision_fractional)
+            return getattr(_self, "fmod")(scaled_divisor)
+
+        response = getattr(_self, "fmod")(divisor)
+
+        return response
+
+    __mod__ = mod
 
     @overloaded.method
     def add(self, _self, other):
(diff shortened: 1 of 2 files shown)


syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -73,6 +73,15 @@ def __init__(
                     "Unsupported arg value for dtype. Use dtype='long' or dtype='int'."
                 )
 
+    def __mod__(self, other):
```
abogaziah, line 76: would you just change "other" to "devisor" for readability

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -73,6 +73,15 @@ def __init__(
...
+        """
+        Define the modulo operation over object instances.
+        """
+        result = self.copy()
```
abogaziah, line 80: is there a reason why you decided to copy the object and not modify the self.child directly and return self?

test/torch/tensors/test_precision.py
```diff
@@ -37,6 +37,13 @@ def test_fix_prec_registration(hook):
         assert hook.local_worker.get_obj(x.id) == x
 
 
+def test_fixed_precision_mod_operation():
```
abogaziah, line 40: I expect that this modification will break the same functionality if the devisor was another FPT or torch.Tensor
please test the case of 
x = torch.tensor([1, 2, 3]).fix_prec()
y =  torch.tensor([3]).fix_prec()
z= (x%y).float_prec()
assert (z == torch.tensor([1.0, 2.0, 0.0])).all()

arturomf94: That's a good shout. Will add that test too.

syft/frameworks/torch/tensors/interpreters/precision.py
arturomf94: For inmutability of `x` in `x % y`. Would you say it's best to return `x`?

test/torch/tensors/test_precision.py
arturomf94: I've added a test for this now.

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -73,6 +74,20 @@ def __init__(
...
 
+    def __mod__(self, divisor: Union[int, torch.Tensor]) -> torch.Tensor:
+        """
+        Define the modulo operation over object instances.
+        """
+        result = self.copy()
+        if isinstance(divisor, int):
+            scaled_divisor = divisor * (self.base ** self.precision_fractional)
+            result.child = result.child % scaled_divisor
+        elif isinstance(divisor, FixedPrecisionTensor):
```
abogaziah, line 85: calling a function on a tensor is supposed to modify the child without explicitly mentioning it, for instance, "result %= divisor" should do the same, and it's more readable.

syft/frameworks/torch/tensors/interpreters/precision.py
abogaziah: no, it would be better to init a new instance :)

syft/frameworks/torch/tensors/interpreters/precision.py
arturomf94: Agree. I'll change that.

arturomf94: Actually, `result %= divisor` ends up being an infinite recursion. So I left it as it was: working with the `.child`. If there's a better way to avoid the recursion without referencing the child please let me know.

abogaziah: hmm.. makes sense, let's keep it that way

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -73,6 +74,20 @@ def __init__(
                     "Unsupported arg value for dtype. Use dtype='long' or dtype='int'."
                 )
 
+    def __mod__(self, divisor: Union[int, torch.Tensor]) -> torch.Tensor:
```
gmuraru, line 77: Q: Can we do it with how we have the ```add```, ```sub``` operator?
Use @overloaded.method

test/torch/tensors/test_precision.py
```diff
@@ -37,6 +37,20 @@ def test_fix_prec_registration(hook):
...
+    # Test mod operation with scalar
+    x = torch.tensor([1, 2, 3]).fix_prec()
+    y = x % 3
+    y = y.float_prec()
+    assert (y == torch.tensor([1.0, 2.0, 0.0])).all()
```
gmuraru, line 45: Q: Could you also add tests if the operators are ```AST``` and ```FPT```?

arturomf94: I tried this, but the tests fail and I don't know how to do a mod operation where AST is the dividend and FTP the divisor... Does this work when both the dividend and the divisor are ASTs?

arturomf94: For instance, consider the following: 

```python
x = torch.tensor([1, 2, 3]).share(bob, alice, crypto_provider=james)
(x % 3).get() 
```

This turns out to be `tensor([1, 2, 3])`, which is not right 😕 ❗ ❓

In fact, it's inconsistent: sometimes I get something like `tensor([4, 2, 3])`. What does this mean? 😱

arturomf94: Lol - never mind. I forgot to use fixed precision I think...

arturomf94: I saw the tests [here](https://github.com/OpenMined/PySyft/blob/aafa70d096acf22d7754da8b7b0037f74d12313a/test/torch/tensors/test_additive_shared.py#L1045) for the mod operation on ASTs and tried to replicate it somehow. Let me know what you think.

PR #3964, opened 2020-08-08
modification to just print the loss tensor

himalayanZephyr:
added the .data attribute to just print the loss tensor.

Tested by running the code on Jupyter notebook.


gmuraru:
LGTM!

PR #3983, opened 2020-08-11
Fix More mypy

madhavajay:
## Description
Fix More mypy and added mypy flag:
    --warn-unreachable

## Affected Dependencies
None

## How has this been tested?
CI

## Checklist

### DIFF
--- a/src/syft/core/store/store_interface.py
+++ b/src/syft/core/store/store_interface.py
@@ -1,9 +1,10 @@
 from abc import ABC
-from typing import Iterable
+from typing import Iterable, List, Type
 
 from ...decorators import syft_decorator
 from ..common.uid import UID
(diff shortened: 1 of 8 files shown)


src/syft/core/store/store_interface.py
```diff
@@ -136,3 +137,23 @@ def clear(self) -> None:
         Clears all storage owned by the store.
         """
         raise NotImplementedError
+
```
tudorcebere, line 140: We should drop these functions, as we were following a dict-like syntax.

madhavajay: Agreed. These are currently here to satisfy the mypy linter. So we can remove them as soon as we are no longer calling them from code.

PR #3996, opened 2020-08-12
Make Sandbox Tutorial more Intuitive

himalayanZephyr:
- Added a code cell to display the populated datasets on a worker
- Changed the initial search from boston housing tags to custom tags
- Added the boston housing dataset search at a later stage

Tested the code by running the entire code as a separate notebook.


Syzygianinfern0:
LGTM!! :tada:

Issue #3901, opened 2020-07-26
 Replicated Sharing Tensor (RST): Matrix multiplicatiosn and Convolutions


gmuraru:
## Feature Description
This builds upon the work done for [this task](https://github.com/OpenMined/PySyft/issues/3900).
Implement *Matrix multiplication* and *Convolutions* (for this there might also be 2 PRs). The multiplication is described in the *Basic Operations* sections from the [FALCON paper](https://arxiv.org/pdf/2004.02229.pdf).

Add tests for *matrix multiplications* and *convolutions*.

Issue #3900, opened 2020-07-26
Replicated Sharing Tensor (RST): Multiplication of shared values


gmuraru:
## Feature Description
Implement the multiplication operation of the Replicated sharing tensor (multiplication of shared values).
To multiply two shared values together x = (x1, x2, x3) and y = (y1, y2, y3), parties locally compute z1 = (x1y1+x2y1+x1y2), z2 = (x2y2+x3y2+x2y3) and z3 = (x3y3 +x1y3 +x3y1). At the end of this, z1,z2 and z3 form a 3-out-of-3 secret sharing of z=x.y . Parties then perform resharing where 3-out-of-3 randomness is used to generate 2-out-of-3 sharing by sending αi +zi to party i−1.

**Notice**: this issue is dependant on the [Correlated Randomness](https://github.com/OpenMined/PySyft/issues/3898)  issue, make sure it's solved first.

RSS operations are explained in [FALCON paper](https://arxiv.org/pdf/2004.02229.pdf). you can use it or any other reference.

Add tests for checking if the behavior is the one expected

PR #4015, opened 2020-08-14
Replicated Sharing Tensor: Multiplication, Matrix multiplication, and Convolution

abogaziah:
## Description
The implementation of the rest of the operations of replicated shared tensor that's necessary to implement falcon protocol.
1- Multiplication: to multiplay a secret x = [x1,x2,x3] by a secret y = [y1,y2,y3] : 

- Parties locally compute z1 = x1y1+x2y1+x1y2, z2 = x2y2+x3y2+x2y3 and z3 = x3y3 +x1y3 +x3y1.
At the end of this, z1,z2, and z3 form a 3-out-of-3 secret sharing of z

- Each party adds 3of3 correlated random number (alpha i) to zi where i is the index of the worker.

- Parties perform resharing by sending a copy of zi to party i+1

2- Matrix multiplication: multiplication algorithm works with matrix multiplication, just by replacing multiplication operator with matrix multiplication

3-  Convolution: convolution is done by converting to one matrix multiplication as explained [here](https://cs231n.github.io/convolutional-networks/)  using [torch unfold](https://discuss.pytorch.org/t/custom-convolution-dot-product/14992/27) 

4- Applied black box principle, i.e. the class should take in a tensor and a set of workers and output a tensor, all other details should be abstracted inside the class, I did that by removing shares_map from the parameters and change other detail around it  

## Checklist
______________________________________________________
- [x] Multiplication #3900 
- [x] Matrix multiplication #3901 
- [x] convolution #3901


syft/frameworks/torch/mpc/przs.py
```diff
@@ -5,7 +5,7 @@
 import torch
 import syft
 
-RING_SIZE = 2 ** 64
+RING_SIZE = 2 ** 5
```
gmuraru, line 8: Q: should this be 5?

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -122,12 +112,120 @@ def private_linear_operation(self, secret, operator):
...
+    def private_unfold(secret, kernel_size, padding):
+        secret = secret.apply_to_shares(torch.Tensor.double)
+        secret = secret.apply_to_shares(
+            torch.nn.functional.unfold, kernel_size=kernel_size, padding=padding
+        )
+        secret = secret.apply_to_shares(torch.Tensor.long)
+        return secret
+
+    @staticmethod
+    def __switch_public_private(value, public_function, private_function, *args, **kwargs):
```
gmuraru, line 219: I really like this function :D

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -11,29 +11,32 @@ def __init__(
...
         return workers
 
     def generate_shares(self, plain_text, number_of_shares=3):
         shares = []
+        plain_text.long()
         for _ in range(number_of_shares - 1):
-            shares.append(torch.tensor(random.randrange(self.ring_size)))
-        shares.append(torch.tensor((plain_text - sum(shares)) % self.ring_size))
+            shares.append(torch.randint(high=self.ring_size // 2, size=plain_text.shape))
+        shares.append((plain_text - sum(shares) % self.ring_size))
```
gmuraru, line 39: Q: ```)``` should be after ```(plain_text - sum(shares))```

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -146,6 +244,37 @@ def get_shares_map(self):
...
+        shares_map = {
+            player: (
+                function(shares_map[player][0], *args, **kwargs),
+                function(shares_map[player][1], *args, **kwargs),
+            )
+            for player in players
+        }
+        return ReplicatedSharingTensor(shares_map)
+
+    def apply_to_shares_(self, function, *args, **kwargs):
```
gmuraru, line 263: Q: Is this used?

syft/frameworks/torch/mpc/przs.py
abogaziah: it should be less than or equal the ring size of RST

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: yeah me too :D it saved a lot of code repetition

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: yes it should

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: No, I thought I will need it for convolution, but I didn't have to do any in-place operations, should I remove it? 
I think I should remove it, but I'm like 99% sure I'll need it in the future and I'll regret removing it 😂

abogaziah: removed it

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -1,39 +1,54 @@
...
+    def share_secret(self, plain_text, workers):
         number_of_shares = len(workers)
         workers = self.__arrange_workers(list(workers))
-        shares = self.generate_shares(secret, number_of_shares)
+        shares = self.generate_shares(plain_text, number_of_shares)
         shares_map = self.__distribute_shares(workers, shares)
-        self.child = shares_map
-        return self
+        PRZS.setup(workers)
+        return shares_map
```
gmuraru, line 33: Q: Shouldn't we return ```self``` here?

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -1,39 +1,54 @@
...
 
 
 class ReplicatedSharingTensor(AbstractTensor):
     def __init__(
-        self, shares_map=None, owner=None, id=None, tags=None, description=None,
+        self, plain_text=None, players=None, owner=None, id=None, tags=None, description=None,
     ):
         super().__init__(id=id, owner=owner, tags=tags, description=description)
+        self.ring_size = 2 ** 32
+        shares_map = self.__generate_shares_map(plain_text, players)
```
gmuraru, line 14: Q: Shouldn't we do this only if we have ```plain_text != None```?
(maybe we can have two ways in which we construct RSS, one when we send the plaintext to the constructor and one using the ```share``` method)

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -1,39 +1,54 @@
...
         self.child = shares_map
-        self.ring_size = 2 ** 5
 
-    def share_secret(self, secret, workers):
+    def __generate_shares_map(self, plain_text, players):
+        if plain_text is None or players is None:
```
gmuraru, line 18: Q: Shouldn't we better raise an exception here? (maybe something like ```BadValue```)

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -1,39 +1,54 @@
...
+            raise ValueError("you must provide 3 players")
+        me = syft.hook.local_worker
+        if me in workers:
+            workers.remove(me)
+            workers = [me] + workers
         return workers
 
     def generate_shares(self, plain_text, number_of_shares=3):
         shares = []
+        plain_text.long()
```
gmuraru, line 48: Q: In case plain_text is a ```float``` this will simply do a "cast" to an ```int``` type, right?

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -79,63 +93,152 @@ def __map_modular_to_real(self, mod_number):
...
+    def __reshare(shares, workers):
+        """convert 3-out-of-3 secret sharing: {player i : share i}
+          to 2-out-of-3 sharing: {player i : (share i, share i+1)}  """
+        shares_map = {}
+        for i in range(len(shares)):
+            pointer = shares[(i + 1) % len(shares)].copy().move(workers[i])
+            shares_map[workers[i]] = (shares[i], pointer)
+        return shares_map
+
+    def conv2d(self, image, padding=0):
```
gmuraru, line 195: Q: Should we have this in the ```RSS``` or in the ```Falcon``` class?

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -46,21 +61,20 @@ def __distribute_shares(workers, shares):
...
-        shares_map = self.get_shares_map()
-        shares = self.__retrieve_shares(shares_map)
+        shares = self.retrieve_shares()
         plain_text_mod = self.__sum_shares(shares)
         plain_text = self.__map_modular_to_real(plain_text_mod)
         return plain_text
 
-    def __retrieve_shares(self, shares_map):
-        pointers = self.__retrieve_pointers(shares_map)
+    def retrieve_shares(self):
```
gmuraru, line 69: Q: Can we make this ```__retrieve_shares```?

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -46,21 +61,20 @@ def __distribute_shares(workers, shares):
...
+        pointers = self.retrieve_pointers()
         shares = []
         for pointer in pointers:
             shares.append(pointer.get())
         return shares
 
-    @staticmethod
-    def __retrieve_pointers(shares_map):
+    def retrieve_pointers(self):
```
gmuraru, line 76: Q: Also this one, ``` __retrieve_pointers?```

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -146,6 +249,30 @@ def get_shares_map(self):
         """
         return self.child
 
+    def __set_shares_map(self, shares_map):
+        self.child = shares_map
+        return self
+
+    def apply_to_shares(self, function, *args, **kwargs):
```
gmuraru, line 256: Q: This is used only internally. Should it be ```public```?

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: No, not anymore, this function was used to kinda init the object but this doesn't happen anymore

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: this is checked in the __generate_shares_map function. if plain_text's None it returns None

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: sometimes you need to make an empty tensor like an empty torch tensor, it should be acceptable I think

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: it will cast it from any torch.dtype to torch.long

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: hmm, it could be implemented both ways, but the paper described it as an operation of the sharing scheme, so let's stick to the paper for now, it can be changed easily later if it is inconvenient

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: I meant to make them public since they are very useful for debugging to see the shares before adding them, it's basically the reconstruction algorithm but you'll have to add the shares manually

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: any reshaping operation like view, unfold, reshape (or any other non-arithmetic operation) will use this function, so it will be used externally

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
gmuraru: Hmm...seems ok - when refactoring we should stick with the same name method (we have ```share``` for ```AST``` and we do the sharing in the constructor for the ```RST```)

gmuraru (approved):

gmuraru:
LGTM! After you merge it, could you give me a ping to refactor [this](https://github.com/OpenMined/PySyft/pull/4474)

Issue #3993, opened 2020-08-12
MPC `reciprocal` using Newthon Raphon Method or Log


marload:
## Feature Description

Reciprocal, currently implemented in FixedPrecisionTensor, is implemented using division. However, the division of the AdaptiveSharingTensor is so slow that it takes too long. If you use NR or Log to implement Reciprocal, we can calculate this much faster. However, if we use NR or Log, it will return an approximate value.

Additionally, if the Reciprocal is faster, we will be able to make the AdditionalSharingTensor faster by referring to the implementation of Crypten.

## Is your feature request related to a problem?

`reciprocal` of FixedPrecisionTensor is very slow.

## What alternatives have you considered?

Using Newton Method

## Additional Context

If the core developers of the Crypto Team think this is worthwhile, I want to implement it. :smile:

PR #4021, opened 2020-08-15
Fast FixedPrecisonTensor Reciprocal Method (150x faster)

marload:
## Description

issue: https://github.com/OpenMined/PySyft/issues/3993

We can use reciprocal 150 times faster on average. Implementing division of AST with reciprocal can make existing very slow division much faster. In addition, many slow operations can be made very quickly.

This implementation referred to the implementation of CrypTen.

https://github.com/facebookresearch/CrypTen/blob/3c4e53aa4910117e4e136691cb09ea7a7e001029/crypten/mpc/mpc.py#L1029

```python
x = torch.tensor([1., 2., 3.]).fix_prec().share(bob, alice, crypto_provider=theo)

# Existing method (using division)
x.reciprocal().get().float_prec() # 27 seconds
# tensor([1.0000, 0.5000, 0.3330])

# New Method (using NR)
x.reciprocal(method="nr").get().float_prec() # 0.18 seconds
# tensor([1.0000, 0.5000, 0.3330])

# New Method (using LOG) / Approximation
x.reciprocal(method="log").get().float_prec() # 0.31 seconds
#  tensor([1.0000, 0.4650, 0.3530])
```

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/test/torch/tensors/test_precision.py
+++ b/test/torch/tensors/test_precision.py
@@ -108,7 +108,6 @@ def test_reciprocal(workers):
 
     tensor = torch.tensor([1.0, 2.0, 3.0])
     x = tensor.fix_prec()
-    real_result = torch.tensor([1.0, 2.0, 3.0]).reciprocal()
 
     result = x.reciprocal().float_prec()
     assert torch.isclose(result, tensor.reciprocal()).all()
--- a/test/torch/tensors/test_precision.py
+++ b/test/torch/tensors/test_precision.py
@@ -106,17 +106,17 @@ def test_methods_for_linear_module(method, parameter):
 def test_reciprocal(workers):
     bob, alice, james = (workers["bob"], workers["alice"], workers["james"])
 
-    x = torch.tensor([1.0, 2.0, 3.0]).fix_prec()
+    tensor = torch.tensor([1.0, 2.0, 3.0])
+    x = tensor.fix_prec()
     real_result = torch.tensor([1.0, 2.0, 3.0]).reciprocal()
 
     result = x.reciprocal().float_prec()
-    assert (result == torch.tensor([1.0, 0.5, 0.333])).all()
-    assert ((real_result - result).abs() <= 0.001).all()
+    assert torch.isclose(result, tensor.reciprocal()).all()
 
     x = torch.tensor([1.0, 2.0, 3.0]).fix_prec()
     result = x.share(bob, alice, crypto_provider=james)
     result = result.reciprocal(method="NR").get().float_prec()
-    assert ((real_result - result).abs() <= 0.001).all()
+    assert torch.isclose(result, tensor.reciprocal()).all()
 
 
 def test_torch_add(workers):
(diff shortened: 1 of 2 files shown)


test/torch/tensors/test_precision.py
```diff
@@ -103,6 +103,22 @@ def test_methods_for_linear_module(method, parameter):
     assert (result == fp_result.float_precision()).all()
 
 
+def test_reciprocal(workers):
+    bob, alice, james = (workers["bob"], workers["alice"], workers["james"])
+
+    x = torch.tensor([1.0, 2.0, 3.0]).fix_prec()
```
gmuraru, line 109: MAYBE: Use 
```
tensor = torch.tensor([1.0, 2.0, 3.0])
x = tensor.fix_prec()

result = x.reciprocal().float_prec()
assert (result == tensor.reciprocal()).all()
```

marload: Because the torch tensor uses the Floating Point, the original method of using division also has an error. That's why you can't use this method. 😄 

```python
tensor = torch.tensor([1.0, 2.0, 3.0])
x = tensor.fix_prec()

result = x.reciprocal().float_prec()
assert (result == tensor.reciprocal()).all() # -> tensor([True, True, False])
# tensor([1.0000, 0.5000, 0.3333]) == tensor([1.0000, 0.5000, 0.3333]) == 
```

gmuraru: Ahh...don't use equal, use: [```torch.isclose```](https://pytorch.org/docs/stable/generated/torch.isclose.html)

marload: I applied this change request and added a docstring 😄

gmuraru:
Re-run the tests. It might be a problem regarding the ```seed``` - we might need to increase the tolerance (this issue was seen before :\)

Syzygianinfern0:
> Re-run the tests. It might be a problem regarding the `seed` - we might need to increase the tolerance (this issue was seen before :)

It ran 3 times and failed all of them :rofl:

marload:
@Syzygianinfern0 @gmuraru 

The error appears to be caused by a small error in the NR. We solved this problem by setting the default value of reciprocal to division. :smile

Syzygianinfern0:
Is the NR scheme buggy? Why did it fail for that?

> @Syzygianinfern0 @gmuraru
> 
> The error appears to be caused by a small error in the NR. We solved this problem by setting the default value of reciprocal to division.

But how does an error tolerance of about `1e-2` (as seen from the test cases you've written) propagate to something like `20 < 1` ?

marload:
NR uses the Approx Exp, so there is a little more error than Division. I think this problem is caused by the error because `tanh` uses `_sigmoid_exp`.

gmuraru:
If we change such that ```sigmoid``` uses ```chebyshev``` by default, does it work?

marload:
Yes, that's right. No errors occur in the `chebyshev`.

marload:
If `chebyshev` is the default, no error occurs.
I just checked. 😄

gmuraru:
I think we can go with ```chebyshev``` as the default one

Syzygianinfern0:
@marload 
A couple of things to consider for future PRs :smile: 
- Squashing basic commits to hold a good commit structure before a push
- Working on a branch different from the master on your fork (makes it easier for you)

Thank you again for your contributions!

PR #4032, opened 2020-08-17
Made tests fault tolerant to missing crypten module for windows

### LINKED ISSUE #4025: Check dependency CrypTen tests
## Description
Currently, CrypTen is available only on Linux and Mac, but if someone installs the PySyft repo and run the tests we also have some CrypTen related tests.
Those should be taken into consideration only if we have ```dependency_check.crypten```.

## How to Reproduce
Run the tests on Windows

## Expected Behavior
All the tests should pass - maybe skip the tests for CrypTen.

Karnak123:
## Description
Resolves #4025 by skipping crypten test files and adding expected to fail in serde tests if operating system is Windows.

## How has this been tested?
My own system was failing some tests and `test_inv_sym` goes on forever, so I ran only crypten and serde tests by temporarily moving other test files to another destination.

## Checklist

### DIFF
--- a/test/serde/msgpack/test_msgpack_serde_full.py
+++ b/test/serde/msgpack/test_msgpack_serde_full.py
@@ -3,6 +3,7 @@ import pytest
 import numpy
 import torch
 from functools import partial
+from os import name as os_name
 
 import syft
 from syft.serde import msgpack
@@ -70,7 +71,8 @@ samples[
 ] = make_fixedprecisiontensor
 samples[syft.frameworks.torch.tensors.interpreters.private.PrivateTensor] = make_privatetensor
 
-samples[syft.frameworks.crypten.model.OnnxModel] = make_onnxmodel
+if os_name != "nt":
+    samples[syft.frameworks.crypten.model.OnnxModel] = make_onnxmodel
 
 samples[syft.generic.pointers.multi_pointer.MultiPointerTensor] = make_multipointertensor
 samples[syft.generic.pointers.object_pointer.ObjectPointer] = make_objectpointer
@@ -98,6 +100,7 @@ samples[syft.workers.virtual.VirtualWorker] = make_virtual_worker
 samples[SerializableDummyClass] = make_serializable_dummy_class
 
 
+@pytest.mark.xfail("os_name == 'nt'")
 def test_serde_coverage():
     """Checks all types in serde are tested"""
     for cls, _ in msgpack.serde.msgpack_global_state.simplifiers.items():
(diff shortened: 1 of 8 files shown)


Karnak123:
The fail shows `FAILED test/torch/tensors/test_precision.py::test_torch_sigmoid_approx[exp-3-0.065]` and I didn't touch the function and neither is it dependent on anything I changes/added, so do I need to worry about it?

gmuraru:
Can't we have a skip method for the entire file (rather than marking each individual test?)

Karnak123:
Not any way I know of, if they were all in a class or module we could have with pytest.mark.skipif(), to skip entire file only pytest.mark.skip is available but this unconditionally skips all tests, so the tests wouldn't run on Linux and Mac too

gmuraru:
> Not any way I know of, if they were all in a class or module we could have with pytest.mark.skipif(), to skip entire file only pytest.mark.skip is available but this unconditionally skips all tests, so the tests wouldn't run on Linux and Mac too

It seems that we can skip a file using the ```conftest.py```. There are more details [here](https://docs.pytest.org/en/latest/example/pythoncollection.html#customizing-test-collection) - did not test it, but looks promising

Karnak123:
Yeah that worked thanks! So now I'm checking if its a Windows system and ignoring all test in `test\crypten`, but still using individual checks for the other two files because only a small part of their checks depend on crypten, should I skip them as files them too or let them be? Currently they are marked as expected to fail

Syzygianinfern0:
> should I skip them as files them too or let them be? Currently they are marked as expected to fail

I guess having it as an xfail is sufficient.

test/serde/msgpack/test_msgpack_serde_full.py
```diff
@@ -98,6 +100,7 @@
 samples[SerializableDummyClass] = make_serializable_dummy_class
 
 
+@pytest.mark.xfail("os_name == 'nt'")
```
Syzygianinfern0, line 103: Could you add the `reason` attribute to these?

Syzygianinfern0 (changes requested):
We can have it merged right after that! :tada:

test/serde/msgpack/test_msgpack_serde_full.py
Karnak123: Of course, added attribute

Syzygianinfern0 (approved):

gmuraru:
There is also a coverage issue - You might need to ignore the crypten related files for coverage if windows is detected

Karnak123:
> There is also a coverage issue - You might need to ignore the crypten related files for coverage if windows is detected

Looked into it, couldn't find any such issues in the tests

iamtrask:
OpenCI seems to be hanging?

Syzygianinfern0:
> OpenCI seems to be hanging?

Yep

Karnak123:
Excuse me? Didn't understand

Karnak123:
Hi, WIP check hasn't still completed, so should I re-commit the code? Or is there anything else I need to change?

gmuraru:
Hmm..Restarted the tests

Karnak123:
I pushed an empty commit, and the test fail is due to 94% test coverage but in the last commit it was passing and I haven't made any changes since so not sure what to do :/

gmuraru:
Ahh...I do not know how this got out of my notification bar from GitHub.
Could you reopen it, please? - I can check it out.

PR #4038, opened 2020-08-18
Pointer docs

tudorcebere:
## Description
Adding documentation and syft decorator on pointers in syft 0.3.0.


src/syft/core/pointer/pointer.py
```diff
@@ -126,17 +195,23 @@ def request_access(self, request_name: str = "", reason: str = "",) -> None:
...
+            data.
+            2. RequestStatus.Pending - your request has not been reviewed yet.
+            3. RequestStatus.Rejected - your request has been rejected.
+
+        :param node: The node that queries the request status.
+        :type node: AbstractNode
+        :param request_id: The request on which you are querying the status.
+        :type request_id: UID
+        """
+        from ..node.domain.service import RequestAnswerMessage
```
gmuraru, line 210: Q: Can't we add this at the top? or dependency resolution conflict?

src/syft/core/pointer/pointer.py
```diff
@@ -1,9 +1,42 @@
...
+
+There are two proper ways of receiving a pointer on some data:
+    1. When sending that data on a remote machine the user receives a pointer.
+    2. When the user searches for the data in an object store it receives a pointer to that data,
+    if it has the correct permissions for that.
+
+After receiving a pointer, one might want to get the data behind the pointer locally. For that the
+user should:
+    1. Request access.
+    2.1 - The data owner has to approve the request.
```
iamtrask, line 20: let's include the actual .request_access() method name here. Possibly with an exxample code snippet!

src/syft/core/pointer/pointer.py
```diff
@@ -13,12 +46,35 @@
 
 
 class Pointer(AbstractPointer):
+    """
+    Pointer is the handler when interacting with remote data.
+
+    Automatically generated subclasses of Pointer need to be able to look up
```
iamtrask, line 52: ... "for more information on how subclasses of pointer are automatically generated..., please see <ast module discussing automatic generation... probably st.klass>"

This notion that "this Pointer class is never directly use but is subclassed" should probably be added to the story at the top.

src/syft/core/pointer/pointer.py
```diff
@@ -112,6 +169,18 @@ def get_protobuf_schema() -> GeneratedProtocolMessageType:
         return Pointer_PB
 
     def request_access(self, request_name: str = "", reason: str = "",) -> None:
+        """Method that requests access to the data on which the pointer points to.
+
```
iamtrask, line 173: let's include an example code snippet here.

src/syft/core/pointer/pointer.py
iamtrask: It's a circular dependency issue at the moment - but it should hopefully be a fixable one.

iamtrask: PR welcome - but be wared - here be dragons

iamtrask (changes requested):
Made a few suggestions - overall looking really good!

iamtrask:
Apologie - one more nit-picky thing. 

https://www.sphinx-doc.org/en/1.7/markup/code.html

Let's do the code example using a code block like this. This will make it print nicely when we run "python setup.py docs"

iamtrask (approved):

PR #4039, opened 2020-08-19
Finished domain test check_access on RequestStatus.Accepted

madhavajay:
## Description
- Finished domain test check_access on RequestStatus.Accepted
- Moved key_emoji to util.py
- Refactored read_permissions and search_permissions to dicts:
  Optional[Dict[VerifyKey, Optional[UID]]]
  to allow check_access given a request_id to find .Accepted requests

## Affected Dependencies

## How has this been tested?
CI

## Checklist


madhavajay:
I have added an Example API notebook with asserts that is also run as a CI job.
The purpose of this notebook is to provide Dev Documentation with working examples that are verified on every PR.

examples/api/start.ipynb
```diff
@@ -0,0 +1,1321 @@
...
+   "outputs": [],
+   "source": [
+    "sy.VERBOSE = False"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Bob decides has some data he wants to share with Alice using PySyft.\n",
```
radusqrt, line 68: add: he - before has

radusqrt (approved):

.github/workflows/tests.yml
```diff
@@ -84,3 +84,33 @@ jobs:
         run: |
           pip install typing-extensions dataclasses
           python setup.py test
+
+  python-nb-examples:
```
iamtrask, line 88: very nice! I'm always excited about this kind of notebook testing.

iamtrask: I can't tell from the code - what is the scope of this? Will it run all notebooks or just some?

iamtrask: Ah nvm

jupyter nbconvert --to script examples/api/*.ipynb

Only notebooks in the api folder.

examples/api/pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
```
iamtrask, line 1: interesting - i didn't know you coudl put this file here and it woudl do things

proto/core/auth/signed_message.proto
```diff
@@ -1,9 +1,12 @@
...
 package syft.core.auth;
 
+import "proto/core/common/common_object.proto";
+
 message SignedMessage {
-  string obj_type = 1;
-  bytes signature = 2;
-  bytes verify_key = 3;
-  bytes message = 4;
+  syft.core.common.UID msg_id = 1;
```
iamtrask, line 7: I'm surprised we needed to add this here. Is the ID different from the id of the message inside SignedMessage?

iamtrask: Well - maybe not too surprised. It makes sense we'd wnat ot have access to the ID without having to deserialize message. I'm hoping that the msg id isn't different than message though.

proto/core/io/address.proto
```diff
@@ -4,17 +4,18 @@ package syft.core.io;
 import "proto/core/io/location.proto";
 
 message Address {
-  string obj_type = 1;
+  string name = 1;
+  string obj_type = 2;
```
iamtrask, line 8: I like the idea that Address has a name - although I'm concerned that it will be more than just for human-readable output. Aka - we shouldn't base any application logic off of name

scripts/pre_commit.sh
```diff
@@ -7,3 +7,13 @@ black src tests
 pre-commit run --all-files
 bandit -r src -ll
 python setup.py test
+
+# run API documentation test notebooks
```
iamtrask, line 11: Nice touch! Glad you added this here too.

src/syft/ast/callable.py
```diff
@@ -41,7 +44,9 @@ def __call__(self, *args, return_callable=False, **kwargs):
                 path=path, index=index + 1, return_callable=return_callable
             )
 
-    def add_path(self, path, index, return_type_name=None):
+    def add_path(
```
iamtrask, line 47: Adding more typing - thank you.

src/syft/ast/globals.py
```diff
@@ -1,4 +1,5 @@
 # from .. import ast # CAUSES Circular import errors
+from typing import Union, List, Optional
```
iamtrask, line 2: Let's get in the habit of putting these imports on separate lines.

src/syft/ast/globals.py
```diff
@@ -7,17 +8,23 @@ class Globals(Module):
...
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__("globals", None, None, None)
 
-    def add_framework(self, attr_name, attr):
+    def add_framework(self, attr_name: str, attr: object) -> None:
         self.attrs[attr_name] = attr
 
-    def add_path(self, path, return_type_name=None, framework_reference=None):
+    def add_path(
```
iamtrask, line 17: these types look good

src/syft/ast/globals.py
```diff
@@ -7,17 +8,23 @@ class Globals(Module):
...
+        path: Union[str, List[str]],
+        index: int = 0,
+        return_type_name: Optional[str] = None,
+        framework_reference: Optional[object] = None,
+    ) -> None:
         if isinstance(path, str):
             path = path.split(".")
 
-        framework_name = path[0]
+        framework_name = path[index]
```
iamtrask, line 27: I.... don't think this is right

iamtrask: The framework_name shoudl always be the first part of the path.

src/syft/ast/module.py
```diff
@@ -25,8 +26,13 @@ def __repr__(self):
 
         return out
 
-    def add_path(self, path, index, return_type_name):
-
+    def add_path(
+        self,
+        path: List[str],
```
iamtrask, line 31: types look right

src/syft/ast/util.py
```diff
@@ -23,7 +23,7 @@
 class_type = type(func_type)
 
 
-def unsplit(list_of_things, separator="."):
+def unsplit(list_of_things: List[str], separator: str = ".") -> str:
     out = ""
```
iamtrask, line 27: yup

src/syft/core/common/message.py
```diff
@@ -69,6 +69,7 @@ def sign(self, signing_key: SigningKey) -> SignedMessageT:
         # for example ReprMessage -> ImmediateSyftMessageWithoutReply.signed_type
         # == SignedImmediateSyftMessageWithoutReply
         return self.signed_type(
+            msg_id=self.id,
```
iamtrask, line 72: i htink i asked about this above - good to see we're copying the ID here

src/syft/core/common/message.py
```diff
@@ -120,14 +122,14 @@ def _object2proto(self) -> SignedMessage_PB:
...
 
-        proto = SignedMessage_PB()
         # obj_type will be the final subclass callee for example ReprMessage
-        proto.obj_type = self.obj_type
-        proto.signature = bytes(self.signature)
-        proto.verify_key = bytes(self.verify_key)
-        proto.message = self.serialized_message
-
-        return proto
+        return SignedMessage_PB(
```
iamtrask, line 126: nice cleanup

src/syft/core/io/address.py
```diff
@@ -110,28 +115,26 @@ def post_init(self) -> None:
 
     @syft_decorator(typechecking=True)
     def key_emoji(self, key: Union[bytes, SigningKey, VerifyKey]) -> str:
-        hex_chars = bytes(key).hex()[-8:]
-        return self.char_emoji(hex_chars=hex_chars)
-
-    @syft_decorator(typechecking=True)
-    def char_emoji(self, hex_chars: str) -> str:
-        base = ord("\U0001F642")
```
iamtrask, line 118: good cleanup

src/syft/core/common/serde/serializable.py
```diff
@@ -14,6 +14,7 @@
 from ....decorators import syft_decorator
 from ....proto.util.json_message_pb2 import JsonMessage
 from ....util import get_fully_qualified_name
+from ....util import random_name
```
iamtrask, line 17: woah? what's this?

src/syft/core/io/address.py
```diff
@@ -151,6 +154,7 @@ def _object2proto(self) -> Address_PB:
             object.
         """
         return Address_PB(
+            name=self.name,
```
iamtrask, line 157: cool - i see this propagated its way through

src/syft/core/node/common/action/run_class_method_action.py
```diff
@@ -37,13 +37,21 @@ def __init__(
         self.kwargs = kwargs
         self.id_at_location = id_at_location
 
-    def execute_action(self, node: AbstractNode, verify_key: VerifyKey) -> None:
-        # print(self.path)
-        # print(self._self)
-        # print(self.args)
-        # print(self.kwargs)
-        # print(self.id_at_location)
+    def intersect_keys(
```
iamtrask, line 40: this should probably be a util not a class method

src/syft/core/node/common/action/save_object_action.py
```diff
@@ -49,7 +49,10 @@ def execute_action(self, node: AbstractNode, verify_key: VerifyKey) -> None:
                     if hasattr(self.obj, "description")
                     else ""
                 ),
-                read_permissions=set([verify_key, node.verify_key]),
+                read_permissions={
+                    node.verify_key: node.id,
+                    verify_key: None,  # we dont have the passed in sender's UID
```
iamtrask, line 54: if we don't pass in the sender's verify_key then they won't be able to download the thing they sent. This is incorrect.

src/syft/core/node/common/service/obj_search_permission_service.py
```diff
@@ -121,9 +121,9 @@ def process(
         verify_key: VerifyKey,
     ) -> None:
         if msg.add_instead_of_remove:
-            node.store[msg.target_object_id].search_permissions.add(verify_key)
+            node.store[msg.target_object_id].search_permissions[verify_key] = msg.id
         else:
-            node.store[msg.target_object_id].search_permissions.remove(verify_key)
+            node.store[msg.target_object_id].search_permissions.pop(verify_key, None)
```
iamtrask, line 126: makes sense for this to be a set instead of list. Good change

src/syft/core/node/device/device.py
```diff
@@ -52,6 +52,9 @@ def __init__(
             verify_key=verify_key,
         )
 
+        # specific location with name
```
iamtrask, line 55: not sure i like the name of the attribute here - in the long run there should be a set of locations but it doesn't have to be that way for now

src/syft/core/node/domain/domain.py
```diff
@@ -112,13 +114,18 @@ def get_request_status(self, message_request_id: UID) -> RequestStatus:
...
                 return RequestStatus.Pending
 
-        # TODO: implement RequestStatus.Accepted
-        # maybe pass through the verify key and check if that exists?
         # check if it was accepted
-        # request_id => object_id is currently not persisted anywhere
-        # if object_id in self.store:
-        #     self.store[object_id] ??
-        # return RequestStatus.Accepted
+        # TODO remove brute search of all store objects
```
iamtrask, line 118: This TODO will be fulfilled with the metadata store.

src/syft/core/node/domain/service/accept_or_deny_request_service.py
```diff
@@ -115,21 +116,28 @@ def process(
...
-                        node.store[req.object_id].read_permissions.add(
+                        node.store[req.object_id].read_permissions[
                             req.requester_verify_key
-                        )
+                        ] = req.id
                         node.requests.remove(req)
 
                         if sy.VERBOSE:
-                            print(f"> Accepting Request:{request_id}")
+                            print(
```
iamtrask, line 133: nice printing

src/syft/core/node/vm/vm.py
```diff
@@ -47,6 +47,9 @@ def __init__(
             verify_key=verify_key,
         )
 
+        # specific location with name
+        self.vm = SpecificLocation(name=self.name)
```
iamtrask, line 51: same comment as DEvice... not sure this is the right attr name for this attribute

iamtrask: mostly because a node can subscribe to many locations

src/syft/core/store/store_interface.py
```diff
@@ -157,3 +158,28 @@ def delete_object(self) -> None:
...
+    def icon(self) -> str:
+        return "🗃️"
+
+    @property
+    def pprint(self) -> str:
+        output = f"{self.icon} ({self.class_name}) {self.peek}"
+        return output
+
+    @property
+    def peek(self) -> str:
```
iamtrask, line 172: I'm not 100% sure i like peek.

src/syft/core/node/common/action/save_object_action.py
iamtrask: Correction - I read this wrong. I was reading the key:value paring as a type hint. This makes sense.

src/syft/util.py
```diff
@@ -117,3 +122,380 @@ def obj2pointer_type(obj):
...
+    "wu",
+    "yalow",
+    "yonath",
+    "zhukovsky",
+]
+
+
+# we could replace these with some favorite AI / Privacy researches and engineers?
+@syft_decorator(typechecking=True)
+def random_name() -> str:
```
iamtrask, line 498: ooo! I love what you did wtih this! I was skeptical at first but by the time i got to this class I love it!

tests/syft/core/node/node_test.py
```diff
@@ -151,14 +155,16 @@ def test_send_message_from_network_client_to_vm() -> None:
     bob_vm = sy.VirtualMachine(name="Bob")
     bob_vm_client = bob_vm.get_client()
     bob_vm.root_verify_key = bob_vm_client.verify_key  # inject 📡🔑 as 📍🗝
-    print(f"> {bob_vm.pprint} {bob_vm.keys}")
-    print(f"> {bob_vm_client.pprint} {bob_vm_client.keys}")
+    if sy.VERBOSE:
```
iamtrask, line 158: thank you for sticking with the sy.VERBOSE flag

iamtrask (approved):
Merging although there's a bug in here which I will fix immediately after merging (see comments in ast module)

.github/workflows/tests.yml
madhavajay: Yes, although since pytest is failing on these, it is currently only running the first one:
```
pip install pytest pytest-cov nbconvert
pip install -r requirements.txt
pip install -e .
jupyter nbconvert --to script examples/api/*.ipynb
> python examples/api/start.py
```

If someone knows why pytest is failing on these tests with this command:
```
$ pytest examples/api
```

But works fine on the normal tests:
```
$ pytest
```

I would love to know. Then we can turn it back to using pytest and anything inside examples/api will be converted to .py and executed as a test.

proto/core/auth/signed_message.proto
madhavajay: The idea was to keep them the same which means as messages flow around you can keep track of them regardless of whether they are signed or not. Thats how its working now, but if its wrong please let me know. I wasn't sure if we had a requirement that all messages had unique IDs since in a way the contents and a signed message are not ==.

proto/core/io/address.proto
madhavajay: 100% agree, we do not want people relying on these, but i have found during debugging when a node or client collapses into an Address it loses so much detail its hard to know what it is, if it retains a "name" and its 4x Locations its easy to figure it out even after serde and storage somewhere remotely.

src/syft/ast/globals.py
madhavajay: I think its time I do the isort fix that will make all of these problems automatically go away! :) 👍

src/syft/ast/globals.py
madhavajay: Yeah this was tough, so the old code had different interfaces in globals vs module which mypy complained about. So i changed it so they matched (which is fine since we try to enforce kwargs anyway), but set the default index to 0 for globals.

```
# globals.py
def add_path(
        self,
        path: Union[str, List[str]],
        index: int = 0,
        return_type_name: Optional[str] = None,
        framework_reference: Optional[object] = None,
    ) -> None:
        if isinstance(path, str):
            path = path.split(".")

        framework_name = path[index]

# module.py
def add_path(
        self,
        path: List[str],
        index: int,
        return_type_name: Optional[str] = None,
        framework_reference: Optional[object] = None,
    ) -> None:
        if path[index] not in self.attrs:

            attr_ref = getattr(self.ref, path[index])
```

Without this index will be unused but also required to match the parent signature.... so not sure what the best solution here is. I was hesitant to refactor too much without solid tests in place.

src/syft/core/node/vm/vm.py
madhavajay: Isn't this the implicit one created when a Node is created for itself?
```
bob_vm = sy.VirtualMachine(name="Bob's VM")
assert bob_vm.name == "Bob's VM"
assert bob_vm.vm.name == "Bob's VM"
```

tests/syft/core/node/node_test.py
madhavajay: I think we should change this to a logging library and set up some logging levels for everything so we can continue to build out more meaningful logging in a non disruptive way.

src/syft/core/node/common/action/save_object_action.py
madhavajay: Yeah sorry, this is ugly, we are overloading the permissions with a separate concept which is a log of requests being accepted vs rejected. This was the quickest way to get check "accepted" working. I don't like it but introducing a whole separate thing or not providing check accept also seemed wrong right now.

src/syft/core/node/device/device.py
madhavajay: I guess that means we should be thinking more about the switch to has-a rather than is-a for Node / Client <- Address.

PR #4043, opened 2020-08-19
Add wrap to remote execution

gmuraru:
## Description
Add the ring size as a parameter to the PRZS

## How has this been tested?
- Tests for PRZS

## Checklist

### DIFF
--- a/syft/generic/utils.py
+++ b/syft/generic/utils.py
@@ -47,6 +47,6 @@ def remote(func, location):
             location, *command, return_ids=response_ids, return_value=return_value
         )
 
-        return response
+        return response.wrap()
 
     return remote_exec


abogaziah:
this will break some tests, I don't know why they didn't run.
if you can fix the broken tests that would be cool, if not, I'll wrap the pointers after calling it temporarily until you make this change and fix what it broke :)

gmuraru:
> this will break some tests, I don't know why they didn't run.
> if you can fix the broken tests that would be cool, if not, I'll wrap the pointers after calling it temporarily until you make this change and fix what it broke :)

I will create an issue around this and close this PR - seems like a good issue to get along our codebase

PR #4049, opened 2020-08-20
Finished all mypy untyped-defs

madhavajay:
Finished all mypy untyped-defs  🎉😅

- Renamed location to client on pointer to force discussion / change
- Documented places where Client import is needed but circular import
- Formatted pre-commit-config.yaml

## Description
We can now enforce 100% mypy checking across the code base. (except for the decorators which I need more time to figure out the error).

## Affected Dependencies
None

## How has this been tested?
CI

## Checklist


madhavajay:
**Please Note:** I have renamed "location" to "client" in Pointer because the code currently uses a client to dispatch messages but serializes to an "Address" which is going to bite us at some point so we should figure out what we want to do here. This then opens a can of circular import worms with Client being used in lots of its own super() classes to dispatch messages. Perhaps we need an interface which promises to have "dispatch" methods but doesnt itself inherit from Client.

tudorcebere (approved):

iamtrask:
Incredible!!! You're a monstor!

PR #4054, opened 2020-08-20
[WIP] Add more operations for torch

gmuraru:
## Description
In this PR I am trying to cover more operations for the torch framework and also add tests as operations are added.
The ideal scenario is to test each operation using all types and check the returned result.
 ```quant``` and ```complex``` can not be tested for the moment.
 * ```complex``` not implemented
 * ```quant``` tensors are created using a different method and not with the constructor

## Status
-  [x] - Add basic operations (add, sub, mul) - tested on all the tensors
-  [ ] - Check why div does not work

### DIFF
--- a/tests/syft/lib/torch/tensor_test.py
+++ b/tests/syft/lib/torch/tensor_test.py
@@ -102,7 +102,10 @@ def test_torch_no_read_permissions() -> None:
     with pytest.raises(Exception) as exception:
         local_x = ptr.get()
 
-    assert str(exception.value) == "You do not have permission to .get() this tensor. Please submit a request."
+    assert (
+        str(exception.value)
+        == "You do not have permission to .get() this tensor. Please submit a request."
+    )
 
     x = th.tensor([1, 2, 3, 4])
 


tests/syft/lib/torch/tensor_test.py
```diff
@@ -1,24 +1,58 @@
...
+
+from syft.core.pointer.pointer import Pointer
+from syft.lib.torch.tensor_util import TORCH_STR_DTYPE
+
 import syft as sy
 import torch as th
 import pytest
+from itertools import product
 
+TYPES_TEST = [e for e in TORCH_STR_DTYPE.keys() if not e.startswith("complex")]
```
gmuraru, line 14: Need to remove the complex type because is still WIP in torch

gmuraru, line 14: Q: @tudorcebere @iamtrask do we want to test all types of tensors that we can send and do operations with them?

tudorcebere: Would be really cool!

iamtrask: Absolutely

iamtrask: Ideally we'd also test them in the context of backpropagation if we can.

PR #4467, opened 2020-08-24
Added PythonPrimitive StorableObject supporting python return types

madhavajay:
## Description
- Added allowlist["torch.Tensor.is_floating_point"]
- PythonPrimitive ser/de tests dont pass because of StorableObject err?
- torch.Tensor.is_floating_point test passes and there are the same number of failing tests as before

I also played around with pytest-parallel.
```
pip install pytest-parallel
pytest --workers auto -k 'test_all_allowlisted_tensor_methods_work_remotely_on_all_types'
```

## Affected Dependencies
None

## How has this been tested?
Local testing and pre_commit hook for linting.

## Checklist


src/syft/lib/python/primitive.py
```diff
@@ -0,0 +1,122 @@
...
+def isprimitive(value: Any) -> bool:
+    # check primitive types first
+    if type(value) in [type(None), bool, int, float]:
+        return True
+
+    # if the type is a collection lets recursively search
+    # set cant be easily converted to json
+    if type(value) in [tuple, list]:
+        for sub in value:
+            is_subprimitive = isprimitive(value=sub)
```
gmuraru, line 30: Can't we also have primitves mixed with non primitives?

src/syft/core/node/common/action/run_class_method_action.py
```diff
@@ -85,9 +86,25 @@ def execute_action(self, node: AbstractNode, verify_key: VerifyKey) -> None:
 
         result = method(resolved_self.data, *resolved_args, **resolved_kwargs)
 
-        # ensure that result has the right id
-        # TODO: overload all methods to incorporate this automatically
-        result.id = self.id_at_location
+        if isprimitive(value=result):
+            # Wrap in a PythonPrimitive
+            primitive = PythonPrimitive(value=result, id=self.id_at_location)
```
gmuraru, line 91: MAYBE:
```
result = PythonPrimitive(value=result, id = self.id_at_location)
```
and remove
```result = StorableObject```
since we have one at line 110

File: src/syft/lib/python/primitive.py (hunk 1 above)
Comment (madhavajay) [reply src/syft/lib/python/primitive.py:30]: yes, there are a few Tuple(Tensor, Tensor) return types I have seen, but we need to figure out how best to mix these.

Comment (madhavajay) [reply src/syft/lib/python/primitive.py:30]: Also for args / kwargs definitely, but this was just a quick stab at return types

File: src/syft/core/node/common/action/run_class_method_action.py [hunk 3]
```diff
@@ -85,9 +86,25 @@ def execute_action(self, node: AbstractNode, verify_key: VerifyKey) -> None:
...
+            result = StorableObject(
+                id=self.id_at_location,
+                data=primitive,
+                read_permissions=result_read_permissions,
+            )
+        else:
+            # TODO: overload all methods to incorporate this automatically
+            if hasattr(result, "id"):
```
gmuraru, line 99: Q: At this point we should have an id for all the results, right?
I think the best approach will be to throw an exception since not having an id is something that we did not expect

src/syft/core/node/common/action/run_class_method_action.py
```diff
@@ -85,9 +86,25 @@ def execute_action(self, node: AbstractNode, verify_key: VerifyKey) -> None:
...
+                result.id = self.id_at_location
+
+            # QUESTION: There seems to be return value tensors that have no id
+            # and get auto wrapped? So is this code not correct?
```
gmuraru, line 103: I vote for throwing exceptions

src/syft/core/node/common/action/run_class_method_action.py
madhavajay: So PythonPrimitive IS an instance of StorableObject, so that code wont execute.

madhavajay: So I think its normally done magically by this:
```
aggressive_set_attr(
    obj=th.nn.parameter.Parameter,
    name="serializable_wrapper_type",
    attr=PyTorchParameterWrapper,
)
```

But we arent injecting that into the primitive classes.

src/syft/core/store/storeable_object.py
```diff
@@ -200,6 +203,8 @@ def pprint(self) -> str:
         output = f"{self.icon} ({self.class_name}) ("
         if hasattr(self.data, "pprint"):
             output += self.data.pprint  # type: ignore
+        elif self.data is not None:
```
gmuraru, line 206: MAYBE: Remove this since we have the line above?

src/syft/core/node/common/action/run_class_method_action.py
madhavajay: So, it turns out that one of the tests fails on the else of that code, but I have commented out the exception I added.

madhavajay: Try uncommenting the ```
raise Exception(f"Result has no ID. {result}")
```

I have to finish for the day so I didnt get time to investigate whats going on but something is going on here.

File: src/syft/core/store/storeable_object.py (hunk 5 above)
Comment (madhavajay) [reply src/syft/core/store/storeable_object.py:206]: They aren't the same thing. The line above only checks for the pprint property. There are a lot of Tensor types that dont have any pretty print methods yet, so they dont print anything.

File: tests/syft/lib/torch/tensor/tensor_remote_method_api_suite_test.py [hunk 6]
```diff
@@ -232,13 +239,18 @@ def test_all_allowlisted_tensor_methods_work_remotely_on_all_types(
...
-            target_result[target_result != target_result] = 0
-
-            # Step 14: Ensure we got the same result locally (using normal pytorch) as we did remotely
-            # using Syft pointers to communicate with remote torch objects
-            assert (local_result == target_result).all()
+            if isprimitive(value=target_result):
+                assert local_result == target_result
+            else:
+                # type(target_result) == torch.Tensor
+                local_result[local_result != local_result] = 0
```
gmuraru, line 248: Q: Why we need this?
(can't we have only the assert form the end?)

src/syft/core/node/common/action/run_class_method_action.py
gmuraru: Hmm...if we add the ```serializable_wrapper_type``` to the ```PythonPrimitive``` we still can't remove the result = ```StorableObject``` (from this ```if```).

(sorry if this is a stupid question)

src/syft/core/node/common/action/run_class_method_action.py
iamtrask: I vote for this as well

src/syft/lib/python/primitive.py
```diff
@@ -0,0 +1,122 @@
...
+            )
+
+        return obj
+
+    @staticmethod
+    def get_data_protobuf_schema() -> GeneratedProtocolMessageType:
+        return PythonPrimitive_PB
+
+    @staticmethod
+    def get_protobuf_schema() -> GeneratedProtocolMessageType:
```
iamtrask, line 109: Don't overload this - the protobuf schema for StorableObject is always StorableObject_PB

iamtrask: Did it not work when you didn't overload it? (see uppercase_tensor.py as a reference)

tests/syft/lib/python/primitive_test.py
```diff
@@ -0,0 +1,144 @@
...
+    assert not isprimitive(value=d)
+
+    assert type(e) is bytes
+    assert not isprimitive(value=e)
+
+    assert type(bad_keys) is dict
+    assert not isprimitive(value=bad_keys)
+
+
+# TODO: Fix
```
iamtrask, line 125: I suspect this was broken because of the PB issue identified in the comment above

iamtrask (changes requested):
One serialization change but otherwise looks good

tests/syft/lib/torch/tensor/tensor_remote_method_api_suite_test.py
madhavajay: We can't access primitives as array index tensor style and they dont have NaNs to convert to 0 so this just fails on Non tensor types. The != = 0 is short hand for converting NaN to 0 because NaN != NaN

src/syft/lib/python/primitive.py
madhavajay: Seems okay without it but i was having issues with the serde tests so maybe thats why I was fiddling with it.

tests/syft/lib/python/primitive_test.py
madhavajay: Thanks for the tip. Im going to investigate.

src/syft/core/node/common/action/get_object_action.py
```diff
@@ -45,10 +47,24 @@ def _object2proto(self) -> GetObjectResponseMessage_PB:
             the other public serialization methods if you wish to serialize an
             object.
         """
+
+        ser = self.obj.serialize()
+        # TODO: Fix this hack
```
iamtrask, line 52: Yikes - let's put this TODO on the top fo the pile for future work. Thank you for documenting it here with a TODO when you found it.

src/syft/core/node/common/action/run_class_method_action.py
iamtrask: agreed - we should throw an error here. good job

src/syft/lib/torch/allowlist.py
```diff
@@ -482,7 +482,7 @@
 allowlist["torch.Tensor.tan"] = "torch.Tensor"
 # allowlist['torch.Tensor.tan_'] = #
 allowlist["torch.Tensor.tanh"] = "torch.Tensor"
-# allowlist['torch.Tensor.tanh_'] = #
+allowlist["torch.Tensor.tanh_"] = "torch.Tensor"
```
iamtrask, line 485: tanh_ returns None not torch.Tensor

iamtrask (approved):

src/syft/lib/torch/allowlist.py
madhavajay: I spent some time hunting through the Pytorch code which is difficult but most of the functionality terminates in ATen.

https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/UnaryOps.cpp
https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/BinaryOps.cpp

All the _ functions I can find seem to follow the same pattern:

```c
// UnaryOps.cpp
Tensor tanh(const Tensor& self) { return unary_op_impl(self, at::tanh_out); }
Tensor& tanh_(Tensor& self) { return unary_op_impl_(self, at::tanh_out); }
```

```c
// BinaryOps.cpp
Tensor div(const Tensor& self, const Tensor& other) {
  if (isIntegralType(self.scalar_type(), /*includeBool=*/ true)
      && isIntegralType(other.scalar_type(), /*includeBool=*/ true)) {
    TORCH_CHECK(false,
      "Integer division of tensors using div or / is no longer supported, ",
      "and in a future release div will perform true division as in Python 3. ",
      "Use true_divide or floor_divide (// in Python) instead.");
  }

  Tensor result;
  auto iter = TensorIterator::binary_op(result, self, other);
  div_stub(iter.device_type(), iter);
  return iter.output();
}

Tensor& div_(Tensor& self, const Tensor& other) {
  return native::div_out(self, self, other);
}
```

They are the same as their non _ versions except that they take a non const reference and return a reference. So they all do return something and it seems that when this translates to python the result is it returns the "self" which was used for the operation.

Screen Shot 2020-08-26 at 9 28 28 am

Do we want to do anything different to this?

tudorcebere (approved):

PR #4471, opened 2020-08-24
Add unsqueeze_

### LINKED ISSUE #4458: Add torch.Tensor.unsqueeze_ to allowlist and test suite
# Description

This issue is a part of Syft 0.3.0 Epic 2: https://github.com/OpenMined/PySyft/issues/3696

In this issue, you will be adding support for remote execution of the torch.Tensor.unsqueeze_
method or property. This might be a really small project (literally a one-liner) or
it might require adding significant functionality to PySyft OR to the testing suite
in order to make sure the feature is both functional and tested.

## Step 0: Run tests and ./scripts/pre_commit.sh

Before you get started with this project, let's make sure you have everything building and testing
correctly. Clone the codebase and run:

```pip uninstall syft```

followed by

```pip install -e .```

Then run the pre-commit file (which will also run the tests)

```./scripts/pre_commit.sh```

If all of these tests pass, continue on. If not, make sure you have all the
dependencies in requirements.txt installed, etc. 

## Step 1: Uncomment your method in the allowlist.py file

Inside [allowlist.py](https://github.com/OpenMined/PySyft/blob/syft_0.3.0/src/syft/lib/torch/allowlist.py) you will find a huge dictionary of methods. Find your method and uncomment the line its on. At the time
of writing this Issue (WARNING: THIS MAY HAVE CHANGED) the dictionary maps from the 
string name of the method (in your case 'torch.Tensor.unsqueeze_') to the string representation
of the type the method returns.

## Step 2: Run Unit Tests

Run the following:

```python setup.py test```

And wait to see if some of the tests fail. Why might the tests fail now? I'm so glad you asked!

https://github.com/OpenMined/PySyft/blob/syft_0.3.0/tests/syft/lib/torch/tensor/tensor_remote_method_api_suite_test.py

In this file you'll find the torch method test suite. It AUTOMATICALLY loads all methods
from the allowlist.py file you modified in the previous step. It attempts to test them.

# Step 3: If you get a Failing Test

If you get a failing test, this could be for one of a few reasons:

### Reason 1 - The testing suite passed in non-compatible arguments

The testing suite is pretty dumb. It literally just has a permutation of possible
arguments to pass into every method on torch tensors. So, if one of those permutations
doesn't work for your method (aka... perhaps it tries to call your method without
any arguments but torch.Tensor.unsqueeze_ actually requires some) then the test will
fail if the error hasn't been seen before.

If this happens - don't worry! Just look inside the only test in that file and look
for the huge lists of error types to ignore. Add your error to the list and keep
going!!!

*WARNING:* make sure that the testing suite actually tests your method via remote
execution once you've gotten all the tests passing. Aka - if the testing suite
doesn't have ANY matching argument permutations for your method, then you're
literally creating a bunch of unit tests that do absolutely nothing. If this is the
case, then ADD MORE ARGUMENT TYPES TO THE TESTING SUITE so that your argument
gets run via remote execution. DO NOT CLOSE THIS ISSUE until you can verify that
torch.Tensor.unsqueeze_ is actually executed remotely inside of a unit tests (and not
skipped). Aka - at least one of the test_all_allowlisted_tensor_methods_work_remotely_on_all_types
unit tests with your method should run ALL THE WAY TO THE END (instead of skipping
the last part.)

*Note:* adding another argument type might require some serialization work if
we don't support arguments of that type yet. If so, this is your job to add it
to the protobuf files in order to close this issue!

### Reason 2 - torch.Tensor.unsqueeze_ returns a non-supported type

If this happens, you've got a little bit of work in front of you. We don't have 
pointer objects to very many remote object types. So, if your method returns anything
other than a single tensor, you probably need to add support for the type it returns
(Such as a bool, None, int, or other types). 

*IMPORTANT:* do NOT return the value itself to the end user!!! Return a pointer object
to that type!

*NOTE:* at the time of writing - there are several core pieces of Syft not yet working
to allow you to return any type other than a torch tensor. If you're not comfortable
investigating what those might be - skip this issue and try again later once
someone else has solved these issues.

### Reason 3 - There's something else broken

Chase those stack traces! Talk to friends in Slack. Look at how other methods are supported. 
This is a challenging project in a fast moving codebase!

And don't forget - if this project seems to complex - there are plenty of others that
might be easier.

### PR DESCRIPTION
**Pull request by gmuraru**
## Description
**Attention**: This needs to be merged after [this](https://github.com/OpenMined/PySyft/pull/4470)
Fixes #4458 
Added the possibility to use ```unsqueeze_``` as remote operation in the syft 0.3.0 version.

## How has this been tested?
- The tests from the allow commands list

## Checklist

### CODE REVIEW

File: src/syft/lib/torch/allowlist.py [hunk 1]
```diff
@@ -502,7 +502,7 @@
 allowlist["torch.Tensor.triu"] = "torch.Tensor"
 # allowlist['torch.Tensor.triu_'] = #
 allowlist["torch.Tensor.true_divide"] = "torch.Tensor"
-# allowlist['torch.Tensor.true_divide_'] = #
```
iamtrask, line 505: Do inline operations return torch.Tensor? I thought they return None, yes?

iamtrask (changes requested):
Quick Q inline

src/syft/lib/torch/allowlist.py
madhavajay: Seems like they always return "self"

Screen Shot 2020-08-25 at 9 57 29 am
Screen Shot 2020-08-25 at 9 58 48 am

madhavajay:
LGTM!

madhavajay:
> Quick Q inline

sure

madhavajay:
Going to merge and continue on my branch.

PR #4474, opened 2020-08-25
Falcon - Select share + evaluating

gmuraru:
## Description
This PR implements the share selection from the Falcon paper and the **evaluation** part.
Fixes the first point and the last from from #3852

## How has this been tested?
- Added test to validate it works

## Checklist


syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -82,13 +88,13 @@ def retrieve_pointers(self):
         return pointers
 
     def __sum_shares(self, shares):
-        return sum(shares) % self.ring_size
+        return sum(shares) % self.field
```
gmuraru, line 91: @abogaziah changed the name to field to be consistent with the ```AdditiveSharingTensor``` + it seems also in CrypTen they use the same name attribute

syft/frameworks/torch/mpc/falcon.py
```diff
@@ -0,0 +1,82 @@
...
+        If the bit b_share is 1, y_sh is returned
+
+        Args:
+            x_share (ReplicatedSharingTensor): the first share to select
+            y_share (ReplicatedSharingTensor): the second share to select
+            b_share (ReplicatedSharingTensor): the bit to take into consideration
+                when selecting between x or y
+
+        Return:
+            x_share or b_share depending on the value of b_share
```
omershlo, line 40: typo: x_share or *y_share*

syft/frameworks/torch/mpc/falcon.py
```diff
@@ -0,0 +1,82 @@
...
+    def _share_random_tensor(
+        field: int, workers: List[BaseWorker]
+    ) -> (ReplicatedSharingTensor, ReplicatedSharingTensor):
+        """
+        TODO: Need to move this in Primitive Store:
+        Share a random tensor with the owners of some shares
+        """
+        c = torch.randint(2, size=(1,))
+        c_share_L = c.share(*workers, protocol="falcon", field=field)
+        c_share_2 = c.share(*workers, protocol="falcon", field=2)
```
omershlo, line 20: I couldn't find the secret sharing scheme you are using (can you share a link?), but if we want to be consistent with the protocol it should be  
1) replicated secret sharing , same as for `b` 
2) it should be maliciously secure secret sharing

syft/frameworks/torch/mpc/falcon.py
```diff
@@ -0,0 +1,82 @@
...
+from typing import List
+
+
+class Falcon(object):
+    @staticmethod
+    def _share_random_tensor(
+        field: int, workers: List[BaseWorker]
+    ) -> (ReplicatedSharingTensor, ReplicatedSharingTensor):
+        """
+        TODO: Need to move this in Primitive Store:
```
omershlo, line 15: this function is returning two different sharing which is not indicated from the name of the function and also makes it less general. why not return a single sharing and call the function twice?

syft/frameworks/torch/mpc/falcon.py
```diff
@@ -0,0 +1,82 @@
...
+    def select_share(
+        b_share: ReplicatedSharingTensor,
+        x_share: ReplicatedSharingTensor,
+        y_share: ReplicatedSharingTensor,
+    ) -> ReplicatedSharingTensor:
+        """ Performs select share protocol
+        If the bit b_share is 0, x_sh is returned
+        If the bit b_share is 1, y_sh is returned
+
+        Args:
```
omershlo, line 33: as a general comment: The sharing used in ReplicatedSharingTensor (https://github.com/OpenMined/PySyft/pull/3856/files) is for semi-honest adversary and falcon is secure against malicious adversary...

syft/frameworks/torch/mpc/falcon.py
```diff
@@ -0,0 +1,82 @@
...
+        diff = y_share - x_share
+        xor_b_c = b_share.xor(c_share_2)
+
+        ones_share = ReplicatedSharingTensor.one_shares(players, field)
+        if xor_b_c.reconstruct().item():
+            d_share = ones_share - c_share_L
+        else:
+            d_share = c_share_L
+
+        selected_share = diff * d_share + x_share
```
omershlo, line 57: according to the paper the multiplication here should be done using \pi_mult (secure multiplication). 
I am not sure how this is suppose to work otherwise.

syft/frameworks/torch/mpc/falcon.py
```diff
@@ -0,0 +1,82 @@
...
+            A RST that is (-1)^beta * x
+        """
+        field = x_share.field
+        players = x_share.players
+        shape = x_share.shape
+
+        pow_sh = (
+            ReplicatedSharingTensor.one_shares(players, field=field, shape=shape) - beta_share * 2
+        )
+        result = pow_sh * x_share
```
omershlo, line 80: here again, \pi_mult should be used (I might be missing something here)

syft/frameworks/torch/mpc/falcon.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Return:
+            A RST that is (-1)^beta * x
+        """
+        field = x_share.field
+        players = x_share.players
+        shape = x_share.shape
+
+        pow_sh = (
+            ReplicatedSharingTensor.one_shares(players, field=field, shape=shape) - beta_share * 2
```
omershlo, line 78: I don't think that you need to share a 1 for that. a much more efficient way will be to do (see Falcon paper, Linear operations): 
party1 share =  1-2[beta1]
party2 share =  -2[beta2]
party3 share =  -2[beta3]

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: I do not agree with that, field stands for the whole arithmetic field ]0, Z [, which makes it an incorrect and misleading name for the variable that represents Z only. also, the paper uses "ring size" so I think we should stick to it

syft/frameworks/torch/mpc/falcon.py
gmuraru: Done

syft/frameworks/torch/mpc/falcon.py
gmuraru: Yep - currently the RST and the Falcon protocol are "combined" inside the RST (the \pi_mul operation is encapsulated in the RST) - if we add more protocols over RST we can have something like we have in the ```aditive_sharing_tensor.py```
```
@crypto_protocl("fss")
def add(self, other):
    # here we define how our operation over the AST should work if we use the FSS protocol
```

syft/frameworks/torch/mpc/falcon.py
gmuraru: Hmm..in that case I think I can completely remove this function - doing that now

syft/frameworks/torch/mpc/falcon.py
```diff
@@ -0,0 +1,82 @@
...
+from syft.workers.base import BaseWorker
+
+from typing import List
+
+
+class Falcon(object):
+    @staticmethod
+    def _share_random_tensor(
+        field: int, workers: List[BaseWorker]
+    ) -> (ReplicatedSharingTensor, ReplicatedSharingTensor):
```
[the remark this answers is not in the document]
gmuraru: Removed this function

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -0,0 +1,73 @@
+import torch
```
gmuraru, line 1: I renamed the file and it seems it got all the changes

syft/frameworks/torch/mpc/przs.py
```diff
@@ -5,7 +5,7 @@
 import torch
 import syft
 
-RING_SIZE = 2 ** 32
+MAX_LIMIT = 2 ** 32
```
gmuraru, line 8: This is not really the ring size - is the maximum value for the random value (for the common randomness)

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -33,6 +33,18 @@ def __share_secret(self, plain_text, workers):
...
+        )
+
+    @staticmethod
+    def zero_shares(players, field, shape=(1,)):
+        return ReplicatedSharingTensor(
+            plain_text=torch.zeros(shape, dtype=torch.int64), players=players, field=field
+        )
+
     @staticmethod
     def __arrange_workers(workers):
```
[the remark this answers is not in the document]
gmuraru: Will create a ticket for adding annotations to all the elements in the falcon related implementation

syft/frameworks/torch/mpc/falcon.py
gmuraru: Yep - I left a comment there - basically doing this I think becomes a "little" intrusive with the RST - but it is worth thinking if we can rewrite this - hopefully, we will finish falcon soon and we will start benchmarking it and see where we can "get some juice"

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -33,6 +33,18 @@ def __share_secret(self, plain_text, workers):
         PRZS.setup(workers)
         return shares_map
 
+    @staticmethod
+    def one_shares(players, field, shape=(1,)):
```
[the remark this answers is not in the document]
gmuraru: Made an issue for adding types annotations for all stuff related to falcon [here](https://github.com/OpenMined/PySyft/issues/4505)

syft/frameworks/torch/mpc/falcon/falcon.py
```diff
@@ -11,3 +14,71 @@ def conv2d(filters: syft.ReplicatedSharingTensor, image, padding=0):
...
+    if xor_b_c.reconstruct().item():
+        d_share = ones_share - c_share_L
+    else:
+        d_share = c_share_L
+
+    selected_share = (y_share - x_share) * d_share + x_share
+    return selected_share
+
+
+def evaluate(
```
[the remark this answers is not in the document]
abogaziah: if you noticed this routine determines the negativity of x, if beta is an even number this will return x, if beta is odd it will return - x, you can choose a name that explains this functionality

syft/frameworks/torch/mpc/falcon/falcon.py
```diff
@@ -11,3 +14,71 @@ def conv2d(filters: syft.ReplicatedSharingTensor, image, padding=0):
...
+    x_share: ReplicatedSharingTensor, beta_share: ReplicatedSharingTensor
+) -> ReplicatedSharingTensor:
+    """Performs the computation (-1)^beta * x
+
+    Args:
+        x_share (ReplicatedSharingTensor): share to "evaluate"
```
abogaziah, line 68: please remove the "share" suffix

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -0,0 +1,73 @@
...
+    @staticmethod
+    def __private_unfold(image, kernel_size, padding):
+        return image.unfold(kernel_size, padding)
+
+    @staticmethod
+    def xor(
+        value: Union[int, ReplicatedSharingTensor], other: Union[int, ReplicatedSharingTensor]
+    ) -> ReplicatedSharingTensor:
+        """
+        Compute the XOR value between value and other
```
abogaziah, line 34: please remove the docstring, xor is self-documented it doesn't need comments, "you should see every comment as a failure to express your intuition with code"

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -0,0 +1,73 @@
...
+
+        values, other can have the following types:
+        (ReplicatedSharingTensor ring size 2, ReplicatedSharingTensor ring size 2)
+        (Public bit {0, 1}, RST) or (RST, Public bit {0, 1})
+
+        Returns:
+            The XOR computation between value and other
+        """
+
+        if not any(isinstance(x, ReplicatedSharingTensor) for x in (value, other)):
```
abogaziah, line 46: this expression will work on ints, this expression adds not-needed overhead

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -34,6 +34,18 @@ def __share_secret(self, plain_text, workers):
         PRZS.setup(workers)
         return shares_map
 
+    @staticmethod
+    def one_shares(players, ring_size, shape=(1,)):
```
abogaziah, line 38: please remove this, we need the interface to be as small as possible, you can just use torch.ones().share()

abogaziah (changes requested):

syft/frameworks/torch/mpc/falcon/falcon_helper.py
gmuraru: We should not use ```FalconHelper``` for simple ints. We should use ```int_a``` ^ ```int_b```

syft/frameworks/torch/mpc/falcon/falcon_helper.py
gmuraru: We need to do docstring because if we move to ```syft_0.3.0``` that will fail if we do not have every function documented.

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
gmuraru: Yep - tried to follow the model from ```AST```, but we might refactor that

syft/frameworks/torch/mpc/falcon/falcon.py
gmuraru: Will do

syft/frameworks/torch/mpc/falcon/falcon.py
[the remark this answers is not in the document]
gmuraru: Will rename it

syft/frameworks/torch/mpc/falcon/falcon.py
```diff
@@ -11,3 +14,71 @@ def conv2d(filters: syft.ReplicatedSharingTensor, image, padding=0):
...
+    # each worker add a random number at an offline step
+    c = torch.randint(high=2, size=(1,))
+    c_share_2 = c.share(*players, protocol="falcon", field=2)
+    c_share_L = c.share(*players, protocol="falcon", field=ring_size)
+
+    xor_b_c = FalconHelper.xor(b_share, c_share_2)
+
+    # TODO: Think of a method to do this better (1 - c_share) without being to
+    # "intrusive" with the operation
+    ones_share = ReplicatedSharingTensor.one_shares(players, ring_size)
```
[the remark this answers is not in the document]
gmuraru: Will do

syft/frameworks/torch/mpc/falcon/falcon.py
```diff
@@ -11,3 +14,71 @@ def conv2d(filters: syft.ReplicatedSharingTensor, image, padding=0):
     output_size = (image_height - filter_height + 2 * padding) + 1
     result = result.view(-1, channels_out, output_size, output_size)
     return result
+
+
+def select_share(
```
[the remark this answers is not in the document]
gmuraru: Done

syft/frameworks/torch/mpc/przs.py
gmuraru: Yep...it is. will redo

syft/frameworks/torch/mpc/falcon/falcon_helper.py
gmuraru: It is a check to make sure we are doing everything correctly - we could remove this after we finish the implementation.

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -0,0 +1,135 @@
...
+        ones = torch.ones(size=(1,)).share(*players, protocol="falcon", field=ring_size)
+        if xor_b_c.reconstruct().item():
+            d = ones - c_L
+        else:
+            d = c_L
+
+        selected_val = (y - x) * d + x
+        return selected_val
+
+    def negate_cond(
```
gmuraru, line 114: Thinking of a conditional negate - I do not know if this is the best name for it

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -0,0 +1,135 @@
+import torch
+import syft
+
+from syft.frameworks.torch.tensors.interpreters.replicated_shared import ReplicatedSharingTensor
+
+from typing import Union
+
+
+class FalconHelper:
+    XOR_EXPECTED_RING_SIZE = 2
```
abogaziah, line 10: why is this a global variable thought it's only in xor?

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -0,0 +1,135 @@
...
+    @staticmethod
+    def xor(
+        value: Union[int, ReplicatedSharingTensor], other: Union[int, ReplicatedSharingTensor]
+    ) -> ReplicatedSharingTensor:
+        """
+        Compute the XOR value between value and other.
+        If value and other are both ints we should use the "^" operator.
+
+        Args:
+            value (ReplicatedSharingTensor or Int): RST with ring size of 2 or Int value in {0, 1}
```
abogaziah, line 38: value should be always RST, other could be RST for private xor, or long tensor or int for public xor

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if not any(isinstance(x, ReplicatedSharingTensor) for x in (value, other)):
+            raise ValueError("One of the arguments should be a RST")
+
+        if all((isinstance(x, ReplicatedSharingTensor) for x in (value, other))) and not all(
+            (x.ring_size == FalconHelper.XOR_EXPECTED_RING_SIZE for x in (value, other))
+        ):
+            raise ValueError("If both arguments are RST then they should be in ring size 2")
+
+        # Make sure we have a ReplicatedSharingTensor as value
```
abogaziah, line 53: remove comments before merging to the codebase

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -0,0 +1,135 @@
...
+            x if b == 0 else y
+        """
+
+        ring_size = x.ring_size
+        players = x.players
+
+        # This should be generated by a crypto provider or create a 0 RST shared value and make
+        # each worker add a random number at an offline step
+        # (Idea): Use the correlated randomness shared at the beginning
+        c = torch.randint(high=2, size=(1,))
```
abogaziah, line 97: should the size be 1 or x.shape?

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -0,0 +1,135 @@
...
+            b (ReplicatedSharingTensor): RST of a bit
+
+        Return:
+            x if b == 0 else y
+        """
+
+        ring_size = x.ring_size
+        players = x.players
+
+        # This should be generated by a crypto provider or create a 0 RST shared value and make
```
abogaziah, line 94: remove comments

syft/frameworks/torch/mpc/falcon/falcon_helper.py
abogaziah: I know that negate_con doesn't mean conditional negative, do not use abbreviations just use the whole word until you come up with a better name

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -0,0 +1,135 @@
...
+            beta (ReplicatedSharingTensor): the reconstructed value should be in {0, 1}
+
+        Return:
+            A RST that is (-1)^beta * x
+        """
+        ring_size = x.ring_size
+        players = x.players
+        shape = x.shape
+
+        pow_sh = (
```
abogaziah, line 130: remove abbreviations

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -0,0 +1,135 @@
...
+            torch.ones(size=shape).share(*players, protocol="falcon", field=ring_size) - beta * 2
```
abogaziah, line 131: we can use public multiplication and addition here to save some communication rounds: (beta*-2)+1

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -0,0 +1,135 @@
...
+        c_2 = c.share(*players, protocol="falcon", field=2)
+        c_L = c.share(*players, protocol="falcon", field=ring_size)
+
+        xor_b_c = FalconHelper.xor(b, c_2)
+
+        # TODO: Think of a method to do this better (1 - c_share) without being to
+        # "intrusive" with the operation
+        ones = torch.ones(size=(1,)).share(*players, protocol="falcon", field=ring_size)
+        if xor_b_c.reconstruct().item():
+            d = ones - c_L
```
abogaziah, line 107: we can use public operations here: d= -1*c_L +1

abogaziah (changes requested):
thank you for the changes, it improved significantly 🎉 just few more changes

syft/frameworks/torch/mpc/falcon/falcon_helper.py
gmuraru: Yep - we can do. I had a feeling that it would require multiple multiple rounds of communications if we reorder the operations, but the public multiplication is done locally.

syft/frameworks/torch/mpc/falcon/falcon_helper.py
gmuraru: The same for this.

syft/frameworks/torch/mpc/falcon/falcon_helper.py
gmuraru: Will remove

syft/frameworks/torch/mpc/falcon/falcon_helper.py
gmuraru: Hm, I made it like this to be able to have ```xor(int, RST)``` - I can remove it

syft/frameworks/torch/mpc/falcon/falcon_helper.py
gmuraru: Reading that part of the paper again - yeah it should be ```x.shape```

syft/frameworks/torch/mpc/falcon/falcon_helper.py
gmuraru: Will do

syft/frameworks/torch/mpc/falcon/falcon_helper.py
abogaziah: I'm working on RST  to enable executing 1-c_L directly, it should be ready shortly

syft/frameworks/torch/mpc/falcon/falcon_helper.py
gmuraru: Hmm..this is not a global variable, it is accessed only using the class, but I think you are right, we can move it inside the xor method

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -21,8 +24,35 @@ def __private_unfold(image, kernel_size, padding):
...
+
+        Args:
+            value (ReplicatedSharingTensor): RST with ring size of 2 or integer value in {0, 1}
+            other (int): integer with value in {0, 1}
+
+        Returns:
+            The XOR computation between value and other
+        """
+
+        if not isinstance(value, ReplicatedSharingTensor) or value.ring_size != 2:
```
gmuraru, line 42: I have all those types in place - which look pretty bad - let's keep them in place only to make sure we define the protocol correctly and then remove them

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -178,8 +178,8 @@ def __public_linear_operation(self, plain_text, operator):
             shares_map[players[0]][1],
         )
         shares_map[players[-1]] = (
-            operator(shares_map[players[-1]][-1], remote_plain_text[-1]),
```
gmuraru, line 181: P1 has share x1, x2; P2 has x2,x3 and P3 has x3 and x1 - when adding a constant we will have
P1 has share x1 + c, x2, P2 has share x2,x3 (nothing changed) and P3 has x3 and x1 + c

test/torch/mpc/falcon/test_falcon_helper.py
```diff
@@ -1,24 +1,121 @@
...
+    bob, alice, james = (workers["bob"], workers["alice"], workers["james"])
+
+    x = torch.tensor([0]).share(bob, alice, james, protocol="falcon", field=3)
     y = torch.tensor([1]).share(bob, alice, james, protocol="falcon", field=2)
-    assert (FalconHelper.xor(x, y).reconstruct() == torch.tensor(1)).all()
-    assert (FalconHelper.xor(x, x).reconstruct() == torch.tensor(0)).all()
-    assert (FalconHelper.xor(y, y).reconstruct() == torch.tensor(0)).all()
-    assert (FalconHelper.xor(y, x).reconstruct() == torch.tensor(1)).all()
+
+    with pytest.raises(ValueError) as e:
```
abogaziah, line 33: last request: let's move the assertion inside the body of the code, the tests should only test the interface and helper functions, we can use assertions in code during construction it would make it easier to find and remove them before moving it to production
for example, put (assert x.ring_size == 2) inside the xor function instead of if statements and remove the corresponding test case 
also better manage whitespace, related expressions should be in a continuous chunk

abogaziah (commented):
Great work! just two final notes on the code

test/torch/mpc/falcon/test_falcon_helper.py
gmuraru: It is not ok to have asserts in the "real" code because they can be removed if you run with the ```-O``` or ```-OO``` flag.

test/torch/tensors/test_replicated_shared.py
```diff
@@ -11,6 +11,18 @@ def test_sharing(workers):
     assert type(secret.child) == dict
 
 
+def test_share_float(workers):
```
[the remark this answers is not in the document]
gmuraru: Hmm...in that case when generating the shares we should raise an exception is the passed Tensor is a different type than long, right?

test/torch/mpc/falcon/test_falcon_helper.py
```diff
@@ -1,24 +1,121 @@
+import pytest
+
 import torch
 from syft.frameworks.torch.mpc.falcon.falcon_helper import FalconHelper
 
+import itertools
```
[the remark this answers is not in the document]
gmuraru: Yep - I will manually create the ```TEST_VALS``` list with all values we except

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -48,7 +48,7 @@ def __arrange_workers(workers):
 
     def generate_shares(self, plain_text, number_of_shares=3):
```
gmuraru, line 49: If ```generate_shares``` is public shouldn't we move the ```__validation``` function here?
I should also remove the line from 51?

test/torch/tensors/test_replicated_shared.py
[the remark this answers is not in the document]
abogaziah: I added that in my last PR

test/torch/mpc/falcon/test_falcon_helper.py
abogaziah: yes that's the point, you should remove the asserts before pushing to production, asserts makes it easier to do that if you want your checks not to be moved to production to improve performance

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: it's not part of the interface, I left it public just to test it

abogaziah (approved):
Awesome!! thank you for the PR and the changes ❤🎉

gmuraru:
> Awesome!! thank you for the PR and the changes

We make a good team! :+1:

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -21,8 +24,33 @@ def __private_unfold(image, kernel_size, padding):
...
+        Args:
+            value (ReplicatedSharingTensor): RST with ring size of 2 or integer value in {0, 1}
+            other (int): integer with value in {0, 1}
+
+        Returns:
+            The XOR computation between value and other
+        """
+
+        assert (
+            isinstance(value, ReplicatedSharingTensor) and value.ring_size == 2
```
gmuraru, line 43: Q: This looks kind of *ugly* but this is how black formatted it :\

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -21,8 +24,33 @@ def __private_unfold(image, kernel_size, padding):
...
+        ), "First argument should be a RST with ring size 2"
+        assert any(
```
gmuraru, line 45: Q: Do you prefer this to be split into 3 asserts? (one for each type?) @abogaziah

abogaziah: there's no difference, anyone would work

syft/frameworks/torch/mpc/falcon/falcon_helper.py
abogaziah: no problem :D

PR #4479, opened 2020-08-25
Fix `Node.known_child_nodes` to only return Clients

Jasopaum:
## Description
`known_child_nodes` was returning all objects in the memory store.

## Affected Dependencies

## How has this been tested?

## Checklist


tests/syft/core/node/node_test.py
```diff
@@ -78,6 +85,30 @@ def test_register_vm_on_device_succeeds() -> None:
...
+    bob_vm_2.root_verify_key = bob_vm_client_2.verify_key  # inject 📡🔑 as 📍🗝
+
+    bob_phone = sy.Device(name="Bob's iPhone")
+    bob_phone_client = bob_phone.get_client()
+    bob_phone.root_verify_key = bob_phone_client.verify_key  # inject 📡🔑 as 📍🗝
+
+    bob_phone_client.register(client=bob_vm_client)
+
+    assert len(bob_phone.known_child_nodes) == 1
+    assert bob_vm.address.vm in (obj.data.vm for obj in bob_phone.known_child_nodes)
```
gmuraru, line 104: Q: Can't we simply check the address?
```bob_vm.address``` in bob_phone.known_child_nodes```?

**Review by gmuraru — APPROVED** [review]
LGTM!

File: src/syft/core/store/store_memory.py [hunk 2]
```diff
@@ -32,7 +32,7 @@ def __init__(self) -> None:
 
     def get_objects_of_type(self, obj_type: type) -> Set[AbstractStorableObject]:
         results = set()
-        for key, obj in self._objects.items():
```
gmuraru, line 35: NITPICK: Return {obj for obj in self._objects.values() if isinstance(obj, obj_type)} <-- only that it makes the code more concise.

tests/syft/core/node/node_test.py
Jasopaum: Good point!
I was having some problems because the adresses in known_child_nodes was composed of not only the vm but also the device... Not sure if it's a bug or what we want though.
What do you think about that?

src/syft/core/node/common/node.py
```diff
@@ -286,14 +286,7 @@ def known_child_nodes(self) -> List[Address]:
         if sy.VERBOSE:
             print(f"> {self.pprint} Getting known Children Nodes")
         if self.child_type_client_type is not None:
-            nodes = []
-            for key in self.store.keys():
-                value = self.store[key]
-                nodes.append(value)
-
-            # TODO: Make this work again
-            # nodes = self.store.get_objects_of_type(obj_type=self.child_type_client_type)
```
iamtrask, line 295: I'm not sure why this was commented out previously but in theory the child types shouldn't be Address types they should be self.child_type_client_type types. Meaning DomainNode will have type DeviceClient and DeviceNode will have child type VMClient.

Since they all subclass from Address this PR probably passes tests but in the long run it could cause issues. That being said, we also need to put a TODO where we explicitly say that a client is a child as this is a special case. But this requires us to finish the metadatastore first. So, for now let's just make sure we're using the narrowest type we can.

The main risk with the current functionality is that if a DomainNode has a client to another DomainNode it will think that that client is its child. I *think* our security implementation is still robust against this but it could certainly create strange behavior.

iamtrask (changes requested):

src/syft/core/node/common/node.py
Jasopaum: That's what I was thinking!
But I think that we now only have `Adresses` in the store and not the client types anymore (I found a comment somewhere telling that). So the information about the client's type seems lost... Or maybe I missed something?

src/syft/proto/util/json_message_pb2.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
 # source: proto/util/json_message.proto
-"""Generated protocol buffer code."""
+
```
madhavajay, line 4: Sorry, this is really annoying but protoc 3.13.0 which just got released seems to include these """Generated protocol buffer code.""" strings again. Can you make sure you have the latest protoc just so we can keep the noise down on these where possible.

tests/syft/core/node/node_test.py
madhavajay: Yes sorry this is all a work around for the fact that we cant serialize clients so to ensure we can disptach in a node service message we keep a python object reference to the clients in a dictionary on the node and then look it up from its UID later. Like here:
```
# heritage_update_service.py
for node_client in node.known_child_nodes:
    try:
        # TODO: Client (and possibly Node) should subclass from StorableObject
        location_id = node_client.data.target_id.id
        msg.address = node_client.data
        try:
            in_memory_client = node.in_memory_client_registry[location_id]
```
 I am interested to know how we are supposed to do this properly so that we can make the changes now to avoid further confusion. One I idea I had was we could operate a a single localhost flask server (like duet) in solo mode and then post to localhost/uid for all the different "clients" so that way there are no references to clients hanging around only UID addresses and a loopback connector which they all listen to on their /UID url stub?

src/syft/core/node/common/node.py
Jasopaum: Do you think `Address` could have a method/property to tell of which type it is (Network, Domain, ...)?

tests/syft/core/node/node_test.py
Jasopaum: I'm not 100% on all this part yet...
But why do we need to store the whole client? Couldn't we send a message only with the Address?

src/syft/core/node/common/node.py
iamtrask: I think we should simply store the Client object in the store. It's designed to cache all the information needed to be able to communicate with a specific location. Without all the other information, storing the address has no use (even if we know the type)

Jasopaum: The Clients are stored in a `in_memory_client_registry` attribute on Nodes, but it's advertised as a temporary hack...
According to what I understood, Clients are not stored in the main store because they can't be serialized, am I right?

iamtrask (approved):
I think this is a strong approach for now. Will we make this fancier in the future? Probably. But we have big fish to fry and not a long time to do it. This is solid.

src/syft/core/node/common/node.py
madhavajay: Yes, to send a message you must have a python reference to a client / connection object which you can call "send_immediate_msg_with_reply" (and its other variations) on, however the serialized version of a client will not unserialize back to a full graph of the connection / route. We could remove this hack by running a solo version of duet with a flask webserver and post messages back to localhost/{uid} so that no actual reference to a client running in memory is required to dispatch, as all nodes could be listening to their address url stub on localhost? I am interested to know how we should handle this more long term.

PR #4484, opened 2020-08-26
Falcon module structure & design 

abogaziah:
## Description
In this PR we initiate the file structure of the classes the Falcon protocol will use, establish their interfaces, and how they will communicate.
 
I have made three new classes:
1- Flacon tensor: extends ReplicatedSharingTensor to implement the tensor operations that are specific to Falcon protocol and not related to the replicated secret sharing scheme
2- Falcon: that's actually a module, not a class, it will overload the torch.nn.Functional module, hence it will reimplement its functions like conv2d, maxpool, batchnorm, etc..
3- FalconHelper: will include subroutines and helper functions of Falcon, we made it to hide details to make levels of abstractions .since Falcon isn't a class, we can't do that in the Falcon module and that's the job of this class 

## Checklist


iamtrask:
(I cancelled your tests owing to another process I was experimenting. If you needed these tests to merge promptly please do re-run and my apologies for the delay)

syft/frameworks/torch/mpc/falcon/falcon_tensor.py
```diff
@@ -0,0 +1,29 @@
+from syft.frameworks.torch.tensors.interpreters.replicated_shared import ReplicatedSharingTensor
+from operator import xor
+
+
+class FalconTensor(ReplicatedSharingTensor):
```
gmuraru, line 5: I thought more on this and having a ```FalconTensor``` (having the suffix tensor on it) it means we can apply all operations from torch to this, but ```xor``` is not a method from a tensor.

My idea is to have the ```xor``` functionality in the ```FalconHelper``` (receives 2 RST and applies the xor).

gmuraru: LE: Although, since we are writing a ```FalconTensor``` and xor was written under ```basic operations``` I think it might be a good idea.

abogaziah: that's an excellent point!

syft/frameworks/torch/mpc/falcon/faclon_helper.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @staticmethod
+    def __switch_public_private(value, public_function, private_function, *args, **kwargs):
+        if isinstance(value, (int, float, torch.Tensor, syft.FixedPrecisionTensor)):
+            return public_function(value, *args, **kwargs)
+        elif isinstance(value, syft.ReplicatedSharingTensor):
+            return private_function(value, *args, **kwargs)
+        else:
+            raise ValueError(
+                "expected int, float, torch tensor, or ReplicatedSharingTensor"
```
gmuraru, line 35: NITPICK: One space at the end - but I can fix that in my PR

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -295,14 +258,3 @@ def shape(self):
     @property
     def players(self):
         return self.__get_players()
-
-    def __repr__(self):
```
gmuraru, line 299: Q: Why remove those?

syft/frameworks/torch/mpc/falcon/falcon.py
```diff
@@ -0,0 +1,13 @@
+import syft
+from .faclon_helper import FalconHelper
```
gmuraru, line 2: NITPICK: We can remove the ```.```

gmuraru (approved):

syft/frameworks/torch/mpc/falcon/falcon.py
abogaziah: I think it won't work, you can either import an installed module or specify the directory

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: I have no idea :D I added them back

Issue #4477, opened 2020-08-25
Duplicated method name in base.py


gkaissis:
`syft/workers/base.py` contains two identically named `register_obj` methods:

https://github.com/OpenMined/PySyft/blob/99190dfa85d86b1d6542a7ee0454b2758c289605/syft/workers/base.py#L189

and 

https://github.com/OpenMined/PySyft/blob/99190dfa85d86b1d6542a7ee0454b2758c289605/syft/workers/base.py#L553

In best python tradition, only the latter one actually gets called, but we should fix that.

varun68:
@gkaissis ..I would like to give this a shot..how would you like me to approach this bug?

gkaissis:
Could you please see to the failing tests and perhaps @cereallarceny or @gmuraru have a look? This was a subject of discussion recently owing to the issues we had with websocket training

cereallarceny:
Hello! Just letting you know that we are no longer planning on supporting anything on the 0.2.x product line and that all work should be ported over to 0.3.x, which is considered a complete rebuild of PySyft. Because of that, I'll be closing this issue. If you feel this is a mistake, or if the issue actually applies to 0.3.x as well, please feel free to ping me on Slack and I'll reopen the issue.

PR #4498, opened 2020-08-28
Remove duplicate method in base.py

steph-en-m:
Remove the duplicate `def register_obj(self, obj):` method

This resolves [#4477](https://github.com/OpenMined/PySyft/issues/4477)
There were two duplicate methods in `syft/workers/base.py` and this pull request fixes this by omitting one of the methods.

### DIFF
--- a/syft/workers/base.py
+++ b/syft/workers/base.py
@@ -186,9 +186,6 @@ class BaseWorker(AbstractWorker):
         """
         return self.object_store.get_obj(obj_id)
 
-    def register_obj(self, obj):
-        self.object_store.register_obj(self, obj)
-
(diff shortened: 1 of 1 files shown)


gmuraru:
You might need to rerun the linter for this (there is a new black version)

cereallarceny:
Hello! Just letting you know that we are no longer planning on supporting anything on the 0.2.x product line and that all work should be ported over to 0.3.x, which is considered a complete rebuild of PySyft. Because of that, I'll be closing this issue. If you feel this is a mistake, or if the issue actually applies to 0.3.x as well, please feel free to ping me on Slack and I'll reopen the issue.

PR #4499, opened 2020-08-28
Minor code fixes

steph-en-m:
Some f-strings in `syft/exceptions.py` weren't properly implemented and this pull request adds a correct implementation of multi-line f-strings to improve logged messages/exceptions

### DIFF
--- a/syft/exceptions.py
+++ b/syft/exceptions.py
@@ -98,7 +98,7 @@ class TensorsNotCollocatedException(Exception):
         if isinstance(tensor_a, sy.PointerTensor) and isinstance(tensor_b, sy.PointerTensor):
             message = (
                 f"You tried to call {attr} involving two tensors which are not on the same machine!"
-                " One tensor is on {tensor_a.location} while the other is on {tensor_b.location}."
+                f" One tensor is on {tensor_a.location} while the other is on {tensor_b.location}."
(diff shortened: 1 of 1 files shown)


tudorcebere:
Hello @steph-en-m!

Thanks for the PR, one mention is that we are updating our `black` version for linting. After the update, please rerun linting and commit your changes. Otherwise, LGTM! :+1:

steph-en-m:
Do you have any idea when the updates(black) will be completed?

syft/workers/base.py
```diff
@@ -186,9 +186,6 @@ def get_obj(self, obj_id: Union[str, int]) -> object:
         """
         return self.object_store.get_obj(obj_id)
 
-    def register_obj(self, obj):
```
gmuraru, line 189: Q: I think this needs to be put back (I think there was another PR merged with this change)

steph-en-m: If I've checked properly, the duplicate method still exists.

cereallarceny:
Hello! Just letting you know that we are no longer planning on supporting anything on the 0.2.x product line and that all work should be ported over to 0.3.x, which is considered a complete rebuild of PySyft. Because of that, I'll be closing this issue. If you feel this is a mistake, or if the issue actually applies to 0.3.x as well, please feel free to ping me on Slack and I'll reopen the issue.

PR #4501, opened 2020-08-29
[WIP] Add support for FixedPrecisionTensor chain [Replicated Sharing Tensor (RST)]

### LINKED ISSUE #3931: Replicated Sharing Tensor (RST): Add support for FixedPrecisionTensor chain
Replicated Sharing Tensor currently can work if the plain text type is float or int, we want it to be able to work given a FixedPrecisionTensor as plain text.
When this issue is solved this test case should work properly:
```
a = a.fix_perc().share(bob, alice, james, protocol = "falcon")
b = a.reconstruct().float_perc()
assert (a==b).all()
```
also, it should be tested on arithmetic operation (addition, multiplication, and matrix multiplication)

aanurraj:
## Description
Added support for the FPT chain in RST and updated add and sub to support these operations in FPT chain. 

On completion this PR closes #3931 

## Affected Dependencies
None

## How has this been tested?
for [Replicated Sharing Tensor (RST)] FixedPrecisionTensor chain 
> Test cases are added for .share while using precision
> Test cases are added for add and sub  

## Checklist

### DIFF
--- a/syft/frameworks/torch/tensors/interpreters/replicated_shared.py
+++ b/syft/frameworks/torch/tensors/interpreters/replicated_shared.py
@@ -23,7 +23,7 @@ class ReplicatedSharingTensor(AbstractTensor):
                 return plain_text.child
             else:
                 dtype = plain_text.dtype if hasattr(plain_text, "dtype") else type(plain_text)
-                raise ValueError(f"expected torch.int64 but got {dtype} ")
+                raise ValueError(f"Expected torch.(int64/long) but got {dtype}")
         else:
             return None
 
@@ -48,7 +48,7 @@ class ReplicatedSharingTensor(AbstractTensor):
 
     def generate_shares(self, plain_text, number_of_shares=3):
         shares = []
-        plain_text.long()
+        plain_text = torch.tensor(plain_text, dtype=torch.long)
         for _ in range(number_of_shares - 1):
             shares.append(torch.randint(high=self.ring_size // 2, size=plain_text.shape))
         shares.append((plain_text - sum(shares)) % self.ring_size)
@@ -164,7 +164,7 @@ class ReplicatedSharingTensor(AbstractTensor):
             return private_function(value, *args, **kwargs)
         else:
             raise ValueError(
-                "expected int, float, torch tensor, or ReplicatedSharingTensor"
+                "expected int, float, torch tensor, or ReplicatedSharingTensor "
                 "but got {}".format(type(value))
             )
 
@@ -178,8 +178,8 @@ class ReplicatedSharingTensor(AbstractTensor):
             shares_map[players[0]][1],
         )
         shares_map[players[-1]] = (
-            operator(shares_map[players[-1]][-1], remote_plain_text[-1]),
-            shares_map[players[-1]][1],
+            shares_map[players[-1]][0],
+            operator(shares_map[players[-1]][1], remote_plain_text[-1]),
         )
         return ReplicatedSharingTensor().__set_shares_map(shares_map)
 
--- a/syft/frameworks/torch/mpc/falcon/falcon.py
+++ b/syft/frameworks/torch/mpc/falcon/falcon.py
@@ -1,5 +1,6 @@
 import syft
-from .falcon_helper import FalconHelper
+
+from syft.frameworks.torch.mpc.falcon.falcon_helper import FalconHelper
 
 
 def conv2d(filters: syft.ReplicatedSharingTensor, image, padding=0):
--- a/syft/frameworks/torch/mpc/falcon/falcon_helper.py
+++ b/syft/frameworks/torch/mpc/falcon/falcon_helper.py
@@ -1,6 +1,9 @@
 import torch
 import syft
 
+from syft.frameworks.torch.tensors.interpreters.replicated_shared import ReplicatedSharingTensor
+from typing import Union
+
 
 class FalconHelper:
     @classmethod
@@ -21,8 +24,33 @@ class FalconHelper:
         return image.unfold(kernel_size, padding)
 
     @staticmethod
-    def xor(value, other):
-        return value + other - (value * 2 * other)
+    def xor(
+        value: ReplicatedSharingTensor, other: Union[int, ReplicatedSharingTensor, torch.LongTensor]
+    ) -> ReplicatedSharingTensor:
+        """
+        Compute the XOR value between value and other.
+        If value and other are both ints we should use the "^" operator.
+
+        Args:
+            value (ReplicatedSharingTensor): RST with ring size of 2 or integer value in {0, 1}
+            other (int): integer with value in {0, 1}
+
+        Returns:
+            The XOR computation between value and other
+        """
+
+        assert (
+            isinstance(value, ReplicatedSharingTensor) and value.ring_size == 2
+        ), "First argument should be a RST with ring size 2"
+        assert any(
+            [
+                isinstance(other, ReplicatedSharingTensor) and other.ring_size == 2,
+                isinstance(other, int) and other in {0, 1},
+                isinstance(other, torch.LongTensor) and ((other == 0) + (other == 1)).all(),
+            ]
+        ), "Second argument should be RST (with ring size of 2)/Integer/LongTensor values in {0, 1}"
+
+        return value + other - 2 * value * other
 
     @staticmethod
     def __switch_public_private(value, public_function, private_function, *args, **kwargs):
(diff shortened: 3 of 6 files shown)


syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -929,17 +924,26 @@ def share(
...
-                    field=field,
-                    dtype=dtype,
-                    crypto_provider=crypto_provider,
-                    owner=self.owner,
-                )
-                .on(self.copy(), wrap=False)
-                .share_secret(*owners)
-            )
+            else:
+                if self.type() == "torch.FloatTensor":
```
abogaziah, line 928: this condition is not reachable, you have already checked it previously.

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -1103,3 +1107,5 @@ def numpy_tensor(self):
                 "on a wrapper. Add NumpyTensor to the chain by hand if you want "
                 "this functionality.",
             )
+
+    reconstruct = get
```
abogaziah, line 1111: how is that useful?

test/torch/tensors/test_replicated_shared.py
```diff
@@ -180,3 +180,142 @@ def test_workers_arrangement(workers):
     me, bob, alice = (workers["me"], workers["bob"], workers["alice"])
     x = torch.tensor(7).share(bob, alice, me, protocol="falcon")
     assert x.players[0] == me
+
+
+def test_fixed_precision_and_sharing(workers):
```
abogaziah, line 185: fixed pointed arithmetic is not test thought it's the purpose of that issue, numbers after decimal point should

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -306,3 +309,7 @@ def __str__(self):
             for v in self.child.values():
                 out += "\n\t-> " + str(v)
         return out
+
+
+### Register the tensor with hook_args.py ###
```
abogaziah, line 314: what is that?

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
```diff
@@ -67,6 +68,8 @@ def reconstruct(self):
         plain_text = self.__map_modular_to_real(plain_text_mod)
         return plain_text
 
+    get = reconstruct
```
abogaziah, line 71: how is that useful?

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -194,6 +194,9 @@ def add(self, _self, other):
             # we swap operators so that we do the same operation as above
             _self, other = other, _self.wrap()
 
+        elif isinstance(other, ReplicatedSharingTensor) and isinstance(_self, torch.Tensor):
```
abogaziah, line 197: this is tight coupling which should be avoided, but why would you swap them anyway?

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -234,6 +237,9 @@ def sub(self, _self, other):
             # we swap operators so that we do the same operation as above
             _self, other = -other, -_self.wrap()
 
+        elif isinstance(other, ReplicatedSharingTensor) and isinstance(_self, torch.Tensor):
```
abogaziah, line 240: why would you mul -1?

syft/frameworks/torch/tensors/interpreters/native.py
aanurraj: this was added temporarily as few of the tests are written using reconstruct(), which fails as it gets called on native.py, in future, it will be good to go with .get() instead of reconstruct()

syft/frameworks/torch/tensors/interpreters/native.py
aanurraj: Sorry, fixed in latest commit

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
aanurraj: its been added so that the  _hook_method_args of overload_method can know Put back SyftTensor on the tensors found in the response. if this was not used the output of add or sub would have (Wrapper)>[ReplicatedSharingTensor] instead of (Wrapper)>FixedPrecisionTensor>[RaplicatedSharingTensor] after the operation.

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
aanurraj: a.reconstruct() <-- this is a wrapper
and in native we do not have the reconstruct method
so it fails, thus as I mentioned earlier we should use .get() universally in future.

syft/frameworks/torch/tensors/interpreters/precision.py
aanurraj: because putting "-" was giving an error (i can't remember the error, for now, I'll tell you soon though.

syft/frameworks/torch/tensors/interpreters/precision.py
aanurraj: the reason can be found in the above code for AST, the swap has been done for the same reason. 
now about

syft/frameworks/torch/tensors/interpreters/native.py
abogaziah: I disagree, you "get" a remote tensor but you "reconstruct" a shared tensor, calling everything get is the easy way, not the correct one, 
can you explain further? give an example of a test case that fails?

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
abogaziah: this is not the right place, it breaks the level of abstraction torch hook is a lower-level abstraction so RST should not know it exists.

syft/frameworks/torch/tensors/interpreters/native.py
aanurraj: Screenshot 2020-08-30 at 12 17 27 AM

if that is not included  this error will prompt up with "reconstruct"

test/torch/tensors/test_replicated_shared.py
aanurraj: fixed

syft/frameworks/torch/tensors/interpreters/precision.py
aanurraj: Screenshot 2020-08-30 at 12 31 19 AM
here is the error!

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
aanurraj: @gmuraru can you help me this issue?

gmuraru: Yep - this is needed because when we add the wrapper we check what tensors exists - if we do not have this here it would not know how to create the ```(Wrapper)>RST```

gmuraru: A better approach to this (for all syft framework) it should be to add something "invasive" like:
```
1. get all classes that inherit from AbstractTensor
2. call this ```default_register_tensor``` on all of them
```
But I think another issue might be in order to do this -- doing one now.

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
gmuraru: Hmm...thinking of this. Could we add a reconstruct method in "native" and have a check -- we can do the reconstruct only if we have ```AdditiveSharingTensor``` or ```ReplicatedSharingTensor```

syft/frameworks/torch/tensors/interpreters/native.py
aanurraj: fixed in the latest commit!

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
aanurraj: fixed in latest commit!

syft/frameworks/torch/tensors/interpreters/replicated_shared.py
aanurraj: we are doing this in other issue, right? @gmuraru @abogaziah

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -929,17 +929,23 @@ def share(
...
-                    protocol=protocol,
-                    field=field,
-                    dtype=dtype,
-                    crypto_provider=crypto_provider,
-                    owner=self.owner,
-                )
-                .on(self.copy(), wrap=False)
-                .share_secret(*owners)
-            )
+            else:
```
gmuraru, line 932: I think you can remove the else here - if it enters the throw

aanurraj: done :)

abogaziah:
update black and run it on your fork, or sync it with upstream/master, changed files currently 69, it should be around 5 when you do that successfully

PR #4506, opened 2020-08-31
Add protobuf serde for RunFunctionOrConstructorAction

Jasopaum:
## Description
Add protobuf serde methods on RunFunctionOrConstructorAction.
Without it we can't do things as `client.torch.zeros_like(ptr)`.

We still can't serialize Python types on the main, that's why I added `torch.zeros_like` which uses a tensor arg to the allowlist. We can wait for Tudor's work on Python types to merge, though.

## Checklist


src/syft/core/node/common/action/function_or_constructor_action.py
```diff
@@ -33,29 +41,144 @@ def __init__(
...
         for arg in self.args:
-            if isinstance(arg, Pointer):
-                # QUESTION: Where is get_object definied?
-                r_arg = node.store.get_object(id=arg.id_at_location)
-                resolved_args.append(r_arg)
-            else:
-                resolved_args.append(arg)
+            if not isinstance(arg, Pointer):
+                raise ValueError(
+                    f"args attribute of RunFunctionOrConstructorAction should only contain Pointers. "
```
iamtrask, line 68: Is this true? We will be able to pass in tuples no? (aka - not pointers to tuples)

src/syft/core/node/common/action/function_or_constructor_action.py
```diff
@@ -33,29 +41,144 @@ def __init__(
...
         resolved_kwargs = {}
         for arg_name, arg in self.kwargs.items():
-            if isinstance(arg, Pointer):
-                r_arg = node.store.get_object(id=arg.id_at_location)
-                resolved_kwargs[arg_name] = r_arg
-            else:
-                resolved_kwargs[arg_name] = arg
+            if not isinstance(arg, Pointer):
+                raise ValueError(
+                    f"kwargs attribute of RunFunctionOrConstructorAction should only contain Pointers. "
```
iamtrask, line 82: (see above)

src/syft/core/node/common/action/function_or_constructor_action.py
```diff
@@ -33,29 +41,144 @@ def __init__(
...
+            result_read_permissions = self.intersect_keys(
+                result_read_permissions, r_arg.read_permissions
+            )
+            resolved_kwargs[arg_name] = r_arg.data
 
         result = method(*resolved_args, **resolved_kwargs)
 
+        if isprimitive(value=result):
+            # Wrap in a PyPrimitive
+            result = PyPrimitive(data=result, id=self.id_at_location)
```
iamtrask, line 96: Would we not have to wrap the actual primitive type instead of the parent class?

iamtrask: Maybe it would be a good idea to add a convenience function to primitive types to make this easier? Or a utility function?

iamtrask: CC: @tudorcebere

src/syft/core/node/common/action/function_or_constructor_action.py
```diff
@@ -33,29 +41,144 @@ def __init__(
...
+        else:
+            # TODO: overload all methods to incorporate this automatically
+            if hasattr(result, "id"):
+                result.id = self.id_at_location
+
         if not isinstance(result, StorableObject):
-            result = StorableObject(id=self.id_at_location, data=result)
+            result = StorableObject(
```
iamtrask, line 103: Nice!

src/syft/core/node/common/action/run_class_method_action.py
```diff
@@ -42,19 +42,17 @@ def __init__(
...
         # if the request_id is different for some reason we still want to keep it,
         # so only intersect the keys and then copy those over from the main dict
         # into a new one
         intersection = set(left.keys()).intersection(right.keys())
-        intersection_dict = {}
-        for k in intersection:
-            intersection_dict[k] = left[k]  # left and right have the same keys
-
-        return intersection_dict
+        # left and right have the same keys
```
iamtrask, line 54: Nice!

iamtrask (changes requested):
Overall looking great! Requested a couple small changes :)

src/syft/core/node/common/action/function_or_constructor_action.py
Jasopaum: That's correct, I hastily replicated what was done in `RunClassMethodAction`
Do we also want "primary" types as int, etc. to do stuff as `t.repeat(2)` without having to send the `2`?

Jasopaum: What case do you think I should implement for now? Only the tuple/list argument?

src/syft/core/node/common/action/function_or_constructor_action.py
Jasopaum: Yep, I think this will change with Tudor's work!
This is based on Madhava's primitive type handling

PR #4510, opened 2020-08-31
Add flag to set whether a tensor is publicly searchable - my_tensor.send(client, searchable=True) 

iamtrask:
## Description
In this PR, I added the ability to set whether or not you want a tensor you send to a remote machine to be searchable/discoverable by other workers (note that searchability is decoupled from the ability to download. It only entails the ability to see the metadata of a tensor)

## Affected Dependencies
None

## How has this been tested?
I added one unit test which uses this functionality in the context of Duet.

## Checklist


tests/syft/grid/duet_test.py
```diff
@@ -1,60 +1,39 @@
...
-
-#     assert xp.id_at_location == x.id
-
-#     yp = xp + xp
-
-#     y = yp.get()
-#     assert ((x + x) == y).all()
-
-#     duet.stop()
+    duet.stop()
```
Jasopaum, line 39: nit: Maybe we could use a pytest feature (I don't remember how it's called right now) to instantiate and clean the Duet for each test (this can be done later, of course!)

src/syft/core/node/common/service/obj_search_service.py
```diff
@@ -194,9 +195,18 @@ def process(
         results: List[Pointer] = list()
 
         for obj in node.store.get_objects_of_type(obj_type=object):
+
+            # if this tensor allows anyone to search for it, then one of its keys
+            # has an All() class in it.
+            contains_all_in_permissions = False
+            for key in obj.search_permissions.keys():
+                if isinstance(key, All):
```
Jasopaum, line 203: nit: I *think* `if any(isinstance(key, All) for key in obj.search_permissions)` is a bit clearer

Jasopaum (approved):
Do you have any plans about how we'll implement more complex search policies if we need to?

Also, I see that protobuf files are modified, do you have the latest protoc version?

Otherwise, LGTM 🚀

tests/syft/grid/duet_test.py
iamtrask: Agreed - we only have 2 duet tests for now so i think we can cleanup later. But if you know how to do this that'd be awesome.

src/syft/core/node/common/service/obj_search_service.py
iamtrask: I'l make this adjustment!

PR #4512, opened 2020-09-01
Added isort config to setup.cfg

madhavajay:
## Description
- Added isort to scripts/pre_commit.sh
- Manually checked every file and cleaned up old import comments

You can now automatically fix all imports with 1 command:
```
$ isort .
```

This is also now part of the scripts/pre_commit.sh script so running this will sort all the imports and then use black to format so you shouldn't need to manually take care of anything.

## Affected Dependencies
None

## How has this been tested?
Local Tests and CI

## Checklist


docs/conf.py
```diff
@@ -45,8 +48,9 @@
     pass
 
 try:
-    import sphinx
+    # third party
```
Jasopaum, line 51: Just out of curiosity, are these comments necessary for isort to work?

src/syft/proto/grid/service/signaling_service_pb2.py
```diff
@@ -2,6 +2,7 @@
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
 # source: proto/grid/service/signaling_service.proto
 """Generated protocol buffer code."""
+# third party
```
Jasopaum, line 5: I don't know if it's a big deal or not but maybe we could avoid formatting generated files?

Jasopaum (approved):
Didn't spot anything wrong 👍

iamtrask (approved):

docs/conf.py
madhavajay: Its not necessary to add comment headings, its optional, but in the rare situations where we do conditional imports there can be some strange issues, but generally it still marks them.

src/syft/proto/grid/service/signaling_service_pb2.py
madhavajay: We are already modifying these files to change the pb2 imports, by formatting them with the same formatter we use on all our other python code we avoid the issue of having to add "blocklist" files to our formatting systems which need to be maintained. Since we have the formatting tool and rules defined as part of the pre-commit hooks and CI these should always be the same for everyone so there will be 0 noise.

Happy to look at removing it but this actually seemed easier and without consequence.

PR #4521, opened 2020-09-02
Alternative Syft Primitive Int implementation with metaclass

madhavajay:
## Description
- Alternative Syft Primitive Int implementation with metaclass
- All the mypy and linting fixes needed to pass pre-commit / CI
- existing failing float test set to @pytest.mark.xfail

So I wanted to play around and see what boiler plate could be reduced using a metaclass.
The code is in int.py and is called SyInt instead of Int.
You can test it easily by doing:

```
# int_test.py
# from syft.lib.python.int import Int
from syft.lib.python.int import SyInt as Int
```

It passes all the same tests.

Its also going to be easy to delete as it doesnt change any of the existing code yet.

## Pros
- overridden dunder methods can be listed and all applied in one go with Sy Primitive return types and persistent ID's.
- easy to debug what dunder methods are called see commented out print statements
- a lot of this can probably be generalised between multiple different primitive types by getting the shadow_type from the bases or the type of type(_value) on the concrete primitive class
- no need for the primitive factory

## Cons
- super() isnt bound to the instance so the run_ops code is a bit more verbose to get the same effect as:
```
# Int
res = super().__add__(other)
return PrimitiveFactory.generate_primitive(value=res)
```
vs
```
# SyInt
@staticmethod
    def make_operand_method(op: str, symbol: str) -> Callable:
        def run_operand(a: "SyInt", b: Any) -> Union["SyInt", bool]:
            ...
```
But you only have 1 method and a list of _dunder properties which is a lot less code to test and read.

I didn't make the comparison functions return Bools with IDs because I wasn't sure what we wanted here, but i can see that if all operations and their return values are network agnostic then wrapped primitive types probably need to nearly always return more wrapped primitive types so they always have IDs to refer to. Thats any easy change and just requires removing the bool stuff from the signature and this code:
```
# these methods just want a boolean comparison result
if op in ["__gt__", "__ge__", "__lt__", "__le__"]:
    return bool(result)
```

## Affected Dependencies
None

## How has this been tested?
Locally / CI

## Checklist


madhavajay:
This experiment was with metaclass for the PyPrimitives instead of pure subclass with possible merits. We can look back at this later if we decide to investigate further.

PR #4529, opened 2020-09-03
Implement a Syft List class

Jasopaum:
## Description
Implement a Syft List class

## Affected Dependencies
None

## How has this been tested?
I will add unit tests for this class.

## Checklist


madhavajay:
Great work. I played around with this locally and it works great. It's cool that Python gives us UserList to do this in such a clean way.

I fixed a few mypy / linting issues I saw after running scripts/pre_commit.sh here if you want to merge those in:
https://github.com/OpenMined/PySyft/tree/python-lists-linting

The only thing I started thinking was, will there be any naming clash issues with the existing:
```
from typing import List
```

Perhaps we should prefix our Collections and Primitives so they are immediately clear rather than causing possible confusion about where they come from. I do prefer the shorter written version without a prefix but I can also see how easily someone could get mixed up with the native types if they see these:

```python
Int()
List()
Bool()

vs 

SyInt()
SyList()
SyBool()
```

Jasopaum:
> Great work. I played around with this locally and it works great. It's cool that Python gives us UserList to do this in such a clean way.
> 
> I fixed a few mypy / linting issues I saw after running scripts/pre_commit.sh here if you want to merge those in:
> https://github.com/OpenMined/PySyft/tree/python-lists-linting
> 
> The only thing I started thinking was, will there be any naming clash issues with the existing:
> 
> ```
> from typing import List
> ```
> 
> Perhaps we should prefix our Collections and Primitives so they are immediately clear rather than causing possible confusion about where they come from. I do prefer the shorter written version without a prefix but I can also see how easily someone could get mixed up with the native types if they see these:
> 
> ```python
> Int()
> List()
> Bool()
> 
> vs 
> 
> SyInt()
> SyList()
> SyBool()
> ```

Thanks, I'll take your commit 🙂 

As for the naming, I myself had this syft.List / typing.List clash... So I do agree that we could prefix our types with "Sy". Maybe after all the current PRs on primitive types are merged?

tudorcebere (approved):
Really nice work!

PR #4535, opened 2020-09-06
Use register for default types

gmuraru:
## Description
Use a mechanism to register the default types (only a POC - can not be merged).
Fix some warnings regarding tests

## Affected Dependencies
Register primitives using a factory method (this should remove circular dependencies)

## How has this been tested?
- Tests should be work

## Checklist


madhavajay:
Is this to help automatically boxing python primitives into SyPrimitive?

iamtrask (approved):
I like the style in which this is done

src/syft/lib/python/primitive_factory.py
```diff
@@ -24,44 +26,74 @@ def isprimitive(value: Any) -> bool:
...
+    @syft_decorator(typechecking=True)
+    def register_primitive(
+        python_primitive: Union[
+            Type[int],
+            Type[float],
+            Type[bool],
+            Type[complex],
+            Type[list],
+            Type[str],
+            None,
```
madhavajay, line 85: Looks like we can use Type[None] to get typing.Type[NoneType]

madhavajay: Actually, I remember trying to resolve this and just got into a loop of mypy errors that make no sense, so feel free to leave it as is.

madhavajay:
@gmuraru I think maybe we can do this type of thing as discussed on Slack:
https://github.com/OpenMined/PySyft/pull/4637

Then maybe we can just close this for now and re-discuss in future if theres a desire to go down this path again?

PR #4547, opened 2020-09-10
Added property support for getset_descriptor properties of Callable

madhavajay:
## Description
Added property support for getset_descriptor properties of Callable

## Affected Dependencies
None

## How has this been tested?
Locally and CI

## Checklist


madhavajay:
You can now do:
```python
tpointer = tensor.send(alice_client)

tpointerTM = tpointer.t()
tpointerTP = tpointer.T

tm = tpointerTM.get()
tp = tpointerTP.get()

assert (tm == tp).all()
```

src/syft/ast/callable.py
```diff
@@ -76,9 +75,14 @@ def add_path(
                 if isinstance(attr_ref, module_type):
                     raise Exception("Module cannot be attr of callable.")
                 else:
+                    is_property = False
+                    if type(attr_ref).__name__ == "getset_descriptor":
+                        is_property = True
```
Jasopaum, line 80: nit: what about `is_property = type(attr_ref).__name__ == "getset_descriptor"`

Jasopaum: Not sure it's that easy to  read though so as you wish!

Jasopaum (commented):
Looks clean 👍

proto/lib/python/none.proto
```diff
@@ -5,5 +5,5 @@ import "proto/core/common/common_object.proto";
 
 
 message None {
-    syft.core.common.UID id_at_location = 2;
+    syft.core.common.UID id = 2;
```
iamtrask, line 8: wow good call - not sure when this got changed

src/syft/ast/klass.py
```diff
@@ -91,8 +98,23 @@ def run_class_method(
...
+        def ga(__self: Any, name: str) -> Any:
+            # we need to override the __getattribute__ of our Pointer class
+            # so that if you ever access a property on a Pointer it will not just
+            # get the wrapped run_class_method but also execute it immediately
+            # object.__getattribute__ is the way we prevent infinite recursion
+            attr = object.__getattribute__(__self, name)
+
+            # if the attr key name is in the props list from above then we know
+            # we should execute it immediately and return the result
+            if name in props:
```
iamtrask, line 110: where is "props" coming from? Surely this should be on __self?

iamtrask: I fear that using a global-ish variable here might cause some issues where multiple pointers (or in this case pointer types) end up accidentally sharing the same list of properties.

iamtrask (changes requested):

src/syft/ast/klass.py
```diff
@@ -91,8 +98,23 @@ def run_class_method(
             if attr_path_and_name is not None:
                 attrs[attr_name] = get_run_class_method(attr_path_and_name)
 
+        def ga(__self: Any, name: str) -> Any:
```
gmuraru, line 101: Q: Can we use another name for this?

proto/lib/python/none.proto
gmuraru: Q: Can't we use "1" here?

src/syft/ast/klass.py
```diff
@@ -91,8 +98,23 @@ def run_class_method(
             if attr_path_and_name is not None:
```
gmuraru, line 98: NITPICK: If attr_path_and name (without is not None) (I know this PR has nothing to do with this :D)

tests/syft/lib/torch/tensor/tensor_remote_method_api_suite_test.py
```diff
@@ -217,15 +218,22 @@ def test_all_allowlisted_tensor_methods_work_remotely_on_all_types(
     # Step 5: Test to see whether this method and arguments combination is valid
     # in normal PyTorch. If it this is an invalid combination, abort the test
     try:
+        is_property = False
         # if this is a valid method for this type in torch
-        valid_torch_command = True
-        if args is not None:
-            target_result = target_op_method(*args)
+        if type(target_op_method).__name__ in ["builtin_function_or_method", "method"]:
+            valid_torch_command = True
```
gmuraru, line 224: Q: Can't we move this outside ```if```. In this case we will remove the ```valid_torch_command = True``` in else

src/syft/ast/klass.py
madhavajay: Yes you are right this looks a bit sketchy and should be tested better, but to be on the safe side ill move it to be implicitly stored on the new class.

src/syft/ast/klass.py
madhavajay: Good idea, fixed.

tests/syft/lib/torch/tensor/tensor_remote_method_api_suite_test.py
```diff
@@ -254,10 +262,14 @@ def test_all_allowlisted_tensor_methods_work_remotely_on_all_types(
         assert op_method
 
         # Step 10: Execute the method remotely
-        if argsp is not None:
-            result = op_method(*argsp)
+        if not is_property:
```
gmuraru, line 265: NITPICK: Can we switch does?
```
if is_property:
    result = op_method
elif argsp is not None:
    result = op_methods(*argsp)
else:
    result = op_method()
```

src/syft/ast/klass.py
madhavajay: I have previously commented this code with:
# QUESTION: Could path_and_name be None?

But i'm not sure if thats what you are referring to?

tests/syft/lib/torch/nn/parameter/parameter_remote_method_api_suite_test.py
```diff
@@ -264,10 +272,14 @@ def test_all_allowlisted_parameter_methods_work_remotely_on_all_types(
         assert op_method
 
         # Step 10: Execute the method remotely
-        if argsp is not None:
-            result = op_method(*argsp)
+        if not is_property:
```
gmuraru, line 275: NITPICK: Also here the same reordering

tests/syft/lib/torch/nn/parameter/parameter_remote_method_api_suite_test.py
```diff
@@ -225,15 +226,22 @@ def test_all_allowlisted_parameter_methods_work_remotely_on_all_types(
...
     # in normal PyTorch. If it this is an invalid combination, abort the test
     try:
+        is_property = False
         # if this is a valid method for this type in torch
-        valid_torch_command = True
-        if args is not None:
-            target_result = target_op_method(*args)
+        if type(target_op_method).__name__ in ["builtin_function_or_method", "method"]:
+            valid_torch_command = True
+            if args is not None:
```
gmuraru, line 233: NITPICK: if args

src/syft/ast/klass.py
gmuraru: Ah...I am referring to:
if attr_path_and_name <-- returns false if attr_path_and_name is None, "" (empty string), [] (empty list) and 0.
I am ok to leave it like it is now, I was just asking if it would make sense to make it shorter.

gmuraru: **LE**: Thinking more into this. I think we should be precise on what we check against -- we need to *clearly* check it against None

tests/syft/lib/torch/tensor/tensor_remote_method_api_suite_test.py
madhavajay: Good spot! Thanks.

tests/syft/lib/torch/tensor/tensor_remote_method_api_suite_test.py
madhavajay: Yes thats much better. Fixed.

src/syft/ast/klass.py
iamtrask: Good call!

src/syft/ast/klass.py
madhavajay: This was a good catch since it prompted me to look at the code closer and remove a lot of unnecessary branching and checks since args will always be a list / tuple even if its just an empty one.

src/syft/lib/torch/allowlist.py
```diff
@@ -614,19 +620,18 @@
...
 
 # SECTION - Parameter methods
-# allowlist["torch.nn.Parameter"] = "torch.nn.Parameter"
-# allowlist["torch.nn.parameter.Parameter"] = "torch.nn.parameter.Parameter"
-# allowlist["torch.nn.Parameter.t"] = "torch.nn.Parameter"
-# allowlist["torch.nn.Parameter.__abs__"] = "torch.nn.Parameter"
-# allowlist["torch.nn.Parameter.__add__"] = "torch.Tensor"
-# allowlist["torch.nn.Parameter.__and__"] = "torch.nn.Parameter"
+# torch.nn.Parameter is a subclass of torch.Tensor
+# we need the main contstructor Class but everything else is
```
gmuraru, line 627: TYPO: ```contstructor``` -> ```constructor```

tests/syft/lib/torch/nn/parameter/parameter_remote_method_api_suite_test.py
```diff
@@ -190,28 +196,33 @@ def test_all_allowlisted_parameter_methods_work_remotely_on_all_types(
...
     # Step 4: Create the arguments we're going to pass the method on
     if _args == "None":
-        args = None
+        args = []
     elif _args == "self":
-        args = [th.nn.Parameter(self_tensor)]
+        args = [self_tensor]
     elif _args == "0":
-        args = 0
+        args = [PrimitiveFactory.generate_primitive(value=0)]
```
gmuraru, line 207: Q: Shouldn't we also be able to pass Python values?

madhavajay: We need to discuss the best way to handle this, either inject .serialize onto the primitives or detect them at the call site and wrap them where .serialize normally happens.

Until then, we need this to be able to test properties which send back other types than Tensor. As soon as we have the desired changes elsewhere we can just remove these from the test suite.

iamtrask (approved):

PR #4565, opened 2020-09-17
Added torch.device and CWrapper in generic

madhavajay:
## Description
- Allows inheriting from types which cannot be inherited from
- Added upcast to SyPrimitives

## Updates
- Fixed issues with CWrapper and id property
- Fixed test failing due to new methods on PyPrimitive
- Fixed PyPrimitive return types in allowlist
- Fixed issue with print / repr(torch.nn.Parameter) throwing exception
- Fixed main return_type_name being squashed by methods return type
- Note: __id and __getattribute__ causes a Class property lookup
  which then fails. The solution is not to use __id but _id

## Affected Dependencies

## How has this been tested?
Locally, WIP failing test

## Checklist


src/syft/core/node/common/action/function_or_constructor_action.py
```diff
@@ -104,7 +104,24 @@ def execute_action(self, node: AbstractNode, verify_key: VerifyKey) -> None:
             )
             resolved_kwargs[arg_name] = r_arg.data
 
-        result = method(*resolved_args, **resolved_kwargs)
+        # when we call original constructors like torch.device we will get errors
```
iamtrask, line 107: looks great!

also - didn't know this was called "upcast"ing - im a fan

src/syft/lib/generic.py
```diff
@@ -16,6 +16,19 @@
 from .util import replace_classes_in_module
 
 
+class CWrapper:
```
iamtrask, line 19: need more documentation on this class - please fully explain how it works

src/syft/lib/python/bool.py
```diff
@@ -49,6 +49,10 @@ def id(self) -> UID:
         """
         return self._id
 
+    @syft_decorator(typechecking=True, prohibit_args=True)
```
iamtrask, line 52: really go crazy on documentation for this one - it's a complex/abstract feature

iamtrask: (same for the other ones - can probably be mostly copy psate)

iamtrask (changes requested):
most docs

src/syft/core/node/common/action/function_or_constructor_action.py
madhavajay: Yeah I had to google it to double check... I was like... its not down casting... so whats the opposite?
There is probably a better mathy name for it somewhere like something to do with "contravariance" or something.

src/syft/lib/generic.py
madhavajay: Yup I will document next, still getting it to work unfortunately, hit some weird issues with property / method pointer return types and getattr on __id on the CWrapper.

madhavajay:
The generic CWrapper / ShadowWrapper approach is no longer being used so this PR can be closed for now.

PR #4581, opened 2020-09-22
Run isort and black on all files and added coverage package.

shubham3121:
## Description

- Run isort and black on all files.
- Fix pytest run command in scripts/pre_commit.sh
- Added coverage package to have a detailed view of the test coverage in html format. 
(Refer to [package docs](https://coverage.readthedocs.io/en/coverage-5.3/) and [sample coverage report](https://nedbatchelder.com/files/sample_coverage_html/index.html) for reference.)

## Affected Dependencies
Added coverage package.

## How has this been tested?
- Describe the tests that you ran to verify your changes. - Ran existing tests. 
- Provide instructions so we can reproduce. pytest -m fast --cov=src/ --cov-report html

## Checklist


tests/syft/lib/python/string/string_utils_test.py
```diff
@@ -1343,7 +1343,7 @@ def test_formatting_c_limits(self):
     def test_floatformatting(self):
         # float formatting
         for prec in range(100):
-            format = "%%.%if" % prec
```
madhavajay, line 1346: I think this is there to test the % __mod__ infix operator for string formatting which is still allowed and is valid in Python 3. These tests are actually from the Python code base so we try to only change superficial aspects where possible to ensure we dont accidentally miss something.

scripts/pre_commit.sh
```diff
@@ -8,8 +8,8 @@ isort .
 black src tests
 pre-commit run --all-files
 bandit -r src -ll
-pytest -m fast
+pytest -m fast --cov=src/ --cov-report html
```
madhavajay, line 11: Is it possible we can combine this with .coveragerc file we have which omits the protos?

requirements.txt
```diff
@@ -18,6 +18,7 @@
 aiortc
 bandit
 black
+coverage
```
madhavajay, line 21: Is this is already a dependency of the pytest-cov plugin below?

scripts/pre_commit.sh
madhavajay: Or possibly the settings in setup.cfg for pytest-cov, and then we can add the htmlcov dir to .gitignore?
```
[tool:pytest]
# Options for py.test:
# Specify command line options as you would do when invoking py.test directly.
# e.g. --cov-report html (or xml) for html/xml output or --junitxml junit.xml
# in order to write a coverage file that can be read by Jenkins.
# TODO (gmuraru) change cov-fail-under as soon as possible
addopts =
    --cov syft --cov-report term-missing --cov-fail-under 82 --no-cov-on-fail
    --verbose
norecursedirs =
    dist
    build
    .tox
testpaths = tests
```

tests/syft/lib/python/string/string_utils_test.py
shubham3121: @madhavajay the change is also a valid syntax and  was the result of running black on the files. The change won't break anything for Python 3.

madhavajay: We are trying not to change the tests from the official python repo as much as possible which means ignoring formatting / linting issues with the copied tests since the originals have these issues in them already. In this case the % mod infix operator is also being tested so it might be best to leave this test as is.

madhavajay:
I will close this PR for the time being as the majority of the changes are either integrated already or not required.

PR #4598, opened 2020-09-25
Fix get permissions

gmuraru:
## Description
In case we create a new object remotely, we set no permission for it.
This PR adds a list of default read permissions in case there is no permissions added (there is not args/kwargs).

The list might be extended.

## Affected Dependencies
Currently, we use only ```torch.cuda.is_available``` (for showcase).
It might impose a **security risk**. Cuda machine vs non-cuda machine, it might mean that a machine has more data?

## How has this been tested?
* Added tests

## Checklist

### DIFF
--- a/src/syft/ast/globals.py
+++ b/src/syft/ast/globals.py
@@ -18,13 +18,6 @@ class Globals(Module):
     def __init__(self) -> None:
         super().__init__("globals")
 
-    def add_framework(
-        self,
-        attr_name: str,
-        attr: Union[Callable, CallableT],
-    ) -> None:
-        self.attrs[attr_name] = attr
-
     def add_path(
         self,
         path: Union[str, List[str]],
--- a/src/syft/core/node/common/action/function_or_constructor_action.py
+++ b/src/syft/core/node/common/action/function_or_constructor_action.py
@@ -3,7 +3,6 @@ from typing import Any
 from typing import Dict
 from typing import List
 from typing import Optional
-from typing import Set
 from typing import Tuple
 from typing import Union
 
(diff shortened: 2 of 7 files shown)


src/syft/core/node/common/action/function_or_constructor_action.py
```diff
@@ -75,7 +78,7 @@ def intersect_keys(
     def execute_action(self, node: AbstractNode, verify_key: VerifyKey) -> None:
         method = node.lib_ast(self.path)
 
-        result_read_permissions = None
+        result_read_permissions: Union[None, Dict[VerifyKey, UID]] = None
```
iamtrask, line 81: ok - so this is a list of things which we always allow people to execute?

src/syft/core/node/common/action/function_or_constructor_action.py
```diff
@@ -24,6 +25,8 @@
 from ...abstract.node import AbstractNode
 from .common import ImmediateActionWithoutReply
 
+DEFAULT_AVAILABLE_READ_PERMISSIONS: Set[str] = {"torch.cuda.is_available"}
```
iamtrask, line 28: HARD BLOCK: nothing torch specific should be in src/syft/core path.

iamtrask (changes requested):

src/syft/ast/globals.py
```diff
@@ -18,13 +18,6 @@ class Globals(Module):
     def __init__(self) -> None:
         super().__init__("globals")
 
-    def add_framework(
```
gmuraru, line 21: Q: Removed this.

I did not find it used for anything.

src/syft/lib/python/__init__.py
```diff
@@ -103,3 +103,15 @@ def create_python_ast() -> Globals:
         klass.create_storable_object_attr_convenience_methods()
 
     return ast
+
+
+DEFAULT_READ_PERMISSION: TypeList[str] = [
```
gmuraru, line 108: Q: Should I leave only ```SyNone```?
IS the only one that doesn't take any argument

tudorcebere (approved):

iamtrask (approved):

PR #4619, opened 2020-09-29
Remote execution of torch hooks

LaRiffle:
## Description
This provides a mechanism to set hook functions (during pytorch backprop) on remote tensors and models.

This allows to register torch hooks on remote tensors. Such operation
        is tricky because you can't really send the hook function to a remote
        party, as python functions are not serializable within PySyft. So you
        need to keep it attached to the PointerTensor.
        On the other hand, the PointerTensor cannot watch for gradient update
        or be triggered natively by torch when backpropagation happens. That's
        why we actually remotely set a hook that we call a callback hook whose
        function is only to call back the pointer during the backpropagation
        to effectively run the hook function.
        So the workflow is: the remote hook is triggered by pytorch, a message
        is sent back to the pointer owner which has the hook function, then
        the hook function is run remotely on the remote gradient, and a termi-
        nation message is returned to the gradient owner.

## How has this been tested?

## Checklist

### DIFF
--- a/syft/frameworks/torch/fl/dataset.py
+++ b/syft/frameworks/torch/fl/dataset.py
@@ -270,7 +270,7 @@ def dataset_federate(dataset, *workers, **kwargs) -> "FederatedDataset":
             to the workers
     """
     # Compat with the old signature without the (*)workers
-    if isinstance(workers[0], tuple):
+    if isinstance(workers[0], (list, tuple)):
         workers = workers[0]
 
     logger.info(f"Scanning and sending data to {', '.join([w.id for w in workers])}...")
--- a/syft/frameworks/torch/fl/dataset.py
+++ b/syft/frameworks/torch/fl/dataset.py
@@ -257,7 +257,7 @@ class BaseDataset(AbstractSendable):
         return BaseDatasetPB
 
 
-def dataset_federate(dataset, workers):
+def dataset_federate(dataset, *workers, **kwargs):
     """
     Add a method to easily transform a torch.Dataset or a sy.BaseDataset
     into a sy.FederatedDataset. The dataset given is split in len(workers)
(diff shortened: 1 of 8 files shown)


syft/frameworks/torch/fl/dataset.py
```diff
@@ -257,12 +257,22 @@ def get_protobuf_schema():
...
     part and sent to each workers
+
+    Args:
+        dataset (Dataset): the dataset to federate across workers
+        *workers (BaseWorker): the workers receiving parts of the dataset
+        **kwargs (dict): any kwargs to use when sending parts of the dataset
+            to the workers
     """
+    # Compat with the old signature without the (*)workers
+    if isinstance(workers[0], tuple):
```
gmuraru, line 273: Q: Can this be a list?

syft/frameworks/torch/tensors/interpreters/native.py
```diff
@@ -89,6 +89,50 @@ def trigger_origin_backward(grad):
...
+        """
+        Register a Torch hook (that is triggered when the `self` tensor has a gradient
+        update) which will send back to a location a specific message whose arguments
+        are updated with a reference to `self`.
+
+        Args:
+            message (Message): the message to send back
+            location (BaseWorker): the worker to which the message should be sent
+        """
+        location = self.owner.get_worker(location)
```
gmuraru, line 102: Q: Could you add tests for those?

syft/frameworks/torch/fl/dataset.py
```diff
@@ -315,6 +325,8 @@ def workers(self):
 
         return list(self.datasets.keys())
 
+    locations = workers
```
gmuraru, line 328: Q: Is this used?

syft/frameworks/torch/fl/dataset.py
LaRiffle: Yup! Will change

syft/frameworks/torch/fl/dataset.py
LaRiffle: It's an alias which is more intuitive, we always have .location(s), except for this one

syft/frameworks/torch/tensors/interpreters/native.py
LaRiffle: Yeah... :)

LaRiffle: @gmuraru done :)

gmuraru (approved):

Issue #3883, opened 2020-07-20
Make sure all tests on AST are performed with all crypto protocols


LaRiffle:
## Description
We want to make sure all tests on AdditiveSharingTensors are performed not only on secureNN but also on FSS crypto protocol and future ones like FALCON

## Infos
Blocked by #3882 
Open to contributions!

ramesht007:
Hey, @LaRiffle can I work on this issue

LaRiffle:
Hey sure yes!
But some elements are required before addressing this issue: I need to merge stuff related to function secret sharing to have it behave in the exact same way than secureNN.
I will let you know when this issue is unblocked!

ramesht007:
ok sure 😀

LaRiffle:
Hey @ramesht007 !
It's ready to go! :)

ramesht007:
ok 👍

gmuraru:
@ramesht007 are you still working on this? :D

ramesht007:
Hey, @gmuraru yup I'm still working on this, but will not be able to give any update as I have my exams till 10th September. After that I will start working on it with full force.🙂

aanurraj:
work is done! thanks, @ramesht007 closing the issue

PR #4646, opened 2020-10-13
Added tests for AST using all crypto protocols

ramesht007:
## Description
Added tests for AST with crypto protocols.
issue #3883


test/torch/tensors/test_additive_shared.py
```diff
@@ -474,6 +474,7 @@ def test_public_mul(workers, protocol):
     # assert (z == (t * y)).all()
 
 
+@pytest.mark.parametrize("protocol", ["snn", "fss"])
 def test_div(workers):
```
LaRiffle, line 478: You should do slightly more change:
```suggestion

@pytest.mark.parametrize("protocol", ["snn", "fss"])
def test_div(workers, protocol):
```
And then when you share, add the `protocol` keyword:
```
x = t.fix_prec(dtype="long").share(bob, alice, crypto_provider=james, protocol=protocol)
```
Do this for all the functions you have modified and it will work!

LaRiffle (changes requested):
Great contibution!
But you need to make some more changes :)

test/torch/tensors/test_additive_shared.py
ramesht007: 👍

ramesht007:
Hi, i have done the changes 
Please have a look

LaRiffle (approved):
🙌

PR #4648, opened 2020-10-13
Reference count on the store that keeps the obj

gmuraru:
## Description
Add a reference count on the store side where we are keeping the objects.
In case 2 have 2 clients and they both refer to the same object, deleting a pointer should not delete the object on the server.

## How has this been tested?
- Re-enabled a test where the ```GarbageCollectObjectAction``` arrived after creating a new object (with the same object).
  - if the GarbageCollectObjectAction arrives before doing ```ptr = x.send(alice)``` then we delete the object from the server and we recreate it when we send it
  - if the GarbageCollectObjectAction arrives after doing ```ptr=x.send(alice0``` then we have 2 reference counts on the same object (even if they are held by the same client) then comes the message and only decrements the reference count.

## Checklist

### DIFF
--- a/src/syft/core/node/common/action/save_object_action.py
+++ b/src/syft/core/node/common/action/save_object_action.py
@@ -37,7 +37,7 @@ class SaveObjectAction(ImmediateActionWithoutReply, Serializable):
 
     def execute_action(self, node: AbstractNode, verify_key: VerifyKey) -> None:
         # save the object to the store
-        obj=StorableObject(
+        obj = StorableObject(
             # ignoring this - the fundamental problem is that we can't force the classes
             # we want to use to subclass from something without creating wrappers for
             # everything which are mandatory for all operations. It's plausible that we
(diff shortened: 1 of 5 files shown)


src/syft/core/node/common/action/save_object_action.py
```diff
@@ -37,37 +37,37 @@ def __init__(
...
-                ),
-                search_permissions={All(): None}
-                if self.anyone_can_search_for_this
-                else {},
-                read_permissions={
-                    node.verify_key: node.id,
-                    verify_key: None,  # we dont have the passed in sender's UID
-                },
-            )
+        obj=StorableObject(
```
gmuraru, line 40: Simply moved this down to be able to do better debugging

src/syft/core/store/store_memory.py
```diff
@@ -63,7 +68,9 @@ def values(self) -> ValuesView[AbstractStorableObject]:
     def store(self, obj: AbstractStorableObject) -> None:
         # TODO: obj should be just "object" and the attributes
         #  of StoreableObject should be put in the metadatastore
-        self._objects[obj.id] = obj
+        with self._lock:
```
gmuraru, line 71: Another idea to do this is to use the ```cython``` extension and dig into the native ```reference count``` for an object.
*PROS*: it might be faster to directly access that code than also keeping a reference count in the application level
*CONS*: We depend on the ```c``` implementation

*Why I choose this*: Since we do not have yet a benchmarking in place, I think that premature optimization might not be our best choice

PR #4649, opened 2020-10-13
Improve allowlist tests

madhavajay:
## Description
The goal of this PR is to make the allowlist tests easier to work with by improving the following:
- Improve speed
- Improve debug / error output
- Improve configurability
- Remove over reliance on exception catching

This is WIP until the allowlist has all current functionality re-enabled and supported by the new test.

## Affected Dependencies
allowlist Tensor and Parameter suite tests

## How has this been tested?
Locally and in CI

## Checklist


tests/syft/lib/allowlist_test.py
```diff
@@ -0,0 +1,495 @@
...
+                # second check if the rule is a list of rules in which case we
+                # need to check for the presence of our current combination value
+                if isinstance(skip_rule[key], Iterable):
+                    # we cant use x not in y because bool matches for 0 and 1 in lists
+                    presence = False
+                    for item in skip_rule[key]:
+                        if combination_dict[key] == item and type(
+                            combination_dict[key]
+                        ) == type(item):
+                            presence = True
```
Jasopaum, line 117:
```
presence = any(
    combination_dict[key] == item and
    type(combination_dict[key]) == type(item)
    for item in skip_rule[key]
)
```
?

tests/syft/lib/allowlist_test.py
```diff
@@ -0,0 +1,495 @@
...
+    # skip this particular test combination
+    return applies
+
+
+BASIC_OPS = list()
+BASIC_OPS_RETURN_TYPE = {}
+# here we are loading up the true allowlist which means that we are only testing what
+# can be used by the end user
+for method, return_type_name_or_dict in allowlist.items():
+    if "torch.Tensor." in method:
```
Jasopaum, line 156: if method.starts_with("torch.Tensor.")?

tests/syft/lib/allowlist_test.py
```diff
@@ -0,0 +1,495 @@
...
+    )
+
+    skipped_combinations = set()
+    for combination in combinations:
+        for skip_rule in skip:
+            if check_skip(
+                combination=combination, skip_rule=skip_rule, lib_version=TORCH_VERSION
+            ):
+                # we use str(combination) so that we can hash the entire combination
+                skipped_combinations.add(str(combination))
```
Jasopaum, line 243: Why not directly doing `TEST_DATA.append(combination)` here if the condition is False?

Jasopaum (commented):
Love the conf file for the tests and the profile thing ❤️

.gitignore
```diff
@@ -20,4 +20,4 @@ docs/_modules
 .coverage
 .vscode/
 **/pip-wheel-metadata
-
+allowlist_test_errors.jsonl
```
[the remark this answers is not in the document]
madhavajay: jsonl is a kind of hybrid format of json and line based log files:
https://jsonlines.org/

The idea is that you can write like a normal log in parallel and read and sort lines which are all still "valid json" inside the line.

I mostly just picked it for the convenience of json.dumps and appending to a file from parallel threads.

tests/syft/lib/allowlist_test.json
```diff
@@ -0,0 +1,311 @@
...
+        "skip": [
+          { "data_types": ["bool"], "max_version": "1.5.1" },
+          { "data_types": ["float16"], "max_version": "1.5.1" },
+          { "min_version": "1.6.0" }
+        ]
+      },
+      "__floordiv__": {
+        "profile": "tensor_float_division_method_v150",
+        "skip": [
+          { "data_types": ["bool"], "max_version": "1.6.0" },
```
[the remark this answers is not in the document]
madhavajay: Yes, i'm still getting confused by these names myself, especially since the condition itself is a kind of negative so it gets into the realm of double negatives etc.

scripts/pre_commit.sh
```diff
@@ -1,15 +1,16 @@
...
-black src tests
-pre-commit run --all-files
-bandit -r src -ll
-pytest -m fast
-
-# run API documentation test notebooks
-./scripts/nb_test.sh
-pytest examples/api --cov-fail-under 0
\ No newline at end of file
+./scripts/build_proto.sh && isort . && black src tests && pre-commit run --all-files &
```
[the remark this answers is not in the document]
madhavajay: Yes, and so far it takes my run time from 30 seconds down to 17, without any issue, but if there are problems we can go back to the old way, I just wanted to try to get that number down so that people are more inclined to run it more frequently.

tests/syft/lib/allowlist_test.py
```diff
@@ -0,0 +1,495 @@
...
+#               self_tensor and None represents no input
+# is_property:  this helps understand if the attribute can be called or not
+# return_type:  this helps verify that the return type matches the expected return type
+
+# this handles instances where the allow list provides more meta information
+def get_return_type(support_dict: Union[str, Dict[str, str]]) -> str:
+    if isinstance(support_dict, str):
+        return support_dict
+    else:
+        return support_dict["return_type"]
```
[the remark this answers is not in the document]
madhavajay: Yes thats still the current format for the actual python file allowlist, although we might change it more once we deal with Union return types. I guess we could convert the whole file to use dict keys instead but then its just adding a lot of unnecessary cruft when there is not version requirements and the return type is a single type.

tests/syft/lib/allowlist_test.py
```diff
@@ -0,0 +1,495 @@
...
+        "inputs": combination[3],
+    }
+
+    # we will assume the skip rule applies and try to prove it doesn't by finding a
+    # condition which we cannot satisfy
+    applies = True
+    for key in skip_rule.keys():
+        if key in combination_dict:
+            # do this differently
+            if key.endswith("_version"):
```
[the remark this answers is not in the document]
madhavajay: Yes good catch, this should go above the key in combination_dict check and continue before the check.

tests/syft/lib/allowlist_test.py
```diff
@@ -0,0 +1,495 @@
...
+    # if there is a min_version for the skip rule and our lib is over that version
+    # then skipping does not apply
+    if "min_version" in skip_rule and lib_version < version.parse(
+        skip_rule["min_version"]
+    ):
+        applies = False
+
+    # if we were unable to fail the skip rules we should have applies = True and will
+    # skip this particular test combination
+    return applies
```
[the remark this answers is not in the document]
madhavajay: I guess we could return in the final non loop version checks, but in the earlier loop we need to check all conditions as some might match. My concern with doing an early return would be that more rules might get added and we might exit early before checking all of them. Unfortunately this kind of thing is always a bit confusing as its nested boolean logic that goes from disproving to proving etc, so definitely open to some rethinking to make it less complex.

tests/syft/lib/allowlist_test.py
madhavajay: Yes, nice refactor! Thanks.

tests/syft/lib/allowlist_test.py
madhavajay: Good point.

tests/syft/lib/allowlist_test.py
madhavajay: Yeah this is a bit annoying. I think my logic was this:
There are multiple skip rules at:
> for skip_rule in skip:
So the possible times we "dont skip" is more than 1, so we would need to use a set to add the combination to but the combination contains a nested list which cant be hashed, so instead we just use another set of just the str representation to keep track of which ones shouldn't be used and then do a second loop for appending the real combination with full intact data structure.

Definitely feels sub optimal, and I would love a suggestion on a good alternative. I considered changing the nested list to a tuple but JSON doesn't have a tuple representation so I figured i'd just go this route for now.

tests/syft/lib/allowlist_test.json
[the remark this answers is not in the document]
madhavajay: I have changed these to lte_version and gte_version lets see if that makes more sense.

scripts/pre_commit.sh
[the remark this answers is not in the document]
madhavajay: I am keen to hear if it works for other people since I haven't been able to test it on any other setups like Linux. I assume the set -e should still trigger a failure if one of the backgrounded tasks fails but maybe im wrong.

.github/workflows/tests.yml
```diff
@@ -83,7 +83,7 @@ jobs:
           pip install -r requirements.txt
           pip install -e .
           pip freeze | grep torch
-          pytest -m fast --cov-fail-under 0
+          pytest -m fast --cov-fail-under 0 -n auto
```
tudorcebere, line 86: the hero that we needed

tudorcebere (approved):

.github/workflows/tests.yml
madhavajay: Yes im shocked it works so well, i tried pytest-parallel and it was slow and buggy but pytest-xdist works like a dream!

src/syft/lib/python/primitive_factory.py
```diff
@@ -64,33 +65,25 @@ def generate_primitive(
...
-        from .list import List
-        from .none import SyNone
-        from .string import String
 
         if isinstance(value, PyPrimitive):
             return value
 
         if isinstance(value, bool):
-            return Bool(value=value, id=id)
+            return python.Bool(value=value, id=id)
```
gmuraru, line 73: Great! Thank you for adding those!

scripts/compatibility_report.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# check if we are inside a pipenv virtualenv
+pipenv --venv
+if test $? -eq 0
+then
+    array=( "0.5" "0.6" "0.6.1" "0.7" )
+    for i in "${array[@]}"
```
gmuraru, line 8: NITPICK: Use a more suggestive name - maybe ```version``` than ```i```

scripts/compatibility_report.sh
```diff
@@ -0,0 +1,19 @@
...
+    do
+        yes | pip uninstall torchvision
```
gmuraru, line 10: MAYBE: Use ```pip uninstall -y torchvision```

scripts/pre_commit.sh
gmuraru, line 14: Hmm...Only the last command will run in the background. You run ```build_proto.sh``` and if successful, you run ```isort```, if success you run...and at the end, you run ```pre-commit run --all-files``` in background.

You use "set -e", this means that you can remove the ```&&```
and simply do:
```
./scripts/build_proto.sh
isort .
black src tests
pre-commit run --all-files  <--- this does not need to run in the background since is the last command
```
if any of the above commands exists with a non zero status code, the script will exit (since you use ```set -e```)

What was the intention here?

scripts/pre_commit.sh
```diff
@@ -1,15 +1,16 @@
...
 
-./scripts/build_proto.sh
+# start tests in parallel (with additional parallelism)
+pytest -m fast -n auto &
+
+# check bandit in parallel
+bandit -r src -ll &
+
+# run API documentation test notebooks in parallel
+./scripts/nb_test.sh && pytest examples/api --cov-fail-under 0 &
```
gmuraru, line 11: Only the last command will run in background

gmuraru (changes requested):

scripts/compatibility_report.sh
madhavajay: Much better! That should work on windows now too. 👍

scripts/pre_commit.sh
madhavajay: Do you mean on each line? The ones written left to right with && are dependant on each other so not sure I can do much about that. I tried running each line in a subshell $() but then its out of the virtualenv, any idea on how to improve this? It's already halved the run time on my machine so it's at least an improvement.

scripts/pre_commit.sh
madhavajay: I guess its just to make it clear they are a single unit, the pre-commit run is the slowest but because its at the end there isnt any advantage, however the slower tests above are run first and backgrounded so that this can at least start as soon as the proto, isort and black are done.

Later if we add another step and it goes after this the pre-commit run --all-files won't block it.

My two goals are:
- as fast as possible
- as clear as possible

Im very interested in any optimal way to re-order things to make it faster. I tried a few different combinations but this was the fastest I came up with.

madhavajay:
Part of this PR will be to trim down the target support for complex, qint and 1.4.0 to be discussed as separate tickets and features at a later date.

scripts/pre_commit.sh
gmuraru: Ahh..Ok. I thought the intention was here to run both commands in parallel.

scripts/pre_commit.sh
```diff
@@ -1,15 +1,16 @@
...
-bandit -r src -ll
-pytest -m fast
-
-# run API documentation test notebooks
-./scripts/nb_test.sh
-pytest examples/api --cov-fail-under 0
\ No newline at end of file
+./scripts/build_proto.sh && isort . && black src tests && pre-commit run --all-files &
+
+wait
```
gmuraru, line 16:
```  wait [n ...]
        Wait  for each specified process and return its termination sta-
        tus.  Each n may be a process ID or a job  specification;  if  a
        job  spec  is  given,  all  processes in that job's pipeline are
        waited for.  If n is not given, all currently active child  pro-
        cesses  are  waited  for,  and  the return status is zero.  If n
        specifies a non-existent process or job, the  return  status  is
        127.   Otherwise,  the  return  status is the exit status of the
        last process or job waited for.```

Taken from [here](https://stackoverflow.com/questions/13296863/difference-between-wait-and-sleep).
It specifies that using wait without any process, the exist status would be 0.
Maybe we should record each pid and at the end wait in 'reverse' order on how much we expected them to run?

File: tests/syft/lib/allowlist_test.py (hunk 9 above)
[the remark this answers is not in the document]
Comment (Jasopaum) [reply tests/syft/lib/allowlist_test.py:161]: But you break the loop as soon as you set `applies` to False, right?
Maybe I don't fully understand what you did 🤔

File: tests/syft/lib/allowlist_test.py (hunk 3 above)
Comment (Jasopaum) [reply tests/syft/lib/allowlist_test.py:313]: So you mean that several items in combinations can be the same combination?

Comment (madhavajay) [reply tests/syft/lib/allowlist_test.py:313]: It's a nested loop so its combination * skip_rules.
```
for combination in combinations:
    # skips are temporary
    for skip_rule in skip:
```
So appending based on the  check_skip could add more than 1 entry of the same combination.

Comment (Jasopaum) [reply tests/syft/lib/allowlist_test.py:313]: And what if we break the skip loop as soon as the combination needs to be skipped?

File: tests/syft/lib/allowlist_test.py (hunk 9 above)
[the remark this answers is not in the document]
Comment (madhavajay) [reply tests/syft/lib/allowlist_test.py:161]: I think you are totally right. I have added two returns where the breaks are and then checked the full tests twice:
```
before change
============ 9430 passed, 6355 warnings in 183.18s (0:03:03) ============
after change
============ 9430 passed, 6355 warnings in 184.55s (0:03:04) ============
```

So thats good enough for me.

File: scripts/pre_commit.sh (hunk 15 above)
Comment (madhavajay) [reply scripts/pre_commit.sh:16]: You are right, looks like i'm using it wrong. I have removed this for now to avoid confusion that it does anything.

File: tests/syft/lib/allowlist_test.py (hunk 3 above)
Comment (madhavajay) [reply tests/syft/lib/allowlist_test.py:313]: yup that works:
```
before change
============ 9430 passed, 6355 warnings in 184.55s (0:03:04) ============
after change
============ 9430 passed, 6355 warnings in 190.64s (0:03:10) ============
```

The tests just passed so the next commit will reset the clock for 23 minutes from now, so ill hold off until there are any final requests.

Jasopaum (approved):

gmuraru (approved):

PR #4674, opened 2020-10-19
Use id_at_location

gmuraru:
## Description
Currently, we use the id of the object as a way to register an object as a remote location.
This brings some advantages, but also some downsides.
The advantage is that we will record only once if we send that item multiple times.
```
ptr = x.send(alice)
ptr = x.send(alice)
```
Since we send the same object, we keep only 1 instance of that object on alice.

The problem comes (in the above example) when we delete ```ptr``` - the second assignment will generate a call to the ```__del__``` method from ```PointerTensor``` (this call can come at any time).
If we have the same id, we might execute the second statement and after that, we delete the first ```ptr``` --> in this case we will delete the underlying data, where ```ptr``` points to.

With this PR we generate another UUID  for each sent item - in this scenario, we know that when the first ```ptr``` will get deleted it will delete an object that is identified with a UUID that is different than the second one

## Affected Dependencies
How the objects are kept at the destination

## How has this been tested?
- Added a test where we overwrite the ```ptr``` with the same object

## Checklist

### DIFF
--- a/src/syft/ast/klass.py
+++ b/src/syft/ast/klass.py
@@ -164,10 +164,12 @@ class Class(Callable):
                 id_ = UID()
                 self.id = id_
 
+            id_at_location = UID()
+
             # Step 1: create pointer which will point to result
             ptr = getattr(outer_self, outer_self.pointer_name)(
                 client=client,
-                id_at_location=self.id,
+                id_at_location=id_at_location,
                 tags=self.tags if hasattr(self, "tags") else list(),
                 description=self.description if hasattr(self, "description") else "",
             )
@@ -184,6 +186,7 @@ class Class(Callable):
             client.send_immediate_msg_without_reply(msg=obj_msg)
 
             # STep 4: return pointer
+            # Step 4: return pointer
             return ptr
 
         def send_to(self: Any, client: Any, searchable: bool = False) -> Pointer:
--- a/src/syft/ast/klass.py
+++ b/src/syft/ast/klass.py
@@ -185,7 +185,6 @@ class Class(Callable):
             # Step 3: send message
             client.send_immediate_msg_without_reply(msg=obj_msg)
 
-            # STep 4: return pointer
             # Step 4: return pointer
             return ptr
 
--- a/src/syft/core/node/common/action/save_object_action.py
+++ b/src/syft/core/node/common/action/save_object_action.py
@@ -37,36 +37,34 @@ class SaveObjectAction(ImmediateActionWithoutReply, Serializable):
 
     def execute_action(self, node: AbstractNode, verify_key: VerifyKey) -> None:
         # save the object to the store
-        node.store.store(
-            obj=StorableObject(
-                # ignoring this - the fundamental problem is that we can't force the classes
-                # we want to use to subclass from something without creating wrappers for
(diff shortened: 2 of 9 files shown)


src/syft/core/node/common/action/save_object_action.py
```diff
@@ -37,37 +37,35 @@ def __init__(
...
-                ),
-                search_permissions={All(): None}
-                if self.anyone_can_search_for_this
-                else {},
-                read_permissions={
-                    node.verify_key: node.id,
-                    verify_key: None,  # we dont have the passed in sender's UID
-                },
-            )
+        obj = StorableObject(
```
gmuraru, line 40: Made it like this to be easier for debugging (using ```peda```)

src/syft/ast/klass.py
```diff
@@ -184,6 +186,7 @@ def send(self: Any, client: Any, searchable: bool = False) -> Pointer:
             client.send_immediate_msg_without_reply(msg=obj_msg)
 
             # STep 4: return pointer
+            # Step 4: return pointer
```
Jasopaum, line 189: I think this line is duplicated

gmuraru: Removed

tests/syft/core/common/serializable_test.py
```diff
@@ -1,5 +1,21 @@
...
+
 # from syft.core.common.uid import UID
-# from syft.core.common.serde.serializable import Serializable
+from syft.core.common.serde.serializable import Serializable
+from syft.core.common.serde.serialize import _serialize
+
 # import syft as sy
+
+
+def test_object_with_no_serialize_wrapper() -> None:
```
gmuraru, line 11: Those are added only to increase test coverage

tests/syft/lib/torch/tensor/tensor_remote_and_serde_test.py
```diff
@@ -20,11 +20,12 @@ def test_torch_remote_tensor_register() -> None:
 
     assert len(alice.store) == 1
 
-    # TODO: Fix this from deleting the object in the store due to the variable
-    # ptr being assigned a second time and triggering __del__ on the old variable
-    # ptr used to be assigned to (Even though its new assignment is the same object)
-    # ptr = x.send(alice_client)
-    # assert len(alice.store) == 1  # Same id
+    ptr = x.send(alice_client)
```
gmuraru, line 23: Fix the test

madhavajay (approved):
LGTM

tests/syft/lib/torch/tensor/tensor_remote_and_serde_test.py
```diff
@@ -20,11 +20,12 @@ def test_torch_remote_tensor_register() -> None:
 
     assert len(alice.store) == 1
 
-    # TODO: Fix this from deleting the object in the store due to the variable
-    # ptr being assigned a second time and triggering __del__ on the old variable
-    # ptr used to be assigned to (Even though its new assignment is the same object)
-    # ptr = x.send(alice_client)
```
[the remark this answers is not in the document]
gmuraru: That line is at line ```23``` (it is the same assignment that is done on ```18```)

tudorcebere (approved):

PR #4699, opened 2020-10-22
Add Blocking Permission Request

### LINKED ISSUE #4661: Add Blocking Permission Request to MNIST demo
## Feature Description
To allow access to metrics during training or testing, we require permissions.
We need to implement an intentional blocking pause while permissions are requested so that the iteration doesn't fail or overload the remote system with requests.

If the remote party denies the request (possibly by accident) we should try to make this easy to recover rather than destroying the whole training loop state.

This is related to this ticket:
https://github.com/OpenMined/PySyft/issues/4660

## Definition of Done
The training / test loops should be runnable and block on permission requests. The Duet partner should be able to approve these requests and training / testing should continue. Rejection should not break the state of the loop allowing it to be re-requested.

aanurraj:
## Description
This issue fixes #4661 
As described in #4661 we need to add blocking or delay if the request made is being accidentally rejected, so the normal working of the training loop doesn't get disturbed.

## Work Done

- Instead of adding a parameter just to `.request()` the parameter is actually added to `.get()` this saves us from the overhead to deal with the exception in the training loop one can simply do
`duet.store[0].get(request_block=True)` in the training loop.

- The request is made internally in `.get()` when requested and it currently has a default delay of 20 seconds to delay with permission if the permission is accidentally denied or delayed the current call of the function will be skipped like
Screenshot 2020-10-22 at 9 43 48 PM
if everything works fine we will get the requested data 
Screenshot 2020-10-22 at 9 47 21 PM

### TODOs

- Currently, the code looks dirty can be improved.
- The delay can be made such that as soon as the permission is granted it should be reflected (need not to wait for 20 seconds by default).
- The time can be added as a parameter too. 

## Affected Dependencies
None

## How has this been tested?
- Currently locally tested.

## Checklist


src/syft/core/pointer/pointer.py
```diff
@@ -145,22 +146,40 @@ def __init__(
...
+                obj_msg = GetObjectAction(
+                    obj_id=self.id_at_location,
+                    address=self.client.address,
+                    reply_to=self.client.address,
+                )
+
+                response = self.client.send_immediate_msg_with_reply(msg=obj_msg)
+
+                return response.obj
+        else:
```
aanurraj, line 173: This and the above block of code can be improved [Need Suggestions}

gmuraru: Maybe have an internal function that simply doe:
```
obj_msg = GetObjectAction(
                    obj_id=self.id_at_location,
                    address=self.client.address,
                    reply_to=self.client.address,
                )

                response = self.client.send_immediate_msg_with_reply(msg=obj_msg)

                return response.obj
```

src/syft/core/pointer/pointer.py
```diff
@@ -243,7 +262,8 @@ def request(
         request_name: str = "",
         name: str = "",
         reason: str = "",
-    ) -> None:
+        block: bool = False,
```
gmuraru, line 265: I do not think we need this here, we need only for get. So I think we can remove this?

aanurraj: we still need to make the internal call from .get() to get the status, if the call is accepted or rejected

madhavajay:
@aanurraj I made some changes here: https://github.com/OpenMined/PySyft/pull/4722

If you want you can merge / cherry pick or we can continue to iterate on it via that shared branch.

Issue #4671, opened 2020-10-18
Fix broken float tests


madhavajay:
## Description
There are two tests marked `@pytest.mark.xfail` in id_test.py which need to be fixed.

## How to Reproduce
Remove the annotations marked `@pytest.mark.xfail` then run the fast tests:
```
$ pytest -m fast
```

## Expected Behavior
Tests should pass

## Definition of Done
PR should be posted with passing test suite and `@pytest.mark.xfail` annotations removed.
If there is a good reason these tests cannot pass that should be documented and then explained here.

rajathpatel23:
Can I work on this? @madhavajay

gmuraru:
Sure @rajathpatel23. Assigned it to you :+1:

gmuraru:
It is the ```syft_0.3.0``` branch

rajathpatel23:
Thank you @gmuraru

madhavajay:
@rajathpatel23 Yes, this is for 0.3 but contributions are welcome. If you checkout syft_0.3.0 and then branch from there and make this change that would be amazing. Just a note, the "slow" tests are currently failing on the 0.3.0 branch which i am fixing now in another PR, so don't be surprised if that happens. This task only applies to the fast tests.

Also this might not be the easiest task for starting on the 0.3 code base, so sorry if I labelled it that way, its hard to know if the work has been done yet or it's just still left off. Feel free to try or if you don't want to thats also okay. 😊

If you have any questions just write them here or ping me on slack and I can try to help.

rajathpatel23:
@madhavajay , cool I would like to give it a try for next couple of days, if I do not make progress we can handover it to someone, does that work ?

rajathpatel23:
@madhavajay I was not able to make progress on this, we can assign it somone.

PR #4711, opened 2020-10-26
Fix float tests

madhavajay:
## Description
- Fixing .hex float test.
Addresses this ticket: https://github.com/OpenMined/PySyft/issues/4671
- Speeding up fast tests by marking anything under 1 sec as slow
- Improving Pre Commit script so that it reports when a parallel item fails

Problem:
- .hex in python on floats returns a very specific encoding which is not compatible with our hex representation of the binary encoding of our type.
```
>>> a = float(42.5)
>>> a.hex()
'0x1.5400000000000p+5'
```

Question:
- In the interest of preventing confusion I have removed .hex and left .to_hex as some other types in the python library have .hex methods.
- Should we remove the other methods like .json .binary .proto and leave their .to_x aliases?
- Or should we solve this another way, perhaps only setting .hex if it doesn't exist and just document that .hex is our method but in cases where it already exists use .to_hex?

## Affected Dependencies
None

## How has this been tested?
Locally and CI

## Checklist


tudorcebere (approved):
LGTM

gmuraru (approved):

PR #4713, opened 2020-10-26
FPT supports mul or div operation with float scalar

marload:
## Description
We can multiply or divide the FPT by the Float Scalar.

```python
x = torch.tensor([1, 2, 3.]).fix_precision()
z = x / 2.1
```

## Affected Dependencies
None

## How has this been tested?

## Checklist


syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -281,6 +281,9 @@ def mul_and_div(self, other, cmd):
         which is inherent to these operations in the fixed precision setting
         """
         changed_sign = False
+        if isinstance(other, float):
+            other = torch.tensor(other).fix_precision().child
```
gmuraru, line 285: Q: Shouldn't we use the same precision as the ```self``` (if one has 4 and the other has 5, it takes automatically the lowest precision?)

marload: Your right!
`other` tensor must have same precision as `self`

gmuraru:
@marload had a talk with @LaRiffle and I think the best solution for this is to throw an exception in case we try to multiply/divide ```fix_precision``` with a "normal tensor" such that anyone who tries to multiply/divide would have to explicitly convert to ```fix_precision```

marload:
> @marload had a talk with @LaRiffle and I think the best solution for this is to throw an exception in case we try to multiply/divide `fix_precision` with a "normal tensor" such that anyone who tries to multiply/divide would have to explicitly convert to `fix_precision`

Are you saying that we should raise the exception when `torch.tensor(...).fix_precision() / torch.tensor(...)`

gmuraru:
> > @marload had a talk with @LaRiffle and I think the best solution for this is to throw an exception in case we try to multiply/divide `fix_precision` with a "normal tensor" such that anyone who tries to multiply/divide would have to explicitly convert to `fix_precision`
> 
> Are you saying that we should raise the exception when `torch.tensor(...).fix_precision() / torch.tensor(...)`?

Yep - but let's see what @LaRiffle has to say about this.

LaRiffle:
Hey!
That's right, but because of the several cases we have faced where you try to monkey patch other libs, operating internal float tensors with our FPT inputs, I think we actually need this, even if technically from a mathematic viewpoint it doesn't make sense.
I would also add @marload support for operating real python floats with FPT, what do you think about this?

syft/frameworks/torch/tensors/interpreters/precision.py
```diff
@@ -281,6 +281,10 @@ def mul_and_div(self, other, cmd):
         which is inherent to these operations in the fixed precision setting
         """
         changed_sign = False
+        if isinstance(other, float):
+            other = torch.tensor(other).fix_prec(precision_fractional=self.precision_fractional)
```
LaRiffle, line 285: Maybe this is better to have all arguments of `self`
```suggestion
            other = torch.tensor(other).fix_prec(**self.get_class_attributes())
```

LaRiffle (changes requested):
I'll clarify my last comment because I misread your PR: I think we need FPT to operate with:
- Python Float (converted to FPT)
- FloatTensors (converted to FPT)
- int, Int/LongTensor (not converted)

But the PR is great :)

gmuraru:
Yep - I am ok with both cases, I think it is that tradeoff game :D - 
* Throwing an exception would be more frustrating for the developer (because it would have to always do an FPT)
* Having Syft doing this automatically would be easier for the developer, but it might introduce some bugs.

LaRiffle:
> Yep - I am ok with both cases, I think it is that tradeoff game :D -
> 
> * Throwing an exception would be more frustrating for the developer (because it would have to always do an FPT)
> * Having Syft doing this automatically would be easier for the developer, but it might introduce some bugs.

Yeah exactly, it will hide bugs, but also now the code is more robust than 6 months ago, that's why I've changed my mind ;)

marload:
@gmuraru @LaRiffle 
Thank you for your review!
Can you check a new commit?

gmuraru:
Looking now! Sorry for the delay.

test/torch/tensors/test_precision.py
```diff
@@ -294,6 +294,13 @@ def test_torch_sub_():
 def test_torch_mul(workers):
     bob, alice, james = (workers["bob"], workers["alice"], workers["james"])
 
+    x = torch.tensor([1, 2, 3.0]).fix_precision()
```
gmuraru, line 297: Q: Could you also add a test with:
```
x = torch.tensor([1, 2, 3.0]).fix_precision()
z = x * 2.1231  <-- more digits after "."
assert (z.float_precision() == torch.tensor([2.123, 4.246, 6.369])).all()
```

gmuraru (changes requested):
Looks great! Only if you can add one more tests to make sure we take the precision from the FPT  we are doing an operation on

marload:
Thanks!
I added new test!

aanurraj (approved):
Looks Good :)

gmuraru (approved):
Great job!

LaRiffle (approved):

PR #4749, opened 2020-10-29
Add torch.Tensor.angle to allowlist and test suite

PlamenHristov:
## Description
Added torch.Tensor.angle to allowlist to close [this issue](https://github.com/OpenMined/PySyft/issues/4185).

## Affected Dependencies
None

## How has this been tested?

## Checklist


madhavajay:
@PlamenHristov awesome! 👍 ❤️ Thanks so much for the PR.
We have added torch==1.7.0 to the syft_0.3.0 branch CI so its worth keeping that in mind now.
What I like to do is test it for the latest version and then check some of the older versions locally or in CI.

If you have the project setup in a virtualenv you can use the scripts/compatibility_report.sh to automatically test every single torch version.

Its configured for pipenv, but if you know another way to detect virtualenvs an update would be cool. Otherwise you can disable the check while using it.

I usually disable all the allowlist entries im not testing so that its faster and then do about 10-20 at a time and look at the results.

At the moment i'm thinking its a good idea to include an lte_version upper bound on any not_available rules so that if a future version of torch supports something we will test it automatically.

I actually need to go through the existing lte_version 1.6.0 not available rules and re-run them on 1.7.0. Currently I just bumped the 1.6.0 string to 1.7.0 so that it skips those tests on 1.7.0 torch as well because I wanted to quickly make sure nothing obvious breaks with 1.7.0, but my guess is some of the previously not supported op / dtype combos will be added and weirdly occasionally sometimes it seems they also get dropped.

Also note, this is all for CPU, so we really need a separate GPU datatype test but thats something I will look into in a few weeks.

For now, I think if you can add the lte_version "1.6.0" to the bool not_available rule that should do it.
Then it might fail on 1.7.0 and we can fix it if thats the case, or feel free to install 1.7.0 and test it now.

Thank you again for the PR and I look forwards to seeing more allowlist fixes from you. 😊

tests/syft/lib/allowlist_test.json
```diff
@@ -720,6 +720,13 @@
           { "data_types": ["float16"], "lte_version": "1.5.1" }
         ]
       },
+      "angle": {
+        "profile": "tensor_method_noinput",
+        "not_available": [
+          { "data_types": ["bool"] },
```
madhavajay, line 726: Add an "lte_version": "1.6.0" to the bool rule so that it gets automatically checked in 1.7.0 later.
I think its a good idea we never add blanket rules without version limits so we don't forget to later test these if new versions are released.

madhavajay (approved):
LGTM!

PlamenHristov:
Hey @madhavajay,

Fist of all, thank you so much for the lovely feedback! I'm super excited to work on other stuff as well.

Regarding the testing on torch 1.7.0. that's where I did my testing and I think everything was passing normally,
The reason I added the bool as ignored, is because I don't think it makes sense (as in mathematically) to ask for the angle of a boolean.

Let me know your thoughts. Again super excited to work on some other stuff.

madhavajay:
@PlamenHristov Yes, I totally agree that there are instances where it seems obvious that a particular data type doesn't make sense, but there are over 400 functions so rather than deciding on a case by case basis its a lot easier if we just keep upper limits on all the not_available rules so we don't accidentally forget to test certain data type support in future versions. I have been meaning to add a "description" field, for this exact purpose so that we can leave notes on why functionality is or isn't available which would be the perfect in this instance. For now it can stay as it is. 👍

Jasopaum (approved):

Issue #4745, opened 2020-10-28
KeyError / Fix the FQN issue on print(duet.store)


madhavajay:
## Description

Sending a searchable SyPrimitive type seems to cause an exception when printing the store.
```
l = sy.lib.python.List()
l.send(duet, searchable=True)
print(duet.store)
```

## Definition of Done
Fixed.

madhavajay:
This is happening with any and all classes where the fqn changes depending on import context, so we need something more permanently corrective.

madhavajay:
Still happens on classes like:
```
>>> torch.nn.modules.Sequential
<class 'torch.nn.modules.container.Sequential'>
>>> torch.nn.Sequential
<class 'torch.nn.modules.container.Sequential'>
```

madhavajay:
This also happens inside the new sy.module:
```
# module.py
def __setattr__(self, name: str, value: Union[Any, "Module"]) -> None:
        # bug where torch.nn.modules isnt the full name on some imports
        # TODO: fix this properly
        if "torch.nn" in full_name_with_qualname(klass=type(value)):
            modules = self.__dict__.get("_modules")
            if modules is not None:
                modules[name] = value
        else:
            object.__setattr__(self, name, value)

```

Issue #4729, opened 2020-10-28
Add sy.Module


madhavajay:
## Description
Create sy.Module and update MNIST to use it.
- The user should only have to do:
```
class Net(sy.Module):
    def __init__(self) -> None:
        super(Net, self).__init__()
        self.conv1 = nn.Conv2d(1, 32, 3, 1)
        self.conv2 = nn.Conv2d(32, 64, 3, 1)
        self.dropout1 = nn.Dropout2d(0.25)
        self.dropout2 = nn.Dropout2d(0.5)
        self.fc1 = nn.Linear(9216, 128)
        self.fc2 = nn.Linear(128, 10)

    def forward(self, x):
        x = self.conv1(x)
        x = F.relu(x)
        x = self.conv2(x)
        x = F.relu(x)
        x = F.max_pool2d(x, 2)
        x = self.dropout1(x)
        x = torch.flatten(x, 1)
        x = self.fc1(x)
        x = F.relu(x)
        x = self.dropout2(x)
        x = self.fc2(x)
        output = F.log_softmax(x, dim=1)
        return output
```

If possible add a function or property like:
```
model.is_remote
```
Which would return a boolean True if all the layers are Pointers and False if all the layers are not pointers.

We also need to decide what functionality we will support and track within the sy.module so we can correctly filter the __setattr__ for classes that match say torch.nn.Module and their Pointers.

## Definition of Done
The MNIST Duet Notebook's are updated to use the above functionality.

madhavajay:
https://github.com/OpenMined/PySyft/pull/4741

PR #4750, opened 2020-10-29
Mnist duet nn sequential

madhavajay:
## Description
Even more cool features for the MNIST notebooks.

## How has this been tested?

## Checklist


gmuraru (approved):

src/syft/core/common/module.py
```diff
@@ -23,7 +27,9 @@ def __init__(self) -> None:
 
     # this is how we catch the modules being set during subclass init
     def __setattr__(self, name: str, value: Union[Any, "Module"]) -> None:
-        if "torch.nn.modules" in full_name_with_qualname(klass=type(value)):
+        # bug where torch.nn.modules isnt the full name on some imports
```
gmuraru, line 30: Q: For what case is this happening?

src/syft/core/common/module.py
```diff
@@ -77,3 +83,81 @@ def cpu(self) -> "Module":
         for _, module in self.modules.items():
             module.cpu()
         return self
+
+    # zero them so we know they are copied
+    def zero_layers(self) -> None:
+        for _, m in self.modules.items():
```
gmuraru, line 89: NITPICK: Why not use for m in self.modules.values()?

src/syft/core/common/module.py
```diff
@@ -77,3 +83,81 @@ def cpu(self) -> "Module":
...
+            if hasattr(m, "weight"):
+                m.weight.requires_grad_(False).zero_()
+            if hasattr(m, "bias"):
+                m.bias.requires_grad_(False).zero_()
+
+    # easy way to check the weights have changed
+    def sum_layers(self) -> None:
```
gmuraru, line 96: Q: If we use this for debugging maybe it would be better to add a proper name?
like ```def debug_sum_layers``` - is this meant to be used by a dev?

src/syft/core/common/module.py
```diff
@@ -77,3 +83,81 @@ def cpu(self) -> "Module":
...
+        timeout_secs: int,
+        delete_obj: bool = False,
+        skip_layers: List[str] = [],
+    ) -> None:
+        # loop over models module pointers
+        for n, m in remote_model.modules.items():
+            try:
+                if n in skip_layers:
+                    log = f"> Skipping: {n}"
+                    print(log)
```
gmuraru, line 121: Q: Shouldn't we remove the prints? and simply leave the  ```logger.debug```
We can change the debug level from ```__init__.py```

src/syft/core/common/module.py
```diff
@@ -77,3 +83,81 @@ def cpu(self) -> "Module":
...
+                        # iterate through the key, values
+                        # weights and biases should be in there
+                        for key, value in state_dict.items():
+                            key_str = key.upcast()
+                            # if the local models module has the same property
+                            # e.g. .weight or .bias
+                            if hasattr(local_m, key_str):
+                                # if the downloaded value is not a Parameter
+                                # its a tensor so we need to convert it
+                                if not issubclass(type(value), torch.nn.Parameter):
```
gmuraru, line 151: Q: Can't we use
```if not isinstance(valu,e torch.nn.Parameter)```?

File: src/syft/core/store/store_memory.py [hunk 6]
```diff
@@ -59,19 +60,17 @@ def keys(self) -> KeysView[UID]:
     def values(self) -> ValuesView[AbstractStorableObject]:
         return self._objects.values()
 
-    @syft_decorator(typechecking=True)
-    def store(self, obj: AbstractStorableObject) -> None:
```
gmuraru, line 63: MAYBE: The ```MemoryStore``` class implements the ```ObjectStore``` which has the ```store``` method (which is ```NotImplemented````).
I am keen to keep the method ```store``` as it is (and also add a ```delete``` method maybe?). We use is as a dict because it is basically a dictionary wrapped inside a class. 

If we were to add more storage types we should adhere to an interface:
* In memory DB (Redis/Memchached)
* File System Based (SQL/NO-SQL DBs)

I think the question is:
Should we adhere to store[object.id_at_location] = object to save objects and del store[object.id_at_location] to remove them, or should we have ```store``` and ```delete``` as methods (and inside those methods we do the "internal" stuff depending on the implementation).

I like the second approach because using it like ```store[object.id_at_location]``` and ```del[object.id_at_location]```` alludes to the store being a dictionary - and might not necessarily be one.

If we go ahead with using the ```dictionary``` methods to delete/add an object then I think we should remove the ```store``` method from the ```ObjectStore```

src/syft/grid/duet/__init__.py
```diff
@@ -297,12 +295,23 @@ def join_duet(
...
+                    # with only its new server_id
+                    if (
+                        "server_id" in loopback_config
+                        and "client_id" not in loopback_config
+                    ):
+                        target_id = str(loopback_config["server_id"])
+                    else:
+                        time.sleep(0.5)
+            except Exception as e:
+                print(e)
```
gmuraru, line 313: MAYBE: Use the debug?

src/syft/grid/duet/__init__.py
```diff
@@ -297,12 +295,23 @@ def join_duet(
...
         loopback_config = {}
-        with open(get_loopback_path(), "r") as f:
-            loopback_config = json.loads(f.read())
-            if "server_id" in loopback_config:
-                target_id = loopback_config["server_id"]
-            else:
-                raise Exception("No loopback file. start the Duet Data Owner first")
+        target_id = ""
+        while target_id == "":
+            try:
```
gmuraru, line 300: Q: What exception can be thrown here?

src/syft/grid/duet/signaling_client.py
```diff
@@ -22,6 +22,9 @@ def __init__(
         # Use Signaling Server metadata
         # to build client route
         conn = conn_type(url=url)  # type: ignore
+
+        print(conn._get_metadata())
```
gmuraru, line 26: MAYBE: Use debug?

src/syft/grid/duet/webrtc_duet.py
```diff
@@ -241,76 +257,94 @@ async def pull(self) -> None:
...
+                AnswerPullRequestMessage(
+                    address=self.signaling_client.address,
+                    target_peer=target_id,
+                    host_peer=self.signaling_client.duet_id,
+                    reply_to=self.signaling_client.address,
+                )
             )
-        )
+        except Exception as e:
+            log = f"Got an exception in Duet send_offer. {e}"
```
gmuraru, line 287: NITPICK: For logging it is better to use a ```lazy``` string evaluation technique. Like
```logger.info("Something happened %s", var_string)```
^^^ In this scenario if the logger level is above ```info``` this would not be evaluated.

Using ```f-strings``` they would always be evaluated.

^^ This is premature optimization, but I think some flake flags or black might complain about this.

File: src/syft/grid/example_nodes/network.py [hunk 11]
```diff
@@ -20,33 +21,56 @@
...
 
 @app.route("/", methods=["POST"])
-def process_network_msgs() -> str:
-    json_msg = request.get_json()
-    obj_msg = _deserialize(blob=json_msg, from_json=True)
+def process_network_msgs() -> flask.Response:
+    data = flask.request.get_data()
+    obj_msg = _deserialize(blob=data, from_bytes=True)
     if isinstance(obj_msg, SignedImmediateSyftMessageWithReply):
+        print(
```
gmuraru, line 53: Q: Shouldn't we use ```log.info```?

src/syft/lib/python/int.py
```diff
@@ -372,8 +372,21 @@ def conjugate(self) -> SyPrimitiveRet:
...
-    def to_bytes(self, length: int, byteorder: str, signed: bool = True) -> bytes:
-        return super().to_bytes(length, byteorder, signed=signed)
+    # method signature override
+    def to_bytes(
+        self,
+        length: Optional[int] = None,
+        byteorder: Optional[str] = None,
+        signed: Optional[bool] = True,
+    ) -> bytes:
+        if length is not None and byteorder is not None and signed is not None:
```
gmuraru, line 382: MAYBE:
Rewrite this to:
```if length and byteorder and signed is not None:```

File: tests/syft/core/node/domain/domain_test.py [hunk 13]
```diff
@@ -106,3 +112,22 @@ def test_domain_request_accepted() -> None:
...
+
+
+def test_domain_is_for_me_exception() -> None:
+    domain_1 = Domain(name="remote domain")
+
+    with pytest.raises(Exception):
+        msg = SyftMessage()
+        domain_1.message_is_for_me(msg)
+
+    # @syft_decorator(typechecking=True)
```
gmuraru, line 124: Q: Could we add a TODO + issue for this?

tests/syft/grid/duet_test.py
```diff
@@ -1,78 +1,77 @@
 # third party
-import pytest
```
gmuraru, line 2: Q: Nothing from here works?

tests/syft/lib/torch/parameter_test.py
```diff
@@ -29,6 +29,31 @@ def test_parameter_vm_remote_operation() -> None:
...
+    y = xp + xp
+
+    assert len(alice.store._objects) == 2
+
+    y.get_copy()
+
+    # no deletion of the object
+    assert len(alice.store._objects) == 2
+
+    # TODO: put thought into garbage collection and then
```
gmuraru, line 50: Q: What should this do?

If you delete ```xp``` you should have one more element into the ```store``` (which is ```y```).
If you also delete ```y``` you should have 0.

Right? :smile:

gmuraru (changes requested):
Only some comments and questions

src/syft/core/common/module.py
madhavajay: Yeah this is weird, and its part of the FQN bug which we need to address separately. I haven't pinned this down to an exact situation but it seems to depend on when you import the library, for example this didn't happen in the tests but it did happen in my jupyter notebook. I think we need to decide exactly what we will support in something like Sy Module so that we can track and filter these the same way that the pytorch Module does. I have added this requirement to the existing ticket: https://github.com/OpenMined/PySyft/issues/4729

madhavajay: And the related bug:
https://github.com/OpenMined/PySyft/issues/4745

src/syft/core/common/module.py
madhavajay: Thanks, fixed.

src/syft/core/common/module.py
madhavajay: Yes this is likely temporary, so renaming is a good idea.

src/syft/core/common/module.py
madhavajay: I haven't gotten the logging system to filter the print statements in a way that im happy with yet, so I have some areas where the printing is explicitly designed to appear in the notebook, but we need to figure out a solid way to make this consistent. I think what we need to do is add our own custom logging levels and use those so that we arent restricted by the normal order of logging importance. I have made a ticket to work on this independently. https://github.com/OpenMined/PySyft/issues/4755

src/syft/core/common/module.py
madhavajay: I wasn't sure if there were cases where nn.Parameter might get subclassed. Maybe we should be changing the part which gives us Tensors instead of Parameters, but i think thats blocked behind the availability of methods and serialization for all the module allowlist methods. For now, I have updated this to `isinstance` as per your suggestion.

src/syft/core/store/store_memory.py
madhavajay: Totally agree, I just wanted to make sure something wasn't happening due to code using both different code paths to mutate the internal dict and the use of UID as the dict key versus passing in an object with a UID set on the StorableObject You are right, the interface should be abstracted from the concept of a dict. I think the naming of "store" as the verb could be changed to "save" so that the code doesn't use: store.store(obj). Also the client store interface exposes the objects as a list which kind of requires indexing so perhaps we could make a distinction there. I have created a ticket, let me know if you think the task requirements need modification: https://github.com/OpenMined/PySyft/issues/4756

src/syft/grid/duet/__init__.py
madhavajay: The file open and json.loads can probably both throw exceptions depending on what happens with the file.

src/syft/grid/duet/__init__.py
madhavajay: Do you mean `logger.debug()`?

src/syft/grid/duet/signaling_client.py
madhavajay: Good spot, I think thats an accidental left over print statement from when we were testing the change from json to protobuf for metadata, so i will just remove it.

src/syft/grid/duet/webrtc_duet.py
madhavajay: Unfortunately, I dont think the logging library acts like print by supporting default printf string formatting like that.
```
var_string = "test"
sy.logger.info("Something happened %s", var_string)
# 2020-11-03 12:53:11.089 | INFO     | __main__:<module>:2 - Something happened %s
```

It does support the sort of .format syntax:
```
sy.logger.info("Something happened {x}", x=var_string)
sy.logger.info("Something happened {}", var_string)
```
So I guess that might be better?

Alternatively they have their own lazy, but it requires wrapping things in functions / lambdas.
https://github.com/Delgan/loguru#lazy-evaluation-of-expensive-functions
```
sy.logger.opt(lazy=True).critical("Something happened {}", lambda: var_string)
```

That starts to make the logging more verbose, so we could make a helper that goes in our global utils without circular import issues to make this a bit nicer, but it feels like this needs a bit more thought. I will add it to the new logging task.

src/syft/grid/example_nodes/network.py
madhavajay: Good question, this is the CLI service running on the web server and we will be adding a Docker container for people to easily deploy, so for ease of debugging printing out to the stdout is probably helpful. But there might be some preferred logging configuration to ensure log retention and auditing of connections within certain networks. For now I guess this is a lightweight example so perhaps this should be addressed as part of Grid?

src/syft/lib/python/int.py
madhavajay: I guess two things:
1) are there values of length and byteorder that are valid but evaluate to False, like 0
2) would it be better to forward the incorrect input to the correct method so the error is meaningful, this code is already a bit suspect but but means that if any of those args or kwargs are passed in the original method will be called.

I think it would be better to avoid this by renaming again, but are we confident we wont come across a type that uses the new attr like "to_binary" again down the track. Perhaps we should prefix with sy?
```
a = Int(1)
a.sy_bytes()
```

Or now that we don't really have multiple encoding types perhaps we should just change this back to "serialize?"
```
a = Int(1)
a.serialize(to_bytes=True)
```

gmuraru (approved):

tests/syft/core/node/domain/domain_test.py
madhavajay: This has been removed.

tests/syft/grid/duet_test.py
madhavajay: This has been removed, Duet has been removed we only have WebRTCDuet now

tests/syft/lib/torch/parameter_test.py
madhavajay: I copied the test above, it seems like the commented out part for delete doesn't work as expected, but it was already like that, perhaps we need to fix that up?

iamtrask (approved):

tests/syft/lib/torch/parameter_test.py
gmuraru: The assert from line 54 should be ```assert len(alice.store_objects) == 1``` since we still have y

madhavajay: Good spot, these are re-enabled and fixed now.

src/syft/lib/python/int.py
madhavajay: Issue here: https://github.com/OpenMined/PySyft/issues/4759

src/syft/grid/duet/__init__.py
gmuraru: Yep

PR #4783, opened 2020-11-05
Demo

madhavajay:
## Description
Features added for the demo, including updated symodule.

## Affected Dependencies
None

## How has this been tested?
Locally and CI

## Checklist


proto/core/store/store_object.proto
```diff
@@ -17,6 +17,10 @@ message StorableObject {
   google.protobuf.Any data = 4;
   string description = 5;
   repeated string tags = 6;
+  // fix by making Dict not StorableObject
```
gmuraru, line 20: Q: Dict should be a storable object, right? :\ (because we can also use it as an object)

src/syft/ast/globals.py
```diff
@@ -18,6 +18,22 @@ class Globals(Module):
     def __init__(self) -> None:
         super().__init__("globals")
 
+    def __call__(
```
gmuraru, line 21: Q: Did we had a callable class for this? (what is the usecase?)

src/syft/core/node/common/client.py
```diff
@@ -383,10 +383,20 @@ def __len__(self) -> int:
 
     def __getitem__(self, key: Union[str, int]) -> Pointer:
         if isinstance(key, str):
+            matches = 0
+            match_obj: Optional[Pointer] = None
             for obj in self.store:
-                if key == str(obj.id_at_location.value):
+                if key in str(obj.id_at_location.value).replace("-", ""):
```
gmuraru, line 389: Q: Why is that replace needed?

src/syft/core/node/common/client.py
```diff
@@ -383,10 +383,20 @@ def __len__(self) -> int:
...
                     return obj
-            raise KeyError("No such request found for string id:" + str(key))
+                if key in obj.tags:
+                    matches += 1
+                    match_obj = obj
+            if matches == 1 and match_obj is not None:
```
gmuraru, line 394: MAYBE:
```
if matches > 1
   raise KeyError(...)
elif match_obj is not None:
   return match_obj
```

src/syft/core/node/domain/client.py
```diff
@@ -65,13 +65,67 @@ def __getitem__(self, key: Union[str, int]) -> RequestMessage:
...
+        timeout_secs: int = -1,
+        element_quota: Optional[int] = None,
+    ) -> Dict[str, Any]:
+        handler_opts: Dict[str, Any] = {}
+        if action not in ["accept", "deny"]:
+            raise Exception("Action must be 'accept' or 'deny'")
+        handler_opts["action"] = action
+        handler_opts["print_local"] = bool(print_local)
+        handler_opts["log_local"] = bool(log_local)
+        if name is not None and name != "":
```
gmuraru, line 121: MAYBE:
Simple``` if name```

src/syft/core/node/domain/client.py
```diff
@@ -65,13 +65,67 @@ def __getitem__(self, key: Union[str, int]) -> RequestMessage:
...
+            handler_opts["name"] = str(name)
+        handler_opts["timeout_secs"] = max(-1, int(timeout_secs))
+        if element_quota is not None:
+            handler_opts["element_quota"] = max(0, int(element_quota))
```
gmuraru, line 125: Q: Why cast to int?

src/syft/core/node/domain/domain.py
```diff
@@ -151,56 +153,114 @@ def get_request_status(self, message_request_id: UID) -> RequestStatus:
...
+
+    def _deny(self, request: RequestMessage) -> None:
+        logger.debug(f"Calling deny on request: {request.id}")
+        request.destination_node_if_available = self
+        request.deny()
+
+    def _try_deduct_quota(
+        self, handler: Dict[Union[str, String], Any], obj: Any
+    ) -> bool:
+        if "action" in handler and handler["action"] == "accept":
```
gmuraru, line 204: MAYBE:
```
if handler.get("action", None) == "accept"
```

src/syft/core/node/domain/domain.py
```diff
@@ -151,56 +153,114 @@ def get_request_status(self, message_request_id: UID) -> RequestStatus:
...
+            element_count = self._count_elements(obj=obj)
+            result = handler["element_quota"] - element_count
+            if result >= 0:
+                # the request will be accepted so lets decrement the quota
```
gmuraru, line 208: Q: What represents the quota for an element?

src/syft/core/node/domain/domain.py
```diff
@@ -151,56 +153,114 @@ def get_request_status(self, message_request_id: UID) -> RequestStatus:
...
+        if "element_quota" in handler:
+            if not self._try_deduct_quota(handler=handler, obj=obj):
+                logger.debug(
+                    f"> Rejecting {request} element_quota={handler['element_quota']}"
+                )
+                self._deny(request=request)
+                handled = True
+
+        # if not rejected based on quota keep checking
+        if "action" in handler and not handled:
```
gmuraru, line 251: MAYBE: The same as above

src/syft/core/store/store_disk.py
```diff
@@ -44,25 +74,37 @@ def __len__(self) -> int:
...
     @syft_decorator(typechecking=True)
+    def __contains__(self, item: UID) -> bool:
+        return str(item.value) in self.db
+
+    @syft_decorator(typechecking=True, prohibit_args=False)
+    def delete(self, key: UID) -> None:
+        try:
+            obj = self.get_object(key=key)
+            if obj is not None:
+                self.db.__delitem__(str(key.value))
```
gmuraru, line 99: MAYBE: I think we should not use the ```__delitem__``` or any internal methods.
I would vote to go with ```del self.db[str(key.value)]``` and delete the ```__delitem__``` from down bellow.

src/syft/core/store/store_interface.py
```diff
@@ -127,7 +116,17 @@ def __setitem__(self, key: UID, value: StorableObject) -> None:
...
     @syft_decorator(typechecking=True)
     def __delitem__(self, key: UID) -> None:
+        self.delete(key=key)
+
+    @syft_decorator(typechecking=True, prohibit_args=False)
+    def delete(self, key: UID) -> None:
         """
+        We should write custom deletion code so we can check if the item exists
+        and then ensure deletion is called at a single place and that full __delitem__
+        is used where possible instead of the weaker del ref count.
```
gmuraru, line 126: I think we should use ```del```.
Using ```__delitem__``` seems more like a hack we are doing because we can not find the problem :(

src/syft/core/store/store_memory.py
```diff
@@ -69,16 +69,27 @@ def __getitem__(self, key: UID) -> AbstractStorableObject:
...
+                self._objects.__delitem__(key)
+            else:
+                logger.critical(f"{type(self)} __delitem__ error {key}.")
+        except Exception as e:
+            logger.critical(f"{type(self)} Exception in __delitem__ error {key}. {e}")
+
     @syft_decorator(typechecking=True, prohibit_args=False)
     def __delitem__(self, key: UID) -> None:
-        del self._objects[key]
+        self.delete(key=key)
```
gmuraru, line 92: Q: Can't we delete this method? (```__delitem__```)
I think we should always use:
store.delete(key) <-- and this should delete our object

gmuraru (commented):
I am still voting for having multiple PRs for more changes :(
Took a first look, with only a "surface level" understanding of what is going on - and left some comments.

proto/core/store/store_object.proto
madhavajay: StorableObject means description, tags, permissions though, surely we don't want that on every dict we send?

src/syft/ast/globals.py
madhavajay: So the parent is callable, but the signature needs to be augmented so that the new obj_type param flows through. This is a fix for the FQN lookup for stuff like torch.nn.modules.conv.Conv2d vs torch.nn.Conv2d

src/syft/core/node/common/client.py
madhavajay: In the pandas output we are replacing the - so that you can double click the text and copy and paste it easily. We should refactor all of this to a more robust "search" method which does stuff like partial matching, case insensitive and stripping weird characters etc.

src/syft/core/node/common/client.py
madhavajay: Yeah i think thats a little clearer. 👍

src/syft/core/node/domain/client.py
madhavajay: There is some weak validation before hand but you are right since we are then doing a direct comparison if there is a "truthy" value at that point its probably safe to assume we can compare it and fail it if doesnt match, actually heres a great place where we should trip and lowercase.

src/syft/core/node/domain/client.py
madhavajay: I guess im paranoid that a non int sneaks through here and breaks the code preventing the quota from working. We should add the type checking decorator too but I don't trust python to deliver an int there unless we force it to.
Heres 1 horrible hack example:
```
>>> max(0, float("inf"))
inf
```

vs

```
>>> max(0, int(float("inf")))
  File "<stdin>", line 1, in <module>
OverflowError: cannot convert float infinity to integer
```

src/syft/core/node/domain/domain.py
madhavajay: Yes! Thats much cleaner, thanks.

src/syft/core/node/domain/domain.py
madhavajay: It is the total number of "elements" you can request from a handler rule, so you could limit it to 100 and get no requests automatically approved which exceed 100 nelements or len(), which are cumulative between requests. Its not perfect since a nested List would not work, but for Tensors which are the main concern, the nelement should work.

iamtrask (approved):

PR #4787, opened 2020-11-06
Implement a make_searchable method on Pointer

### LINKED ISSUE #4762: Ability to Toggle Searchability on Data
## Description
Some data cant be set to searchable because the method only exists on send, and other situations would benefit from the ability to potentially toggle it if the user has the correct permissions.

## Definition of Done
All Pointers can be used to toggle the searchability flag through a method assuming the user has the permission to do so.

Jasopaum:
## Description
Fixes #4762

## How has this been tested?
Manually tested and unit tests.

## Checklist


madhavajay (approved):
I tested this and it works as promised. ❤️ I also trained MNIST for a bit just to make sure all the existing permissions and requests seem to work fine.

If the DS tries to do it, it fails but the error shows up on the DO side. I think we have another example where the error is retuned. I think for this we need to maintain a separate channel of async error messages being sent back so that messages with no reply don't all need to also wait for a reply in case they create errors. We should match this up with the message_ids or something to give the user a better idea of what caused the error.

I would be cool if you could also change from searchable back to un-searchable, so you can toggle the searchability of a pointer. Perhaps we could also have a @property bool which you can just set to trigger the change where a key isnt required? Or at least the bool to toggle it. However that can wait till later and we can make a ticket so, I think this is ready for PR.

vvmnnnkv (approved):

jackhart:
Hey all,

I think this PR made a bug that I noticed when using `some_tensor.send(client, searchable=True)`.  


`SaveObjectAction` instantiates VerifyAll, while this new method returns the class.   This new line will return false with a VerifyAll object, rather than class:

> `               contains_all_in_permissions = any(
>                     key is VerifyAll for key in obj.search_permissions.keys()
>                 )`

PR #4789, opened 2020-11-06
Fixing storable object serialization and deserialization

tudorcebere:
## Description
It is a hack (at least an obvious one) until I figure out how to properly fix the current problems in ours wrappers.


madhavajay:
@tudorcebere im not sure what the status of this is, but I have updated it and added the new torch.return_types.* and the tests pass here: https://github.com/OpenMined/PySyft/tree/tudor/storable_fix

Issue #3967, opened 2020-08-08
FALCON: Private Compare


abogaziah:
## Feature Description
This function evaluates the bit x > r where r is public and
the parties hold shares of bits of x in Zp. Algorithm 1 shows
the implementation of the private compare function

![image](https://user-images.githubusercontent.com/33666625/89720197-f7f97580-d9cf-11ea-8379-a5c49787a283.png)

## Additional Context
Private compare is described in section 3.3 in [FALCON paper](https://arxiv.org/pdf/2004.02229v1.pdf)

gmuraru:
@omershlo talked with Sameer Wagh and the if-else from step 7 should be switched

omershlo:
note that a new version of the Falcon is available as they promised. 
Besides line 7 there are few other changes to the protocol (which is now at subsection 3.2)
<img width="644" alt="Screen Shot 2020-09-14 at 15 34 21" src="https://user-images.githubusercontent.com/2446179/93086599-08bc9b80-f6a0-11ea-9a64-c3375789d34b.png">

https://arxiv.org/pdf/2004.02229.pdf

PR #4792, opened 2020-11-08
[WIP]: Falcon - Private Compare

PlamenHristov:
## Description
this PR implements the private compare algorithm including the preprocessing material
it should close issue #3967

## Checklist


test/torch/mpc/falcon/test_falcon.py
```diff
@@ -26,3 +28,98 @@ def test_conv2d_private(workers):
...
+def test_private_compare(workers):
+    l = 2 ** 5
+    L = 2 ** l
+    p = 37
+
+    players = (workers["alice"], workers["bob"], workers["james"])
+    x_bit_sh = (
+        FalconHelper.decompose(torch.LongTensor([13]), L).share(*players, protocol="falcon").child
+    )
+    r_bit = (
```
omershlo, line 42: `r` is public in the protocol, no need to secret share it

test/torch/mpc/falcon/test_falcon.py
```diff
@@ -26,3 +28,98 @@ def test_conv2d_private(workers):
...
     ).all()
+
+
+def test_private_compare(workers):
+    l = 2 ** 5
+    L = 2 ** l
+    p = 37
+
+    players = (workers["alice"], workers["bob"], workers["james"])
+    x_bit_sh = (
```
omershlo, line 39: `x` is in `Zp`, not `L`

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -67,6 +82,47 @@ def xor(
...
+        """
+        return: x if beta = 0 | -x if beta = 1
+        """
+        beta_ = 1 - beta * 2
+        # The order is important if the ring sizes are different.
+        # The left operand's ring size takes precedence.
+        return x * beta_
+
+    @staticmethod
+    def private_compare_preprocess(players, p=7, k=10):
```
omershlo, line 98: This method is never used, 
it can be useful and save space to have such a method (see the repeated preprocess code in tests )

syft/frameworks/torch/mpc/falcon/falcon.py
```diff
@@ -23,3 +28,23 @@ def conv2d(filters: syft.ReplicatedSharingTensor, image, padding=0):
...
+def private_compare(x_bit_sh, r_bit, beta_b, m):
+    # 2)
+    u = FalconHelper.determine_sign(x_bit_sh - r_bit, beta_b)
+    # 3)
+    w = FalconHelper.xor(x_bit_sh, r_bit)
+    # 4)
+    wc = w.flip(-1).cumsum(-1).flip(-1) - w
+    c = u + wc + 1
+    # 6)
+    # TODO: Figure out how to do prod() privately?
```
omershlo, line 42: what seem to be the issue with prod()? 
btw, some hints can be taken from the source code used in the paper: 
https://github.com/snwagh/falcon-public/blob/7556f911baa52b320ff76852deadf20036d64b51/src/Functionalities.cpp#L1238

syft/frameworks/torch/mpc/falcon/falcon_helper.py
```diff
@@ -67,6 +82,47 @@ def xor(
             result = result.wrap()
         return result
 
+    @staticmethod
+    def determine_sign(
+        x: ReplicatedSharingTensor, beta: ReplicatedSharingTensor
+    ) -> ReplicatedSharingTensor:
+        """
+        return: x if beta = 0 | -x if beta = 1
```
omershlo, line 90: I am not sure I understand how this is done here, 
according to Line 2 in the protocol `beta` should be secret shared (over p, not 2 !!)

syft/frameworks/torch/mpc/falcon/falcon.py
gmuraru: Hmm..I think the problem here is that the reconstruction happens on the orchestrator and then we do the ```prod```.
And we should do "prod" (while it is still secretly shared) and then "reconstruction"?

PlamenHristov: What George said.

test/torch/mpc/falcon/test_falcon.py
PlamenHristov: That `p` is just a placeholder tbh. I'll delete it.

syft/frameworks/torch/mpc/falcon/falcon.py
omershlo: I understand that - but I am asking - what is the issue with doing prod on shares ?

test/torch/mpc/falcon/test_falcon.py
PlamenHristov: Yeah that's fair.

PlamenHristov:
Btw for anyone review, feedback is very much appreciated, but do keep in mind it's still work in progress.

syft/frameworks/torch/mpc/falcon/falcon_helper.py
PlamenHristov: So if the result is in `Z2` the algorithm won't work, since some of the `u` values need to be negative to take `c`=0 , which is effectively what the private compare is looking for

syft/frameworks/torch/mpc/falcon/falcon_helper.py
PlamenHristov: We do need the `private_compare_preprocess` to generate the `beta` and `m` mask

syft/frameworks/torch/mpc/falcon/falcon.py
PlamenHristov: Yeah that's the thing. I'm not doing it on the shares but on the reconstructed value.

I need to figure out how to do it in on the shares.

Issue #4767, opened 2020-11-03
Optional Authentication for Clients


harshkasyap:
## Description
https://github.com/harshkasyap/PyGrid/blob/dev/examples/model-centric/01-Create-plan.ipynb as mentioned here, authentication is optional. So, I was curious about running the notebooks without auth_token. However, at https://github.com/harshkasyap/PyGrid/blob/dev/examples/model-centric/02-ExecutePlan.ipynb, I could trace that https://github.com/OpenMined/PySyft/blob/master/syft/federated/fl_client.py auth_token looks like a required parameter.

## Additional Context
Inputs from @vvmnnnkv auth_token shouldn't be required param. Note that server_config shouldn't have auth as well in this case, otherwise, you won't authorize

vvmnnnkv:
Good catch! 👍

harshkasyap:
It looks, it could have been done very easily. I just could not think of this while raising the issue.
**client = FLClient(url=gridAddress, auth_token=None, verbose=True)**

I believe this issue can be closed without a change required. Now, the concern is whether we want auth_token as an optional parameter or go like this way only. @vvmnnnkv your thoughts?

cereallarceny:
Hello! Just letting you know that we are no longer planning on supporting anything on the 0.2.x product line and that all work should be ported over to 0.3.x, which is considered a complete rebuild of PySyft. Because of that, I’ll be closing this issue. If you feel this is a mistake, or if the issue actually applies to 0.3.x as well, please feel free to ping me on Slack and I’ll reopen the issue.

PR #4793, opened 2020-11-09
Making auth_token as optional parameter

harshkasyap:
Making auth_token as an optional parameter

#4767

### DIFF
--- a/syft/federated/fl_client.py
+++ b/syft/federated/fl_client.py
@@ -4,7 +4,7 @@ from syft.federated.fl_job import FLJob
 
 
 class FLClient:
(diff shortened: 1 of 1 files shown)


cereallarceny:
Hello! Just letting you know that we are no longer planning on supporting anything on the 0.2.x product line and that all work should be ported over to 0.3.x, which is considered a complete rebuild of PySyft. Because of that, I’ll be closing this issue. If you feel this is a mistake, or if the issue actually applies to 0.3.x as well, please feel free to ping me on Slack and I’ll reopen the issue.

PR #4812, opened 2020-11-12
Union support v2

tudorcebere:
## Description
Adding support un union types when generating the ast.

## Checklist


src/syft/lib/misc/union.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+
+        # searching in the cache to not solve the union type multiple times
+        if name in union_cache:
+            return name
+
+        target_type = UnionGenerator(name, tuple(), {})
+        union_cache.add(name)
+
+        # adding the type and it's union types to a lazy dict that will be solved
```
gmuraru, line 59: Q: Why is this a lazy dict? (it looks like a dict, right?)

tests/syft/lib/misc/union_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # the union respects Bool specific operations that don't interfere
+    # with the List method set
+    assert hasattr(union_type, "__float__")
+    assert hasattr(union_type, "__pow__")
+
+    # inspecting that function generation is respected in the pointer type as
+    # well, we care only about the pointer type. The union is acting as a
+    # proxy to the original type, it's not going to be created.
+    union_pointer_function_set = set(
```
gmuraru, line 35: Q: Do you need to convert it to a set?

tests/syft/lib/misc/union_test.py
```diff
@@ -0,0 +1,63 @@
...
+    assert "__radd__" in bool_list_function_set
+
+    # function not present in the List pointer
+    assert "__radd__" not in list_function_set
+
+    # function present in the List function set
+    assert "__radd__" in dir(sy.lib.python.List)
+
+    # the __radd__ is denied from the List pointer function set
+    # thus, neither the union pointer should have it due to anomalous behaviors
```
gmuraru, line 61: NITPICK: ```anomalous behaviors possbile``` --> ```possible anamalous behaviors```

src/syft/lib/misc/union.py
```diff
@@ -0,0 +1,71 @@
...
+        return UnionGenerator.from_qualnames(list(union_types))
+
+
+class UnionGenerator(type, metaclass=IndexableTrait):
+    """
+    UnionGenerator is a metaclass used to generate on the spot union types
+    from their qualnames.
+
+    Eg. UnionGenerator["syft.lib.python.Int", "syft.lib.python.Float"] will generate
+    in the current module the IntFloatUnion type that will have no functions added
```
gmuraru, line 25: Q: Shouldn't be ```FloatIntUnion```? (since it is sorted)

src/syft/lib/misc/union.py
tudorcebere: It's lazy in the sense that it delays the computation of the actual union types until you need to add them in the AST, not when requested. I a not sure I have better naming.

src/syft/lib/misc/union.py
tudorcebere: Totally right.

tests/syft/lib/misc/union_test.py
```diff
@@ -0,0 +1,64 @@
...
+    assert "__radd__" in bool_list_function_set
+
+    # function not present in the List pointer
+    assert "__radd__" not in list_function_set
+
+    # function present in the List function set
+    assert "__radd__" in dir(sy.lib.python.List)
+
+    # the __radd__ is denied from the List pointer function set
+    # thus, neither the union pointer should have it due to possible anamalous
```
madhavajay, line 62: spelling: anamalous  -> anomalous

tests/syft/lib/misc/union_test.py
```diff
@@ -0,0 +1,64 @@
...
+    assert "__float__" in union_pointer_function_set
+    assert "__pow__" in union_pointer_function_set
+
+    # dummy example found out in the codebase of the behavior of a denied function
+    # in the union type -> `__radd__`
+
+    # function present in the Bool pointer
+    assert "__radd__" in bool_list_function_set
+
+    # function not present in the List pointer
```
madhavajay, line 55: Two questions:
1) was __radd__ ommited by accident on List?
2) does that mean Union types have Intersection functionality rather than Union functionality. Maybe we should call them Intersection types?

madhavajay:
@tudorcebere this is really awesome work!! 👍 
I have added a PR here to your PR which enables this for multiple ops in the allowlist and fixes a small bug with the name cache which only revealed itself after adding more than 1 identical Union type.
https://github.com/OpenMined/PySyft/tree/madhava/4812-fix

I think this is ready to merge once that fix and the tests from my PR are added.

madhavajay (approved):
Works great on the remaining torch.Tensor ops from the allowlist.
We should start thinking about how to use this for collections and adding some predefined Unions like Numeric or Optionals and maybe Failable types which return an Exception which can be raised on the callers side.

tests/syft/lib/misc/union_test.py
tudorcebere: It's not really intersection either. It's a Union \ {functions that are denied from one of the underlying pointer types}.

tudorcebere: Regarding radd, it's left out by accident (it's not usefull at all right now, as we don't support reversed operators and lists have addition only with other lists, so it's a direct add there)

madhavajay: Lucky example then. I guess we can address it in the separate Binary Reverse Operators Issue we discussed.

PR #4827, opened 2020-11-18
Fix StorableObject lossing .data .description .tags after serde

xutongye:
## Description
Fixes OpenMined#4791

Issue OpenMined#4791 says that in `PySyft/syft_0.3.0 branch`, a `StorableObject` will loss it's `.data`, `.description` and `.tags` attributes if you serialize it into bytes and then deserialize it back. But as what I found, this issue also happens in `PySyft/master` branch, so I direct this PR to master branch. And I found that this issue also exists when you serialize a `StorableObject` to `proto`, not only when to bytes.

What I found is that after serialization, the `.description` and `.tags` information disappear in the bytes or proto. And in the descrialization process, the `.data` information is dropped.

Why does the  `.description` and `.tags` information disappear in the serialization process? The problem is in the `_object2proto` method of `StorableObject`. In this function, at "Step 4: save the description into proto", it checks `if hasattr(self.data, "description")`. And as `hasattr(self.data, "description")` is `False`, no `description` attribute is set to proto. So, what I think it should be checking is `if hasattr(self, "description")`. And this is also the thing happens at "Step 5: save tags into proto if they exist". So that's the reason why the  `.description` and `.tags` information disappear in the serialization process, and how to fix it.

Then why the `.data` information is dropped in the deserialization process? The problem is in the `get_data_protobuf_schema` method of `StorableObject`. In another function `_proto2object`, this function `get_data_protobuf_schema` is called to get the schematic_type of `StorableObject`. As `get_data_protobuf_schema` returns `None`, the `data` information is dropped. So what I do is I change this `get_data_protobuf_schema`  function to let it return a `StorableObject_PB`. 

So that's how I fix this bug.

But there is one thing I am not sure about. These codes that I found leading to this bug, they seem to be written that way on purpose. Because they do not seem like typos. So I am not sure what consequences changing them.

## Affected Dependencies

## How has this been tested?
- Add a test function in `storable_test.py'
- Assert that the '.id', '.data', '.description' and '.tags' are the same between the initial obj and after serde it.

## Checklist
- [ x] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [x ] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [x ] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [x ] My changes are covered by tests


tudorcebere:
LGTM, we'll be working on refactoring the Store API soon, but this brings a quality of life fix until then. One small linting error and you are good to go.

tudorcebere (approved):

madhavajay:
Hi @xutongye thank you so much for the quick fix and contribution. I apologise for the delay in responding we have been a little busy with the switch over. As @tudorcebere mentioned we have some refactors coming for this part of the code as it stands I don't think our code actually initialises StorableObject without it being from a subclass using the `construct_new_object` `staticmethod`. But this itself is not how we want it to be so we fully acknowledge this is confusing and ugly.

Regarding your patch. The only thing left to update is the linting errors.
You can see them here:
https://github.com/OpenMined/PySyft/pull/4827/checks?check_run_id=1441012237
```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

src/syft/core/store/storeable_object.py:106: error: unused 'type: ignore' comment
src/syft/core/store/storeable_object.py:111: error: unused 'type: ignore' comment
src/syft/core/store/storeable_object.py:150: error: Left operand of 'and' is always true
src/syft/core/store/storeable_object.py:157: error: Statement is unreachable
Found 4 errors in 1 file (checked 204 source files)
```

We highly recommend contributors run the local checks before pushing so they know their code will pass the lint checks on CI.

I have an open PR which aims to fix the contribution guide to make this clearer:
https://github.com/OpenMined/PySyft/pull/4837

In this case the fix to mypy's complaints are a little more complex so I have provided what might be a workable solution via a .patch file:
https://pastebin.com/THaT2zv0

xutongye:
> Hi @xutongye thank you so much for the quick fix and contribution. I apologise for the delay in responding we have been a little busy with the switch over. As @tudorcebere mentioned we have some refactors coming for this part of the code as it stands I don't think our code actually initialises StorableObject without it being from a subclass using the `construct_new_object` `staticmethod`. But this itself is not how we want it to be so we fully acknowledge this is confusing and ugly.
> 
> Regarding your patch. The only thing left to update is the linting errors.
> You can see them here:
> https://github.com/OpenMined/PySyft/pull/4827/checks?check_run_id=1441012237
> 
> ```
> mypy.....................................................................Failed
> - hook id: mypy
> - exit code: 1
> 
> src/syft/core/store/storeable_object.py:106: error: unused 'type: ignore' comment
> src/syft/core/store/storeable_object.py:111: error: unused 'type: ignore' comment
> src/syft/core/store/storeable_object.py:150: error: Left operand of 'and' is always true
> src/syft/core/store/storeable_object.py:157: error: Statement is unreachable
> Found 4 errors in 1 file (checked 204 source files)
> ```
> 
> We highly recommend contributors run the local checks before pushing so they know their code will pass the lint checks on CI.
> 
> I have an open PR which aims to fix the contribution guide to make this clearer:
> #4837
> 
> In this case the fix to mypy's complaints are a little more complex so I have provided what might be a workable solution via a .patch file:
> https://pastebin.com/THaT2zv0

Hi @madhavajay , thinks for pointing me to the new contributing guide, it helps a lot. 
I did not run pytest and pre-commit checks to my previous commits, so a lot of tests failed when I push new commits to PRs. So, according to your new contributing guide,  I reinstalled all these things. Feels good to pass all the checks. 
But there is one thing, I can run `pytest -m fast` well, but when I run `pytest -m fast -n auto`, error happened:
```python
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: -n
  inifile: /mnt/d/github/PySyft/setup.cfg
  rootdir: /mnt/d/github/PySyft
```
Should I worry about this? Or can you please give some hints how to fix this?

xutongye:
> LGTM, we'll be working on refactoring the Store API soon, but this brings a quality of life fix until then. One small linting error and you are good to go.

Sure, thanks.

madhavajay:
Hi @xutongye, the issue is due to the missing pytest-xdist package, which provides the `-n auto` option.
The package is in requirements.txt but is not in setup.cfg as an install_requires= dependency.

While it probably should be added to the testing= section of the setup.cfg from my understanding that won't install it using `pip install -e .` and we are using pytest directly to give a bit more control over how we run our tests.

Currently this is covered in the CONTRIBUTING during the `pipenv` install step but it should be made more clear with alternative steps that there are development dependencies in the `requirements.txt`. Its a pity there is no good standard here since both `poetry` and `pipenv` allow for dev dependencies but do not share a format. 😕

I will add it to setup.cfg testing= now just to make things a little more robust.

madhavajay:
Related PR: https://github.com/OpenMined/PySyft/pull/4860

PR #4843, opened 2020-11-23
Moving  Module and VerifyKeyWrapper

tudorcebere:
Small refactor to get rid of `torch` in core and other quality of life moving to reduce the change of circular imports. More to come!

## Affected Dependencies

## How has this been tested?

## Checklist


src/syft/lib/torch/module.py
```diff
@@ -94,13 +94,11 @@ def setup(self, torch_ref: Any) -> None:
         self.duet = None
         if "syft" in full_name_with_qualname(klass=type(torch_ref)):
             log = "> Creating remote model"
-            print(log)
```
madhavajay, line 97: I think these two prints were intentionally left in so they come out on the stdout of the notebook. But this raises an important question. Should we at the minimum create an alias for this like console() or something similar so that its clear we intend to output this somehow to the user not as a debug print statement?

madhavajay (approved):
LGTM!

PR #4859, opened 2020-11-25
Add warnings for implicit CUDA-to-CPU shift, closing #4804

### LINKED ISSUE #4804: Print warning if calling fix_precision() on a CUDA-tensor
## Description
Currently when calling fix_precision() on a CUDA tensor will automatically first shift that tensor to the CPU to then call fix_precision() on it. It would be useful to issue a warning whenever this happens to warn the user that he might have intended to let the tensor be implicitly shifted to the CPU. 

## Are you interested in working on this improvement yourself?
If nobody is interested in the next week I could do it. 

## Additional Context
As discussed with @gmuraru and @LaRiffle

NiWaRe:
## Description
Closes #4804. 
I added the warning logs directly to the places where the implicit CUDA shifting occurs. This way we won't have a problem finding CUDA warnings once we introduce solutions to use CUDA with encryption (if one day we do). 

As a note to the issue - `fix_precision()` doesn't cause an implicit shift to the CPU but two other operations (see comments in code for more on that).

## Affected Dependencies
- /syft/frameworks/torch/tensors/interpreters/paillier.py needs to import `logging` 
- /syft/frameworks/torch/tensors/interpreters/additive_shared.py needs to import `logging` 

## How has this been tested?
- I checked for implicit shifting when SMPC-encrypting using `AdditiveSharingTensor` and when HE-encrypting using `PaillierTensor` on Google Colab, building my changed syft repo from source. 
- The warning is successfully outputed (using `logging.warning()` ) only if I try to encrypt CUDA tensors. 

## Checklist


syft/frameworks/torch/tensors/interpreters/additive_shared.py
```diff
@@ -344,6 +345,10 @@ def generate_shares(self, secret, n_workers, random_type):
         """
         random_type = torch.LongTensor if random_type == torch.int64 else torch.IntTensor
         if not isinstance(secret, random_type):
+            if "cuda" in secret.type():
```
gmuraru, line 348: Q: Can't we use something like ```secret.device``` or hmm...use ```isinstance``` (I am not sure what to check against with isinstance)?

gmuraru (approved):

syft/frameworks/torch/tensors/interpreters/additive_shared.py
NiWaRe: Yes `isinstance` was a bit too narrow to capture multiple cuda types, and yes using the `device` property could also work. Initially, I thought that maybe also numpy arrays could wind up there and they don't have a `device` property (`type()` is more python native/general), but yes actually these are always tensors in these specific places, so I could also change it, what do you think?

gmuraru: Hmm, I think it would make more sense to treat them as tensors (I am sorry for the late reply :( )

NiWaRe: Alright, I'll change it to use the `device` property then.

NiWaRe: Out of curiosity, what is the benefit of that compared to the current version? @gmuraru

gmuraru: From my point of view if only tensors should reach this point:
* it is a safety mechanism to catch bugs
* it is also (I think) more intuitive to treat all the values as tensors since this file is in **_tensors_**/interpreters/

But, I do not have a strong opinion about this, this also looks good :+1:

PR #4864, opened 2020-11-27
Added Union support to allowlist_test

madhavajay:
## Description
- Added item, max, median, min, stride with Union types
- Added comments about List types which are really Tuples
- Added torch.return_types.* namedtuple types for max, median and min
- Fixed bug in UnionGenerator where short name was cached not fqn

## Affected Dependencies
None

## How has this been tested?
Locally and CI

## Checklist

### DIFF
--- a/src/syft/lib/misc/union.py
+++ b/src/syft/lib/misc/union.py
@@ -52,12 +52,12 @@ class UnionGenerator(type, metaclass=IndexableTrait):
             return name
 
         target_type = UnionGenerator(name, tuple(), {})
-        union_cache.add(name)
 
         # adding the type and it's union types to a lazy dict that will be solved
         # when the ast is created.
         lazy_pairing[target_type] = union_types
         qualname = target_type.__qualname__
+        union_cache.add(qualname)
         return qualname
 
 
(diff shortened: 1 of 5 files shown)


src/syft/lib/misc/union.py
```diff
@@ -52,12 +52,12 @@ def from_qualnames(union_types: List[str]) -> str:
             return name
 
         target_type = UnionGenerator(name, tuple(), {})
-        union_cache.add(name)
 
         # adding the type and it's union types to a lazy dict that will be solved
         # when the ast is created.
         lazy_pairing[target_type] = union_types
         qualname = target_type.__qualname__
+        union_cache.add(qualname)
```
tudorcebere, line 60: Hmmm, you may by right, but why? Is there a case where there are created same unions with different paths?

tudorcebere (approved):
LGTM! Got only 1 question.

src/syft/lib/misc/union.py
madhavajay: It was broken because it cached the name not the qual name and the resultant name couldn't resolve the full Union class path, but it didn't happen until we created two of the same Union which my PR required in the allowlist.

```
-        union_cache.add(name)
+        union_cache.add(qualname)
```

tudorcebere: OOH, cool!

PR #4878, opened 2020-12-01
Fixed issue where static classes can't be invoked via RPC

madhavajay:
## Description
- Fixed issue where static classes can't be invoked via RPC
- If a class init fails because its not available we can store a None
- When executing the static method we can detect the None and remove it

## Affected Dependencies
None

## How has this been tested?
Locally and on PSI branch

## Checklist


gmuraru:
This is for static classes? or static methods?

madhavajay:
> This is for static classes? or static methods?

We need both but they are different problems I guess, in this instance we need to enable the path to the static class without a constructor, so the issue is, allowing static functions on any level of the module / class hierarchy.

Here you can see an example of how the code in this PR fails:
Screen Shot 2020-12-04 at 9 34 17 am

We create a None and send back a correct Pointer which is fine, but if any other triggers the store to regenerate the pointer such as searching the store we will get a SyNone.

Screen Shot 2020-12-04 at 9 27 45 am

Screen Shot 2020-12-04 at 9 28 12 am

madhavajay:
Closed in favour of the new AST Refactor.

PR #4912, opened 2020-12-15
[WIP] allow user pass in function to handle exchange ids in Duet initialization(OpenMined#4909)

xutongye:
## Description
Currently, to initialize duet connection, we have two options:
- manually insert server/client ID
- loopback

But we may need some more flexible way. Because, for example, we may want to use some cryptography technologies to exchange the server/client ID. This PR aims to introducing a flexible way to exchange server/client ID. 

The main thing this PR introduces is to add a callable argument `exchangeIds_method: Callable` to `sy.launch_duet` and `sy.join_duet` function. Users can pass in any function as long as it takes a string (id_to_send) as argument and returns a string (id_received). Inside `sy.launch_duet` and `sy.join_duet`, this passed in function will be called to send an ID and get an ID.

So now there are three ways to exchange ids:
1. `loopback=True`, this will be the loopback mode as it was
2. `loopback=False` [default], and no `exchangeIds_method` passed in, this will be the manual way
3. `loopback=False` [default], and an `exchangeIds_method` is passed in, this will be a user defined way

## Affected Dependencies
The default manual way is a bit different. The old way was that the client needs to pass in server ID when call `.join_duet`. The new way is that the client needs not to do that any more. Instead, you are asked to input the server ID with an input prompt. 

## How has this been tested?
Open jupyter notebook and run the following three examples: 
1. first example:
```python
#-------------------------
# Server side
import syft as sy
duet = sy.launch_duet()

#-------------------------
# Client side
import syft as fy
duet = sy.join_duet()
```
2. second example:
```python
#-------------------------
# Server side
import syft as sy
duet = sy.launch_duet(loopback=True)

#-------------------------
# Client side
import syft as fy
duet = sy.join_duet(loopback=True)
```
3. third example:
```python
#-------------------------
# Server side
import syft as sy
def any_func_1(id_to_send: str) -> str:
    '''
    some code
    '''
    return id_received

duet = sy.launch_duet(exchangeIds_method=any_func_1)

#-------------------------
# Client side
import syft as fy
def any_func_2(id_to_send: str) -> str:
    '''
    some code
    '''
    return id_received
duet = sy.join_duet(exchangeIds_method=any_func_2)
```

## Checklist


madhavajay:
This is awesome work! 👍🏼

src/syft/grid/duet/__init__.py
```diff
@@ -3,18 +3,20 @@
...
 
 # third party
 import nest_asyncio
 import requests
 
 # syft relative
+from . import loopback_exchangeIds
+from . import manual_exchangeIds
 from ...core.node.domain.domain import Domain
+from .manual_exchangeIds import bcolors
```
madhavajay, line 19: I think we should keep this 1 level above in its own file called `bcolors` and possibly refactor the "♫♫♫ text" into a separate set of strings so it's easier to change or translate and the code is cleaner.

src/syft/grid/duet/__init__.py
```diff
@@ -295,6 +224,7 @@ def join_duet(
     target_id: str = "",
     network_url: str = "",
     loopback: bool = False,
+    exchangeIds_method: Callable = manual_exchangeIds.client_get_then_send,
```
madhavajay, line 227: Same here, change to 🐍 `snake_case` for variables.
The place for 🐫 `CamelCase` would be for `ClassNames`.

madhavajay (changes requested):
Looking really good, just a few small adjustments. 👍🏼

madhavajay:
I think it would be great if we went one step further and refactored the manual and loopback functions into a nice interface like this:
```python
import os
from typing import Any as TypeAny
from typing import Union as TypeUnion

class DuetCredentialExchanger:
    def __init__(self, credential: TypeAny) -> None:
        self.credential = credential

    def run(self) -> TypeAny:
        raise NotImplementedError


class OpenGridTokenManualInputExchanger(DuetCredentialExchanger):
    # join=True is the Client and join=False is the Server
    def __init__(self, credential: str, join: bool = False) -> None:
        self.credential = credential

    def run(self) -> str:
        ...
    
class OpenGridTokenFileExchanger(DuetCredentialExchanger)
    def __init__(self, credential: str, join: bool = False, file_path: Union[str, os.PathLike]) -> None:
        self.credential = credential
        self.file_path = file_path

    def run(self) -> str:
        ...


```

madhavajay:
@wip-abramson I believe "These are the Droids you are looking for?" 🧙🏼‍♂️

src/syft/grid/duet/__init__.py
```diff
@@ -3,18 +3,20 @@
...
 from typing import Optional
+from typing import Callable
 
 # third party
 import nest_asyncio
 import requests
 
 # syft relative
+from . import loopback_exchangeIds
+from . import manual_exchangeIds
```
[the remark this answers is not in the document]
xutongye: resolved in a new commit

src/syft/grid/duet/__init__.py
xutongye: - bcolors: resolved in a new commit
- refactor the "♫♫♫ text": TODO

src/syft/grid/duet/__init__.py
xutongye: resolved in a new commit

xutongye:
> I think it would be great if we went one step further and refactored the manual and loopback functions into a nice interface like this:
> 
> ```python
> import os
> from typing import Any as TypeAny
> from typing import Union as TypeUnion
> 
> class DuetCredentialExchanger:
>     def __init__(self, credential: TypeAny) -> None:
>         self.credential = credential
> 
>     def run(self) -> TypeAny:
>         raise NotImplementedError
> 
> 
> class OpenGridTokenManualInputExchanger(DuetCredentialExchanger):
>     # join=True is the Client and join=False is the Server
>     def __init__(self, credential: str, join: bool = False) -> None:
>         self.credential = credential
> 
>     def run(self) -> str:
>         ...
>     
> class OpenGridTokenFileExchanger(DuetCredentialExchanger)
>     def __init__(self, credential: str, join: bool = False, file_path: Union[str, os.PathLike]) -> None:
>         self.credential = credential
>         self.file_path = file_path
> 
>     def run(self) -> str:
>         ...
> ```

done in a new commit @madhavajay

madhavajay:
We can add some additional checks in this PR to make sure all these connection permutations are tested.
https://github.com/OpenMined/PySyft/pull/4889/files

madhavajay (approved):
Looks good!

tudorcebere:
LGTM!

PR #4926, opened 2020-12-17
added location and locations printouts to remote and SMPC-encrypted nn.Module

NiWaRe:
## Description
This adds a new `__str__()` function to `torch.nn.Module` which adds a location printout to every shared model. In the case of an SMPC-encrypted model it outputs all the different locations. Should serve as a small convenience function.

## Affected Dependencies
None.

## How has this been tested?
- Tested the output of normal tensors, a remote tensor, a SMPC-encrypted tensor, a normal nn.Module, a nn.Module with FixedPrecisionTensor weights, a remote nn.Module and an SMPC-encrypted nn.Module.

## Checklist

### DIFF
--- a/syft/frameworks/torch/hook/hook.py
+++ b/syft/frameworks/torch/hook/hook.py
@@ -851,6 +851,26 @@ class TorchHook(FrameworkHook):
 
         self.torch.nn.Module.location = location
 
+        # print the location
+        native___str__ = self.torch.nn.Module.__str__
+
+        def model___str__(nn_self):
+            out = ""
+            # shared_encryption tensor doesn't have a location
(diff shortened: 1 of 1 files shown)


NiWaRe:
There are these three other issues that also exist, I could take a look or, given that I already have another task, leave it for now. What do you think @LaRiffle @gmuraru? 

* Further possible tasks 
  1. add 'locations' property to additiveSharedModels (same as 'location' for normal remote models)
  2. suppress 'location' property in normal tensors 
  3. FixedPrecisionTensor.child has a location property (should only have if AdditiveSharedTensor as child)

LaRiffle (approved):
Perfect! Thanks for the fix :)

Issue #4898, opened 2020-12-10
Add the device type to Tensor serialization


LaRiffle:
## Feature Description
We would like to send the information about is the data should be stored in the CPU or the GPU in the protobuf file: https://github.com/OpenMined/PySyft/blob/dev/proto/lib/torch/tensor.proto

madhavajay:
I believe this is done?

Issue #4899, opened 2020-12-10
Check in data2object whether the device of the sent tensor is compatible with current machine.  


NiWaRe:
## Description
In deserialization before creating a tensor check whether the current machine can support the device type that was sent - if a CUDA tensor was sent make sure that if the machine hasn't got a GPU the tensor device is set to CPU first. 

## Are you interested in working on this improvement yourself?
- Yes, I am.

## Additional Context
Should be solved at the same time as #4898

NiWaRe:
Solved by #5053.

PR #4939, opened 2020-12-23
[WIP] Add device info to msg and add checks in deserialization 

NiWaRe:
## Description
Solves #4898 and #4899 - adds an `on_cuda` field of type bool to msg, saves the current device type as on_cuda bool before serialization, and checks during deserialization whether CUDA should be used and/or if the current machine has CUDA. 

## Affected Dependencies
None

## How has this been tested?
- Still has to be done

## Checklist


NiWaRe:
Possible changes: 

- In serialization by default shift on CPU and in deserialization only check for CUDA. 
- In serialization, an extra feature could be implemented which allows the sender to specify CUDA even if he/she doesn't have CUDA him/herself. 

Also, I still have to think about testing.

madhavajay:
@NiWaRe This is awesome! 👍🏼 

Do you think this would be even better if we supported `torch.device` properly then just added that to the tensor protobuf as well?
```
>>> torch.randn((2,3), device=torch.device('cuda:1'))
>>> torch.device(1)
device(type='cuda', index=1)
```

That way we can support the optional ordinal of the cuda device as well, and easily expand to other device types if they get added in future.

All we need is to add a protobuf wrapper and then set it onto the real torch.device class like so:
```
# syft/lib/torch/__init__.py
from . import parameter  # noqa: 401
from . import uppercase_tensor  # noqa: 401
from . import device  # noqa: 401  < ---- add this
```

```
# hypothetical device.py
aggressive_set_attr(
    obj=device_type, name="serializable_wrapper_type", attr=DeviceWrapper
)

```

NiWaRe:
Thanks, @madhavajay! 
Sry for the late response I was pretty busy with exam prep lately. 
Your suggestion sounds like a great idea! I have to admit that I'm just getting started with the protobuf serialization.
So to add the device type directly into the serialization do you mean to create a file device.py that looks similar to the current uppercase_tensor.py? Smt like that: 

```
# device.py
import [...]

device_type = torch.device(1)

class DeviceWrapper(StorableObject): 
    def __init__(self, device : object): 
        super().__init__(
           [...]
        )
        self.device = device
    def _data_object2proto(self) -> Tensor_PB:
       proto = Tensor_PB()
       # leave as bool to be more efficient than a string? 
       proto.on_cuda = getattr(self.device, "type", "cpu") == "cuda"
       # new int in tensor.proto
       proto.index = getattr(self.device, "index", 0)
    
   [...]

    @staticmethod
    def _data_proto2object(proto: Tensor_PB) -> th.device:
        tensor = protobuf_tensor_deserializer(proto.tensor)
        [...]

        if proto.on_cuda is True and th.cuda.is_available():
            type = "cuda"
        else:
            type = "cpu" 
        index = proto.index

        device = torch.device(type, index)

        return device

aggressive_set_attr(
    obj=device_type, name="serializable_wrapper_type", attr=DeviceWrapper
)
```

Tbh I don't quite understand what's going on in the original uppercase_tensor.py would you mind clarifying some things? :) 

- why is the file called 'uppercase' ? ^^
- the purpose of the file: creating a wrapperclass for serialization and deserialization and setting that as an attribute of a normal tensor? Why do we do that and how is that then applied to every torch tensor that is created? Is it because somewhere else we call e.g. the `_data_object2proto` function on a tensor before sending it? 
- concerning this DeviceWrapper Class - given that I didn't misunderstand you completely ^^ - why couldn't we just define an field `int index = 4`? And how will be the Wrapper be used as an attribute? (about same questions as the previous one)

Still struggling a bit with the bigger picture, thanks for your help :)

madhavajay:
@NiWaRe your implementation looks correct! 👍🏼 
Good questions:

> why is the file called 'uppercase' ? ^^
> the purpose of the file: creating a wrapperclass for serialization and deserialization and setting that as an attribute of a normal tensor?

It's called uppercase because its the torch.Tensor class not the torch.tensor function.

> Why do we do that and how is that then applied to every torch tensor that is created? Is it because somewhere else we call e.g. the _data_object2proto function on a tensor before sending it?

The `aggressive_set_attr` is modifying the class to have an attribute `serializable_wrapper_type` which points to the Wrapper class. When syft finds a foreign object and wants to serialize it checks if this attribute exists and if it does it uses that class and its methods to do the serde.

> concerning this DeviceWrapper Class - given that I didn't misunderstand you completely ^^ - why couldn't we just define an field int index = 4? And how will be the Wrapper be used as an attribute? (about same questions as the previous one)

The advantage of separating the Device from the Tensor as its own type means that users can construct a `Device` and then re-use that same pointer / variable as input on other Tensors or methods which take a device as input. If Device ever changes we can update it separately from Tensor itself. As we want to cover as much of the Torch API as possible it makes sense for us to support `device` so adding it now would be really appreciated. 😊

Let me know if you have any more questions.

NiWaRe:
@madhavajay thanks for the answers! 
Sorry for not progressing much lately (exams coming up at university ^^), I saw that @xutongye also did a PR to resolve #4898, that means this isn't useful anymore right?

madhavajay:
@NiWaRe No worries, thank you for the PR anyway. I hope the exams are going well. 😊

PR #5006, opened 2021-01-12
Fix asyncio CLI / Interactive issues

madhavajay:
## Description
This PR fixes the issues with the event loop in CLI mode and addresses some other related issues.

## Affected Dependencies
None

## How has this been tested?
CLI, Jupyter and CI

## Checklist

### DIFF
--- a/src/syft/core/common/event_loop.py
+++ b/src/syft/core/common/event_loop.py
@@ -7,6 +7,7 @@ from typing import Dict as TypeDict
 from typing import Optional
 
 # third party
+from loguru import logger
 import nest_asyncio
 
 # syft relative
@@ -34,7 +35,6 @@ try:
 except RuntimeError:
     pass
 
-
 # no event loop found, lets create one so that future asyncio imports use the same loop
 if loop is None:
     loop = asyncio.new_event_loop()
@@ -45,10 +45,7 @@ asyncio._set_running_loop(loop)
 
 # https://github.com/erdewit/nest_asyncio
 # patch the event loop to allow nested event loops
-try:
-    nest_asyncio.apply(loop)
-except Exception as e:
-    print(f"Failed to patch event loop with nest_asyncio {e}")
+nest_asyncio.apply(loop)
 
 
 def loop_in_thread(loop: TypeAny) -> None:
@@ -61,10 +58,10 @@ class EventLoopThread:
 
     def __init__(self, loop: TypeAny) -> None:
         if "loop" not in self.__shared_state:
-            print("Starting Event Loop")
+            logger.info("Starting Event Loop")
             self.__shared_state["loop"] = loop
         if "thread" not in self.__shared_state:
-            print("Starting Event Loop Thread")
+            logger.info("Starting Event Loop Thread")
             # daemon=True needed to allow REPL exit() to terminate the thread
             # stdlib
             import threading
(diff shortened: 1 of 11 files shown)


src/syft/core/common/event_loop.py
```diff
@@ -0,0 +1,115 @@
...
+# set the event loop so its the same everywhere
+asyncio._set_running_loop(loop)
+
+
+# https://github.com/erdewit/nest_asyncio
+# patch the event loop to allow nested event loops
+try:
+    nest_asyncio.apply(loop)
+except Exception as e:
+    print(f"Failed to patch event loop with nest_asyncio {e}")
```
gmuraru, line 51: Q: Shouldn't we raise the exception?

src/syft/core/common/event_loop.py
```diff
@@ -0,0 +1,115 @@
...
+except ImportError:  # pragma: no cover
+    # stdlib
+    from asyncio.events import _get_running_loop as get_running_loop  # pragma: no cover
+
+# first lets get any existing event loop, this will throw an exception in if there is
+# no event loop or will return none in older versions of python
+loop = None
+try:
+    loop = get_running_loop()
+except RuntimeError:
```
gmuraru, line 34: Q: Print a warning? with the error?

src/syft/core/common/event_loop.py
```diff
@@ -0,0 +1,115 @@
...
+        self.shutdown()
+
+
+event_loop_thread: Optional[EventLoopThread] = None
+
+# allow threads to be disabled in REPL which will require an event loop to be started
+# before running import syft, an example use case is the SCONE container which has
+# threads disabled
+thread_env = os.environ.get("SYFT_USE_EVENT_LOOP_THREAD", "1")
+SYFT_USE_EVENT_LOOP_THREAD = not (thread_env == "0" or thread_env.lower() == "false")
```
gmuraru, line 100: NIT: ```thread_env.lower() not in {"0", "false"}```

gmuraru (approved):

src/syft/core/common/event_loop.py
madhavajay: I think this only happens in python 3.7+ if there is no loop and in python 3.6 it returns None, so we just try to get it and if it fails we create a new one anyway.

src/syft/core/common/event_loop.py
madhavajay: I added this to try to catch an exception which happens sometimes in python 3.6.
```
Exception in callback Task._wakeup(<Future finis...2.253.59.127'>)
handle: <Handle Task._wakeup(<Future finis...2.253.59.127'>)>
  File "/Users/madhavajay/.pyenv/versions/3.6.9/lib/python3.6/asyncio/events.py", line 145, in _run
    self._callback(*self._args)
  File "/Users/madhavajay/.pyenv/versions/3.6.9/lib/python3.6/asyncio/tasks.py", line 276, in _wakeup
    self._step()
  File "/Users/madhavajay/.local/share/virtualenvs/PySyft-lHlz_cKe/lib/python3.6/site-packages/nest_asyncio.py", line 159, in step
    step_orig(task, exc)
  File "/Users/madhavajay/.pyenv/versions/3.6.9/lib/python3.6/asyncio/tasks.py", line 260, in _step
    self.__class__._current_tasks.pop(self._loop)
KeyError: <_UnixSelectorEventLoop running=True closed=False debug=False>
```

It seems to be related to nest_asyncio patching asyncio and possibly due to us using the event loop in a new thread. The try except doesn't catch it nor does one inside the thread code. I tried a bunch of ways to catch or prevent this exception but I cant. It also seems to have no negative effect on the application running so since its for python 3.6 im just going to leave it for now.

I have removed the try except because theres nothing for it to catch anyway.

PR #5017, opened 2021-01-15
Improve request handler

### LINKED ISSUE #4799: Request Handler Improvements
## Description
Request Handlers have IDs and can be referenced for deletion by ID.
Request Handlers can be updated by ID and some care is taken to prevent conflicting / duplicate request handlers.
Request Handlers can be displayed with Pandas in a table, including how much longer they will be active.
Active Request Handler count is displayed in the Dashboard update status with Message Count and Objects etc.
Request Handlers should be able to respond to data with DO specified tags or some kind of derived tagging after ops.

## Definition of Done
The above functionality works and is tested.

xutongye:
## Description

This may close #4799 .

---
#### 1. Replace `RequestMessage.name` with `.object_tags`
Currently, `RequestMessage ` has a `.name` attribute. And this `.name` attribute are used to match requesthandlers and requests. But this is not very reasonable. Beacuse the `.name` of a request is given by DS, when `some_ptr.request` is called. But the DS can give a meaningless or even misleading `.name`.

So, the first thing this PR does is replace `RequestMessage.name` with `.object_tags`, which is the same as `.tags` attribute of the object this request asks for. This is more reasonable and safe, because the `.tags` of object is given by DO, and the DO can make sure that it's meaningful.

#### 2. improve `duet.requests.handlers`
This PR introduce some new API's to manage request handlers:
- We can call `duet.requests.handlers`, like `duet.store` and `duet.requests`. 
- We can call `duet.requests.handlers.pandas`, like `duet.store.pandas` and `duet.requests.pandas`
- When call `duet.requests.remove_handler`, you pass in a `key`, which can be an `index:int`, or a `tag:str`, or an `id:str`.

#### 3. handler["tags"]
Currently, a request handler has a `name`, and it's used to match a handler with a request. This PR replace it with `tags`, and math a handler and a request if `set(handler["tags"])==set(request.object_tags)`.  And a handler will match all requests if it has `tags=[]`.

#### 4. Show `object_type` in `.pandas`
When call `duet.store.pandas`, it would be good to know the type of the underlying object. And it would be good to know the type of the requested object when call `duet.requests.pandas`.

This PR let you see this "object_type" information when call `.pandas`:
- When call `duet.store.pandas`, there is a new column named `object_type`
- When call `duet.requests.pandas`, there is a new column named `Requested Object's type`

#### 5. "Request Handlers: #" in duet live status
The number of request_handlers is shown in duet's live status

#### 6. "remaining time (s)" in `handlers.pandas`
When call `duet.requests.handlers.pandas`, the "remaining time (s)" information of each handler is shown.

#### 7. Handle requests to results of `__len__`
The remote  iteration operation needs the DO to automatically accept requests to results of `__len__()` of some iterable objects. To do that, on DO side:
1. When send an iterable object to duet, pass in some tags, for example `tags=["tag_x"]`
2. Then add a handler with `action="accept"` and `tags=["tag_x", "__len__"]`, this means that the DO will accept all requests to the results of `__len__()` called on objects whose tags is ["tag_x"]

More comming...

## Affected Dependencies
- The API of `Pointer.request` method changed. There is no `name` argument anymore.
- The API of `Pointer.get` method changed. There is no `name` argument anymore.
- The API of `Module.get` method changed. There is no `name` argument anymore.
- The things displayed when `duet.requests.pandas` is called changed. The "name" column is removed, and a "Requested Object's tags" column is added.

## How has this been tested?
1. Two jupyter notebooks are added to test all of the features this PR intrduces.
    - `examples/experimental/request_handlers/DO.ipynb`
    - `examples/experimental/request_handlers/DS.ipynb`

## Checklist


src/syft/core/node/domain/client.py
```diff
@@ -70,62 +71,60 @@ def __getitem__(self, key: Union[str, int]) -> RequestMessage:
...
+        ]
+        return pd.DataFrame(request_lines)
+
     def add_handler(
         self,
         action: str,
         print_local: bool = False,
         log_local: bool = False,
-        name: Optional[str] = None,
+        tags: List[str] = [],
```
tudorcebere, line 93: Not safe/a good practice: https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

src/syft/core/node/domain/client.py
```diff
@@ -70,62 +71,60 @@ def __getitem__(self, key: Union[str, int]) -> RequestMessage:
...
-    ) -> None:
-        handler_opts = self._validate_options(
-            action=action,
-            print_local=print_local,
-            log_local=log_local,
-            name=name,
-            timeout_secs=timeout_secs,
-            element_quota=element_quota,
-        )
+    def remove_handler(self, key: Union[str, int]) -> None:
```
tudorcebere, line 109: Love it!!

src/syft/core/node/domain/client.py
```diff
@@ -70,62 +71,60 @@ def __getitem__(self, key: Union[str, int]) -> RequestMessage:
...
+            id_str = str(handler["id"].value).replace("-", "")
+            self.remove_handler(id_str)
 
     def _validate_options(
         self,
         action: str,
         print_local: bool = False,
         log_local: bool = False,
-        name: Optional[str] = None,
+        tags: List[str] = [],
```
tudorcebere, line 124: Same problem as above.

src/syft/core/node/domain/service/request_handler_service.py
```diff
@@ -213,13 +213,18 @@ def _object2proto(self) -> GetAllRequestHandlersResponseMessage_PB:
             the other public serialization methods if you wish to serialize an
             object.
         """
+        # For handler["created_time"], it's a large number. In order to keep it's precision
```
tudorcebere, line 216: I don't think we need such precision + 64 bits might be enough? Can't we make this time relative to the process start or the current time (only seconds precision should be enough). If not, encode it as a string directly?

src/syft/core/node/domain/service/request_handler_service.py
```diff
@@ -238,13 +243,15 @@ def _proto2object(
             This method is purely an internal method. Please use syft.deserialize()
             if you wish to deserialize an object.
         """
+        handlers = [upcast(value=Dict._proto2object(proto=x)) for x in proto.handlers]
```
tudorcebere, line 246: Same problem as above

src/syft/core/node/domain/service/request_message.py
```diff
@@ -46,7 +46,8 @@ def __init__(
         address: Address,
         requester_verify_key: VerifyKey,
         owner_address: Address,
-        name: str = "",
+        object_tags: List[str] = [],
```
tudorcebere, line 49: same problem as above

src/syft/core/node/domain/service/request_message.py
```diff
@@ -207,4 +211,11 @@ def process(node: AbstractNode, msg: RequestMessage, verify_key: VerifyKey) -> N
 
         # using the local arrival time we can expire the request
         msg.set_arrival_time(arrival_time=time.time())
+
+        # At receiving a request from DS, we clear it's object_tags, and re-set it as the
+        # tags of the requested object. Because the DS may give fake tags.
+        while msg.object_tags != []:
```
tudorcebere, line 217: I prefer

```
while len(msg.object_tags):
...
```

tudorcebere (changes requested):
Amazing work @xutongye! Added a few comments here and there.

src/syft/core/node/domain/service/request_handler_service.py
xutongye: `Float` uses a 32-bit float type in proto message. And only 24-bit are used for precision. So a value of `time.time()` will loss some precision. Please try the following and you will see it:
```python
t = time.time()
ptr = sy.lib.python.Float(t).send(duet)
print(t)
print(ptr.get())
```
But add and sub something like 1.6e9 truly looks ugly. Yes, encode it as str is a good idea, I will do that.

src/syft/core/node/domain/client.py
xutongye: Good point. I will change it to 
```python
tags: Optional[List[str]] = None,
```

src/syft/core/node/domain/client.py
xutongye: Same as above.

src/syft/core/node/domain/service/request_handler_service.py
xutongye: Same as above.

src/syft/core/node/domain/service/request_message.py
xutongye: same as above.

src/syft/core/node/domain/service/request_message.py
xutongye: Yes, your way looks better. I will change it.

PR #5021, opened 2021-01-17
add paper link

devkumar24:
add a direct link to the paper as a .pdf file and fix a typo.


cereallarceny:
There is no typo and this is not necessary.

Issue #4846, opened 2020-11-24
Support Homomorphic Encryption


madhavajay:
## Description
This ticket relates to implementing some kind of MVP for Homomorphic Encryption integration for educational purposes for the 0.4.0 Milestone.

## Definition of Done
A notebook is created in examples/homomorphic-encryption or something similar which demonstrates how HE could be used with Duet sufficiently to achieve the educational goals.

## Contacts
- @madhavajay
- @youben11 
- @bcebere 

## TODO
- [x] Initial Hello World Duet Notebooks
- [x] Support LMW (🐧🍎💠)
- [x] Serialization and Deserialization of all types
- [x] Fix issue with Context required for deserialization of Vectors
- [x] Support @property setters in Syft - https://github.com/OpenMined/PySyft/issues/4918
- [x] Support the Enum tenseal.SCHEME_TYPE's in Syft - https://github.com/OpenMined/PySyft/issues/4919
- [x] Support 100% of TenSEAL API
- [x] Flesh out Notebook Examples

madhavajay:
Related PR: https://github.com/OpenMined/PySyft/pull/4853/files

madhavajay:
This is doneskis. ❤️

PR #5044, opened 2021-01-20
TenSEAL integration 

bcebere:
## Description
 - update to tenseal 0.3.0a4
 - use lazy TenSEAL context.
 - full TenSEAL API support.
 - ...

https://github.com/OpenMined/PySyft/issues/4846

## Checklist


examples/homomorphic-encryption/TenSEAL_Syft_Data_Owner.ipynb
```diff
@@ -7,7 +7,7 @@
    "outputs": [],
    "source": [
     "%%capture\n",
-    "!pip install tenseal==0.3.0a1"
+    "!pip install tenseal==0.3.0a2"
```
youben11, line 10: 0a4 maybe?

examples/homomorphic-encryption/TenSEAL_Syft_Data_Scientist.ipynb
```diff
@@ -7,7 +7,7 @@
    "outputs": [],
    "source": [
     "%%capture\n",
-    "!pip install tenseal==0.3.0a1"
+    "!pip install tenseal==0.3.0a2"
```
youben11, line 10: Here as well

youben11 (changes requested):

src/syft/lib/tenseal/ckks_tensor.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @staticmethod
+    def get_wrapped_type() -> type:
+        return ts.CKKSTensor
+
+    @staticmethod
+    def construct_new_object(
+        id: UID,
+        data: StorableObject,
+        description: Optional[str],
```
gmuraru, line 57: NITPICK: Shouldn't we also set:
```description = None``` if that is marked as ```Optional```

Also for the other ones

File: src/syft/lib/tenseal/__init__.py [hunk 4]
```diff
@@ -13,9 +13,13 @@
 from ...ast import add_modules
 from ...ast.globals import Globals
 from ..misc.union import UnionGenerator
-from .ckks_vector import CKKSVector  # noqa: 401
 from .context import ContextWrapper  # noqa: 401
 
+from .ckks_vector import CKKSVector  # noqa: 401
```
gmuraru, line 18: Q: The ```noqa: 401``` isn't it used for modules that are unused?
Can't we delete it since we are using the modules?

src/syft/lib/tenseal/bfv_vector.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @staticmethod
+    def get_wrapped_type() -> type:
+        return ts.BFVVector
+
+    @staticmethod
+    def construct_new_object(
+        id: UID,
+        data: StorableObject,
+        description: Optional[str],
```
gmuraru, line 57: The same "Q" from ```src/syft/lib/tenseal/ckks_tensor.py```

tests/syft/lib/tenseal/tenseal_bfvvector_test.py
```diff
@@ -0,0 +1,217 @@
...
+
+# syft absolute
+import syft as sy
+
+from typing import Any
+
+
+@pytest.fixture(scope="function")
+def context() -> Any:
+    import tenseal as ts
```
gmuraru, line 12: Q: Why don't we import this at the top of the file? (and also the ```sy.load_lib``` there?)

tests/syft/lib/tenseal/tenseal_ckkstensor_test.py
```diff
@@ -0,0 +1,374 @@
...
+
+
+def _almost_equal_number(v1: Any, v2: Any, m_pow_ten: int) -> bool:
+    upper_bound = pow(10, -m_pow_ten)
+
+    return abs(v1 - v2) <= upper_bound
+
+
+def _almost_equal(vec1: Any, vec2: Any, precision_pow_ten: int = 1) -> bool:
+    import tenseal as ts
```
gmuraru, line 17: Q: Can't we import this at the top of the file?

tests/syft/lib/tenseal/tenseal_context_test.py
```diff
@@ -0,0 +1,129 @@
...
+    v1 = [0, 1, 2, 3, 4]
+    enc_v1 = ts.ckks_vector(context, v1)
+
+    ctx_ptr = context.send(duet, searchable=True)
+    enc_v1_ptr = enc_v1.send(duet, searchable=True)
+
+    assert ctx_ptr.is_private().get() == True  # noqa: E712
+    assert ctx_ptr.has_galois_keys().get() == False  # noqa: E712
+    assert ctx_ptr.has_secret_key().get() == True  # noqa: E712
+    assert ctx_ptr.has_public_key().get() == True  # noqa: E712
```
gmuraru, line 81: Q: Is this better then simply leaving ```ctx_ptr.has_public_key().get()```? (without the True) or is there the case where ```ctx_ptr.has_public_key().get()``` returns something that is an object that is ```!= True```

tests/syft/lib/tenseal/tenseal_bfvvector_test.py
gmuraru: Thinking more into this. Can't we do something with ```pytest.importorskip("tenseal")```

tests/syft/lib/tenseal/tenseal_ckkstensor_test.py
```diff
@@ -0,0 +1,374 @@
+# third party
+import pytest
+
+# syft absolute
+import syft as sy
+
+from typing import Any
+
+
+def _almost_equal_number(v1: Any, v2: Any, m_pow_ten: int) -> bool:
```
gmuraru, line 10: Q: Why don't use the ```math.isclose``` it has ```absolute``` and ```relative``` tolerance?

madhavajay:
@bcebere this is super awesome. I removed the unused Union. Also I noticed that there is no serde for the Enum types. We have support for this now, so I wonder if it's worth adding. It would mean that they can be sent back and forth or remotely created in a store which isn't a Memory Store like the SQLite store which would be handy.

For example:
```
st_ckks = ts.SCHEME_TYPE.CKKS
st_ckks.send(duet, searchable=True)
```

src/syft/lib/tenseal/__init__.py
madhavajay: To do that I think the code should change from:
```
        ("tenseal.CKKSTensor", "tenseal.CKKSTensor", ts.CKKSTensor),
```

```
        ("tenseal.CKKSTensor", "tenseal.CKKSTensor", CKKSTensor),
```

Then the `# noqa: 401` can be removed.

madhavajay: Oh, I see they're actually different. I wonder if this should be named differently to prevent confusion, not a big deal either way. I guess we normally call these wrappers:
```
class CKKSVector(StorableObject):
```

```
class CKKSVectorWrapper(StorableObject):
```

Since the serde is handled transparently using the serializable_wrapper_type property on the real class the user should never interact with these Wrappers anyway.

tests/syft/lib/tenseal/tenseal_bfvvector_test.py
madhavajay: Good idea, this seems to work for me:
```
# stdlib
from typing import Any

# third party
import pytest

# syft absolute
import syft as sy

ts = pytest.importorskip("tenseal")
sy.load_lib("tenseal")
```

Then as long as all the tests correctly specify their minimum requirements the pytest import will skip and the load_lib will skip.

tests/syft/lib/tenseal/tenseal_ckkstensor_test.py
madhavajay: Theres also approx from pytest.
`from pytest import approx`

madhavajay (approved):
Very solid PR! 👍🏽

youben11 (approved):
Awesome Bogdan!

PR #5088, opened 2021-02-02
Result of func/method/property/attribute inherites upstream tags

### LINKED ISSUE #5076: Tags for pointers generated mid-computation
## Feature Description
With the new mechanism for requests, where I reference the tag of the variable I am requesting, it has become more difficult to pull back pointers which are the product of some computation. Let's say I want to pull back 'pred_ptr' from

`pred_ptr = model_ptr(x_ptr)
`
In order to request pred_ptr I need to reference its tag, however it currently has no tag as it was generated through some computation on the Data Scientist side. I propose the following new features to get around this;

- pointers should _inherit the tags_ of the variables they are generated from further up in the computation graph. If pointers of are the product of multiple tagged pointers, we make each tag present go with the permissions of the most restrictive tag . In this case that would mean all tags must have been given the OK by the Data Owner to be pulled back by the Data Scientist.
- There may be times where the Data Scientists needs to reference a newly created variable to the Data Owner by some unique tag. In this circumstance it'd be good to be able to tag a pointer.

## Is your feature request related to a problem?
I need to auto-allow the retrieval of the result of a computation in a loop but I can't do it bc the resultant pointer is not reference-able in a handler.

## Data Scientist
Screenshot 2021-01-31 at 13 34 38

## Data Owner
Screenshot 2021-01-31 at 13 36 59

## What alternatives have you considered?
At the moment my workaround is just to add a handler with no tags which accepts everything. This isn't ideal though as there should be some access control in the notebook.

xutongye:
## Description
This may close #5076 .

---
### Motivation
Consider the following example:
```python
# DO side
torch.Tensor([1, 2, 3]).send(duet, searchable=True, tags=["a"]) 
torch.Tensor([4, 5, 6]).send(duet, searchable=True, tags=["b"])

# DS side
a = duet.store["a"]
b = duet.store["b"]
c = a + b
c.get(reason="some reason") 
```
Then what should the DO do to accept the request to `c`? One way is to call `duet.requests[0].accept()`, this is the manual way. Another way, which is a automatical way, is to use request handlers, in this way, the DO can call `duet.requests.add_handler(action="accept", tags=??)`. Wait, what's the tags for `c`? No, there are no tags for `c`.

This is the problem this PR aims to solve.

#### 1. Fix a bug.
A bug is encountered and then get fixed. The bug is the following: 
```python
# DO side
torch.Tensor([1,2,3]).send(duet, searchable=True, tags=["a"]
torch.Tensor([4,5,6]).send(duet, searchable=True, tags=["b"]
# DS side
duet.store["a"].tags # you get ["a"]
duet.store["b"].tags # you still get ["a"]
```
The reason for this bug is that if the key passed in to duet.store[key] is very short, for example key="a", then there are chances that it can be matched with any id string very easily, so it return the first element in duet.store. This is fixed.
#### 2. obj.property/attribute inherites tags of obj, plus name of property/attribute
For example:
```python
# DO side
torch.Tensor([1,2,3]).send(duet, searchable=True, tags=["a"])
# DS side
a = duet.store["a"]
ptr = a.requires_grad # ptr points to an object whose tags=["a", "rerquires_grad"]
```

#### 3. obj.method(args) inherites tags of obj, tags of args, plus name of method
```python
# DO side
torch.Tensor([1,2,3]).send(duet, searchable=True, tags=["a"])
Int(2).send(duet, searchable=True, tags=["b"])
# DS side
a = duet.store["a"]
b = duet.store["b"]
ptr = a.pow(b) # ptr points to an object whose tags=["a", "b", "pow"]
```
#### 4. func(args) inherites tags of args, plus name of func
```python
# DO side
torch.Tensor([1,2,3]).send(duet, searchable=True, tags=["a"])
torch.Tensor([4,5,6]).send(duet, searchable=True, tags=["b"])
# DS side
a = duet.store["a"]
b = duet.store["b"]
ptr = a + b # ptr points to an object whose tags=["a", "b", "__add__"]
```

## Affected Dependencies
No.

## How has this been tested?
- examples/experimental/tongye/inherit_tags/

## Checklist


madhavajay:
@xutongye I am not sure what happened but when ever I run the examples the tags are empty.
I also tried adding a test:
```python
def test_tag_heritance() -> None:
    bob = sy.VirtualMachine(name="Bob")
    root_client = bob.get_root_client()

    ten = th.tensor([1, 2])

    ptr = ten.send(root_client, tags=["a"])
    assert ptr.tags == ["a"]

    result_ptr = ptr.T
    assert result_ptr.tags == ["a", "T"]
```

Do you also have the problem of empty `tags`?

madhavajay:
@xutongye I fixed this by cherry picking it over the top of `dev`: https://github.com/OpenMined/PySyft/pull/5159

PR #5105, opened 2021-02-03
Removing syft_decorator

tudorcebere:
## Description
Removed syft_decorator from the codebase

## Affected Dependencies
Typeguard no longer needed.

## Checklist


iamtrask:
Please document why syft_decorator is being removed and how we will enforce types in the end user API at runtime (as well as internal API) moving forward.

iamtrask (changes requested):
Please do not merge until code is added to this PR which provides the following guarantees which are essential to the long-term readability and health of the codebase:

- requires type hints in every method in the codebase
- requires that kwargs be used for all methods - especially the end-user API
- enforces that arguments passed into methods are correct at runtime by the end user

madhavajay:
I have added our custom flake8-kwarger linter plugin to resolve these requirements.

**- requires type hints in every method in the codebase**
✅ this is already covered by mypy
**- requires that kwargs be used for all methods - especially the end-user API**
✅ see the new flake8-kwarger https://github.com/madhavajay/flake8-kwarger
**- enforces that arguments passed into methods are correct at runtime by the end user**
❎ python * and PEP 3102 checks the kwarg names at runtime but not the types.
We are trying to avoid the runtime type checking since its expensive and Python doesn't normally do this.
For all our internal code we have mypy static checking and I think this is the right trade-off.

This first commit https://github.com/OpenMined/PySyft/pull/5105/commits/ed133f711632202fe7a38577ab4b70d6b3530aba shows an example of how it works.

With * included in the VirtualMachine init for example the following happens.
```python
>>> import syft as sy
>>> sy.VirtualMachine("Bob")
  File "<stdin>", line 1, in <module>
TypeError: __init__() takes 1 positional argument but 2 were given
>>> sy.VirtualMachine(name="Bob")
VirtualMachine: Bob: <UID: 51ade0963f28484b8d22284bc421b295>
```

tudorcebere:
I am into the solution from @madhavajay, it's the best tradeoff in terms of speed and functionality + we can extend it to do whatever we want in the future.

madhavajay:
@iamtrask I have detailed the remaining work on the lint checker here:
https://github.com/OpenMined/PySyft/issues/5135

Since its a static check that will always run in CI, I see no harm in merging this PR while its hot and prevent it going stale and being harder to merge.

Can you mark your "Change request" as approved so we can merge?

PR #5089, opened 2021-02-02
Add attributes to allowlist which can be accessed through objects (closes #5031 and #5032 )

avinsit123:
## Description

Closes #5031 
Closes #5032 

Currently, the problem with adding attributes in `allowlist` dictionary is that only those attributes can be added which are static and only accessible by class. This is not possible  if you will try to access any attribute of `torchvision.transforms.Normalize` like `inplace`  or `mean`. You get an error as shown below:-

These attributes can only be accessed when we pass the required positional arguments list as given below. The trick is to pass these positional arguments whenever we call the attributes like `mean` and `std`

### Fix
My proposed fix adds two new attribute to `src/syft/ast/attribute.py`  -  `self.require_pargs` and `self.parg_list` .
`self.require_pargs` is set to `True` is that attribute requires any positional arguments to access its sub-attributes, otherwise by default it is set to `False `. `self.parg_list` contains list of all the positional arguments that need to be passed into the attribute so that we can access its sub-attributes, by Default it is empty. As all components of AST like `Module` , `Class` , `Callable`  inherit from `attribute` class they inherit these 2 parameters also.

We check if `self.require_pargs` is set to `True` and pass `self.parg_list` into the class as shown in the code so that we can access its internal sub-attributes during using `getattr()`.

### How to use
Specified in the code. In case of `torchvision.transforms.Normalize` we need to pass in 2 attributes `mean` and `std` . Add name of attribute( in this case `torchvision.transforms.Normalize`) and list of positional arguments as corresponding entry in this case `[(1, 2, 3), (1, 2, 3)]` . Have a look at `src/syft/lib/torchvision/allowlist.py ` . 

## Affected Dependencies
Added new dict module_pargs in which modules that require positional arguments to access attributes can list their positional arguments.

## How has this been tested?
Wish to get some guidance for added fix and then write tests

## Checklist
- [x] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [x] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [x] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [ ] My changes are covered by tests

PR #5117, opened 2021-02-06
Add attributes to allowlist accessible through objects(closes #5031 and #5032 )

### LINKED ISSUE #5031: TODOs fix - Torchvision - mean and std
## Description
In `src/syft/lib/torchvision/allowlist.py` we have a TODO: `mean and std are actually tuples`. 

## Additional Context
The fix has to be made and a test has to be written.

## Note
This issue will not be assigned to anybody until a PR is opened.

### LINKED ISSUE #5032: TODOs fix - Torchvision - normalize
## Description
In `src/syft/lib/torchvision/allowlist.py` we have a TODO: ` Normalize properties only exists on the object not on the class?`. 

## Additional Context
The fix has to be made and a test has to be written that the methods are working and are added to the allowlist.

## Note
This issue will not be assigned to anybody until a PR is opened.

avinsit123:
## Description

Closes #5031 
Closes #5032 

Currently, the problem with adding attributes in `allowlist` dictionary is that only those attributes can be added which are static and only accessible by class. This is not possible  if you will try to access any attribute of `torchvision.transforms.Normalize` like `inplace`  or `mean`. You get an error as shown below:-

These attributes can only be accessed when we pass the required positional arguments list as given below. The trick is to pass these positional arguments whenever we call the attributes like `mean` and `std`

### Fix
My proposed fix adds two new attribute to `src/syft/ast/attribute.py`  -  `self.require_pargs` and `self.parg_list` .
`self.require_pargs` is set to `True` is that attribute requires any positional arguments to access its sub-attributes, otherwise by default it is set to `False `. `self.parg_list` contains list of all the positional arguments that need to be passed into the attribute so that we can access its sub-attributes, by Default it is empty. As all components of AST like `Module` , `Class` , `Callable`  inherit from `attribute` class they inherit these 2 parameters also.

We check if `self.require_pargs` is set to `True` and pass `self.parg_list` into the class as shown in the code so that we can access its internal sub-attributes during using `getattr()`.

### How to use
Specified in the code. In case of `torchvision.transforms.Normalize` we need to pass in 2 attributes `mean` and `std` . Add name of attribute( in this case `torchvision.transforms.Normalize`) and list of positional arguments as corresponding entry in this case `[(1, 2, 3), (1, 2, 3)]` . Have a look at `src/syft/lib/torchvision/allowlist.py ` . 

## Affected Dependencies
Added new dict module_pargs in which modules that require positional arguments to access attributes can list their positional arguments.

## How has this been tested?
Wish to get some guidance for added fix and then write tests

## Checklist


madhavajay:
@avinsit123 does this PR replace this PR: https://github.com/OpenMined/PySyft/pull/5089

avinsit123:
Yes

Pipfile
```diff
@@ -0,0 +1,41 @@
+[[source]]
```
madhavajay, line 1: Please don't commit the Pipfile as its an opinionated virtualenv / package management file which isn't relevant to the source code repo itself.

src/syft/lib/torchvision/allowlist.py
```diff
@@ -12,10 +12,10 @@
 allowlist["torchvision.transforms.ToTensor"] = "torchvision.transforms.ToTensor"
 allowlist["torchvision.transforms.Normalize"] = "torchvision.transforms.Normalize"
 # TODO: Normalize properties only exists on the object not on the class?
-# allowlist["torchvision.transforms.Normalize.inplace"] = "syft.lib.python.Bool"
+allowlist["torchvision.transforms.Normalize.inplace"] = "syft.lib.python.Bool"
 # TODO: mean and std are actually tuples
-# allowlist["torchvision.transforms.Normalize.mean"] = "syft.lib.python.List"
-# allowlist["torchvision.transforms.Normalize.std"] = "syft.lib.python.List"
+allowlist["torchvision.transforms.Normalize.mean"] = "syft.lib.python.Tuple"
```
madhavajay, line 17: The return type depends on what is passed in. It seems that its allowed to be any scalar or sequence. We need to use the UnionGenerator here.

madhavajay (commented):
I am unsure if this is the right approach.
As I understand it the issue is that the attribute is found on the object not the class?
Do we have a test that proves getting the attribute (or method) on an initialised object isn't just bound to that object indefinitely?

We also have a separate requirement to allow runtime resolving of attributes on objects so i think perhaps a better solution would be through this configuration as part of the already supported allowlist Dict:
```python
allowlist["torch.Tensor.movedim"] = {
    "return_type": "torch.Tensor",
    "min_version": "1.7.0",
    "obj_attr": True
}
```

Then refactor this code to work around that `obj_attr` setting so that things like Opacus can use this setting. Or perhaps this needs to be a "reload" since the target library Opacus monkey patches is actually Torch.
```python
if self.is_static:
    result = method(*upcasted_args, **upcasted_kwargs)
else:
    # in opacus the step method in torch gets monkey patched on .attach
    # this means we can't use the original AST method reference and need to
    # get it again from the actual object so for now lets allow the following
    # two methods to be resolved at execution time
    method_name = self.path.split(".")[-1]
    if method_name in ["step", "zero_grad"]:
        # TODO: Remove this Opacus workaround
        try:
            method = getattr(resolved_self.data, method_name, None)
            result = method(*upcasted_args, **upcasted_kwargs)
        except Exception as e:
            critical(
                f"Unable to resolve method {self.path} on {resolved_self}. {e}"
            )
            result = method(
                resolved_self.data, *upcasted_args, **upcasted_kwargs
            )
    else:
        result = method(resolved_self.data, *upcasted_args, **upcasted_kwargs)
```

tudorcebere:
Hello,

This issue has been addressed by a previous PR. Thanks for your interest in Syft! :+1:

avinsit123:
Hi @tudorcebere could you refer me to the PR as I really wanted to know the solution to this issue. I am not able to find the PR. Thanks

PR #5120, opened 2021-02-07
Add threadsafe stop

gmuraru:
## Description
When running in a Python interactive console, pressing CTRL + D should cancel the interactive console.
This was not happening because the ```run_forever``` call was never stopped.
This was happening because the ```stop``` call should be done in a threadsafe manner ([reference](https://stackoverflow.com/questions/40143289/why-do-most-asyncio-examples-use-loop-run-until-complete)

## Affected Dependencies

## How has this been tested?
- ```python + import syft + CTRL-D``` (should close the console)

## Checklist

### DIFF
--- a/src/syft/core/common/event_loop.py
+++ b/src/syft/core/common/event_loop.py
@@ -74,8 +74,8 @@ class EventLoopThread:
     def shutdown(self) -> None:
         if "loop" in self.__shared_state:
             info("Stopping Event Loop")
-            loop.call_soon_threadsafe(loop.stop)  # type: ignore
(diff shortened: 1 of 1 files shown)


madhavajay:
I see this error sometimes on quitting. I think its unrelated to the change, because this error is now showing when we connect two syft instances as well.

Screen Shot 2021-02-08 at 11 49 46 am

madhavajay (approved):
Assuming that new error is just showing something thats already there, I think we can merge this.

gmuraru:
@madhavajay pushed one more change to interchange two lines.

PR #5124, opened 2021-02-08
TenSEAL Network tests

bcebere:
## Description
This PR explores different configurations for WebRTC. The benchmarks are reported to the encrypted image inference(Tutorial_2_TenSEAL). The current duration is 220 seconds.

 - [1] suggests that 256 KiB sized chunks are ideal, with max buffer size of 1 MiB. This reduces the benchmark to 80 seconds.
 - We add a wrapper for chunks, to send the index too, and use WebRTC in the unordered setup. This reduces the benchmark to 60 seconds.
 - We set chunk size to 256 KB and max buffer to 4MB. This reduces the benchmark to 48 seconds.
 - Sending the GaloisKeys in the context adds 24 seconds overhead on serialization. One workaround is to send the secret key instead, recreate the GaloisKeys on the other end, and drop the secret key. But the DS does another serialization - with the GaloisKeys regenerated -  that cancels the optimization. Fixing this reduces the benchmark to ~27 seconds.
 
Notes:
 - This PR introduces a temporary hack for the TenSEAL integration, sending the secret key to the DS for performance improvements. This should be removed when the serde logic is improved for big objects, like the GaloisKeys. Also, this should **NOT** be used in production - remove the secret key first.
  - WebRTC with unordered packages is unstable for now, but the OrderedChunk improvement is still visible, since we are not sending the END chunk anymore(less data to send).
  - The performance boost allows automating the notebooks for MacOS too.
  - Now it is allowed to configure the chunk size from ENV, so we can benchmark the best combination for Duet.
  - 
Ref:
 [1] https://stackoverflow.com/questions/56327783/webrtc-datachannel-for-high-bandwidth-application

## Affected Dependencies

## How has this been tested?

## Checklist


madhavajay (approved):
Nice work! Looks solid. 👍🏼 
I will see if i can sort out this flapping MacOS issue.

PR #5130, opened 2021-02-09
Only raise AttributeError in __getattribute__

gmuraru:
## Description
This ticket is to solve a log and issue that appears when using static methods (and possibly attributes).
Do not log as critical, traceback and raise in case ```__getattribute__``` raises an ```AttributeError```.
```__getattribute__``` is called first and in case it returns an ```AttributeError``` then it is called the ```__getattr__``` method.

## Affected Dependencies
Tried to call a static method in SyMPC

## How has this been tested?
- Call a static method -- this should log only a warning.

## Checklist


bcebere (approved):
LGTM!

PR #5193, opened 2021-02-19
separate requirements

Borda:
## Description

not very familiar with all packages, so can someone help triangle what is really needed for the package itself and what is development tools (testing/docs)

## Affected Dependencies
List any dependencies that are required for this change.

## How has this been tested?
- Describe the tests that you ran to verify your changes.
- Provide instructions so we can reproduce.
- List any relevant details for your test configuration.

## Checklist
- [ ] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [ ] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [ ] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [x] My changes are covered by tests

PR #5194, opened 2021-02-19
CI: simplify install versions

Borda:
## Description

unify scripts for installing Torch versions and related packages
after merging #5192 we can also drop PT version from the chafe file name

## Affected Dependencies

## How has this been tested?

## Checklist


.github/workflows/pr_tests.yml
```diff
@@ -126,47 +126,25 @@ jobs:
...
-
-      - name: Install PyTorch Windows
-        if: startsWith(runner.os, 'Windows')
-        run: |
-          ./scripts/pytorch_install.ps1 ${{ matrix.torch-version }}
+          python .github/adjust_torch_versions.py ./requirements.txt ${{ matrix.torch-version }}
 
       - name: Install packages
         run: |
-          pip install -r requirements.txt --default-timeout=60
```
madhavajay, line 141: We had timeout issues with pip, are you sure the job task timeout will do the same thing?

.github/workflows/pr_tests.yml
```diff
@@ -126,47 +126,25 @@ jobs:
...
-        if: startsWith(runner.os, 'Windows') != true
-        run: |
-          ./scripts/pytorch_install.sh ${{ matrix.torch-version }}
-
-      - name: Re-install PyTorch Windows
-        if: startsWith(runner.os, 'Windows')
+      - name: Install supported library packages
         run: |
-          ./scripts/pytorch_install.ps1 ${{ matrix.torch-version }}
+          pip isntall -r requirements/supported_libs.txt
```
madhavajay, line 146: The reason we xargs the supported_libs is that they aren't available on all OSes or Python versions, so this allows them to silently fail to install and then those tests will simply not run.

.github/workflows/pr_tests.yml
```diff
@@ -216,10 +194,10 @@ jobs:
           ./scripts/nb_test.sh
           pytest examples/api -n auto
 
-      - name: Install supported library packages Linux and MacOS - Ignore Errors
-        if: ${{ matrix.os != 'windows-latest' }}
+      - name: Install supported library packages
         run: |
-          cat requirements/supported_libs.txt | xargs -I {} -n 1 pip install {} --default-timeout=60 || true
+          pip isntall -r requirements/supported_libs.txt
```
madhavajay, line 199: typo: install

.github/workflows/pr_tests.yml
Borda: my bad, thx :]

.github/workflows/pr_tests.yml
Borda: my bad, thx

Borda:
> os.linesep fails because requirements.txt uses \n only

quite curious, I would expect that this shall work since we want to run it on all OS

.github/workflows/pr_tests.yml
```diff
@@ -124,51 +124,34 @@ jobs:
...
-          ./scripts/pytorch_install.ps1 ${{ matrix.torch-version }}
+          python .github/adjust_torch_versions.py ./requirements.txt ${{ matrix.torch-version }}
+          cat ./requirements.txt
 
       - name: Install packages
         run: |
-          pip install -r requirements.txt --default-timeout=60
-          pip install -e . --default-timeout=60
+          echo "attempting to install"
+          cat ./requirements.txt
```
Borda, line 135:
```suggestion
```
this is not needed since you have cat in the step above

.github/workflows/pr_tests.yml
Borda: in combination with `continue-on-error: true` it should
but if you silently accept missing some packages isn't it making the test weak, imagine that none of the packages is installed, you skip them in testing and show as all pass... so you think all is fine, but in fact, you just don't know...

.github/workflows/pr_tests.yml
```diff
@@ -124,51 +124,34 @@ jobs:
...
           pip list
+        timeout-minutes: 20
 
       - name: Run normal tests
         run: |
           pytest -m 'fast or slow' --cov syft --cov-fail-under 80 -n auto
 
-      - name: Install supported library packages Linux and MacOS - Ignore Errors
+      - name: Install supported library packages - Ignore Errors
         if: startsWith(runner.os, 'Windows') != true
```
Borda, line 146:
```suggestion
```

setup.cfg
```diff
@@ -44,7 +44,7 @@ install_requires =
     PyNaCl
     requests
     sqlitedict
-    torch>=1.4
+    torch>=1.4.0
```
Borda, line 47: just a question, why not split the requirements file into mandatory and development and then in setup load the mandatory, then you have to keep updated packages only on one place, something in directions #5193

.github/workflows/pr_tests.yml
madhavajay: That was for debugging, I was trying to figure out why it was failing. Turns out it was the Windows python install not matching the os.linesep and installing CUDA instead of CPU.

setup.cfg
madhavajay: Yeah i saw that separate PR but I guess we can discuss that in the other PR.

.github/workflows/pr_tests.yml
```diff
@@ -142,19 +142,12 @@ jobs:
...
-
-      - name: Install supported library packages Windows - Ignore Errors
-        if: startsWith(runner.os, 'Windows')
-        run: |
-          foreach ($i in cat requirements/supported_libs.txt) {
-            pip install $i --default-timeout=60
-          }
+          pip install requirements/supported_libs.txt
         timeout-minutes: 10
+        continue-on-error: true
```
madhavajay, line 147: The problem I had was that if you pip install a .txt file and one of the packages fails it aborts the whole install. I could not find another way to ignore individual failures except to `xargs`.

Borda: ok, now I understand what you mean, but do you know what package fails on what os?
what we did was filtering od adjustment based on OS / versions...

setup.cfg
Borda: sure :] that is why I made it separate :]

Borda:
@madhavajay I think it is ready to go :]

madhavajay (commented):
> > os.linesep fails because requirements.txt uses \n only
> 
> quite curious, I would expect that this shall work since we want to run it on all OS

The requirements.txt file we are using has \n and our lint checker probably enforces this so `os.linesep` probably won't be used.

PR #5199, opened 2021-02-21
Private attr API refactor

### LINKED ISSUE #5172: API Refactor Private Klass Methods
## Description
As an intermediary step we should immediately rename all private API calls so they, a) dont clash with target classes we patch and b) make it obvious that they will be deprecated soon by making them private.

Check insde `klass.py` and anywhere where we use `aggressive_set_attr`.
Rename the following:
- .serializable_wrapper_type
- .serialize
- .to_proto
- .proto
- .to_bytes

To:
`f"_sy_{name}"`

Remove duplicates and workarounds like 
- .sy_serialize
- ._to_bytes

## Definition of Done
Everything except `send` `tag` and `describe` are renamed and the tests still pass.

xutongye:
## Description
This may close #5172 .

There are some new attrs we attach to object, including
- `serialize`
- `to_bytes`
- `binary`
- `to_proto`
- `proto`
- `serializable_wrapper_type`

These attrs are intented to be used by developers and not by end-users. We want to refacor them, so that they don't confuse end-users.

This work is done in three steps:
1. prefix `_sy_` to their names. Make it clear that they should be used inside by developers.  
2. remove `_sy_to_bytes/binary/to_proto/proto` methods, use `_sy_serialize(...)`.
3. remove `_sy_serialize`, use `sy.serialize`

## Affected Dependencies

## How has this been tested?
- pytest -m fast/libs still all pass

## Checklist

### DIFF
--- a/examples/api/start.ipynb
+++ b/examples/api/start.ipynb
@@ -975,7 +975,7 @@
     }
    ],
    "source": [
-    "assert serialize(repr_msg, to_bytes=True) == sig_msg.serialized_message\n",
+    "assert sy.serialize(repr_msg, to_bytes=True) == sig_msg.serialized_message\n",
     "print(repr_msg.pprint, \" ⬅️ \", sig_msg.pprint)"
    ]
   },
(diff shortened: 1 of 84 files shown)


madhavajay (approved):
Excellent work as always! ❤️

Issue #5185, opened 2021-02-18
Add torch==1.7.1


madhavajay:
## Description
We can add torch==1.7.1 we just need to fix up CI.

## Definition of Done
Add 1.7.1 to CI and update README.md.

This would close: https://github.com/OpenMined/PySyft/issues/4936

marcalph:
I'd ike to try this one out

marcalph:
Hi, 

`torch==1.7.1` does look good thanks  to `autofix_allow_list_test.py` but I have a couple more questions (sorry):
- If I got this right, in the autofix process "skip rules" are added to `allowlist_test.json`,  is this file maintained manually? do I need to refine some of these rules e.g.:

```
 "__ifloordiv__": {
        "profile": "tensor_float_division_method_v150",
        "not_available": [
          {
            "data_types": [
              "bool",
              "float16"
            ],
            "lte_version": "1.7.0"
          },
          {
            "data_types": [
              "bool",
              "float16"
            ],
            "lte_version": "1.7.1", << REMOVE THIS?
            "gte_version": "1.7.1" 
          }
        ]
      }
```
- Other markers in the test suite are almost fine : 
   * marker `fast` runs OK
   * for the `libs` one, where I had to install a couple dependencies should these be added to `requirements.dev.txt`? (namely `opacus`, `openmined_psi`, `python-dp`) 
   * for the `slow` one , I'm still looking into an error with `test_duet`

- On the CI side, to get this straights is it just about changing the *torch-version* var in the matrices for the `pr_tests.yml` and `merge_test.yml` workflows?

Thanks!

madhavajay:
@marcalph good questions.

Regarding the `allowlist_test.json` file, it used to be handled manually but now the script `autofix_allow_list_test.py` automatically updates it. That is why you will see a rule like:
```
```

For the tests if you check the `pr_tests.yml` file you can see how they run. The `libs` tests installs the packages inside `requirements/supported_libs.txt`. For the failing `test_duet` in `slow` are you on macOS? There is a situation where that tests sometimes stalls on mac.

As far as torch version matrix, yes, I have removed older torch versions from `dev` so your PR should add `1.7.1` to the existing `1.6.0` and `1.7.0`.

marcalph:
@madhavajay Thank you so much for your input!

I've updated the matrices now, so apart from duet things look fine.
[...]

PR #5223, opened 2021-02-26
torch 1.7.1

marcalph:
## Description
add support for torch==1.7.1

## How has this been tested?
pytest markers  
- [x] torch
- [x] fast
- [x] slow >> ongoing
- [x] libs

## Checklist


marcalph:
about #5185

madhavajay:
@marcalph The torch tests are failing (they don't run in the pr_tests.yml in CI only on merge).
I ran them on Python 3.8 and Torch 1.7.1 and there are hundreds of ops which have not been updated.
My guess is there are some issues with `scripts/autofix_allowlist_test.py`. The other thing which needs to be done is to confirm that there is no new functionality we need to add.

https://github.com/pytorch/pytorch/releases

I read the notes and diffed both `dir(torch)` and `dir(torch.Tensor)` and they are identical.

If you can fix the script so it generates a working .json file to pass the tests for `1.7.1` that would be greatly appreciated.

marcalph:
i will!

marcalph:
@madhavajay 
I must have done something wrong the first time... Cannot figure out what, so I've restarted with a fresh install just in case. 
The `allowlist_test.json` was missing some "not_available" rules, I've updated it.
Things should be better : 
pytest_torch_mark

About the duet_test that was failing in slow, I'll try something tomorrow.

marcalph:
@madhavajay 
Finally! The **duet_test** was hitting the wrong route... 
Seems a bit late though...

madhavajay:
I updated the compatibility report script and ran it.
It looks like a lot of the time you are just ignoring the tests rather than running them.

This rule on a pretty important op like `dot` is effectively saying that op is not supported on `1.7.1` for any data types so don't test it.


```json
{
  "dot": {
    "profile": "default",
    "data_types": ["common"],
    "tensors": ["tensor1"],
    "not_available": [
      {
        "inputs": [false, true, 0, 1, [0], [1]],
        "lte_version": "1.7.0"
      },
      {
        "data_types": ["bool", "bfloat16"],
        "lte_version": "1.7.0"
      },
      {
        "data_types": ["float16"],
        "lte_version": "1.5.1"
      },
      {
        "data_types": [
          "bool",
          "uint8",
          "int8",
          "int16",
          "int32",
          "int64",
          "float32",
          "float64",
          "float16",
          "bfloat16"
        ],
        "lte_version": "1.7.1",
        "gte_version": "1.7.1"
      }
    ]
  }
}
```

Screen Shot 2021-03-08 at 12 00 08 pm

There are many other examples like this in the compatibility report where 1.7.1 has been set to just ignore the tests.
This isn't really the solution we were looking for.

Also it seems that `pin_memory` is broken?
Screen Shot 2021-03-08 at 12 02 23 pm

PR #5225, opened 2021-02-27
Duet example - RL notebooks

### LINKED ISSUE #4794: Add Reinforcement learning Duet Notebooks
## Description
Add two notebook's which reflect the Reinforcement learning example split into DO (Data Owner) and DS (Data Scientist):
https://github.com/pytorch/examples/blob/master/reinforcement_learning/

## Definition of Done
The partially runnable Reinforcement learning notebook should be in the examples/duet/reinforcement_learning folder and a README.md should be added in the parent examples/duet directory with a link to the original example and our notebook.

Koukyosyumei:
## Description
closes #4794 

This notebook demonstrates reinforcement learning with duet. Although I confirmed that this notebook works as I expected,  it only works in loopback mode and training is really slow because data owner and data scientist have to frequently share state, action, and reward. Also, I can't introduce the ci test to this notebook. If you have a better idea, please ping me!

## Affected Dependencies
no

## Checklist


madhavajay:
@Koukyosyumei Great work, interesting solution. I tried adding this to an integration test but it seems to hang or fail, possibly due to the additional async code. If you could give it another shot now that the code is in place that would be nice, otherwise we can merge without a test and come back and re-implement this with Plans.

The code I added does a find and replace of the notebooks to change the `epoch` and `steps` to make the test faster, but I might have gotten something wrong with the checkpoints, or its just hanging on the first one?

Koukyosyumei:
@madhavajay 
Thank you for your work! I tested the notebook you modified, and it worked! The checkpoints seem fine to me.

examples/experimental/tongye/Example_Remote_DataLoader.ipynb
```diff
@@ -0,0 +1,259 @@
+{
```
tudorcebere, line 1: Is this file related to the PR?

scripts/nb_duet_test.py
```diff
@@ -57,6 +57,7 @@
     list(Path("examples/homomorphic-encryption").rglob("*.ipynb"))
     + list(Path("examples/duet/dcgan").rglob("*.ipynb"))
     + list(Path("examples/duet/super_resolution").rglob("*.ipynb"))
+    # + list(Path("examples/duet/reinforcement_learning").rglob("*.ipynb"))
```
tudorcebere, line 60: Can we add to notebook to the CI?

tudorcebere (changes requested):

scripts/nb_duet_test.py
madhavajay: It doesn't work, thats why its commented out. It uses asyncio in the notebook so I suspect that is colliding with the asyncio sleeping in the integration testing system.

Koukyosyumei:
I think supporting gym is better, and I guess it makes the integration test work because we don't need asyncio sleeping.
I created a new PR #5457

Issue #5082, opened 2021-02-01
Add support for slice objects


tudorcebere:
## Description
We don't have support for `slice` objects. We would like to add slices in our primitive supported types.

TODO:
- [x] Add some serde tests
       The proto only takes ints so you might need to change that to support None as well.
- [ ] Add the tests from CPython: https://github.com/python/cpython/blob/3.9/Lib/ctypes/test/test_slicing.py
- [ ] Add the CPython test filename to the LICENSE.txt file
- [x] Make sure Slice works on other types like Tensor or Pandas
        If there are issues we might need to override the Pointer with a wrapped __getitem__
- [ ] Make sure that __setitem__ also works: https://stackoverflow.com/questions/32772755/how-do-getitem-setitem-work-with-slices
- [x] Add to `primitive_factory`

Theres a bit of work here so it might make sense if we split it up. Could you start by writing some tests for the `serde` and fixing up the proto?

## Additional Context
We would like to have wrappers for slice objects, to be able to manipulate them.

dnabanita7:
I want to work on this.

madhavajay:
See: https://github.com/OpenMined/PySyft/pull/5228

PR #5228, opened 2021-02-28
Added Slice object in syft

dnabanita7:
## Description
This PR adds a slice object to syft. It is related to [this issue](https://github.com/OpenMined/PySyft/issues/5082). 

## How has this been tested?
It hasn't been tested yet. I will add tests as soon as I get a review on this PR.

## Checklist


src/syft/lib/python/slice.py
```diff
@@ -0,0 +1,204 @@
...
+                start = sys.maxint if step < 0 else 0
+
+            if stop is None:
+                stop = -sys.maxint - 1 if step < 0 else sys.maxint
+
+            res = super().unpack(start, stop, step)
+            return PrimitiveFactory.generate_primitive(value=res)
+
+        def adjustindices(self, length, start, stop, step) -> SyPrimitiveRet:
+            assert step is not 0
```
Param-29, line 151: Error:
```
src/syft/lib/python/slice.py:151:20: F632 use ==/!= to compare constant literals (str, bytes, int, float, tuple)
```

Change to 
`assert step != 0`

src/syft/lib/python/slice.py
```diff
@@ -0,0 +1,204 @@
...
+        self,
+        start: Any = None,
+        stop: Any = None,
+        step: Any = None,
+        id: Optional[UID] = None
+    ):
+        if not start and not stop and not step:
+            start = NULL
+            stop = NULL
+            step = NULL
```
Param-29, line 40: I am not sure why you have used `NULL` instead of `None`. Any specific reason? If not replace it with `None` and that should solve following logs 
```
src/syft/lib/python/slice.py:38:21: F821 undefined name 'NULL'
            start = NULL
                    ^
src/syft/lib/python/slice.py:39:20: F821 undefined name 'NULL'
            stop = NULL
                   ^
src/syft/lib/python/slice.py:40:20: F821 undefined name 'NULL'
            step = NULL
```

and some more similar errors, have a look

src/syft/lib/python/slice.py
```diff
@@ -0,0 +1,204 @@
...
+from ...logger import traceback_and_raise
+from ...logger import warning
+from ...proto.lib.python.slice_pb2 import Slice as Slice_PB
+from .iterator import Iterator
+from .primitive_factory import PrimitiveFactory
+from .primitive_factory import isprimitive
+from .primitive_interface import PyPrimitive
+from .types import SyPrimitiveRet
+from .util import downcast
+from .util import upcast
```
Param-29, line 24: I see an error because these are not used 
```
src/syft/lib/python/slice.py:23:1: F401 '.util.downcast' imported but unused
from .util import downcast
^
src/syft/lib/python/slice.py:24:1: F401 '.util.upcast' imported but unused
from .util import upcast
```

Param-29:
I have added in my comments for pre-tests, hope I was helpful.

src/syft/lib/python/__init__.py
```diff
@@ -73,6 +76,20 @@ def create_python_ast(client: Optional[AbstractNodeClient] = None) -> Globals:
     ]
 
     methods = [
+        # Slice methods - quite there
+        ("syft.lib.python.Slice.__getitem__", "syft.lib.python.Any"),
```
madhavajay, line 80: I don't see getitem and iter on the slice Class. Am I missing something?

proto/lib/python/slice.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package syft.lib.python;
+import "proto/core/common/common_object.proto";
+
+message Slice {
+  repeated bytes start = 1;
```
madhavajay, line 7: Can't these just be normal `ints`?

dnabanita7: yes, that won't be a problem but i just thought if it might help with faster computation

madhavajay: The problem would be that we need to convert the ints into syft.lib.python.Ints then to protos and bytes, where as i'm sure the built in protobuf integer type is extremely fast and lightweight.

madhavajay:
@dnabanita7 if you run the `pre-commit` checks locally you can see the errors which are failing in CI.

Install: https://pre-commit.com/

And then run:
```
$ pre-commit run --all-files
```

proto/lib/python/slice.proto
dnabanita7: Alright!

src/syft/lib/python/__init__.py
dnabanita7: I still get this error. They are present in the slice Class.

dnabanita7:
yes, sure! this PR looks messy. I will do that next time.

dnabanita7:
@madhavajay can somebody else add the tests for the object?

dnabanita7:
I need help here. I am stuck.

Param-29:
Is there a way I could push some code to this PR? I tried a couple of commands, made the changes but during push it says failed because permission denied.

madhavajay:
> @madhavajay can somebody else add the tests for the object?

Its usually preferred if the original author of the code creates the tests as they know the code best.

madhavajay:
> Is there a way I could push some code to this PR? I tried a couple of commands, made the changes but during push it says failed because permission denied.

Unfortunately pushing directly to another branch requires permission from the author which is usually only set on creating the PR and given to the maintainers of the destination repo which in this case is the Syft Core team.

So the other way to do this currently is either, create a PR to this original branch: `dnabanita7:slice`.

If you go to the PR section in GitHub it usually only lets you select your `repo:branch` to compare with the main `PySyft:branch` but if you edit the browser url to change the destination to be another `user:branch` then it works.

Or alternatively, since the most important thing is that the contributions of others are preserved, the easiest way is just to open a new PR against the main repo with the shared history of the branch you branched off.

Either is fine by me.

dnabanita7:
> > @madhavajay can somebody else add the tests for the object?
> 
> Its usually preferred if the original author of the code creates the tests as they know the code best.

Alright! Let me try. Another thing is I can't use ``pytest`` for testing as I get the `` couldn't import VendorLibrary error``.

madhavajay:
> > > @madhavajay can somebody else add the tests for the object?
> > 
> > 
> > Its usually preferred if the original author of the code creates the tests as they know the code best.
> 
> Alright! Let me try. Another thing is I can't use `pytest` for testing as I get the ` couldn't import VendorLibrary error`.

As all code is fully tested on Mac, Linux and Windows in CI, its likely this is an issue with the way you are trying to run the tests. Please provide the exact command you are running and we can try to figure out what is going wrong.

Param-29:
@dnabanita7 
`src/syft/lib/__init__.py`
Here there is an exception class: 
```
class VendorLibraryImportException(Exception):
```

You could edit this class to get a clear message about the exception. ([Link](https://stackoverflow.com/questions/1319615/proper-way-to-declare-custom-exceptions-in-modern-python))

Hope this helps!

madhavajay:
Hi @dnabanita7, thanks for your PR. I have refactored the code a little bit.
I think the remaining things to be done are:
- [ ] Add some serde tests
       The proto only takes ints so you might need to change that to support None as well.
- [ ] Add the tests from CPython: https://github.com/python/cpython/blob/3.9/Lib/ctypes/test/test_slicing.py
- [ ] Add the CPython test filename to the LICENSE.txt file
- [ ] Make sure Slice works on other types like Tensor or Pandas
        If there are issues we might need to override the Pointer with a wrapped __getitem__
- [ ] Make sure that __setitem__ also works: https://stackoverflow.com/questions/32772755/how-do-getitem-setitem-work-with-slices

Theres a bit of work here so it might make sense if we split it up. Could you start by writing some tests for the `serde` and fixing up the proto?

dnabanita7:
> Hi @dnabanita7, thanks for your PR. I have refactored the code a little bit.
> I think the remaining things to be done are:
> 
>     * [ ]  Add some serde tests
>       The proto only takes ints so you might need to change that to support None as well.
> 
>     * [ ]  Add the tests from CPython: https://github.com/python/cpython/blob/3.9/Lib/ctypes/test/test_slicing.py
> 
>     * [ ]  Add the CPython test filename to the LICENSE.txt file
> 
>     * [ ]  Make sure Slice works on other types like Tensor or Pandas
>       If there are issues we might need to override the Pointer with a wrapped **getitem**
> 
>     * [ ]  Make sure that **setitem** also works: https://stackoverflow.com/questions/32772755/how-do-getitem-setitem-work-with-slices
> 
> 
> Theres a bit of work here so it might make sense if we split it up. Could you start by writing some tests for the `serde` and fixing up the proto?

Yes yes sure thing! this looks great now! I am sorry not to be have responded earlier as I had my final exams going on. I will try to add tests asap! Thanks for summarizing this up.

madhavajay:
@dnabanita7 No worries. I hope your exams went well! 😊 Take a look at the changes I made so you can get a feel for the way we are handling Python types. Also I think @tudorcebere is probably the best person to check this PR since he did many of the other types. Something which comes to mind is the `primitive_factory` we probably need to add `Slice` in there as well. I will update the issue / todos.

madhavajay:
@dnabanita7 As soon as the `serde tests` are in I think we should merge this PR and then start each of the remaining tasks as a few separate PRs so that they can be smaller and quicker to parallelize.

tudorcebere:
Looks good @dnabanita7!

I'll merge this as CI passes and then I'll open issues with the remaining steps.

Congrats! :)

madhavajay:
@dnabanita7 Thanks for the serde tests. I added another test to check for the last field `step` which is optional and found that it wasn't working because the proto only had ints and the default int in proto is 0. I have added some code to allow for a boolean to determine if the value is set and should be used and defaulted the slice attribute as None if the boolean is false.

I also added some tests to see if it works with Tensor and Pandas DataFrame which it does, and because of the primitive_factory we can just use the normal python syntax:
```python
t = th.Tensor([1, 2, 3])
t_ptr = t.send(alice_client)

res_ptr2 = t_ptr[0:1]
res2 = res_ptr2.get()
assert t[0:1] == res2
```

Pretty cool!
Thanks for your help, and ill get this merged ASAP.

dnabanita7:
> @dnabanita7 Thanks for the serde tests. I added another test to check for the last field `step` which is optional and found that it wasn't working because the proto only had ints and the default int in proto is 0. I have added some code to allow for a boolean to determine if the value is set and should be used and defaulted the slice attribute as None if the boolean is false.
> 
> I also added some tests to see if it works with Tensor and Pandas DataFrame which it does, and because of the primitive_factory we can just use the normal python syntax:
> 
> ```python
> t = th.Tensor([1, 2, 3])
> t_ptr = t.send(alice_client)
> 
> res_ptr2 = t_ptr[0:1]
> res2 = res_ptr2.get()
> assert t[0:1] == res2
> ```
> 
> Pretty cool!
> Thanks for your help, and ill get this merged ASAP.

Yes, I checked that Tensor and Pandas DataFrame didn't work for step because proto takes ints only.

madhavajay:
> > @dnabanita7 Thanks for the serde tests. I added another test to check for the last field `step` which is optional and found that it wasn't working because the proto only had ints and the default int in proto is 0. I have added some code to allow for a boolean to determine if the value is set and should be used and defaulted the slice attribute as None if the boolean is false.
> > I also added some tests to see if it works with Tensor and Pandas DataFrame which it does, and because of the primitive_factory we can just use the normal python syntax:
> > ```python
> > t = th.Tensor([1, 2, 3])
> > t_ptr = t.send(alice_client)
> > 
> > res_ptr2 = t_ptr[0:1]
> > res2 = res_ptr2.get()
> > assert t[0:1] == res2
> > ```
> > 
> > 
> > Pretty cool!
> > Thanks for your help, and ill get this merged ASAP.
> 
> Yes, I checked that Tensor and Pandas DataFrame didn't work for step because proto takes ints only.

Okay, next time let us know so we can help you fix it. 😊

dnabanita7:
Thank you @madhavajay and @tudorcebere for helping out. Now, I understand the codebase better and syft is so cool!

Issue #5224, opened 2021-02-26
Basic API for building plans


koenvanderveen:
## Description
Currently we can only build plans by explicitly defining a collection of actions and inputs. However, we need a more convenient way to do this. A possible solution is to create a bunch of pointers and trace the actions that are executed on them in a subclass of `Node`. We need a mechanism that allows us to define a function or method and automatically:

1) create a TraceNode
2) gather all inputs to the function and creates pointers for them at the TraceNode
3) pass all pointers through the function, while recording the actions passed to TraceNode
4) extract list of actions and gather them in a `Plan`, together with the inputs

## Are you interested in working on this improvement yourself?
- Yes, I am.

PR #5230, opened 2021-03-01
Plan API

koenvanderveen:
## Description
Aims to implement #5224 

## Checklist


tests/syft/core/node/common/plan/plan_test.py
```diff
@@ -203,7 +209,118 @@ def test_plan_batched_execution() -> None:
...
+    local_model = MLP(th)  # type: ignore
+
+    @make_plan
+    def train(  # type: ignore
+        xs=th.rand([64 * 3, 1, 28, 28]),
+        ys=th.randint(0, 10, [64 * 3, 10]),
+        params=List(local_model.parameters()),
+    ):
+
+        model = local_model.send(ROOT_CLIENT)
```
iamtrask, line 301: local_model should be passed in as an argument, no? Otherwise the plan might never be able to find it when the plan is used in the wild somewhere.

tests/syft/core/node/common/plan/plan_test.py
```diff
@@ -203,7 +209,118 @@ def test_plan_batched_execution() -> None:
...
+    alice = sy.VirtualMachine(name="alice")
+    alice_client = alice.get_root_client()
+
+    @make_plan
+    def mul_plan(inp=th.zeros((3)), inp2=th.zeros((3))) -> th.Tensor:  # type: ignore
+        return inp * inp2
+
+    t1, t2 = th.tensor([1, 2, 3]), th.tensor([1, 2, 3])
+    plan_pointer = mul_plan.send(alice_client)
+    res = plan_pointer(inp=t1, inp2=t2)
```
iamtrask, line 242: can this plan be run multiple times? let's add a test to be sure.

iamtrask (changes requested):
Really excellent progress! A couple questions in the tests.

tests/syft/core/node/common/plan/plan_test.py
koenvanderveen: Yes it can, there is already a nb in this pr doing that. Will add a test

tests/syft/core/node/common/plan/plan_test.py
koenvanderveen: Yes that would be better. However, I ran into problems with sending sy.modules, by passing them as arguments, because it relies on the object2proto function, which is not implemented for sy.modules

koenvanderveen: So i think one of my next projects might be to make it serializable. If you have any ideas here let me know

koenvanderveen:
@madhavajay what do you need from me to wrap this up?

madhavajay:
> @madhavajay what do you need from me to wrap this up?

Its high on the list sorry, just getting swamped.

koenvanderveen:
> > @madhavajay what do you need from me to wrap this up?
> 
> Its high on the list sorry, just getting swamped.

np

xutongye:
Great work. I pulled this PR and added a notebook to demo a use case with Opacus+Plan, #5333 . From the process I create the notebook using the Plan API, I feel it very necessary to support `for` loop in Plan. Without supporting to `for` loop, the Plan object created is very very large when I define a full training process as a Plan.

koenvanderveen:
sry for the mess above, accidentally pushed to this branch -_-'

tests/syft/core/node/common/plan/plan_test.py
madhavajay: We are working on serializable nn.Module and then a refactor of Sy.Module for this.

madhavajay (approved):
This is excellent @koenvanderveen! I have played around the with the notebooks and copied PlanBuilder out into the code and added a test.

I noticed a few small things which I think we should address separately to not hold this up and all the new PRs that are depending on it.

I think we can't meaningfully refactor the Plans API until we start using it anyway.

PR #5236, opened 2021-03-02
refactor torch.return_types

### LINKED ISSUE #4945: Refactor Namedtuple and remove ValuesIndices
## Description
To support the namedtuple return types for things like `tensor.cummax` or `tensor.eig` we created a ValuesIndicies proto which is a string key => tensor mapping.

We should create a real namedtuple typle which can contain a normal tuple of any values plus the keys and automatically generate the type on return.

It would be nice to be able to find where the torch namedtuples are generated but in the worst case we can still initialize the objects and curse their classes like in `add_torch_return_types()` if we have to.

## Definition of Done
Namedtuples exist and can be used normally as well as the new implementation replaces the specific ValuesIndicies code.

xutongye:
## Description
This may close #4945 .

Remove `ValuesIndices` and support `torch.return_types`.
1. serde
2. send/get

## Affected Dependencies
No `ValuesIndices` anymore.

## How has this been tested?
- all operators in `allowlist.py` whose return type is `torch.return_types.xx` can still pass torch test.

## Checklist


src/syft/lib/misc/__init__.py
```diff
@@ -51,7 +51,13 @@ def solve_ast_type_functions(path: str) -> KeysView:
     def solve_real_type_functions(path: str) -> Set[str]:
         parts = path.split(".")
         klass_name = parts[-1]
-        return set(dir(getattr(sys.modules[".".join(parts[:-1])], klass_name)))
+        # A way to walkaround the problem we can't `import torch.return_types` and
+        # get it from `sys.modules`. TODO: a general solution to this.
```
madhavajay, line 55: Can we create an issue in GitHub and then put the URL to that issue here where the TODO is so that we won't forget?

src/syft/lib/torch/return_types.py
```diff
@@ -0,0 +1,130 @@
...
+import torch
+
+# syft relative
+from ...generate_wrapper import GenerateWrapper
+from ...lib.util import full_name_with_name
+from ...proto.lib.torch.returntypes_pb2 import ReturnTypes as ReturnTypes_PB
+from ..torch.tensor_util import protobuf_tensor_deserializer
+from ..torch.tensor_util import protobuf_tensor_serializer
+
+# add a fake module "return_types" to torch
```
madhavajay, line 18: Add a TODO and issue in GitHub here to also possibly work around this?

tests/syft/lib/python/namedtuple/namedtuple_test.py
```diff
@@ -4,44 +4,25 @@
 # syft absolute
 from syft import serialize
 from syft.core.common.serde.deserialize import _deserialize
-from syft.lib.python import ValuesIndices
 
 
-def test_torch_valuesindices_serde() -> None:
+def test_torch_return_types_serde() -> None:
     x = torch.Tensor([[1, 2], [1, 2]])
```
madhavajay, line 10: Can we add a test for each type so we are confident they all work?

madhavajay:
@xutongye this is an awesome job! ❤️
My only thought is that is it now no longer possible to remotely construct one of these types?
```
eigenvalues_ptr = duet.python.ValuesIndices.eigenvalues()
```

I don't know if thats hugely important but it would be nice if this functionality was somehow still possible.

src/syft/lib/misc/__init__.py
xutongye: sure

src/syft/lib/torch/return_types.py
xutongye: agree

tests/syft/lib/python/namedtuple/namedtuple_test.py
xutongye: will do

xutongye:
> @xutongye this is an awesome job! ❤️
> My only thought is that is it now no longer possible to remotely construct one of these types?
> 
> ```
> eigenvalues_ptr = duet.python.ValuesIndices.eigenvalues()
> ```
> 
> I don't know if thats hugely important but it would be nice if this functionality was somehow still possible.

We can partially do that. Look at `returntypes_test.py`.

madhavajay (approved):
Great job! This was a pretty messy part of the code base which is much much better now.

Issue #5186, opened 2021-02-18
Add all of torchvision to AST


madhavajay:
## Description
Finish adding all of torchvision to AST with tests and version checks.

## Definition of Done
Just like `torch` this would contain the entire torchvision API and support different versions.

Param-29:
I'd like to work on the issue, I am still not sure about the testing methodology.

Param-29:
PR merged (#5250) Issue can be closed 👍

PR #5250, opened 2021-03-04
Torch vision api's added

Param-29:
## Description
Solves issue #5186 
Added list of datasets from https://pytorch.org/vision/stable/datasets.html
Added list of transforms from https://pytorch.org/vision/stable/transforms.html
## Affected Dependencies
None
## How has this been tested?

## Checklist


Param-29:
Any idea what is causing failure in python-nb-notebooks?

madhavajay:
@Param-29 this is awesome work!! Can we work on a test for all of these just to make sure they work as expected.
I also think we need to start thinking about introducing a special way to provide these from the DO side easily to load into the store, or from the DS side without passing in a remote path (maybe by only using an approved path from the DO side.

Param-29:
@madhavajay I am working on tests for these now, I didn't really understand these 
> we need to start thinking about introducing a special way to provide these from the DO side easily to load into the store, or from the DS side without passing in a remote path (maybe by only using an approved path from the DO side.

madhavajay:
> @madhavajay I am working on tests for these now, I didn't really understand these
> 
> > we need to start thinking about introducing a special way to provide these from the DO side easily to load into the store, or from the DS side without passing in a remote path (maybe by only using an approved path from the DO side.

Great news on the tests! :)

DM me on Slack and I can speak more about the remote path stuff. It doesn't need to be done in this PR but its something that we should think about adding soon.

Param-29:
Following are the list of tests 

**Transforms** 
- [x] ToPILImage 
- [x] RandomErasing
- [x] Normalize
- [x] RandomOrder
- [x] RandomChoice
- [x] GaussianBlur
- [x] TenCrop
- [x] Scale **warning, depricated**
- [x] Resize
- [x] RandomVerticalFlip
- [x] RandomSizedCrop **warning, depricated**
- [x] RandomRotation
- [x] RandomResizedCrop
- [x] RandomPerspective
- [x] RandomHorizontalFlip
- [x] RandomGrayscale
- [x] RandomCrop
- [x] RandomApply
- [x] RandomAffine
- [x] Pad
- [x] Grayscale
- [x] CenterCrop
- [x] ColorJitter
- [x] FiveCrop

**Datasets**

I couldn't test many of these datasets as they are dependent on huge data-sets which need to be downloaded, following are the classes I tested

- [x] EMNIST
- [x] CIFAR10
- [x] CIFAR100
- [x] FakeData
- [x] FashionMNIST
- [x] Flickr8k

**ImageNet's support is removed so I'll remove that.**

**[WIP]**

I am still facing issues adding functional, any idea how could I add the functional module?
I was hoping something like this would work but it didn't
```
allowlist["torchvision.transforms.functional.adjust_brightness"] = "torchvision.transforms.functional.adjust_brightness"
```  
https://pytorch.org/docs/stable/torchvision/transforms.html#functional-transforms

src/syft/lib/torchvision/__init__.py
```diff
@@ -5,6 +5,7 @@
 
 # third party
 from packaging import version
+import torch as tr
```
gmuraru, line 8: Q and NITPICK: Could we leave this as simple ```torch``` or ```th```?

Param-29: Yes, I'll make the change

Param-29:
#5258: I have also added in some features which are available in the latest torchvision (in torch 1.8)

Param-29:
@madhavajay  **PR ready to be merged**, do let me know if you think anything else needs to be added.  


**Functional transformers**

- [x] adjust_brightness 
- [x] adjust_contrast
- [x] adjust_gamma
- [x] adjust_hue
- [x] adjust_saturation
- [x] adjust_sharpness
- [x] affine 
- [x] autocontrast
- [x] center_crop
- [x] crop
- [x] equalize
- [x] erase
- [x] five_crop
- [x] gaussian_blur
- [x] hflip
- [x] invert
- [x] normalize
- [x] pad
- [x] perspective
- [x] pil_to_tensor
- [x] to_grayscale

setup.cfg
```diff
@@ -53,8 +53,8 @@ install_requires =
 # The usage of test_requires is discouraged, see `Dependency Management` docs
 # tests_require = pytest; pytest-cov
 # Require a specific Python version, e.g. Python 2.7 or >= 3.4
-python_requires = >=3.6
```
madhavajay, line 56: Moving this down a line looks like a mistake?

src/syft/lib/torchvision/__init__.py
```diff
@@ -13,6 +14,7 @@
 from .allowlist import allowlist
 
 TORCHVISION_VERSION = version.parse(tv.__version__)
+TORCH_VERSION = version.parse(th.__version__)
```
madhavajay, line 17: Shouldnt we match the torchvision capabilities to the torchvision version? Occasionally there is more than 1 torchvision version for a torch version like 0.8.0 and 0.8.1 due to a bug fixes.

src/syft/lib/torchvision/allowlist.py
```diff
@@ -20,3 +21,219 @@
...
+    "min_version": "1.7.0",
+}
+allowlist["torchvision.transforms.functional.rotate"] = "torch.Tensor"
+allowlist["torchvision.transforms.functional.solarize"] = {
+    "return_type": "torch.Tensor",
+    "min_version": "1.8.0",
+}
+allowlist["torchvision.transforms.functional.ten_crop"] = "torch.Tensor"
+allowlist["torchvision.transforms.functional.to_grayscale"] = "torch.Tensor"
+# Not adding because of datatype doubt torchvision.transforms.functional.to_pil_image
```
madhavajay, line 235: It might be better to leave this in but commented out so that its easy to spot what needs to be done to fix it.
e.g.

```python
torchvision.transforms.functional.to_pil_image(a)
<PIL.Image.Image image mode=RGBA size=736x700 at 0x11599C908>
```
```
# allowlist["torchvision.transforms.functional.to_pil_image"] = "PIL.Image.Image"
```

madhavajay:
@Param-29 How did you run the tests? I don't see any test files?

setup.cfg
Param-29: Yeah, I am not sure what caused it, I'll make the change

src/syft/lib/torchvision/__init__.py
Param-29: I tried comparing the Torchvision version but I was facing some issues, I'll re-check

Param-29 (commented):
> @Param-29 How did you run the tests? I don't see any test files?
@madhavajay 
I checked by running a custom jupyter-notebook with code like this:
`tens_client = torchvision.transforms.functional.adjust_contrast(tens,0.2).get(alice)` 

I tried finding tests for `/tensor/allowlist.py` and `/tensorvision/allowlist.py` to get an idea of what is expected, but couldn't find any. Can you guide me to the location where these tests can be found?

src/syft/lib/torchvision/allowlist.py
```diff
@@ -4,6 +4,7 @@
 
 allowlist: Dict[str, Union[str, Dict[str, str]]] = {}  # (path: str, return_type:type)
 
+allowlist["torchvision.__version__"] = "syft.lib.python.String"
```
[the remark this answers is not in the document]
madhavajay: I uncommented it and it works. 😊

madhavajay:
> > @Param-29 How did you run the tests? I don't see any test files?
> 
> @madhavajay
> 
> I checked by running a custom jupyter-notebook with code like this:
> `tens_client = torchvision.transforms.functional.adjust_contrast(tens,0.2).get(alice)`
> 
> I tried finding tests for `/tensor/allowlist.py` and `/tensorvision/allowlist.py` to get an idea of what is expected, but couldn't find any. Can you guide me to the location where these tests can be found?

`/tests/syft/lib/allowlist_test.py` tests all the torch ops but you will probably not want to re-use that since its very targetted at sending and retrieving stuff to test the torch ops.

I think what would be good is a test file inside `/tests/syft/lib/torchvision/` which basically just creates a `vm` like you did in your notebook and then runs through all these APIs and constructs them and makes sure they don't create errors.

For the dataset downloading if you use a new util class I created we can share the same data dir.
```python
# inside syft/util.py
def get_root_data_path() -> Path:
    # get the PySyft / data directory to share datasets between notebooks
    ...
```

That function should give you the `/PySyft/data` dir from anywhere inside the project.
Also I think we need to change it to this, since I had an issue on Linux with case sensitivity.
```python
-    while os.path.basename(here) != "PySyft" and here != here.parent:
+    while os.path.basename(here).lower() != "pysyft" and here != here.parent:
```

If this test is taking a long time to run due to the downloads of datasets we could try to monkey patch the method, turn off download=True or make sure that the code somehow runs without downloading the datasets in the test environment.

Param-29:
For a lot of functional transformers, they expect a `PIL` image as an input for torch version 1.6. They added in support for `torch.Tensor` for version `1.7`. So I suggest adding a minimum version for all these functional transformers (as I belive, we don't support PIL images currently) 

I would like suggestions on testing the `Datasets` as many of these do not download the dataset (or if they do, its time consuming and would make the tests slow)

Param-29:
@madhavajay I have added in the tests the way we discussed above. Also, any suggestions on testing the `Datasets` without downloading the dataset would be really helpful.

gmuraru:
> @madhavajay I have added in the tests the way we discussed above. Also, any suggestions on testing the `Datasets` without downloading the dataset would be really helpful.

Maybe it might be worth having a dataset in our repo? (and use ```download=False``` when loading the dataset)

Param-29:
@gmuraru If we add dataset in the repo for testing purposes, we would have to add all the datasets which torchvision supports, so as suggested above, I am working on monkey patching the `__init__` for all the classes in Dataset.

madhavajay (approved):
This is great, thank you so much @Param-29!
I have moved two small remaining issues into new issues to deal with later as they arent a big deal and it would be great to get this into the pending release.

PR #5260, opened 2021-03-05
Simple API Refactor

### LINKED ISSUE #5173: Simple API Refactor
## Description
See Design Doc on Notion: https://www.notion.so/openmined/2021-02-10-0-5-API-Refactor-f636ae0ff25f4921a84505ac7bd97ac7

To prevent clashes with existing APIs we should try to prefix any methods we add to Pointers or Classes.

Rename:
- [ ] `ptr.id_at_location` to `ptr.sy_id` (used by SyMPC so we need to think about it
- [x] `searchable` to `pointable`
`t_ptr = t.send(duet, pointable=True)`
- [x] make pointable=True default and pointable=False optional

Add:
- [ ] `var_name` to `send` / `store`
`t_ptr = t.send(duet, pointable=True, var_name="t")`
- [ ] check and prevent duplicates in the store with the same `var_name`
- can we detect the local variable name `t` at the call site?
- if we don't have a mechanism yet add one to `delete` from the store

Add:
- [ ] `duet.store(obj=t)` for the DO side
- [ ] make it so that calling `.send` on your own duet store throws an exception
- for DO this should be:
`duet.store(obj=t)`
- and vica versa for DS
`duet.store(obj=t)` throw exception and require t.send(duet)

Add:
- [ ] syft `ENV` variable `SYFT_MONKEYPATCHED_CLASSES_API` default to `True` for now
if `SYFT_MONKEYPATCHED_CLASSES_API==False` do not add anything to AST enabled classes like:
- `send`
- `tag`
- `description`
- In this case the DS `duet.store(obj=t)` would work as expected

## Definition of Done
These changes are made with tests in place and existing tests updated or removed as needed and passing.
Additionally PyGrid Integration tests should pass, so this might require opening a PR on the PyGrid side to update these API calls if needed.

xutongye:
## Description
Aims to close #5173 .

## Affected Dependencies

## How has this been tested?

## Checklist


madhavajay (approved):
LGTM! I added the old API back as deprecated and forwarded through to the new API and added a deprecated API test.

I also disabled the MacOS Integration tests for the moment because they sometimes hang CI.

PR #5261, opened 2021-03-06
fixed code quality issues

withshubh:
### Description
Hi :wave: 
I ran the [DeepSource](https://deepsource.io) analysis on the forked copy of this repo and found some interesting [code quality issues](https://deepsource.io/gh/withshubh/PySyft/issues/?category=recommended).

##### Motivation and Context
Some of the issues I have fixed boost minor performance and a few removes the anti-patterns in code.

#### Summary of changes

- Remove length check in favour of truthiness of the object
- Remove unnecessary use of comprehension
- Use `callable()` to check if the object is calllable
- Replace ternary syntax with if expression
- Add the bound instance as method parameter
- Fix dangerous default argument
- Add .deepsource.toml

## Checklist


tudorcebere:
Hi @withshubh

Nice fixes here! A few comments:
* don't touch the protobuf files, as they are automatically generated

@madhavajay, is a thing of interest to keep a CI integration test with DeepSource or we should remove the .toml file? The issues found by the tool are super awesome and it might be a nice thing to have.

withshubh:
Hi @tudorcebere :wave: 

Thank You for your review! :sparkling_heart: 

Shall I revert the fix from protobuf file? :thinking: 

Also, Just to let you know all these fixes are made using the DeepSource auto-fix feature. :grin:

tudorcebere:
@withshubh yeah, the protobuf fix should be reverted and there are a few fixes that should be added to the linting, then this should be good to go. :+1:

withshubh:
> @withshubh yeah, the protobuf fix should be reverted and there are a few fixes that should be added to the linting, then this should be good to go.

@tudorcebere 
- Reverted the fix in protobuf file
- Fixed linting issues using pre-commit (mypy in pre-commit is still failing)
- Added exclude pattern for protobuf files in .deepsource.toml file

tudorcebere:
It looks like the fixes generated some mypy issues @withshubh. Could you fix them before merging?

src/syft/core/node/common/node.py
```diff
@@ -262,8 +262,10 @@ def __init__(
     def icon(self) -> str:
         return "📍"
 
-    def get_client(self, routes: List[Route] = []) -> ClientT:
-        if not len(routes):
+    def get_client(self, routes: List[Route] = None) -> ClientT:
```
madhavajay, line 265: We run mypy with `--no-implicit-optional` so this will probably fail in our linter step.
I think the best way to refactor this would be to introduce an Optional or just remove the two if statements since they are not needed.

src/syft/core/node/common/node.py
```diff
@@ -281,7 +283,9 @@ def get_client(self, routes: List[Route] = []) -> ClientT:
             verify_key=None,  # DO NOT PASS IN A VERIFY KEY!!! The client generates one.
         )
 
-    def get_root_client(self, routes: List[Route] = []) -> ClientT:
+    def get_root_client(self, routes: List[Route] = None) -> ClientT:
+        if routes is None:
```
madhavajay, line 287: same as above

src/syft/core/node/device/device.py
```diff
@@ -40,10 +40,12 @@ def __init__(
         device: SpecificLocation = SpecificLocation(),
         vm: Optional[Location] = None,
         device_type: DeviceType = unknown_device,
-        vms: Dict[UID, VirtualMachine] = {},
+        vms: Dict[UID, VirtualMachine] = None,
```
madhavajay, line 43: This used to be an Optional so i think it has just been wrongly changed, but perhaps it can be refactored by removing the default below...

madhavajay: We need to be careful here since the concrete child classes actually pass through None:
```python
def __init__(
        self,
        *,  # Trasterisk
        name: Optional[str] = None,
        network: Optional[Location] = None,
        domain: Optional[Location] = None,
        device: Optional[Location] = None,
        vm: SpecificLocation = SpecificLocation(),
        signing_key: Optional[SigningKey] = None,
        verify_key: Optional[VerifyKey] = None,
    ):
        super().__init__(
            name=name,
            network=network,
            domain=domain,
            device=device,
            vm=vm,
            signing_key=signing_key,
            verify_key=verify_key,
        )
```

So I think these are actually correct.

```python
# must be supplied and defaults to {}
vms: [Dict[UID, VirtualMachine]] = {}
```
If we call this parent constructor with a `None`, we need to then detect that down stream and overwrite it since an explicit None here from the subclass won't use the default...

src/syft/lib/python/dict.py
```diff
@@ -36,7 +36,7 @@ class Dict(UserDict, PyPrimitive):
     # python 3.8 signature includes a new PEP 570 (args, /, kwargs) syntax:
     # https://www.python.org/dev/peps/pep-0570/
 
-    def __init__(*args: Any, **kwargs: Any) -> None:
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
madhavajay, line 39: This is also a super special case where we actually need the self as part of the args to support older python versions.

madhavajay:
@withshubh thanks for the PR, this is certainly interesting. I think if we can add some `# ignore rules` to the areas where deepsource has false positives and make sure that CI is passing then we can look at adding this. How do we actually consume these checks, is it via deepsource or can we get a CI check added to report on these? I would opt for a non failing test initially so that we can evaluate how this goes and if there is any value.

madhavajay:
@withshubh Is it possible to make the changes I suggested above?

withshubh:
Hey @madhavajay 

Apology for late response. I skipped the previous notification.

Yes, you can add the `skipcq` to ignore the issue. Read more [here](https://deepsource.io/docs/how-to/silence-issues.html)

withshubh:
Working on the changes requested 🚀

withshubh:
@madhavajay - made the changes please review! :sparkles:

madhavajay:
Hi @withshubh you missed a bunch of my requested changes and didn't fix CI. I have fixed these issues and reverted a major cause of breaking tests which was the false positive change to the `Dict` init function which has a whole comment block above it explaining why it needs to be special.

Overall I would say that this experience with `deepsource.io` has been disappointing. There were far more false positives than true positives and the fixes themselves were often questionable, lacking "deep" enough understanding of the class hierarchy. Analyzing such a dynamic language as Python is no doubt quite hard, however I am skeptical of the value of this tool right now, over our existing tooling which seems to be providing very high quality true positive checks, and so im going to remove the `.deepsource.toml` file for now.

I would love to be convinced otherwise in future if the tool improves and PRs to include it in our code base, pass the CI test suite without our team having to fix all the mistakes that false positives introduce to our code base.

tudorcebere (approved):
The changes look good to me, but as @madhavajay said, I wouldn't keep the tool around the codebase. 

Thank you for your interest in syft!

PR #5290, opened 2021-03-10
Param Objs not removed

gmuraru:
## Description
Make the ```args```/```kwargs```.

## Affected Dependencies

## How has this been tested?
- Every test should pass

## Checklist

### DIFF
--- a/requirements/supported_libs.txt
+++ b/requirements/supported_libs.txt
@@ -1,5 +1,5 @@
 opacus
 openmined.psi
 python-dp
-sympc @ git+https://github.com/OpenMined/SyMPC@abd30cd33c565fd249cbd66a242e718dba0b66a4
+sympc @ git+https://github.com/OpenMined/SyMPC@main
 tenseal
--- a/requirements/supported_libs.txt
+++ b/requirements/supported_libs.txt
@@ -1,4 +1,4 @@
-opacus
+opacus==0.12.0
 openmined.psi
 python-dp
 sympc @ git+https://github.com/OpenMined/SyMPC@main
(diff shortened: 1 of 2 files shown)


requirements/supported_libs.txt
```diff
@@ -1,5 +1,5 @@
-opacus
+opacus==0.12.0
```
gmuraru, line 1: This is needed because in version 0.13.0 ```opacus``` uses ```torchvision 0.9.0``` which needs ```torch 1.8``` - Created a subsequent [ticket](https://github.com/OpenMined/PySyft/issues/5291) for adding ```opacus``` back when we have ```torch 1.8``` support.

**IMPORTANT**: I saw that we install all the supported libs for every torch version -- I think we should have only one workflow when we test them -- if we continue to do this we would end up testing only torch1.8 (if we solve the torchvision depedency by adding torch1.8 to syft, then ```opacus``` would install each time this version of torch)

madhavajay: We could move to testing the more complex set of versions less frequently like nightly or weekly.

madhavajay (approved):
Good catch.
This underscores more reason why we need to come up with a good GC test suite so we know when this kind of stuff breaks.

requirements/supported_libs.txt
gmuraru: The problem with this I think it is that we should start pinning versions.

madhavajay: Seems like this is purely to work around the MNIST dataset issue that happened the other day, I think we can try `--no-deps` instead and still use the latest opacus with older torch. It works locally in my branch, lets see what CI looks like.

PR #5314, opened 2021-03-17
Remote Dataloader

xutongye:
## Description
In normal use case, a Dataset on data owner's machine will not be allowed to get downloaded by data scientist. The DS will only need to access the pointers to sample/batch. So, in this case, we don't need to serde the whole training data. This PR aims at building a RemoteDataset and RemoteDataloader to support this use case in an efficient way. 

## Affected Dependencies

## How has this been tested?
- tests/syft/core/remote_dataloader/test_remote_dataloader.py
- example/experimental/tongye/train_boston_with_RemoteDataLoader

## Checklist


madhavajay (approved):
This works well, I tested the notebook and also modified the RemoteDataset to support csv files and used that in the CDC notebook and it also worked by passing the result of a full slice [:] to the DataFrame constructor.

I think this is sufficient for now so that with a little pre-processing work we can at least load large tensors or dataframes into the store.

PR #5320, opened 2021-03-18
Order dict __getitem__ fix

gmuraru:
## Description
Treat the case when we have a Tensor in the Ordered Dict (is the same fix as the one with the Dict).

## Affected Dependencies

## How has this been tested?
- Create a remote OrderedDict with a tensor as a value

## Checklist


madhavajay (approved):
Good catch!
We need a long term solution for the `get` method though, seems a bit annoying for dicts.

PR #5321, opened 2021-03-18
Add shape property allowlist

gmuraru:
## Description
This would allow us to implement in SyMPC the possibility to share a model that has the weight remote (for this we need to access the shape of the weights).

For the moment we can not serialize ```torch.Size```, but we can return a pointer to an object and then iterate over it (we do that in SyMPC in this [PR](TODO)).

## Affected Dependencies
SyMPC

## How has this been tested?
- In SyMPC

## Checklist


madhavajay:
@gmuraru why don't we just add the code to turn size into a tuple and then back again on the other side?

```python
>>> a = torch.Tensor([1, 2, 3])
>>> torch.stack((a, a, a))
tensor([[1., 2., 3.],
        [1., 2., 3.],
        [1., 2., 3.]])
>>> tuple(b.size())
(3, 3)
>>> c = tuple(b.size())
>>> torch.Size(c)
torch.Size([3, 3])
>>> x = torch.Size(c)
>>> x == b.size()
True
```

gmuraru:
> @gmuraru why don't we just add the code to turn size into a tuple and then back again on the other side?
> 
> ```python
> >>> a = torch.Tensor([1, 2, 3])
> >>> torch.stack((a, a, a))
> tensor([[1., 2., 3.],
>         [1., 2., 3.],
>         [1., 2., 3.]])
> >>> tuple(b.size())
> (3, 3)
> >>> c = tuple(b.size())
> >>> torch.Size(c)
> torch.Size([3, 3])
> >>> x = torch.Size(c)
> >>> x == b.size()
> True
> ```

I tried this! but I got an error because I did not have access to ```len``` (from a VirtualClient) - it throws my an exception because i do not have the permissions.

Shouldn't all VirtualClients have access to ```len``` an ```iterator``` without having to request access? (because we can call ```get``` without explicitly specifying the permissions - I was expecting the same behavior)

PR #5336, opened 2021-03-22
hide VerifyAll

xutongye:
## Description
We need one and only one global instance of `VerifyAll`, and use this instance everywhere needed. 

This PR tries to "hide" the class `VerifyAll`, and only expose a global instance of it, named `VERIFYALL`. The way I "hide" `VerifyAll` is to define it inside a function, make it a local class, and hard for users to import and use it directly.

## Affected Dependencies

## How has this been tested?

## Checklist


src/syft/core/common/group.py
```diff
@@ -32,15 +32,21 @@ def proto2object(proto: VerifyKey_PB) -> VerifyKey:
...
-        return VerifyAllWrapper_PB(all=Empty_PB())
-
-    @staticmethod
-    def _proto2object(proto: VerifyAllWrapper_PB) -> Type:  # type: ignore
-        return VerifyAll
-
-    @staticmethod
-    def get_protobuf_schema() -> GeneratedProtocolMessageType:
-        return VerifyAllWrapper_PB
+def _create_VERIFYALL() -> Any:
```
tudorcebere, line 35: Why not make it a Singleton just to be sure instead of hiding it in a function?

CC: @madhavajay

xutongye: @madhavajay said that we don't want users call like `isinstance(some_key, VerifyAll)`, but want them to use `some_key is VERIFYALL`.

madhavajay: I think we can make this a singleton as well just make sure its not possible to get the type and then create a different copy. So there are two things here:
✅ 1) hide the class from import
☑️ 2) make it a singleton so you end up with if if you try to make another one from its type constructor

src/syft/util.py
```diff
@@ -104,6 +104,11 @@ def index_syft_by_module_name(fully_qualified_name: str) -> object:
 
     """
     attr_list = fully_qualified_name.split(".")
+
+    # we deal with VerifyAll differently, because we don't it be imported and used by users
+    if attr_list[-1] == "VerifyAll":
+        return type(syft.core.common.group.VERIFYALL)
```
madhavajay, line 110: If it was a singleton perhaps this would let new construction match?

tudorcebere (approved):

PR #5337, opened 2021-03-22
add numpy.ndarray

Koukyosyumei:
## Description
Support numpy.ndarray

## Affected Dependencies
numpy

## Checklist


madhavajay:
@Koukyosyumei awesome work.
Could we add a few unit tests to make sure it works and possibly iterate through the available data types and make sure they also go to their destination and come back as the same type without issue?

The test can just use a `VirtualMachine`:
```python
>>> import syft as sy
>>> vm = sy.VirtualMachine()
>>> client = vm.get_root_client()
>>> import numpy as np
>>> a = np.array([1, 2, 3])
>>> a
array([1, 2, 3])
>>> sy.load("numpy")
>>> x = a.send(client)
>>> y = x.get()
>>> y
array([1., 2., 3.], dtype=float32)
>>> type(y)
<class 'numpy.ndarray'>
>>> assert a == y
```

Koukyosyumei:
@madhavajay 
I'll implement it asap!

Koukyosyumei:
@madhavajay 

Since there are some differences between dtypes of torch.Tensor and numpy.array, I cannot make np.uint16, np.uint32, np.uint64, np.str_, np.object_ work. Also, because syft currently doesn't support complex numbers, np.complex64 and np.complex128 are not available.

So, the available dtypes are as follows.

        np.int8
        np.int16
        np.int32
        np.int64
        np.uint8
        np.float16
        np.float32
        np.float64
        np.bool_

src/syft/lib/numpy/array.py
```diff
@@ -0,0 +1,39 @@
...
+from ...lib.torch.tensor_util import protobuf_tensor_serializer
+from ...proto.lib.torch.tensor_pb2 import TensorData
+
+
+def object2proto(obj: np.ndarray) -> TensorData:
+    tensor = torch.from_numpy(obj).clone()
+    tensor_proto = protobuf_tensor_serializer(tensor)
+    # TODO:
+    # support numpy.object_, numpy.str_, numpy.unicode_,
+    #         numpy.uint16, numpy.uint32, numpy.uint64,
```
madhavajay, line 17: If we add a custom Numpy proto which contains an additional field which is the numpy type as a string.

Then we should be able to do something like this:
```python
import numpy as np
import torch

# our unsupported numpy array
x = np.array([0, 1, 2], dtype=np.uint16)
npdtype = x.dtype

# refactor to a dict of uint -> int types for matching
if npdtype == np.uint16:
    x = x.astype(np.int16)

y = torch.from_numpy(x)

# serialize along side the original dtype
ser_y = y
ser_numpy_dtype = npdtype.name


# deserialize on other side
y = ser_y
numpy_dtype = ser_numpy_dtype

# conver to numpy first
z = y.numpy()

# get numpy type from string
npdtype = np.dtype(numpy_dtype)

# down sample
z = z.astype(npdtype)

assert z.all() == x.all()
assert z.dtype == np.uint16
```

Leave off complex, unicode, str and object for now and just throw exceptions and we can make it seperate tasks to handle them as a proper Numpy Array serde protobuf.

tests/syft/lib/numpy/array_test.py
```diff
@@ -0,0 +1,37 @@
+# third party
+import numpy as np
```
madhavajay, line 2: Make sure to move the import numpy as np and other np calls to inside the test.
That way when the test collection process checks for `numpy` and if it doesn't find it it wont crash collecting the test file. In reality no one will not have numpy but its still how the vendor lib tests are done.

madhavajay:
@Koukyosyumei this is awesome, and a much needed bit of functionality to unblock many different library integrations which return `numpy.ndarray` types.

One final change, can you add `numpy` to `requirements.libs.txt` its not explicitly included anywhere and it would be good to add it to be specific.

tests/syft/lib/numpy/array_test.py
Koukyosyumei: Thank you!

src/syft/lib/numpy/array.py
Koukyosyumei: Please correct me if I'm wrong, could I also ignore uint and raise exceptions or should I make a custom proto which supports uint now?

Koukyosyumei:
@madhavajay 
Thanks for your review and comments! I have one question. Please correct me if I'm wrong, could I also ignore uint and raise exceptions or should I make a custom proto which supports uint now?

src/syft/lib/numpy/array.py
madhavajay: I would add it now since the majority of things we want numpy for is numbers so it would be good if we dont hit any of these by accident and have it not work.

To add the custom proto just create one that looks something like this:
```proto
syntax = "proto3";

package syft.lib.numpy;

import "proto/lib/torch/tensor.proto";

message NumpyProto {
  syft.lib.torch.TensorData tensor = 1;
  string dtype = 2;
}

```

Then run: `scripts/build_proto.sh`.

Koukyosyumei: Thank you for your clear instruction! I implemented this, and it worked! We can now use uint!

madhavajay:
Looks like we need to check the `dtype` as well just in case since different data types still match the == condition.
```python
>>> import numpy as np
>>> x = np.array([1, 2, 3])
>>> y = np.array([1, 2, 3], dtype=np.int32)
>>> x
array([1, 2, 3])
>>> y
array([1, 2, 3], dtype=int32)
>>> x == y
array([ True,  True,  True])
>>> x.dtype == y.dtype
False
```

```python
assert test_array.dtype == received_array.dtype

E           AssertionError: assert dtype('uint16') == dtype('int16')
E             +dtype('uint16')
E             -dtype('int16')
```

Interesting. Let me investigate.

Koukyosyumei:
@madhavajay 
Thank you! I didn't know that. I fixed the process of dtype and added a dtype check to test.

madhavajay:
@Koukyosyumei Nice!
I also went a step further and refactored some more so that we can only support what we know we support, and make sure the `max` and `min` values of all types are tested. The naming conventions in Python usually reserve 🐫 `CamelCase` for class names, and GLOBAL_STATIC type variables as all 🐍 `UPPER_SNAKE_CASE`.

madhavajay (commented):
Awesome work!

madhavajay (approved):
🙌

PR #5342, opened 2021-03-23
Support torch.nn.Module

xutongye:
## Description
Support serde and send of `torch.nn.Module` objects, including pre-defined `Module` types in `torch` like `Linear/Sequential/...` and user defined models subclassed from `nn.Module`.

PS: This PR is directely based on #5230 , because I am considering taking use of `Plan` to support remote call of user defined models.

## Affected Dependencies

## How has this been tested?
- examples/experimental/tongye/torch_nn_Module.ipynb
## Checklist


koenvanderveen:
I wonder whether we really need to serialize the forward method as a plan here. Alternatively, we could just serialize the state of the module, call the .forward method on the module pointer, and trace the actions that originate from that call. So a module in a plan instead of a plan in a module.

madhavajay (approved):
I have added the notebook examples to tests and I think this is good to merge. We should discuss how best to use this in combination with Plans and the existing Sy.Module.

Issue #4850, opened 2020-11-24
Support Zero-knowledge Proofs


madhavajay:
## Description
This ticket relates to implementing some kind of MVP for Zero-knowledge proofs integration for educational purposes for the 0.4.0 Milestone.

## Contacts
- @madhavajay

The library for this will be:
https://github.com/spring-epfl/zksk

## TODO
- [ ] Initial Hello World Notebooks
- [x] Serialization and Deserialization of ZKP types
- [ ] 100% of API in AST
- [x] LMW support
- [ ] Flesh out Notebook Examples

## Definition of Done
A notebook is created in examples/zero-knowledge-proofs or something similar which demonstrates how ZKP could be used with Duet sufficiently to achieve the educational goals.

Param-29:
I'd like to work on this issue.

PR #5347, opened 2021-03-23
[WIP] Integrating zero knowledge proof 

Param-29:
## Description

#4850

## Affected Dependencies

## How has this been tested?

## Checklist


madhavajay:
@Param-29 this is excellent work! I have fixed the conflict with `src/syft/lib/__init__.py` since these are optional libraries we won't load them on startup but make it a separate `sy.load` call.

Theres a few naming changes to make and then it I would love to see the `Peggy` and `Victor` example from the README https://github.com/spring-epfl/zksk in a test file and that should be ready for merge! 🙌

proto/lib/petlib/ec_obj.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package syft.lib.petlib;
+
+message EC2_obj {
```
madhavajay, line 5: This is the EcPt right? Can we just name it `EcPt`?

proto/lib/petlib/ec_obj.proto
```diff
@@ -0,0 +1,10 @@
...
+  uint64 group_nid = 1;
+  string obj_type = 2;
+  bytes vector = 3;
```
madhavajay, line 8: Rename `vector` to `data`.

requirements.txt
```diff
@@ -10,6 +10,7 @@ matplotlib
 nest_asyncio
 packaging
 pandas
+petlib
```
madhavajay, line 13: Move this into the `requirements.libs.txt`.

requirements.txt
```diff
@@ -18,3 +19,4 @@ sqlitedict
 typeguard
 typing-extensions # backport to older python 3
 websockets
+zksk
```
madhavajay, line 22: Move this into the `requirements.libs.txt`.

setup.cfg
```diff
@@ -46,6 +46,8 @@ install_requires =
     sqlitedict
     torch>=1.4.0,<=1.8.0
     torchvision>=0.5,<=0.9
+    zksk
```
madhavajay, line 49: Remove these two, since these are not hard dependencies for the Syft library.

src/syft/lib/petlib/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+from typing import Union
+
+# third party
+import petlib
+
+# syft relative
+from . import ec  # noqa: 401
+from ...ast.globals import Globals
+from .allowlist import allowlist
+
```
madhavajay, line 13: You will need to adhere to the loadable library API, check something like `tenseal/__init__.py`:

```python
LIB_NAME = "tenseal"
PACKAGE_SUPPORT = {
    "lib": LIB_NAME,
    "python": {"max_version": (3, 9, 99)},
}


def create_ast(client: TypeAny) -> Globals:
    ast = Globals(client=client)
```

src/syft/lib/petlib/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+        return support_dict["return_type"]
+
+
+def create_petlib_ast(client: Any = None) -> Globals:
+    ast = Globals(client)
+
+    # most methods work in all versions and have a single return type
+    # for the more complicated ones we pass a dict with keys like return_type and
+    # min_version
+    for method, return_type_name_or_dict in allowlist.items():
```
madhavajay, line 28: You could convert this to the inline version used in supported libraries like `tenseal` since its a bit easier to read and update.

src/syft/lib/petlib/ec.py
```diff
@@ -0,0 +1,50 @@
+# stdlib
+
+# third party
+import petlib
+
+# syft relative
+from ...generate_wrapper import GenerateWrapper
+from ...proto.lib.petlib.ec_obj_pb2 import EC2_obj
```
madhavajay, line 8: Rename `EC2_obj` to `EcPt_PB` to be consistent.

src/syft/lib/petlib/ec.py
```diff
@@ -0,0 +1,50 @@
...
+)
+
+
+def object2proto_group(obj: object) -> EC2_obj:
+    proto = EC2_obj()
+    proto.group_nid = obj.group.nid()  # type: ignore
+    return proto
+
+
+def proto2object_group(proto: EC2_obj) -> petlib.ec.EcGroup:
```
madhavajay, line 39: I don't think these should share a protobuf, maybe just split it out to a seperate protobuf.

src/syft/lib/zksk/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+
+# third party
+import zksk as zk
+
+# syft relative
+from . import secret  # noqa: 401
+from ...ast.globals import Globals
+from .allowlist import allowlist
+
+
```
madhavajay, line 14: Same as above, make this dynamically `sy.load` able.

src/syft/lib/zksk/secret.py
```diff
@@ -0,0 +1,38 @@
+# stdlib
+
+# third party
+import petlib  # noqa: 401
+from petlib.bn import Bn
+import zksk
+
+# syft relative
+from ...generate_wrapper import GenerateWrapper
+from ...proto.lib.petlib.ec_obj_pb2 import EC2_obj
```
madhavajay, line 10: This is a different library so I would just change this entirely to not use the same protobuf.
We already have a generic protobuf for when you want to just send bytes, which `tenseal` is also using.

```proto
message VendorBytes {
  string obj_type = 1;
  string vendor_lib = 2;
  string vendor_lib_version = 3;
  bytes content = 4;
}
```

madhavajay (changes requested):

proto/lib/petlib/ec_obj.proto
Param-29: I have named it `EC2_obj` because I used it for both, `EcPt` and `EcGroup`, I could change it to something else though

src/syft/lib/petlib/__init__.py
Param-29: Sure, I find allowlist a lot cleaner, but if that is what we follow, I'll make the change

Param-29:
I have made the changes you suggested, for running hello-world notebooks, I am facing 2 issues 
You can find the code and the issue in jupyter-notebook `Issue 4850`

Errors:
```
TypeError: 'Expression' object is not iterable
```

Expression object is not expected to be iterable 

```
TypeError: cannot unpack non-iterable EcPtPointer object
```

I couldn't find iterator for `EcPt`

madhavajay:
@Param-29 okay no problem I will take a look as soon as I get a chance. 👍🏼

madhavajay:
> I have made the changes you suggested, for running hello-world notebooks, I am facing 2 issues
> You can find the code and the issue in jupyter-notebook `Issue 4850`
> 
> Errors:
> 
> ```
> TypeError: 'Expression' object is not iterable
> ```
> 
> Expression object is not expected to be iterable
> 
The issue here is that you are trying to construct 2 x `petlib.ec.EcPt` but the allowlist return type is set to a single one.

```python
allowlist["zksk.utils.make_generators"] = "petlib.ec.EcPt"
```

It seems that the return type of `make_generators` is actually a list even if you only ask for one.
```
>>> x = zksk.utils.make_generators(num=1, seed=42)
>>> type(x)
<class 'list'>
```

I have also noticed that the protobuf hasn't been regenerated which is causing issues with the `serde` of the EcPt type. I would highly suggest writing a few light tests which just create, send and get all of these types you are supporting to make sure that all works first.

madhavajay:
@Param-29 I saw `E   ModuleNotFoundError: No module named 'petlib'`. I personally had some issues compiling petlib on macos before due to a dependency on `bplib`. I wonder if its related.

tests/syft/lib/petlib/ec_obj_test.py
```diff
@@ -0,0 +1,35 @@
+# Testing code
+
+# third party
+import petlib
```
madhavajay, line 4: this is breaking CI because petlib doesnt exist until the vendor lib install step, so you need to move it inside the test and make sure the test has a @pytest.vendor(lib="petlib") tag around it.

madhavajay:
@Param-29 I went ahead and finished this off for now, its all working. Thanks for your help. 👍🏼

PR #5363, opened 2021-03-24
Iterator/GC fixes to solve store __delitem__ errors

tudorcebere:
## Description
Fixing a few iterator return types

## Affected Dependencies

## How has this been tested?

## Checklist


gmuraru (approved):

src/syft/lib/python/iterator.py
```diff
@@ -21,6 +21,9 @@ def __init__(self, _ref: Any, max_len: Optional[int] = None):
     def __iter__(self) -> "Iterator":
         return self
 
+    def __len__(self) -> int:
```
madhavajay, line 24: Theres some code below that I wrote:
```python
try:
    if hasattr(_obj_ref, "__len__"):
        max_len = _obj_ref.__len__()
except AttributeError:
    # I am not sure why this happens on some types
    pass
```

I can't remember but I am pretty sure this happens, maybe its on generators which cant give you their length because its not known, or something like that, either way, I think its possible that there is no known `len` sometimes. Should we add some `try` and logging just in case so we can spot it when it blows up?

madhavajay:
Awesome spot! I have created a related ticket because of this find. https://github.com/OpenMined/PySyft/issues/5369

We might have more completed functionality which isn't completely turned on.

src/syft/lib/python/iterator.py
tudorcebere: Maybe we should add generators as a separated supported type instead?

Issue #5045, opened 2021-01-21
Basic Plans TDD


madhavajay:
## Description
Create a series of expected outputs based on the `0.2.x` version of plans that we add as tests to Syft to start the process of developing Custom Tensors and Plans to be compatible with the same expected output needed for Model Centric FL.

Notebook:
https://github.com/OpenMined/PyGrid/blob/dev/examples/model-centric/01-Create-plan.ipynb

## Definition of Done
Several stub functions in Syft which are expected to output the relevant serialized plans, and a list of tests added to syft `0.4` which are marked `@pytest.mark.xfail` which attempt to match the output.

PR #5372, opened 2021-03-25
Plan to Torchscript translation

vvmnnnkv:
## Description
Part of https://github.com/OpenMined/PySyft/issues/5045:
1. Defines Plan with basic MLP model (backward pass is coded by hand)
2. Translate this Plan to torchscript
3. Torchscript produces same output as syft Plan
Note:
4. Need to test torchscript on mobiles

## Affected Dependencies
Added syft_proto dependency to syft, to use old Plan protobuf

## How has this been tested?
- Unit test

## Checklist

### DIFF
--- a/src/syft/core/plan/translation/torchscript/plan.py
+++ b/src/syft/core/plan/translation/torchscript/plan.py
@@ -1,0 +1,86 @@
+# stdlib
+import io
+
+# third party
+from google.protobuf.reflection import GeneratedProtocolMessageType
+from syft_proto.execution.v1.plan_pb2 import Plan as Plan_PB
+import torch as th
+
+# syft absolute
+from syft.core.common.object import Serializable
+from syft.logger import traceback_and_raise
+
+
+class PlanTorchscript(Serializable):
+    """
+    Represents Syft Plan translated to TorchScript
+    """
+
+    def __init__(self, torchscript=None):
+        super().__init__()
+        self.torchscript = torchscript
+
+    @staticmethod
+    def get_protobuf_schema() -> GeneratedProtocolMessageType:
+        """Return the type of protobuf object which stores a class of this type
+
+        As a part of serialization and deserialization, we need the ability to
+        lookup the protobuf object type directly from the object type. This
+        static method allows us to do this.
+
+        Importantly, this method is also used to create the reverse lookup ability within
+        the metaclass of Serializable. In the metaclass, it calls this method and then
+        it takes whatever type is returned from this method and adds an attribute to it
+        with the type of this class attached to it. See the MetaSerializable class for details.
+
+        :return: the type of protobuf object which corresponds to this class.
+        :rtype: GeneratedProtocolMessageType
+
+        """
+        return Plan_PB
+
+    def __call__(self, *args):
+        if self.torchscript:
+            return self.torchscript(*args)
+        else:
+            traceback_and_raise("No torchscript")
+
+    def _object2proto(self) -> Plan_PB:
+        """Returns a protobuf serialization of self.
+
+        As a requirement of all objects which inherit from Serializable,
+        this method transforms the current object into the corresponding
+        Protobuf object so that it can be further serialized.
+
+        :return: returns a protobuf object
+        :rtype: ObjectWithID_PB
+
+        .. note::
+            This method is purely an internal method. Please use object.serialize() or one of
+            the other public serialization methods if you wish to serialize an
+            object.
+        """
+        bin = self.torchscript.save_to_buffer() if self.torchscript else None
+        return Plan_PB(torchscript=bin)
+
+    @staticmethod
+    def _proto2object(proto: Plan_PB) -> "PlanTorchscript":
+        """Creates a ObjectWithID from a protobuf
+
+        As a requirement of all objects which inherit from Serializable,
+        this method transforms a protobuf object into an instance of this class.
+
+        :return: returns an instance of Plan
+        :rtype: Plan
+
+        .. note::
+            This method is purely an internal method. Please use syft.deserialize()
+            if you wish to deserialize an object.
+        """
+        if proto.torchscript:
+            bin = io.BytesIO(proto.torchscript)
+            ts = th.jit.load(bin)
+        else:
+            ts = None
+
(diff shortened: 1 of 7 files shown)


cereallarceny (approved):

src/syft/core/plan/translation/torchscript/plan_translate.py
```diff
@@ -0,0 +1,108 @@
...
+            self._kwarg_names = kwarg_names
+
+        def forward(self, *args: Any) -> Any:
+            kwarg_ptrs: Dict[str, Pointer] = {}
+
+            # Since Syft Plan needs pointers as args,
+            # reverse-map actual arg values to pointers
+            for name, arg in zip(self._kwarg_names, args):
+                found = False
+                for ptr in ROOT_CLIENT.store:
```
koenvanderveen, line 45: It would be great if we could somehow simplify this search. The whole part of Actions working with pointers is also somewhat annoying in the normal `Plan` class. Maybe for now extract this complexity in a function/classmethod to improve readability?

koenvanderveen (approved):

src/syft/core/plan/translation/torchscript/plan_translate.py
```diff
@@ -0,0 +1,108 @@
...
+        so we need to have kwarg names inside torchscript to pass them into Plan.
+        We use nn.Module to store input kwarg names inside torchscript.
+        """
+
+        def __init__(self, kwarg_names: List[str]):
+            _kwarg_names: List[str]
+            super(PlanWrapper, self).__init__()
+            self._kwarg_names = kwarg_names
+
+        def forward(self, *args: Any) -> Any:
```
koenvanderveen, line 38: maybe execute is a better name, as people may associate `.forward` with `torch.nn.module.forward`

koenvanderveen:
In general: looks great!

src/syft/core/plan/translation/torchscript/plan_translate.py
vvmnnnkv: it really is nn.Module :) 
This wrapper is a workaround to store extra state (kwarg names for Plan) that can't be passed as argument into torchscript (note: it's possible to pass dict to torchscript, but dict values need to have same type, i.e. you'd need to flatten model params list into this dict)

src/syft/core/plan/translation/torchscript/plan_translate.py
```diff
@@ -0,0 +1,108 @@
...
+            # reverse-map actual arg values to pointers
+            for name, arg in zip(self._kwarg_names, args):
+                found = False
+                for ptr in ROOT_CLIENT.store:
+                    store_value: Any = PLAN_BUILDER_VM.store[ptr.id_at_location].data
+                    is_list = isinstance(arg, (SyList, list)) and isinstance(
+                        store_value, (SyList, list)
+                    )
+                    is_dict = isinstance(
+                        arg, (dict, Dict, OrderedDict, SyOrderedDict)
```
tudorcebere, line 51: could we define some constants somewhere? Like: LIST_TYPES = (SyList, list), DICT_TYPES = (dict, Dict, OrderedDict, SyOrderedDict)? you can store these in syft/lib/python?

src/syft/core/plan/translation/torchscript/plan_translate.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # Builder VM holds inputs in the store, retrieve actual arg values from there
+    kwarg_names = list(plan.inputs.keys())
+    args = tuple(
+        PLAN_BUILDER_VM.store[plan.inputs[name].id_at_location].data
+        for name in kwarg_names
+    )
+
+    # Remove wrappers, th.jit.trace doesn't like them
+    args = tuple(list(arg) if isinstance(arg, SyList) else arg for arg in args)
```
tudorcebere, line 93: I think you can check for SyPrimitive and then use .downcast() on it? note sure.

src/syft/core/plan/translation/torchscript/plan.py
```diff
@@ -0,0 +1,88 @@
...
+        As a part of serialization and deserialization, we need the ability to
+        lookup the protobuf object type directly from the object type. This
+        static method allows us to do this.
+
+        Importantly, this method is also used to create the reverse lookup ability within
+        the metaclass of Serializable. In the metaclass, it calls this method and then
+        it takes whatever type is returned from this method and adds an attribute to it
+        with the type of this class attached to it. See the MetaSerializable class for details.
+
+        :return: the type of protobuf object which corresponds to this class.
```
jmaunon, line 38: Please, Use Google convention

jmaunon (commented):
@vvmnnnkv , Please update docstring to Google convention (see https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

src/syft/core/plan/translation/torchscript/plan_translate.py
koenvanderveen: Oh wow, it took me a few minutes but now I get it. I would document this extremely clearly, not just in the init but also in the forward method, because it makes sense once you get it but I think it could potentially confuse many people.

src/syft/core/plan/translation/torchscript/plan_translate.py
madhavajay: Yeah it would be nice to have Unions of these be auto generated for these kinds of checks.

cereallarceny (approved):

src/syft/core/plan/translation/torchscript/plan_translate.py
juharris: `OrderedDict` is a `dict` so you don't really need it in the list:
```python
from collections import OrderedDict
isinstance(OrderedDict(), dict)
Out[8]: True
```

`SyOrderedDict` is an `OrderedDict` so it's not really needed either unless you think that might changed in the future but I assume it will always at least be a `dict`.

Looking around a bit online, it looks like you don't really need `Dict` in there either: https://stackoverflow.com/questions/62547848/should-isinstance-check-against-typing-or-collections-abc

so `isinstance(arg, dict)` should be sufficient.

PR #5376, opened 2021-03-26
API of user-defined model & Plan

xutongye:
## Description
This PR is directly based on #5342 . 

### How does the API look like?
```python
class MyModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.fc1 = torch.nn.Linear(4,2)
        self.fc2 = torch.nn.Linear(2,1)

    def forward(self, x=torch.rand(4), torch=torch): # << The only difference is here.
        x = self.fc1(x)
        x = torch.relu(x)
        x = self.fc2(x)
        return x

@make_plan
def train(dl=dl, model=MyModel()):
    optimizer = ROOT_CLIENT.torch.optim.SGD(model.parameters(), lr=1e-1, momentum=0)
    criterion = ROOT_CLIENTtorch.nn.CrossEntropyLoss()
    
    for x, y in dl:
        out = model(x=x) # << use keyword args, not position args
        loss = criterion(out, y)
        loss.backward()
        optimizer.step()
        
    return model

local_model = MyModel()
remote_model = local_model.send(alice_client)
train_ptr = train.send(alice_client)
m_ptr = train_ptr(dl=remote_dl, model=remote_model)
m_get = m_ptr.get()
assert isinstance(m_get, MyModel) # the type of model got back is the class you defined
assert type(m_get.forward).__name__=="method" # the forward of m_get is a method, not a plan.
```
So, as you can see, the way we define our model is almost the same as in pytorch, except two requirements in forward method:
1. It must have a default value for `x`. This is necessary for us to create plan from the forward method.
2. It must have an argument `torch`, and also give default value to it. This is necessary because we want to replace `torch` by `remote_torch` in plan. With `torch` as an argument, we can replace it simply by passing in `remote_torch` to the `torch` argument.

#### 1. How to call child modules of a remote model, i.e., `model_pointer.fc1(x)`?
An user defined model class is not pre-defined in our source code, which means we can't add it to AST. So, the `model_pointer` doesn't have attribute named `fc1` when it's created. We need to attach an attribute `fc1` to  `model_pointer`. 
To do that, one way is to send `local_model.fc1` and get a pointer `fc1_pointer`, and then do `remote_model.fc1=fc1_pointer`. But, this means we send `local_model.fc1` two times, one time when we send `local_model`, second time when we send `local_model.fc1`. Doing so, means `remote_model.fc1_pointer` is pointing to a different remote `nn.Linear` object, i.e., not the child module `fc1` of the remote model. Then, when training, the optimizer is updating the parameters of which `fc1`? So, this is a problem. Maybe we can find a solution to this problem. But let's find a better way, where there is no such a problem.
My solution is, we don't need to send `local_model.fc1` to get `fc1_pointer`, we directly copy the pointer `remote_model` to get `fc1_pointer`, and give an attribute `attribute_name="fc1"` to `fc1_pointer`. With this new attribute `attribute_name="fc1"`, we can directly access the child module `"fc1"` of the remote model. So, `fc1_pointer` is pointing to the child module of the remote model, not a different `Linear` object.
#### 2. How is the forward plan created?
The forward method of our model is sent as a plan object. So, we create a plan from the forward method. So, we need the forward method to have default value for it's `x` argument. And we need the forward method to have a `torch` argument, which is replaced by `remote_torch` when creating the plan object.
One problem is that when making a plan, it needs a function, not a method. But `forward` is a mothod. How we get a function? It's easy, we can call a method from the class name, `MyModel.forward(self=model, x=x, torch=torch)`, then it acts like a function.
#### 3. When is the forward plan created?
We create the forward plan inside `send` method, i.e., the forward plan is created when you call `my_model.send()`. I think this is the good time to create the plan, because this is the time the plan is needed.
#### 4. The forward plan will not be created multiply times.
It's slow to create a plan object, especially when the function is big. So, we store the plan object once it's created the first time, and re-use it when it's needed later.
```python
local_model = MyModel()
assert not hasattr(local_model, "_sy_forward_plan")
remote_model = local_model.send(alice_client) # a plan is created
assert hasattr(local_model, "_sy_forward_plan")
remote_model = local_model.send(alice_client) # _sy_forward_plan is used
```
#### 5. How to transform a plan to a function?
When a model is sent, the forward method becomes a plan on remote side. When a remote model is got back, is the forward attribute of the got back model a plan or a method? It's a mothod again.
When a model is sent, we record the class in a global dict named `CUSTOM_MODELS`, which maps a class name to a class. For example, `CUSTOM_MODELS["MyModel"]=MyModel`.
So, when we get back a remote model, we can get the class of that model from `CUSTOM_MODELS`, because we know the class name from `proto.module_type`. So, we can create exactly a `MyModel` object. And it's forward attribute is a method naturally.

## Affected Dependencies

## How has this been tested?
- examples/experimental/tongye/torch_nn_Module.ipynb
- examples/experimental/tongye/MLP Plan.ipynb

## Checklist


proto/core/pointer/pointer.proto
```diff
@@ -13,4 +13,5 @@ message Pointer {
   repeated string tags = 5;
   string description = 6;
   string object_type = 7;
+  string attribute_name = 8;
```
tudorcebere, line 16: What is attribute name? Can't we deduce it from `points_to_object_with_path`?

src/syft/lib/torch/module.py
```diff
@@ -396,3 +404,81 @@ def debug_sum_layers(self) -> None:
...
+GenerateWrapper(
+    wrapped_type=torch.nn.Module,
+    import_path="torch.nn.Module",
+    protobuf_scheme=Module_PB,
+    type_object2proto=object2proto,
+    type_proto2object=proto2object,
+)
+
+
+class ModelExecutor:
```
tudorcebere, line 462: Would you mind adding:
* a test with a very dummy model (in can have like an internal FC of [32, 2]) with random data/labels to add a test that the forward pass is not breaking?
* add a few lines of docstrings?

tudorcebere (commented):
Legendary work! Added 2 comments overall.

proto/core/pointer/pointer.proto
xutongye: As I understand it, they are different. Let's use an example, 
```python
m = nn.Sequential()
m.add_module("fc1", nn.Linear(4,2))
m.add_module("fc2", nn.Linear(2,1))
m_ptr = m.send(duet)
```
Then, for `m_ptr`, the `points_to_object_with_path` of it would be like `torch.nn.Sequential`. Then, imagine if we want a pointer pointing to `m.fc1`, we can create a Pointer same as `m_ptr` but has `attribute_name=fc1`.

src/syft/lib/torch/module.py
xutongye: sure, will do.

xutongye: sure, will do.

src/syft/lib/torch/module.py
```diff
@@ -456,3 +457,28 @@ def proto2object(proto: Module_PB) -> torch.nn.Module:
...
+    def __call__(
+        self, model: torch.nn.Module, x: torch.Tensor
+    ) -> Union[torch.Tensor, Pointer]:
+        if isinstance(model, Pointer):
+            for name in self.children_names:
+                ModelExecutor.set_pointer_attr(model, name)
+        return self.model_forward(model, x)
+
+    @staticmethod
+    def set_pointer_attr(model: torch.nn.Module, attr_name: str) -> None:
```
koenvanderveen, line 476: Could you help me understand what this is doing?

src/syft/core/plan/plan_builder.py
```diff
@@ -30,14 +30,30 @@ def build_plan_inputs(forward_func: Callable) -> Dict[str, Pointer]:
     return res
 
 
+def map_in2out(inputs: dict, outputs: tuple) -> dict:
```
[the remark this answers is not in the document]
koenvanderveen: Also check if this works if the plan output is a single object (instead of a list)

tudorcebere (approved):

PR #5384, opened 2021-03-28
fix#5365 skip rewrapping iter

### LINKED ISSUE #5365: Improving Dict Iterator creation
## Description
Currently, if you trace it, we are making 4 remote `__len__` requests when creating a Dict/OrderedDict remote iterator. Ideally, we should stick to one. The same goes for any Iterator that is being created from Dict (`.values()`, `.items()`, etc).

Snippet to reproduce:
```
import syft as sy
client = sy.VirtualMachine().get_root_client()
iter(client.syft.lib.python.collections.OrderedDict({1: 1, 2: 2}))
```

with a `print(attr_path_and_name)` in `syft/ast/klass.py` at the top of the function `run_class_method`.

vsquareg:
## Description
Fix #5365

`__iter__` gets wrapped multiple times due to 
https://github.com/OpenMined/PySyft/blob/261869e50852a24b2d76f3b44a5819050acd9eb8/src/syft/ast/klass.py#L338-L342

Fix: skip re-wrapping if already wrapped

## Checklist


src/syft/ast/klass.py
```diff
@@ -232,6 +232,11 @@ def __iter__(self: Any) -> Iterable:
 
     attr_name = "__iter__"
     iter_target = attrs[attr_name]
+
+    # skip if __iter__ has already been wrapped
+    if "__qual_name__" in dir(iter_target) and "wrap_iter" in iter_target.__qualname__:
```
gmuraru, line 237: `
qual_name = getattr(iter_target, "__qual_name__", None)
if qual_name and "wrap_iter" in qual_name:
    return
`

gmuraru (approved):
LGTM!

madhavajay:
@vsquareg is it possible to show this in a test so that we don't regress this behaviour again later?

tudorcebere:
Awesome work @vsquareg!

As @madhavajay said, introduce a PR that checks this behaviour. Could you double-check any other existing data structures for this behaviour as well?

vsquareg:
@tudorcebere All data structures with more than one iterator return type were rewrapping it (Dict, List, ...). Since `wrap_iterator(.)` was a common fn call, this should fix them all.

Working on a test to verify this behaviour, but those don't seem very elegant. Do you have anything in mind?

tudorcebere:
I'll merge this as it is so we won't stack open PRs that are good to go. Indeed, it's quite hard to demonstrate the behaviour without wrapping something internally in our codebase to trace the call. I'll think about this and post and issue if I find something good enough.

PR #5379, opened 2021-03-26
Added Serialized/Deserialized Protocol

cgoxo:
## Description
#5276 
- Added the field `protocol` for session serialization and deserialization in `session_util.py`.
- Introduce `protocol` in `session.proto`.
- Files got reformatted but I have only included `session_pb2.py` since major changes are reflected in it.

## Affected Dependencies
- Files got reformatted under `src/syft/proto/` by running the script `./scripts/build_proto.sh` but I have not included them 

## How has this been tested?
- Building the code with `python3 setup.py install` after making changes 
- Runnung the tests for SyMPC module with `pytest tests/syft/lib/sympc`

## Checklist
- [x] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [x] My changes are covered by tests

PR #5390, opened 2021-03-29
Added Protocol serialize with protobuf files for session

cgoxo:
## Description
Included the modifications of adding protocol to session and the protobuf files for the same
Extension of #5379 

## Affected Dependencies
- The modifications made to include protocol
- `proto/lib/sympc/session.proto`
- `src/syft/lib/sympc/session_util.py`
- The protobuf files of session listed below run by script `build_proto.sh`
- `src/syft/proto/lib/sympc/session_pb2.py`
- `src/syft/proto/lib/sympc/share_tensor_pb2.py`

## How has this been tested?
- Tests for sympc and fast works fine

## Checklist


src/syft/lib/sympc/session_util.py
```diff
@@ -25,12 +25,16 @@ def protobuf_session_serializer(session: Session) -> MPCSession_PB:
         (length_nr_parties + 7) // 8, byteorder="big"
     )
 
+    protocol = session.protocol.__name__
+    protocol_serialized = str.encode(protocol)
```
gmuraru, line 29: Q: @tudorcebere do we need to encode the string or we can leave it simple like `protocol`?

proto/lib/python/slice.proto
```diff
@@ -4,11 +4,11 @@ package syft.lib.python;
 import "proto/core/common/common_object.proto";
 
 message Slice {
-  optional int64 start = 1;
```
gmuraru, line 7: Q: If we add `optional` (like it was before), doesn't work?

I have seen there was a PR that was referring the `slice`. @tudorcebere do you have any idea if this should be removed or not?

tudorcebere: The latest protoc (which we are using) uses this, so it's not a problem.

src/syft/lib/sympc/session_util.py
tudorcebere: As long as it's a str, I think you can leave it be

proto/lib/python/slice.proto
gmuraru: Could you put back the optional field in this case?

src/syft/lib/sympc/session_util.py
gmuraru: Then do not encode it

proto/lib/python/slice.proto
madhavajay: I was testing out optional but it doesnt actually give us what i wanted so i think in the end this might be "optional" lol, aka not needed, because for ints the default is 0 and for bool the default is False, so we only use fields that have their bools set, there are tests so if the tests pass this should be working.

gmuraru (approved):
LGTM! Great job!

PR #5406, opened 2021-03-30
Add simple gc strategies

gmuraru:
## Description
Add a strategy like design pattern to the GC logic.
Currently, it implements two strategies:
 - `simple` - the one that we had before when one pointer goes out of scope a message would be sent to the DO to delete the object.
 - `batched` - send a message when a number of messages have been collected.

TODO:
 - add `benchmarking`

## Affected Dependencies
- GC logic that was used before

## How has this been tested?
- Added tests
- TBD: Add benchmark.

## Checklist


src/syft/core/garbage_collection/garbage_collection.py
```diff
@@ -0,0 +1,23 @@
...
+from ..pointer.pointer import Pointer
+from .gc_strategy import GCStrategy
+
+
+class GarbageCollection:
+    def __init__(self, gc_strategy: GCStrategy) -> None:
+        self._gc_strategy = gc_strategy
+
+    def apply(self, pointer: Pointer) -> None:
+        """ Apply the GCStrategy on the """
```
LaRiffle, line 13: typo in docstring

src/syft/core/garbage_collection/gc_batched.py
```diff
@@ -0,0 +1,99 @@
...
+            threshold_total (int): the threshold after which a message
+                would be sent to all clients to delete all the objects
+                that were recorded
+
+        Return:
+            None
+        """
+        self.client_to_obj_ids = defaultdict(list)
+        self.threshold_client = threshold_client
+        self.threshold_total = threshold_total
```
LaRiffle, line 48: Do we need both in your opinion?

gmuraru: I think we might need the ```threshold_total``` in case the DS is talking with a lot of DOs (and it might accumulate a lot of pointers? even if they don't take as much space as the real data, they might also add up) - and in case we never reach the limit for ```trehshold_client``` we might consume memory on the DS side?

Thinking also about this now...I think I would remove this, this is premature optimization and we should add code only in case we need it (for the moment there is no indication that this might happen).

tudorcebere:
Looks pretty solid!

Before merging/jumping into doing a bit more complicated stuff, would you mind benchmarking this? Like, create 2-3 GC-intensive functions and then profile a client (would love to see even a duet test) with our new GC heuristics, including our current one.

src/syft/core/garbage_collection/gc_batched.py
```diff
@@ -0,0 +1,82 @@
...
+
+# syft relative
+from ..common.uid import UID
+from ..node.common.action.garbage_collect_batched_action import (
+    GarbageCollectBatchedAction,
+)
+from ..pointer.pointer import Pointer
+from .gc_strategy import GCStrategy
+
+if TYPE_CHECKING:
```
gmuraru, line 19: Q: Not sure how to solve the circular dependency more nicely :\

madhavajay (approved):
Small fix but looks great, I can't wait to see how some of these strategies work with some test cases and this should give us more flexibility to try different GC approaches.

tudorcebere (changes requested):
Would love to see the benchmarks and a proper way of picking GC strategies before merging (no hurry)

gmuraru:
@tudorcebere could you take a look to see if this pattern for gc strategies seems better,

tests/syft/core/pointer/garbage_collection/gc_strategies_test.py
```diff
@@ -0,0 +1,91 @@
...
+    assert len(node.store) == 9
+
+    x.send(client, pointable=False)
+
+    assert len(node.store) == 0
+
+
+def test_gc_batched_strategy_gc_constructor(node: sy.VirtualMachine) -> None:
+    client = node.get_client()
+    client.gc = GarbageCollection("gcbatched", 5)
```
gmuraru, line 45: @tudorcebere this is the way that I would think most people would use the gc strategy

tudorcebere: I like this as well, but we need really nice docs on it on how to actually use them/how to implement one. Most likely, people won't be aware of this feature, but it's nice to have the option to do it. (Maybe we want to cut network usage and never free anything, stuff like that).

gmuraru: Yep - I would do that in the ```__init__.py``` file from ```garbage_collection``` dir

gmuraru:
@tudorcebere could you recheck this PR?
The benchmark should be added in another PR - only to keep each PR at a decent level such that people would review it!

tudorcebere (approved):

Issue #5359, opened 2021-03-24
Plan Source Code


madhavajay:
## Description
Add additional string representation of the code of traced plan function to be sent with plan for preview:

```python
@make_plan
def my_plan(data, arg2):
	y = data + data
	z = y * y + arg2

    output = {}
    output['z'] = z
    output['y'] = y
	return output

print(my_plan.code):
"""
@make_plan
def my_plan(data, arg2):
	y = data + data
	z = y * y + arg2

    output = {}
    output['z'] = z
    output['y'] = y
	return output
"""
```

This will allow the DO to look at the code which was used to generate the plan and eye ball what its doing.
Later we will make this more secure by compiling on the DO side.

koenvanderveen:
When we do:
`y = data+data`
Do we store the name "y" in the `StoreObjectAction`? Probably not right? Do we want (and is it feasible) to do this? Otherwise we would get something like:
```
var1 = data+data
var2 = var1*var1 + arg2
```

Also, I dont think we can infer whether multiple operations were chained on one line, so I think we cannot infer whether it was

`z = y * y + arg2`

or

```
z1 = y * y
z = z1 + arg2
```

it would make a bit of difference though, because if we don't have these things we might choose to improve the representation for individual actions, and just make the plan add some boilerplate (@makeplan def ...), and replace e.g. "UID123" -> "var1"

koenvanderveen:
or do we just use something like `inspect.getsource(func)`, that would be pretty easy, but not very secure if we rely on this representation for approving execution

madhavajay:
@koenvanderveen I personally was thinking it would be cool to re-construct the code from our actions and assign variables, and perhaps we might look at some kind of Plan decompilation step later but I think the goal right now is to do `inspect.getsource` so that its verbatim the same code you used to create it. If it worked on the DS side it should work on the DO side, so later we can optionally do the compilation on the DO side to provide some level of confidence in the Plan being untampered.

Issue #5292, opened 2021-03-11
Make Plans Serialization Deterministic


madhavajay:
## Description
When tracing plans some of the metadata of things like Node, Address, Pointer UID are included meaning that each time a Plan is created of the same code the output is different.

It would be good to have a single representation of a Plan no matter when it is created.
We can probably replace all of the Pointers with special PlanPointer types which are the bare minimum needed to stitch the logic together and then during runtime when replacing them with the correct ones generate the UIDs necessary for placement of intermediate values.

There is a test marked xfail to fix for this.

## Definition of Done
The xfail test passes, and when a Plan is created with the same code it always has the same binary representation which can then be hashed and relied on to have not changed.

PR #5413, opened 2021-03-31
[WIP] Plans stripped reconstruction

### LINKED ISSUE #5412: Plan stripped re-construction
## Description
Currently, when a plan is created, it contains a lot of metadata like Pointer UIDs, VM addresses, etc. So, if the same plan is created twice, its representation would be completely different.

Solution: to re-construct the plan without any metadata.

## Are you interested in working on this improvement yourself?
- Yes, I am.

## Additional Context
https://github.com/OpenMined/PySyft/issues/5359#issuecomment-807726213

vsquareg:
## Description
Fix #5412

Currently, when a plan is created, it contains a lot of metadata like Pointer UIDs, VM addresses, etc. So, if the same plan is created twice, its representation would be completely different.

This PR attempts to construct a stripped Plan.

**Why?**
- Allows to build a more user-readable source code (https://github.com/OpenMined/PySyft/issues/5359#issuecomment-807726213)
- Make serialization and deserialization deterministic (#5292) i.e. the representation of a plan is solely dependent on the actions that happen inside it and types of pointers it uses.

## Progress
- [ ] Implement all `Action` types [1/10]
- [x] `strip()`
- [ ] Re-construct from `serialized(plan)`


vsquareg:
Would really appreciate some thoughts and discussions on this. Regarding whether or not this is even required, and if there's a better approach for implementation.

koenvanderveen:
Hey, about this:

> Allows to build a more user-readable source code (#5359 (comment))

Can you explain me how you would like to improve that with the current (or future?) changes?

vsquareg:
> Hey, about this:
> 
> > Allows to build a more user-readable source code (#5359 (comment))
> 
> Can you explain me how you would like to improve that with the current (or future?) changes?

I guess my point was ambiguous, I meant the string representation of the code (say to be sent to the DO for review).
```
@make_plan
def operations(xs = th.rand(5), ys = th.rand(5)):

    ls = xs + ys
    ms = xs * ys

    return ls + ms
```
After stripping the Plan, the above is essentially represented like this:
```
input: xs: inp1, ys: inp2
actions:
- add(inp1, inp2) store in inter1
- mul(inp1, inp2) store in inter2
- add(inter1, inter2) store in inter3
output: inter3
 ```
Although, I suppose it can be done directly from the current code without stripping too.

File: scripts/build_proto.sh [hunk 1]
```diff
@@ -17,11 +17,11 @@ find ${PROTO_IN} -name "*.proto" -print0 | xargs -0 protoc --python_out=${PYTHON
 if [ "$(uname)" == "Darwin" ]; then
     echo "Darwin"
     # note the '' for empty file on MacOS
-    find src/syft/proto -name "*_pb2.py" -print0 | xargs -0 sed -i '' 's/from \(proto.*\) import /from syft.\1 import /g'
```
madhavajay, line 20: Any reason for looking for this files outside of the /proto folder?

vsquareg: No, did not intend to push this change.

src/syft/core/plan/plan_run_class_method_action.py
```diff
@@ -0,0 +1,119 @@
...
+from ...proto.core.node.common.action.plan_run_class_method_pb2 import PlanRunClassMethodAction as PlanRunClassMethodAction_PB
+from ..common.serde.deserialize import _deserialize
+from ..common.serde.serializable import bind_protobuf
+# from ..node.common.action.run_class_method_action import RunClassMethodAction
+from .plan_pointer import PlanPointer
+
+
+@bind_protobuf
+class PlanRunClassMethodAction(Serializable):
+    """
```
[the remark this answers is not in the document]
vsquareg: I thought about it earlier, but to me, it seems that `Plans` is dependent on (and uses) `Actions`, and not the other way round.
Making `RunClassMethodAction` inherit from `PlanRunClassMethodAction`, would somehow inherently imply that `Actions` are dependent on `Plans`.

Do you think I should add the inheritance?

madhavajay:
Closing because this is stale. We will revisit this when we understand how we want to handle Plans in the `0.6.0` branch.

PR #5455, opened 2021-04-09
Forward plan (SyModule)

koenvanderveen:
This PR is an attempt to solve the following challenges around SyModule
- Creating 1 model as source of truth, not a local and a remote model
- The ability to compose multiple SyModules into one bigger model
- A convenient API for defining complex models, load pre-trained weights, and then fine tune the pretrained model using SyModule
- The ability to pass a SyModule as a default argument to a plan 

This PR is based on earlier work by @xutongye around serialization of nn.module and introduces:

1) SyModule
- - the forward method is automatically converted into a plan **after initialization of an object**, by using the state in the SyModule in the Plan.
- - SyModule can contain nn.Parameters, nn.Modules from the ast, and nested SyModules
- - The most tricky part of this PR is the "recompile" step. We currently recompile (relinking the obj state to the plan) the SyModule on **deserialization**, as the state may have changed compared to when the forward `Plan` was constructed. We might want to handle this in a different place, or store a `needs_recompile` attribute and check before execution. The implementation details are in the lowest part of the `proto2object` for SyModule
2) SySequential, inheriting from SyModule which allows to chain multiple SyModules
3) Nested plans, in order to wrap a SyModule.forward call in a plan


examples/experimental/koen/CNN plan.ipynb
```diff
@@ -0,0 +1,1058 @@
+{
```
iamtrask, line 1: <p>would be really cool if this dummy data could be automatically generated based on the input argument variables</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenMined/PySyft/pull/5455/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

iamtrask, line 1: <p>train_pointer is accepting actual variables as input instead of pointers to variables already on the remote machine. This makes me think that the plan is actually being executed locally on pointers instead of the actions actually being sent to the remote machien to be executed there. This would mean that there's still a tremendous amount of network traffic (one message per action) when running a plan as opposed to only saying "hey run this plan on these variables" being the only message actually sent over the network.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenMined/PySyft/pull/5455/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='41'/>

examples/experimental/koen/MLP Plan.ipynb
```diff
@@ -118,6 +118,26 @@
   {
```
iamtrask, line 118: <p>Calling model.send() in the plan feels strange to me. If we're sending the plan to a remote worker and running it there then why would the plan (which is already on the remote machine) send data somewhere else?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenMined/PySyft/pull/5455/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

examples/experimental/koen/Sy Resnet.ipynb
```diff
@@ -0,0 +1,530 @@
+{
```
iamtrask, line 1: <p>glad to se this is working without model.send() although i'm not sure where the model or its parameters are being sent to the remote machine. where is this happening?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OpenMined/PySyft/pull/5455/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

koenvanderveen: <p>both, because model is an argument to the plan it is sent to the plan vm automatically. model is a symodule and therefore uses the object2proto in lib/torch/module line 419</p>

examples/experimental/koen/MLP Plan.ipynb
koenvanderveen: <p>This notebook was experimental and is outdated, please check the Sy Resnet nb</p>

examples/experimental/koen/CNN plan.ipynb
koenvanderveen: <p>Yes, the plan pointer does accept variables (instead of pointers), however I think it would also accept pointers. But it does not mean that those variables are not sent and executed remotely (I am not sure how to interpret "local", is a VM also local?). When we execute the plan pointer by passing variables we don't need a lot of traffic, because we only serialize the arguments, not the actions in the plan. <strong>However, </strong>we currently use a hack for local execution of plans, which actually involves sending them to an ephemeral machine. We should get rid of that hack at some point.</p><p> </p><p>Also, this notebook is outdated </p>

koenvanderveen: <p>Agreed, I do think this would be a nice stand alone ticket so we might leave it our for now?</p>

examples/experimental/koen/Sy Resnet.ipynb
iamtrask: <p>ahhhhh!</p>

iamtrask (approved):
Reviewing the final API (with some notes) - I think this is a very exciting pull request which has made the right end-user API decisions. Well done @koenvanderveen

examples/experimental/koen/CNN plan.ipynb
koenvanderveen: I think a requirement for this is that our API for calling the forward plans exactly matches the api for calling the torch.nn.Module.forward (I think this is a good idea anyway), right now there is a difference:

SyModule.forward
```
out = self.layer(x=x)[0]
```
torch.nn.Module.forward
```
out = self.layer(x)
```

koenvanderveen:
a few thoughts:
- Do we we want to get rid of the old Sy.Module?
- It would be really cool to have a `sy_module = my_torch_module.to_symodule()` API
- as mentioned by Andrew it would be great if we could infer `input_size`
- We should refactor plan.execute_locally and make it such that it does need to serialize and deserialize the plan, as execution of SyModule is very slow for larger models right now

koenvanderveen:
@madhavajay do you need any extra features in this PR? Imo this is ready to be merged, and we can make issues for additional features (auto inferring shapes, performance, local execution, `model.to_symodule()`, etc.)

madhavajay (approved):
Awesome work as usual. I have put the resnet into a test which is disabled and we can use to do performance improvements. Its currently 60 seconds just to build the resnet and training plans so our target is to get that down so we can run this in CI.

I have moved the discussed improvements to a new ticket:
https://github.com/OpenMined/PySyft/issues/5541

PR #5463, opened 2021-04-12
[WIP] Allowlist test with cuda

uid42:
## Description
1.All parameter combinations were tested on the CPU and GPU separately and marked "skip_no_cuda" if no GPU device was available.
2.A CUDA (true/false) entry is added to the not_available rule to indicate whether the parameter is not available on the CPU or CUDA.
3.Exceptions from syft are treated separately and added with "reason:not_supported_syft" to the entry not_available.
4.Fixed some parameter errors in some of the allowlist_test.json functions, such as "inverse" with tensor_cube as the parameter.
5.Add rules to generate sparse tensor and quantized tensor.
6.If an inappropriate _arg causes the error, the not_available entry takes a "reason:bad_input."
7.All entries in the allowlist_test.json that do not equal the current torch version(lte_version=gte_version=torch version) are temporarily removed and regenerated except for untested, added_feature, and deprecated rules.

## Checklist
- [*] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [*] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [*] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [*] My changes are covered by tests


tests/syft/lib/allowlist_test.py
```diff
@@ -59,7 +59,7 @@
 #
 # We then check to see if a explicit configuration is set for the attribute path in
 # allowlist_test.json and if so we use the configuration from there.
-# To keep the custom definitions DRY we use a concept called "profiles" which are just
+# io keep the custom definitions DRY we use a concept called "profiles" which are just
```
madhavajay, line 62: Typo?

uid42: typo...

madhavajay:
@uid42 Sorry for the wait on reviewing this. I booted a K80 GPU to test this and im a little confused.
On my CPU only Laptop this fails to run because the new rules seem to cause some confusion.
On the K80 the same happens. I am running the `autofix_allowlist_test.py` on the K80 now, but it seems to be failing with nearly all the ops. I was under the impression it should be the opposite and that the GPU system should be capable of practically all ops?

Were you able to run the `pytest -m torch -n auto` tests on both CPU and GPU systems with this PR?

uid42:
sorry，so long to see this. can you post some output information here?

tudorcebere:
Closing this due to inactivity. We will make GPU testing a bit different soon.

PR #5474, opened 2021-04-14
[WIP] torchvision mean, std

dnabanita7:
## Description
Added all accepted data types in ``Normalize.mean``  and ``Normalize.std`` which are accepted by ``as_tensor()``. issue #5031 

## How has this been tested?
- ``pre-commit run --all-files``

## Checklist

### DIFF
--- a/src/syft/lib/torchvision/allowlist.py
+++ b/src/syft/lib/torchvision/allowlist.py
@@ -18,9 +18,19 @@ allowlist["torchvision.transforms.Normalize"] = "torchvision.transforms.Normaliz
 # TODO: Normalize properties only exists on the object not on the class?
 # allowlist["torchvision.transforms.Normalize.inplace"] = "syft.lib.python.Bool"
 # TODO: mean and std are actually tuples
-# allowlist["torchvision.transforms.Normalize.mean"] = "syft.lib.python.List"
-# allowlist["torchvision.transforms.Normalize.std"] = "syft.lib.python.List"
+allowlist["torchvision.transforms.Normalize.mean"] = "syft.lib.python.Int"
+allowlist["torchvision.transforms.Normalize.mean"] = "syft.lib.python.Float"
+allowlist["torchvision.transforms.Normalize.mean"] = "syft.lib.python.Tuple"
+allowlist["torchvision.transforms.Normalize.mean"] = "syft.lib.python.List"
+allowlist["torchvision.transforms.Normalize.mean"] = "syft.lib.tensor.IntegerTensor"
+allowlist["torchvision.transforms.Normalize.mean"] = "syft.lib.numpy.ndarray"
 
+allowlist["torchvision.transforms.Normalize.std"] = "syft.lib.python.Int"
+allowlist["torchvision.transforms.Normalize.std"] = "syft.lib.python.Float"
+allowlist["torchvision.transforms.Normalize.std"] = "syft.lib.python.Tuple"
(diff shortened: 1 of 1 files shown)


src/syft/lib/torchvision/allowlist.py
```diff
@@ -18,9 +18,19 @@
...
+allowlist["torchvision.transforms.Normalize.mean"] = "syft.lib.python.List"
+allowlist["torchvision.transforms.Normalize.mean"] = "syft.lib.tensor.IntegerTensor"
+allowlist["torchvision.transforms.Normalize.mean"] = "syft.lib.numpy.ndarray"
+
+allowlist["torchvision.transforms.Normalize.std"] = "syft.lib.python.Int"
+allowlist["torchvision.transforms.Normalize.std"] = "syft.lib.python.Float"
+allowlist["torchvision.transforms.Normalize.std"] = "syft.lib.python.Tuple"
+allowlist["torchvision.transforms.Normalize.std"] = "syft.lib.python.List"
+allowlist["torchvision.transforms.Normalize.std"] = "syft.lib.tensor.IntegerTensor"
+allowlist["torchvision.transforms.Normalize.std"] = "syft.lib.numpy.ndarray"
```
Param-29, line 33: Shouldn't these be UnionGenerator? example:
```
allowlist["torch.Tensor.max"] = UnionGenerator[
    "syft.lib.python.Bool",
    "syft.lib.python.Float",
    "syft.lib.python.Int",
    "torch.return_types.max",
]
```

dnabanita7: Is there any specific need for this? The error is 
`` mean is not an attribute of Normalize``

dnabanita7: also mean, std are not functions. they are attributes of normalize function

dnabanita7: allowlist return type is set to a single one.

dnabanita7:
The error still is ``Attribute error: type object Normalize has no attribute mean``

Param-29:
I tired reading up to get to understand more about what you are trying to solve, please check previous PRs which tried to solve the issue
Based on my understanding issue is, mean and std are not attributes of the class, rather they are attributes of objects of that class
```
>>> cls_ = torchvision.transforms.Normalize
>>> type(cls_)
<class 'type'>
>>> x = torchvision.transforms.Normalize((0.485, 0.456, 0.406), (0.229, 0.224, 0.225))
>>> type(x)
<class 'torchvision.transforms.transforms.Normalize'>
>>> def Diff(li1, li2):
    return (list(list(set(li1)-set(li2)) + list(set(li2)-set(li1))))
... 
>>> Diff(dir(cls_),dir(x))
['std', 'mean', '_non_persistent_buffers_set', 'training', '_load_state_dict_pre_hooks', '_forward_pre_hooks', '_backward_hooks', '_parameters', '_forward_hooks', '_state_dict_hooks', '_modules', '_buffers', 'inplace']
```
This is why directly adding it to the allowlist wouldn't work!!

Although I'd love to checkout how we could make changes so that these attributes can also be added in ast.

tudorcebere:
Hi @dnabanita7, sadly, the problem is quite deep in our AST architecture right now. Namely, we can't have an object reference in our AST to support this kind of ops. We are working on this, but it's blocked right now. I'll close this for now and we can open it back when we have obj references in the AST.

PR #5477, opened 2021-04-15
Fix tutorials SyMPC

gmuraru:
## Description
Fix the SyMPC VMs and Duet tutorials.

At one point we need to make a request to our own duet store -- at that point, we check if the node we respond is the same as the "we" and in that case, we made a request to our own store and we can give ```Request.Accepted```.

The problem is because requesting the ```len``` of an object triggers a ```request``` and then a ```get```.
The same behavior can be seen using:
```
ptr = duet.python.List([1,2,3])
ptr.get() ## This works because we do not request, we simply get the object


BUT, doing this:

# This won't work because we would do a get(request_block=True)
# and we do not have the logic for what happens when
# node X request an object from node X
len(ptr)
```

## How has this been tested?
- Running the tutorial

## Checklist


tests/syft/core/node/node_test.py
```diff
@@ -209,20 +209,3 @@ def test_send_message_from_network_client_to_vm() -> None:
     bob_network_client.send_immediate_msg_without_reply(
         msg=sy.ReprMessage(address=bob_vm.address)
     )
-
-
-@pytest.mark.slow
-@pytest.mark.asyncio
-def test_autoapprove_requests_made_by_root_clients_5015(
```
gmuraru, line 216: It should also approve the requests made by simple clients to be consistent with the case when we create an object and simply "get" it without approval?
Or should we change the behavior for "get"?

tudorcebere (approved):

Issue #5467, opened 2021-04-13
Integrating numpy to syft


cgoxo:
## Integrating numpy to PySyft.
Currently syft supports np.ndarray. The methods, classes and modules are listed below.

#### Array Objects
- [x] Constructing arrays
- [x] Array attributes
- [x] Array methods
- [ ] Arithmetic, matrix multiplication, and comparison operations
- [ ] Special methods
#### Scalers
- [ ] Attributes
- [ ] Methods
#### Data Type Objects
- [ ] Attributes
- [ ] Methods
#### Standard Array Subclasses
- [ ] Character arrays
- [ ] Record arrays
- [ ] Standard container class
#### Masked arrays
- [ ] Methods Listed
#### Constants
- [ ] Methods listed
#### Universal Functions
- [ ] Attributes
- [ ] Methods
- [ ] Math Operation
- [ ] Trigonometric functions
- [ ] Bit-twiddling functions
- [ ] Comparison functions
- [ ] Floating functions 
#### Routines
- [ ] Array Creation Routines
- [ ] Array manipulation routines
- [ ] Binary operations
- [ ] String Operation
- [ ] Datetime Support Functions
- [ ] Business Day Functions
- [ ] Data type routines
- [ ] Creating data types
- [ ] Optionally SciPy-accelerated routines
- [ ] Mathematical functions with automatic domain
- [ ] Floating point error handling
- [ ] Functional programming
- [ ] NumPy-specific help functions
- [ ] Indexing routines
- [ ] Input and output
- [ ] Linear algebra
- [ ] Logic functions
- [ ] Matrix library
- [ ] Miscellaneous routines
- [ ] Padding Arrays
- [ ] Set routines
- [ ] Sorting, searching, and counting
- [ ] Statistics
- [ ] Window functions

madhavajay:
@cgoxo Great work so far, but can we discuss this before any more work is done. I would like to take a more programmatic approach to lib support rather than manually adding ops, tests, serde and return types.

PR #5479, opened 2021-04-18
[WIP]Adding Array objects to numpy in syft

cgoxo:
## Description
#5467 
Planning to add array objects to numpy(i.e Array attributes, Array methods, Arithmetic, matrix multiplication, and comparison operations, Special methods) with proto and tests.

## Affected Dependencies
`src/syft/lib/numpy/__init__.py`

## How has this been tested?

## Checklist


Param-29:
Hi @cgoxo, great job. 

Some of the methods you have added should not be allowed to keep the data private, so I'd suggest giving this a good thought before adding a method/class to `ast`. 

Example (dtype): 

`PySyft/src/syft/lib/torch/allowlist.py`
```
# allowlist["torch.Tensor.dtype"] = SECURITY WARNING: DO NOT ADD TO ALLOW LIST YET - talk to TRASK
```

Param-29:
Hi @cgoxo, great job. 

Some of the methods you have added should not be allowed to keep the data private, so I'd suggest giving this a good thought before adding a method/class to `ast`. 

Example (dtype): 

`PySyft/src/syft/lib/torch/allowlist.py`
```
# allowlist["torch.Tensor.dtype"] = SECURITY WARNING: DO NOT ADD TO ALLOW LIST YET - talk to TRASK
```

Hope this helps you get a better insight.

cgoxo:
Thank you for the response
So, should I make a seperate `allowlist.py` for numpy library and comment out the methods which hinders data privacy?

Param-29:
No, you could just comment the line which you do not want to add, but do keep them so that we could know which methods were commented for security reasons

cgoxo:
Hello @Param-29 I have commented the methods which hinder privacy. 
Just want to know as of now do I have to make changes under `array.py` for serialize/deserialize methods?

Param-29:
We already do support `numpy.ndarray` and all the methods you have added in are for that so I believe serialization and de-serialization are not required, If there is another data type which you want to add support for, then you need to work on serializing and de-serializing it. 

The next step would be to write unit tests to check if all the methods you have added support for run as expected on the virtual machine. Check out `PySyft/tests/syft/lib/` for more. 

Here in `xgboost` you can see how we have tested `.predict()` method. I believe you will have to do something similar, just try creating some kind of `.json` to keep the code clean (like `torchvision`)

.pre-commit-config.yaml
```diff
@@ -15,7 +15,7 @@ repos:
         exclude: ^(tests/syft/lib/allowlist_report.py)
       - id: requirements-txt-fixer
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.8.4
+    rev: 3.9.1
```
Param-29, line 18: Any specific reason for this change?

.pre-commit-config.yaml
```diff
@@ -57,7 +57,7 @@ repos:
         additional_dependencies:
           - flynt
   - repo: https://github.com/pycqa/isort
-    rev: 5.7.0
+    rev: 5.8.0
```
Param-29, line 60: Same as above

.pre-commit-config.yaml
cgoxo: I haven't made that change. This was done by @gmuraru.

gmuraru: hmm..I think you need to rebase? (if this change is on dev, it should not show up in this PR)

Param-29: Agreed, don't try and rebase manually. use git to rebase and pull branch.

src/syft/lib/numpy/__init__.py
```diff
@@ -27,10 +27,83 @@ def create_ast(client: TypeAny = None) -> Globals:
...
+        ("numpy.ndarray.ctypes", "numpy.ndarray"),
+        ("numpy.ndarray.__array_interface__", "numpy.ndarray"),
+        ("numpy.ndarray.__array_struct__", "numpy.ndarray"),
+        # Array methods
+        ("numpy.ndarray.item", "numpy.ndarray"),
+        ("numpy.ndarray.tolist", "numpy.ndarray"),
+        ("numpy.ndarray.itemset", "numpy.ndarray"),
+        ("numpy.ndarray.tostring", "numpy.ndarray"),
+        ("numpy.ndarray.tobytes", "numpy.ndarray"),
+        ("numpy.ndarray.tofile", "numpy.ndarray"),
```
madhavajay, line 58: I think we might need to remove this if it allows you to write data to any path on the remote system.

cgoxo: Sure! On it.

cgoxo:
Hello Sir! I have a couple of doubts regarding this issue
1. Do we need to add methods for `np.tolist` , `np.astype` ?
2. The attribute `dtype` does not support while applying method `send` and it refers the error 
`AttributeError: type object 'numpy.int16' has no attribute 'send'`
3. How to make tests for non mutable methods as the pointer variable does not take the value?  
4. Some methods do have a return type that are not present under `src/lib/python` such as `np.data` , `np.__array_struct__`. Do these methods are required or need to make serde for these classes ?

Sorry for asking so many questions at once

madhavajay:
Hi @cgoxo Great work so far. Lets discuss how we can approach the addition of whole libraries in a more programmatic way, including the answers to some of your questions above.

> * Do we need to add methods for `np.tolist` , `np.astype` ?

Lets discuss a more programmatic approach to these on Slack.

> * The attribute `dtype` does not support while applying method `send` and it refers the error

`dtype` has to be supported. The Type is a type itself, which is different to the instance of said type.

>   `AttributeError: type object 'numpy.int16' has no attribute 'send'`
> * How to make tests for non mutable methods as the pointer variable does not take the value?

Do you have an example of a method that makes no changes and gives no return value?

> * Some methods do have a return type that are not present under `src/lib/python` such as `np.data` , `np.__array_struct__`. > Do these methods are required or need to make serde for these classes ?

Lets discuss.

PR #5480, opened 2021-04-19
Test to prevent regression of .grad and .data on torch.Tensor

### LINKED ISSUE #4689: Add an AttributePointer Class for Tensor.grad and Tensor.data
## Description
Create an AttributePointer class which allows access to attributes which point to other data such as torch.Tensor.grad and torch.Tensor.data.

The goal is to prevent the need for a reply message in sending an AttributePointer for use. To achieve this when a pointer is created several attribute pointers should be created in the remote store immediately so they are available for use.

```
ptr = x.send(alice_client)
# on alice_client the following is created:
# x.grad AttributePointer
# x.data AttributePointer
```

The id_at_location for these AttributePointers will be deterministic hashes based on their parent Tensor id_at_location.
```
grad_ptr = ptr.grad
grad_ptr.id_at_location == hash(ptr.id_at_location + "grad")
```

These can be recursive and should be handled the same way:
```
ptr.grad.grad.grad == hash(hash(ptr.id_at_location + "grad") + "grad")
```

These AttributePointers should be GCed if the parent is GCed.

## Definition of Done
A test that demonstrates the usage of .grad and .data without requiring additional reply messages from the DO side.
A test which demonstrates the child AttributePointer GCing works.

tudorcebere:
## Description
Test to prevent regression on .grad and .data.

Closes #4689 

## Checklist


madhavajay (approved):
👍

PR #5499, opened 2021-04-22
[WIP] Fix remote binary operators 

### LINKED ISSUE #4774: Fix remote binary operators
## Description
Currently, when it comes to binary operators, python executes a finite state machine internally, let's look for example at:

```
a + b
```
0. Check if a and b are the same type, if yes, call `__add__` on a.
1. Check if a has `__add__`, if not, jump to 4.
2. If it has it, call `a.__add__(b)`, if it returns NotImplemented continue, if not return the given value.
4. Check if b has `__radd__` <- the reversed operator if it does not have it, raise an error.
5. if it has it, call "b._radd__(a)". if it returns NotImplemented, raise an error, if not, return the value.

Currently, when doing remote execution, we don't have a mechanism to forward NotImplemented back. This means that all the reversed operators `__radd__`, `__rsub__`, etc. are never actually called. This can become harmful when doing stuff like:

`Int(5) + Float(5.5)`

`Int` will not know how to add himself to `Float`, `Float` knows how to add `Int` to himself, but due to the fact that we are not forwarding `NotImplemented` back, the reversed operator is never called and the operations result in a `NotImplemented`.

vsquareg:
## Description
Fix #4774 

## To Discuss
Possible approaches to fix this:
- [currently implemented] After `RunClassMethodAction`, validate the pointer for binary ops (i.e. set pointer type to what remote client has it as, instead of referencing the AST tree).
This results in two messages being sent to the client.

- Add another `RunClassMethodAction` as a `with_reply` message such that it returns the actual result type after executing the action. Use this for binary ops.
This would reduce the pointer validation message being sent again, but might have other implications I'm missing.

## How has this been tested?
Tests pending.


tudorcebere:
Hi @vsquareg!

Really nice work! As far as I've understood, you are solving the remote type to see if it's a not implemented after executing. This is nice, but it basically requires you to have permission on doing that all the time.

We can diverge from how python behaves and maybe call the add operation for example if `__add__` is called and solve the state machine that python uses for addition. Ping me for a small call in slack to figure it out, just to be sure we are on the same page.

madhavajay:
@vsquareg any chance we can get this updated with the latest changes in `dev`.

PR #5502, opened 2021-04-23
Revert "API of user-defined model & Plan"

tudorcebere:
Reverts OpenMined/PySyft#5376, need to sync with @koenvanderveen and @xutongye  to better understand the approach


madhavajay (approved):
👍🏽

Issue #4764, opened 2020-11-03
Test GC on In-place Operators


madhavajay:
## Description
There are scenarios where GC might have issues with inplace_ tensor operations if the original pointer is cleaned up and the returned pointer after the mutation still points to the same torch memory. We should consider the different possibilities and write tests to evaluate the behaviour.

## Definition of Done
Tests prove this is either not a problem or that it is and we can discuss a possible solution.

gmuraru:
Looked into this issue!

Run the following code:
```
alice = sy.VirtualMachine(name="alice")
alice_client = alice.get_client()
    
x_ptr = x.send(alice_client)
print(alice.store)
   
x_ptr2 = x_ptr.transpose(0, 1)
print(alice.store)
```
How the store looks:
1.
```
{<UID:0ba289de-72a8-415f-ad5f-fa322e104367>: <Storable:tensor([[1, 2, 3],    [4, 5, 6]])>}
``` <--- makes sense to be only one object since we sent one tensor
2.
```
{<UID:0ba289de-72a8-415f-ad5f-fa322e104367>: <Storable:tensor([[1, 2, 3],    [4, 5, 6]])>, <UID:0d01b581-fcde-46ff-b0de-e72cc161d6d6>: <Storable:tensor([[1, 4],    [2, 5],    [3, 6]])>}
``` <--- makes sense to be 2 objects since we called ```transponse``` on the tensor (and one is the transpose of the other)

With inplace methods/ops:
```
alice = sy.VirtualMachine(name="alice")
alice_client = alice.get_client()
        
x = th.tensor([[1,2,3], [4,5,6]])
x_ptr = x.send(alice_client)
print(alice.store)

x_ptr2 = x_ptr.transpose_(0, 1)
print(alice.store)
```
How the store looks:
1. ```{<UID:dbf20faa-d36e-446d-87cf-39b2d09368bd>: <Storable:tensor([[1, 2, 3],    [4, 5, 6]])>}```
2. ```{<UID:dbf20faa-d36e-446d-87cf-39b2d09368bd>: <Storable:tensor([[1, 4],    [2, 5],    [3, 6]])>, <UID:d9acfb0f-b215-4068-b32c-90142d68453a>: <Storable:tensor([[1, 4],    [2, 5],    [3, 6]])>}``` <-- since we used ```transpose_``` (does the inplace transpose + returns the same transposed tensor -- in our case pointer to tensor)

^^^ This might be an acceptable behavior depending on what we want.
[...]

PR #5503, opened 2021-04-23
inplace ops gc test

tudorcebere:
## Description
Solves #4764


tests/syft/core/pointer/garbage_collection/inplace_op_gc_test.py
```diff
@@ -0,0 +1,38 @@
...
+    node: sy.VirtualMachine,
+    client: sy.VirtualMachineClient,
+) -> None:
+
+    object_ptr = obj.send(client, pointable=False)
+    initial_id = object_ptr.id_at_location
+    assert len(node.store) == 1
+
+    func = getattr(object_ptr, op)
+    func(*args)  # inplace op
```
gmuraru, line 33: NITPICK: Maybe also add assert after the operation?

gmuraru (approved):
LGTM!

tests/syft/core/pointer/garbage_collection/inplace_op_gc_test.py
madhavajay, line 33: We need to check the return pointer of the Tensor op because it returns a second pointer to the same value.

madhavajay (changes requested):
LGTM just need the tensor pointer return value checked.

madhavajay (approved):
LGTM!

PR #5554, opened 2021-05-10
feat: move allowlist from syft to sympc

danielorihuela:
## Description
Move the allow list of operations over share tensors from pysyft to sympc.
Decouple list of methods from pysyft, giving the control to sympc.

This PR is related with a [SyMPC issue](https://github.com/OpenMined/SyMPC/pull/162). 

## Affected Dependencies
SyMPC

## How has this been tested?
- I executed sympc related tests of pysyft.

## Checklist
- [x] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A) No, because I cannot


gmuraru (approved):
LGTM

madhavajay:
I agree with this in principal, but we will probably still need some kind of controls on the Syft library to both block ops and override which ones are available at some point. I think it would be a good idea to talk about what we expect out of a third party library API so that we can implement that for situations like this.

PR #5600, opened 2021-05-23
Use uuid session for SyMPC session

gmuraru:
## Description
This PR should be merged with [this one](https://github.com/OpenMined/SyMPC/pull/190).
Keep track of the shares using a `session_uuid`.
When we pass a `ShareTensor` we would pass only the `session_uuid` and the session should already be on the receiver side when it is deserialized.

## Affected Dependencies
- SyMPC

## How has this been tested?
- SyMPC Tests

## Checklist


kamathhrishi (approved):
Looks good to me.

packages/syft/proto/lib/sympc/session.proto
```diff
@@ -7,12 +7,12 @@ import "proto/lib/python/dict.proto";
 
 
 message MPCSession {
-  bytes protocol = 8;
-  bytes uuid = 1;
+  string uuid = 1;
```
rasswanth-s, line 10: I think using bytes object would be optimal  for uuid.

packages/syft/proto/lib/sympc/share_tensor.proto
```diff
@@ -3,9 +3,10 @@ syntax = "proto3";
...
 
 import "proto/lib/torch/tensor.proto";
-import "proto/lib/sympc/session.proto";
+import "proto/lib/python/dict.proto";
 
 message ShareTensor {
   syft.lib.torch.TensorProto tensor = 1;
-  syft.lib.sympc.MPCSession session = 2;
+  syft.lib.python.Dict config = 2;
+  string session_uuid = 3;
```
rasswanth-s, line 11: Same here on session_uuid

packages/syft/proto/lib/sympc/session.proto
gmuraru: Ahh...I remember why I used string. it is because we won't have to do a lot of conversions at the moment and it might be easier to debug until we reach a more "production" like code.

One example is that we keep rank -> session_uuid (in SyMPC) and that session_uuid is a string because it is easier for devs to "debug" it.

rasswanth-s (approved):
Looks good to me.

PR #5609, opened 2021-05-28
Test network utility

vsquareg:
## Description
Basic network diagnostics test utility. 

- [x] CI test
- [x] Check local signaling server
- [x] Loopback


packages/syft/src/syft/grid/duet/__init__.py
```diff
@@ -296,3 +299,66 @@ def join_duet(
     # begin_duet_client_logger(duet.node)
 
     return duet
+
+
+def test_duet_network(loopback: bool = False) -> None:
+    if not loopback:
```
madhavajay, line 305: loopback=True still uses the same signalling server it just exchanges the IDs automatically using a file, so we should still check all of these main url service checks. There is the case where a user can run their own signalling server so it might be nice to be able to pass that in as a parameter and have these checks run against it to make sure that the python process can read the url on the port you supply.

like:
```
sy.test_duet_network(network_url="http://192.168.0.2:5432")
```

madhavajay (changes requested):
Can we add a test which runs this and asserts that they all pass and then we can run that in CI and get a canary alert that our Signalling server has gone down and be sure the code coverage is running?

Also perhaps change the loopback=True branch to still check the live urls are reachable.

madhavajay:
@vsquareg great work, just need a few small tweaks before merge. 😊

packages/syft/src/syft/grid/duet/__init__.py
vsquareg: Not sure how I missed those 😞️. Fixed.

vsquareg:
> Can we add a test which runs this and asserts that they all pass and then we can run that in CI and get a canary alert that our Signalling server has gone down and be sure the code coverage is running?
> 
> Also perhaps change the loopback=True branch to still check the live urls are reachable.

Should I add an explicit test in the CI for this? Or just let it run as it would normally with the blanket pytest action?

tudorcebere:
> > Can we add a test which runs this and asserts that they all pass and then we can run that in CI and get a canary alert that our Signalling server has gone down and be sure the code coverage is running?
> > Also perhaps change the loopback=True branch to still check the live urls are reachable.
> 
> Should I add an explicit test in the CI for this? Or just let it run as it would normally with the blanket pytest action?

Normal sounds good!

madhavajay (approved):
Excellent!

PR #5636, opened 2021-06-03
result permissions v1

koenvanderveen:
## Description
Make sure that a data owner can say: if you run this plan on this data, you can download the result without requesting it

## How has this been tested?
notebook, will add tests later


koenvanderveen:
Tests, linting, cleaning up are todo. Would love early feedback

packages/syft/src/syft/core/node/common/action/run_class_method_action.py
```diff
@@ -269,6 +274,11 @@ def _object2proto(self) -> RunClassMethodAction_PB:
             msg_id=serialize(self.id),
         )
 
+    def has_result_read_permission(self, permissions, method_name, verify_key):
+        _args = [a.id_at_location for a in self.args]
+        _kwargs = {k: v.id_at_location for k, v in self.kwargs.items()}
+        return any([permission.matches(self._self.id_at_location, verify_key, method_name, _args, _kwargs) for permission in permissions])
```
gmuraru, line 280: Q: Shouldn't we check that all the attributes have this permission?

I am thinking at the following scenario:
```
a_ptr = a.send(duet)
# b_ptr is privata data owner by the DataOwner
(a_ptr + b_ptr) # This has read permission
```

^ we could have something like this from a plan perspective, right?

koenvanderveen: In this case, the result permissions would mean:

"if we run __add__ on a_ptr, with as input b_ptr, the result can be downloaded", so it makes sense that `(a_ptr + b_ptr)` is downloadable, would you agree?

gmuraru: I agree, only if both `a_ptr` and `b_ptr` are downloadable.
If `a_ptr` was not downloadable and the result was (and also `b_ptr` was downloadable) this might leak info about `a` - the other party could do `result - b_ptr` and would find `a`

koenvanderveen: Its the other way around, `a_ptr` is downloadable, `b_ptr` is not. However, maybe this is not a good example, because this only works for plans. And for plans, the "result-b=a" logic may or may not hold, the point is, you should think about that when giving permission.

I think the example we should look at is

```
plan_ptr=plan.send(duet) #downloadable
# x_ptr is private
train_ptr.set_result_permissions("__call__", {"x": x_ptr}, ds.verify_key) # set by data owner
y_ptr = plan_ptr(x_ptr) # called by ds
y = y_ptr.get() # called by ds, and no need to request download permission
``` 

It may be that y_ptr reveals some information about x_ptr, but thats beyond the point (the plan could be a machine learning model, or some statistic, it could be anything). However, this is no different from an ad hoc download requests.

If the plan is really doing x_ptr +3, or something else that reveals too much information about the inputds, the data owner should consider denying that request.

madhavajay:
@koenvanderveen is this still relevant or should we close and tackle as part of the future Plan Usage Policy work?

koenvanderveen:
I think it might haven gotten a bit out of date (assumption).  I probably I will have time this month to have a look again

PR #5657, opened 2021-06-14
[WIP] CI datasets caching

### LINKED ISSUE #5619: Add MNIST to Cache in CI
## Description
MNIST is only 10mb but its being downloaded on every CI instance so we should cache.

## Definition of Done
A new cache entry is added to CI for datasets and the folder where they are downloaded is configured and cached.
We could add some code output the cache directory using the `Path.home()` lib so that its consistent with `get_root_data_path`.

vsquareg:
## Description
Fix #5619


madhavajay:
Thanks, this was a great idea, but I have started putting the datasets in a github repo so they are easy to manage but quick to install on CI. I think the cache action has a limit to how much we can cache and we sometimes go over the limit so perhaps we should keep datasets separate. I will close this for now and we can revisit it later if we need it.

PR #5677, opened 2021-06-19
[Experimental] [WIP] Arrow flight

vsquareg:
## Description
_Experimental_

Arrow flight RPC


packages/syft/src/syft/grid/duet/__init__.py
```diff
@@ -230,6 +233,13 @@ def launch_duet(
         begin_duet_logger(my_domain=my_domain)
     info(print=True)
 
+    #TODO (flight): handle outside localhost (hard)
```
tudorcebere, line 236: Q: What does this mean?

packages/syft/src/syft/grid/duet/__init__.py
```diff
@@ -295,4 +305,21 @@ def join_duet(
     info("♫♫♫ > " + bcolors.OKGREEN + "CONNECTED!" + bcolors.ENDC, print=True)
     # begin_duet_client_logger(duet.node)
 
+    #TODO (flight): handle outside localhost (hard)
+    #TODO (flight): get available ports
+
+    flight_args = {
```
tudorcebere, line 311: We will need support for TLS, in Duet, WebRTC validates data through self-signed certificates I believe. It would be interesting to validate this. (Don't address in this PR, but we need to keep an eye opened on this)

packages/syft/src/syft/grid/flight/flight_server.py
```diff
@@ -0,0 +1,63 @@
...
+            ], names=[obj_id_str[3:]])
+        writer.begin(data.schema)
+        writer.write_table(data)
+
+    def add_accessible(self, obj, id_at_location):
+        self.accessible[id_at_location] = obj
+    
+    def retrieve_accessible(self, id_at_location):
+        #TODO (flight): fix this mess (use appropriate arrow types)
+        return self.accessible.get(id_at_location, None).to_pandas()[str(id_at_location.value)].to_numpy()
```
tudorcebere, line 63: Can you implement the Node/Client interfaces using Flight as a connection? It should be easier to modify/test if we implement those interfaces.

PR #5734, opened 2021-06-25
Redoing code cleanup PR

iamtrask:
## Description
The following PR has gotten too distant to conveniently debug so I'm going to re-pull in changes as we go and test.

https://github.com/OpenMined/PySyft/pull/5670

PR #5735, opened 2021-06-26
 Continue work from https://github.com/OpenMined/PySyft/pull/5734

iamtrask:
See https://github.com/OpenMined/PySyft/pull/5734


iamtrask:
@madhavajay at https://github.com/OpenMined/PySyft/pull/5735/commits/46fe370117c41ba495f520ee54cfbd122eaf8111 I couldn't figure out how to import sqlalchemy into tox correctly. Please help.

iamtrask:
@madhavajay and @tudorcebere - check out https://github.com/OpenMined/PySyft/pull/5735/commits/34dba1fbdb76698be78fb66a657f49e02493d1e0 The idea of the recursive serde class is that it can be subclassed and make it so that we don't have to create custom protobufs for all our custom classes as long as its attributes eventually are composed of simple types we do know how to serialize

iamtrask:
In forcing PySyft to serialize all objects it checks into the store, I've discovered a large number of objects which currently aren't serializable or wherein serialization is broken. I've commented out the respective tests.

PR #5820, opened 2021-07-20
[WIP] pip install hagrid (and it's deps automagically) - Part 2

iamtrask:
Continuation of: #5777 

At the time of writing, the [hagrid tool](https://github.com/OpenMined/PySyft/tree/demo_strike_team_branch_4/packages/hagrid) can be locally installed, but it cannot be pip installed.

For this project, we want to make it possible for anyone running Linux or OSX to be able to pip install hagrid and be *ready to go* to launch PyGrid domains in all the environments that hagrid requires.

Note that for Linux, it has become apparent that not all required dependencies will necessarily be easy to include as dependencies in pip. The fallback should be some kind of hand-holding CLI tool that helps people run the right commands to get everything setup properly. Hopefully this isn't necessary. :)

[ch65]


madhavajay:
Closing as I think this was a clubhouse story place holder.

PR #5889, opened 2021-08-08
[SMPC] Add public operations and refactor some code

gmuraru:
## Description
Add a part of the remaining public operations and refactor some of the code.
Private multiplication still needs to be added - but in another PR.

## Affected Dependencies
 - SMPC
 
## How has this been tested?
 - [x] Using a SMPC Notebook
 - [x] Tests

## Checklist


packages/syft/src/syft/core/tensor/smpc/share_tensor.py
```diff
@@ -154,3 +299,12 @@ def _proto2object(proto: ShareTensor_PB) -> "ShareTensor":
...
     def get_protobuf_schema() -> GeneratedProtocolMessageType:
         return ShareTensor_PB
+
+    __add__ = add
+    __radd__ = add
+    __sub__ = sub
+    __rsub__ = rsub
+    __mul__ = mul
+    __rmul__ = mul
+    __matmul__ = matmul
```
gmuraru, line 309: Those need to be implemented (for public operations)

packages/syft/src/syft/core/tensor/smpc/share_tensor.py
```diff
@@ -118,23 +121,165 @@ def generate_przs(
...
+
+        Args:
+            y (Union[int, float, torch.Tensor, np.ndarray, "ShareTensor"]): y - self
+
+        Returns:
+            ShareTensor. Result of the operation.
+        """
+
+        ShareTensor.sanity_check(y)
+        new_self = self.mul(-1)
```
aanurraj, line 212: Q: Any specific reason we did this instead of doing `y - sub`?

gmuraru: I am not sure why that will work.
We got here because we called something like `y - x` (where x is a ShareTensor and possibly y is an integer).
Calling `y-sub` (sub == x?) means we are doing that recursively.

packages/syft/src/syft/core/node/common/action/smpc_action_message.py
```diff
@@ -64,13 +70,35 @@ def filter_actions_after_rank(
...
-            if action.ranks_to_run_action is None:
+
+            should_not_run = (
+                isinstance(action.ranks_not_to_run_action, list)
+                and rank in action.ranks_not_to_run_action
+            )
+
+            should_run = False
+            if (
+                action.ranks_to_run_action == []
```
madhavajay, line 81: Should we just check the length in case we have a different type of list or iterator which doesnt equal empty list []?

madhavajay: Also, is there a reason the default value for ranks_to and ranks_not_to are None and not empty lists? Is it correct that this should run if ranks is an empty list but not if it's None?

madhavajay (approved):
Looks good. I fixed a lint issue and ran the notebooks with 4 stacks successfully then added it to the tox integration test with 2x stacks which is passing locally.

PR #5948, opened 2021-09-02
Added support & tests for np operations (eq, ne, add, sub, repeat, pos) to S.E.P.T, R.E.P.T

IshanMi:
Adding support & tests for np operations (eq, ne, add, sub) to S.E.P.T, R.E.P.T, I.G.T


packages/syft/src/syft/core/tensor/autodp/intermediate_gamma.py
```diff
@@ -257,3 +258,42 @@ def __mul__(self, other: Any) -> IntermediateGammaTensor:
             bias_tensor=bias_tensor,
             scalar_manager=self.scalar_manager,
         )
+
+    def __eq__(self, other: SupportedChainType) -> IntermediateGammaTensor:
+        if is_acceptable_simple_type(other):
+            term_data = self.term_tensor == other  # Need to check if shapes are broadcastable!!
+        elif isinstance(other, IntermediateGammaTensor):
+            # TODO: Check what actually needs to be identical to do an equality comparison
```
madhavajay, line 266: Good question. We really should add `__eq__` to the new serde interface because its crucial that equality is defined on a class so that we can confirm that the serialization works as expected. We probably need to do this through the whole tensor chain since each tensor can have arbitrary attributes with values which need comparison.

packages/syft/tests/syft/core/tensor/autodp/row_entity_phi_test.py
```diff
@@ -0,0 +1,125 @@
+# stdlib
```
madhavajay, line 1: awesome work

packages/syft/tests/syft/core/tensor/autodp/row_entity_phi_test.py
```diff
@@ -0,0 +1,125 @@
...
+    """Generate binary data to test the equality operators with bools"""
+    binary_data = np.random.randint(2, size=(dims, dims))
+    return binary_data
+
+
+def test_eq(row_data: List) -> bool:
+    """Test equality between two identical RowEntityPhiTensors"""
+    reference_tensor = REPT(rows=row_data)
+    second_tensor = REPT(rows=row_data)
+    third_tensor = reference_tensor
```
madhavajay, line 73: This assignment will mean that `third_tensor` and `reference_tensor` will point to the same variable in memory and have the same id() so comparing it later should always be True, was that intended?

packages/syft/tests/syft/core/tensor/autodp/row_entity_phi_test.py
```diff
@@ -0,0 +1,125 @@
...
+
+    assert (
+        reference_tensor == reference_sept
+    ), "REPT and SEPT equality comparison failed"
+    assert row_data == reference_tensor.child, "Error: data & child don't match"
+    assert (
+        type(reference_tensor == reference_sept) == REPT
+    ), "Return type error for equality comparison b/w REPT, SEPT"
+    # assert type(reference_sept == reference_tensor) == REPT, "Return type error for == comparison b/w SEPT, REPT"
+    return True
```
madhavajay, line 96: Test functions don't need to return anything. You can change the return type to -> None as well.

packages/syft/tests/syft/core/tensor/autodp/row_entity_phi_test.py
```diff
@@ -0,0 +1,125 @@
...
+        return tensor1 == tensor2
+
+
+def test_eq_ndarray(row_data: List) -> bool:
+    """Test equality between a SEPT and a simple type (int, float, bool, np.ndarray)"""
+    row_data = row_data[0]
+
+    reference_tensor = REPT(rows=row_data)
+    assert row_data.child == reference_tensor, "Comparison b/w REPT and "  # type: ignore
+    return True
```
madhavajay, line 125: same as above

packages/syft/tests/syft/core/tensor/autodp/row_entity_phi_test.py
```diff
@@ -0,0 +1,125 @@
...
+    ), "REPT and SEPT equality comparison failed"
+    assert row_data == reference_tensor.child, "Error: data & child don't match"
+    assert (
+        type(reference_tensor == reference_sept) == REPT
+    ), "Return type error for equality comparison b/w REPT, SEPT"
+    # assert type(reference_sept == reference_tensor) == REPT, "Return type error for == comparison b/w SEPT, REPT"
+    return True
+
+
+def test_eq_diff_entities(
```
madhavajay, line 99: If we want to make sure you cant compare with different owners, should we also have a test that shows you CAN compare with the same owner and that the `__eq__` passes?

packages/syft/tests/syft/core/tensor/autodp/row_entity_phi_test.py
```diff
@@ -0,0 +1,125 @@
...
+    with pytest.raises(NotImplementedError):
+        return tensor1 == tensor2
+
+
+def test_eq_ndarray(row_data: List) -> bool:
+    """Test equality between a SEPT and a simple type (int, float, bool, np.ndarray)"""
+    row_data = row_data[0]
+
+    reference_tensor = REPT(rows=row_data)
+    assert row_data.child == reference_tensor, "Comparison b/w REPT and "  # type: ignore
```
madhavajay, line 124: You can remove the type ignore if we tell mypy what the type is like this:
```
def test_eq_ndarray(row_data: List) -> None:
    """Test equality between a SEPT and a simple type (int, float, bool, np.ndarray)"""
    sub_row_data: SEPT = row_data[0]

    reference_tensor = REPT(rows=sub_row_data)
    assert sub_row_data.child == reference_tensor, "Comparison b/w REPT and "
```

packages/syft/tests/syft/core/tensor/autodp/row_entity_phi_test.py
IshanMi: Added!

packages/syft/src/syft/core/tensor/autodp/intermediate_gamma.py
```diff
@@ -258,3 +259,52 @@ def __mul__(self, other: Any) -> IntermediateGammaTensor:
...
+                bias_data = self.bias_tensor == other.bias_tensor
+            else:
+                raise Exception(
+                    f"Bias Tensor shapes do not match for __eq__: {self.child} != len{other}"
+                )
+        elif isinstance(other, PassthroughTensor):
+            if (
+                self.child.shape == other.child.shape
+            ):  # also check if shapes are broadcastable
+                term_data = self.child == other.child
```
iamtrask, line 300: Unfortunately I'm not sure that this can be the way in which we handle equality. This checks to see if the polynomial itself is equal but what we actually need to do is check whether the underlying result of the polynomial is equal. One could have two different polnyomials return the same value (they could both return 0, for example)

packages/syft/src/syft/core/tensor/autodp/row_entity_phi.py
```diff
@@ -101,10 +102,20 @@ def __eq__(self, other: Any) -> RowEntityPhiTensor:
                 f"Tensor dims do not match for __eq__: {len(self.child)} != {len(other.child)}"  # type: ignore
             )
 
+    def __ne__(self, other: Any) -> RowEntityPhiTensor:
+        opposite_result = self.__eq__(other)
```
iamtrask, line 106: Probably need to comment this out for now. see __eq__ for comment.

packages/syft/src/syft/core/tensor/autodp/single_entity_phi.py
```diff
@@ -376,55 +377,57 @@ def __repr__(self) -> str:
             f"{self.__class__.__name__}(entity={self.entity.name}, child={self.child})"
         )
 
+    # Check for shape1 = (1,s), and shape2 = (,s) --> as an example
     def __eq__(self, other: SupportedChainType) -> SingleEntityPhiTensor:
-
-        if is_acceptable_simple_type(other) or self.child.shape == other.child.shape:  # type: ignore
-            # if the tensor being compared is also private
-            if isinstance(other, SingleEntityPhiTensor):
```
iamtrask, line 383: Any reason this was deleted?

packages/syft/src/syft/core/tensor/autodp/single_entity_phi.py
```diff
@@ -376,55 +377,57 @@ def __repr__(self) -> str:
...
-                max_vals=max_vals,
-                scalar_manager=self.scalar_manager,
-            )
-        else:
-            raise Exception(
-                f"Tensor shapes do not match for __eq__: {len(self.child)} != {len(other.child)}"  # type: ignore
-            )
-
-    def logical_and(self, other: SupportedChainType) -> SingleEntityPhiTensor:
-        if is_acceptable_simple_type(other) or self.child.shape == other.child.shape:
```
iamtrask, line 407: I'm confused about what's going on here.

packages/syft/src/syft/core/tensor/autodp/intermediate_gamma.py
iamtrask: (we can't do IntermediateGammaTensor until we modify with the clip at teh end of the polynomial as discussed at the last meeting I attended.)

iamtrask (dismissed):

packages/syft/src/syft/core/tensor/autodp/intermediate_gamma.py
IshanMi: I fully agree, I haven't modified the code for IntermediateGammaTensors since before we had that meeting, and reduced the scope of this PR to just the methods for SingleEntity and RowEntity Phi Tensors

packages/syft/src/syft/core/tensor/autodp/single_entity_phi.py
IshanMi: This would fail if self.child and other.child have shapes that are not equal, but still broadcastable, so I rewrote them to check for broadcastability, and if self.entity == other.entity:

[Link to relevant lines](https://github.com/OpenMined/PySyft/blob/a13862e62d75dd1cc3e3f2c64625f9abfb757320/packages/syft/src/syft/core/tensor/autodp/single_entity_phi.py#L381)

packages/syft/src/syft/core/tensor/autodp/single_entity_phi.py
IshanMi: if `other` is an acceptable_simple_type (int, float, bool, np.ndarray) then there are two scenarios:

- if it's a float/int/bool, then the comparison is doable, and no additional checks need to be made
- if it's a numpy array, then we need to check that self.child.shape and other.child.shape are broadcastable. If it's not, we raise an exception and tell the user that __eq__ can't be performed because the shapes are not broadcastable. If they are, we save the result of `self.child == other.child`, and return a SingleEntityPhiTensor whose child equals this result.

Sorry if my changes and thought process isn't clear! I can add more comments inline to explain.

iamtrask: Ah makes sense - actualy what confuse dme was the way Github printed it. The code is clear.

packages/syft/src/syft/core/tensor/autodp/intermediate_gamma.py
IshanMi: Marking as resolved since there's another PR open for equality for IntermediateGammaTensors

iamtrask (approved):

PR #5952, opened 2021-09-03
Fixed Intergration Tests SMPC

rasswanth-s:
## Description
Modified to use `get_copy` in reconstruct.

## How has this been tested?
Locally

## Checklist


iamtrask (approved):

madhavajay:
Is it okay to do this? Will the objects be stuck in the store forever?

rasswanth-s:
> Is it okay to do this? Will the objects be stuck in the store forever?
Yes @madhavajay ,that's a caveat, @iamtrask  and @gmuraru  suggested ideas like using blocking `get` until object appears, If we incorporate it , I think it would solve.

iamtrask:
Yeah it strikes me these would get stuck in the store forever. We need to do .get() instead but handle when subsets fail.

gmuraru:
What is the issue with doing `get`, @rasswanth-s ?
The idea from using `get` here was to fetch the shares locally and then to add them.
If the process of `fetching` the shares would  fail (because maybe one of the parties did not approve the request for `get`) - then we will `resend` the shares we fetch until that point.

rasswanth-s:
Yes @gmuraru , I am not sure completely , why it is not working , I will take a additional look into it.

PR #5954, opened 2021-09-04
SimpleService and NodeRunnableMessageWithReply

iamtrask:
## Description
In offline discussions, it has been raised that our use of Service and Message objects (the latter of which requires creating custom protobufs for each message type) is extremely cumbersome. Instead, it has been proposed that we should simply have Message types which themselves have a method that knows how to execute some functionality when the message is received by a Node. 

This PR is an initial step in this direction. A project which moves all of our services (or rather removes many of them) into this kind of model is a massive project. The desire of this PR is to make such a project incremental. I have created a service called "SimpleService" and installed it on the Domain node. Any subclass of NodeRunnableMessageWithReply can automatically be sent from a Client to a Domain, and NodeRunnableMessageWithReply is a subclass of RecursiveSerde.

Future work will attempt to use this service and message type to port existing functionality or to implement new functionality with ideally greatly reduced effort from our previous setup.

NOTE: I also discovered a bug in 'dev' wherein database tables are not being properly created. I fixed it.
NOTE: I also developed a suspicion that our service-based auth decorators might not be working correctly.


iamtrask:
One of the Syft + Grid Stack Integration Tests are failing and it's non-obvious why. I don't think anything I changed should have affected it.

iamtrask:
I'm merging this now because even though one of hte integration tests fails I'm confident this is fixing more than it's breaking (at present the users page doesn't work)

iamtrask:
Ah the issue in this one is fixed in a different PR apparently?

PR #5955, opened 2021-09-04
syft.register() and domain_client.request_budget()

iamtrask:
## Description
In this pull request, I prototype two pieces of functionality which are MVP for Course 3:

- syft.register() - if the Domain is configured to allow external users to register their accounts without needing approval from the Administrator, they can do so through syft.register().
- domain_client.request_budget() - someone logged in via python can make a request for more budget which will populate into the requests queue of the Domain. If the request is approved (via either the UI or python interface), that user will have the additional budget allocated to their account.


iamtrask:
Merging as failing test is not a regression - merely inheriting the fail from dev.

PR #5971, opened 2021-09-13
Abstraction of MPCTensor for tensor pointer computation.

rasswanth-s:
## Description
The tensor pointers are abstracted at a high level such that when computation is done , it is automatically converted to MPCTensors extendable for arbitrary number of parties.

## How has this been tested?
Created a jupyter notebook detailing the same.

## Checklist
- [x]  I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)


packages/syft/src/syft/grid/client/client.py
```diff
@@ -136,6 +140,8 @@ def login(
         time.sleep(1)  # ok maybe too fancy... but c'mon don't you want to be fancy?
         print("done!")
 
+    sy.parties.add(node)
```
rasswanth-s, line 143: Works only for Hagrid Version

packages/syft/src/syft/core/tensor/tensor.py
```diff
@@ -60,13 +61,14 @@ def share(self, *parties: Tuple[AbstractNodeClient, ...]) -> MPCTensor:
...
-    def simple_add(self, other: Any) -> TensorPointer:
+    def _apply_tensor_op(self, other: Any, op_str: str) -> Any:
         # we want to get the return type which matches the attr_path_and_name
         # so we ask lib_ast for the return type name that matches out
         # attr_path_and_name and then use that to get the actual pointer klass
         # then set the result to that pointer klass
 
-        attr_path_and_name = "syft.core.tensor.tensor.Tensor.__add__"
+        attr_path_and_name = "syft.core.tensor.tensor.Tensor."
+        attr_path_and_name += "__" + op_str + "__"
```
gmuraru, line 71: MAYBE: Use string formating - ```attr_path_and_name = f"syft.core.tensor.Tensor.__{op_str}__"```

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
```diff
@@ -384,42 +384,14 @@ def __apply_op(
         Returns:
             MPCTensor. the operation "op_str" applied on "self" and "y"
         """
+        # relative
+        from .... import parties as mpc_parties
```
iamtrask, line 388: I don't think the design paradigm of creating a global set of parties that every MPC tensor must exactly adhere to is the paradigm we should be considering. Instead, the list of parties should be attached to the MPCTensor and grow as needed. 

For example - what if I was doing federated learning across 2,000 different data providers? If I wanted to do secure aggregation across clusters of 5 at a time, this design decision would preclude me from doing this (I'd be forced to only and always do SMPC against all 2,000 providers).

iamtrask (changes requested):
(See inline comment)

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
rasswanth-s: Modified to Reshare.

packages/syft/src/syft/grid/client/client.py
rasswanth-s: Modified

packages/syft/src/syft/core/tensor/tensor.py
```diff
@@ -108,18 +109,43 @@ def simple_add(self, other: Any) -> TensorPointer:
...
+        Args:
+            other (Union[TensorPointer,MPCTensor,int,float,np.ndarray]): second operand.
 
-        if self.client != other.client:
+        Returns:
+            Tuple[MPCTensor,Union[MPCTensor,int,float,np.ndarray]] : Result of the operation
+        """
+        op = getattr(operator, op_str)
+
+        if isinstance(other, TensorPointer) and self.client != other.client:
```
gmuraru, line 148: Q: I know that this piece was already here, but do you know if this works in case we have:
`mpc_tensor1` between `client1` and `client2` and
`mpc_tensor2` between `client2` and `client3`?

rasswanth-s: I created a notebook and tested the functionality george in the same PR,
Since we have the reconstruct problem, I thought to add hagrid test after a fix for that.
https://github.com/OpenMined/PySyft/blob/ch2858/notebooks/smpc/SMPC%20Tensor%20pointer%20Abstraction.ipynb

rasswanth-s: I tried covering all the edge cases in the notebook, If there are any other edge cases comment , george, I will include it.

rasswanth-s: The previous case did not work for MPCTensors george, only between tensor pointer and MPCTensor, with different subsets of parties,I think now it work for all scenario.

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
```diff
@@ -352,6 +353,78 @@ def __getattribute__(self, attr_name: str) -> Any:
 
         return object.__getattribute__(self, attr_name)
 
+    @staticmethod
+    def reshare(self: MPCTensor, parties: Iterable[Any]) -> MPCTensor:
+        """Reshare a given secret to a superset of parties.
+
+        Args:
+            parties(List[Any]): Input parites List.
```
gmuraru, line 361: TYPO: parties

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
```diff
@@ -352,6 +353,78 @@ def __getattribute__(self, attr_name: str) -> Any:
 
         return object.__getattribute__(self, attr_name)
 
+    @staticmethod
+    def reshare(self: MPCTensor, parties: Iterable[Any]) -> MPCTensor:
```
gmuraru, line 357: OUGHTTO: `staticmethod` should not have a `self` attribute attached to them - could you name it with something else?

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
```diff
@@ -352,6 +353,78 @@ def __getattribute__(self, attr_name: str) -> Any:
...
+                "Input parties for resharing are same as the input parties."
+            )
+
+        shares = [client_map.get(party) for party in parties]
+        for i, party in enumerate(parties):
+            shares[
+                i
+            ] = party.syft.core.tensor.smpc.share_tensor.ShareTensor.generate_przs(
+                rank=i,
+                nr_parties=len(parties),
```
gmuraru, line 385: NITPICK: Could you save this in a variable before the loop and use it?

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
```diff
@@ -352,6 +353,78 @@ def __getattribute__(self, attr_name: str) -> Any:
...
+        shape = self.shape
+        seed_shares = self.seed_shares
+        client_map = {share.client: share for share in self.child}
+        if mpc_parties == parties:
+            raise ValueError(
+                "Input parties for resharing are same as the input parties."
+            )
+
+        shares = [client_map.get(party) for party in parties]
+        for i, party in enumerate(parties):
```
gmuraru, line 380: MAYBE:
```
res_shares = []
for share, party in zip(shares, parties):
    res_shares.append(generate_przs)
```

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
```diff
@@ -352,6 +353,78 @@ def __getattribute__(self, attr_name: str) -> Any:
...
+                # TODO: Should be modified after Trask's Synthetic data PR.
+                raise ValueError("The input tensor pointer should have public shape.")
+            if client not in parties:
+                parties.append(client)
+                self = MPCTensor.reshare(self, parties)
+
+            other = MPCTensor(secret=other, parties=parties, shape=public_shape)
+
+        elif isinstance(other, MPCTensor):
+            p1 = set(self.parties)  # parties in first MPCTensor
```
gmuraru, line 419: BUG?: What happens if we have:
```
self.parties = [p1, p2]
other.parties = [p2, p1]
```
The sets would be equal, right?

But in that case - if we were to do a `private_add` we would have: share_p1 + share_p2 <-- will throw an exception because they are not on the same machine?

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
```diff
@@ -385,42 +458,9 @@ def __apply_op(
...
-            op = getattr(operator, op_str)
-
-            new_shares = [
-                op(share1, share2) for share1, share2 in zip(self.child, shares)
-            ]
-
-            new_shares.append(shares[-1])
 
-            _self = MPCTensor(shares=new_shares, shape=self.mpc_shape, parties=parties)
+        _self, y = MPCTensor.sanity_checks(_self, y)
```
gmuraru, line 463: Q: Could we name those? x, y = MPCTensor.sanity_check(self, y)

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
rasswanth-s: We sort the MPCTensor child nodes, every time we instantiate it george,
https://github.com/OpenMined/PySyft/blob/a7d5ac43534f96b9586fadf7302ddca5b0554d18/packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py#L107
by which we call on the correct mpc tensors,

rasswanth-s: maybe I will also include a sanity check in ops level for robust code.

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
rasswanth-s: Since it is dictionary, we automatically set `None` george if key not available, which automatically fits for the PRZS check in `ShareTensor`,I left it like that.

rasswanth-s: It autmatically choose a Zero Tensor in ShareTensor code.

rasswanth-s: It is actually , ``shares[i]`` george, if that part confuses, black formatted it that way :sweat:

iamtrask (approved):
This PR is beautiful.

iamtrask:
Can we get an integration test for this before we merge? Perhaps including the functionality from the notebook included in the PR?

rasswanth-s:
@iamtrask ,since we had the reconstruct issue, we have to create test with ``pytest.mark.xfail``, I thought that maybe we could do it after fixing,I could also do now .Shall we do after fixing or we will add now itself?

iamtrask:
Sounds good - just be sure to add it to the spreadsheet somehow so we don't forget it.

PR #5986, opened 2021-09-19
Add comparison operators for REPTs and SEPTs

fiza11:
Added support and tests for np operations (le, ge, lt, gt) to S.E.P.T and R.E.P.T.


packages/syft/src/syft/core/tensor/autodp/row_entity_phi.py
```diff
@@ -271,6 +272,123 @@ def transpose(self, *dims: Any) -> RowEntityPhiTensor:  # type: ignore
...
+            new_list = list()
+            for i in range(len(self.child)):
+                new_list.append(self.child[i] <= other)
+
+            return RowEntityPhiTensor(rows=new_list, check_shape=False)
+
+        if len(self.child) == len(other.child):  # type: ignore
+            # tensors have different entities
+            if not (self.entities == other.entities).all():
+                raise Exception("Tensor owners do not match")
```
iamtrask, line 288: best thing to do here would be to instead call self.gamma <= other.gamma (cast both to gamma tensors and run from there, obv that's not implemented yet but you can still delegate the computation to there from here)

iamtrask (approved):
Really beautiful PR.

PR #5987, opened 2021-09-19
Ci optimize path filter

bitsofsteve:
## Description
This PR will enable path filter and only run test where code changes have been made.


.github/workflows/grid-backend_tests.yml
```diff
@@ -7,11 +7,9 @@ on:
       - dev
       - main
       - "0.6.0"
-    paths:
-      - ".github/workflows/grid-*.yml"
-      - "packages/grid/**.py"
-      - "packages/grid/**.txt"
-      - "packages/grid/**.toml"
```
rasswanth-s, line 14: Q: Should we also include `` .txt and .toml to file-filters`` workflow as it is inlcuded here.

.github/workflows/hagrid-pr_tests.yml
```diff
@@ -6,11 +6,6 @@ on:
       - dev
       - demo_strike_team_branch_4
       - "0.6.0"
-    paths:
-      - ".github/workflows/hagrid-*.yml"
-      - "packages/hagrid/**.py"
-      - "packages/hagrid/**.txt"
```
rasswanth-s, line 12: Q: Same question as above

.github/workflows/grid-backend_tests.yml
bitsofsteve: Right @rasswanth-s thanks for the catch! 💚

.github/file-filters.yml
```diff
@@ -0,0 +1,6 @@
+backend:
```
rasswanth-s, line 1: Q: Should we skip an entire job by creating a proxy one instead like one dorny's examples
https://github.com/dorny/paths-filter#examples
Or it is better we do step by step checking?

.github/workflows/syft-pr_tests.yml
```diff
@@ -67,76 +72,25 @@ jobs:
...
-        if: startsWith(runner.os, 'macos')
+        if: startsWith(runner.os, 'macos') && steps.changes.outputs.backend == 'true'
         run: |
           brew install libomp
           pip uninstall python-dp -y
 
       - name: Run normal tests
+        if: steps.changes.outputs.backend == 'true'
         run: |
           pytest -m 'fast or slow' --cov syft --cov-fail-under 70 -n auto --durations=50
```
rasswanth-s, line 89: This is one of the trick , I used in ``SyMPC`` repo for speeding up CI,
https://github.com/OpenMined/SyMPC/blob/634396a9d663eda514f66654a96c23ae2ab7d2de/.github/workflows/tests.yml#L189
Instead of setting auto , we could set a custom like ``8``,  The value 8 really worked for our repo, we could also experiment with the values, I am sure, It would greatly speed the CI

.github/file-filters.yml
rasswanth-s: I think the step method looks optimal , without another CI runner,I think it better like this

rasswanth-s (approved):
LGTM :rocket: some comments

.github/workflows/syft-pr_tests.yml
bitsofsteve: Sweet stuff! @rasswanth-s thanks for the tip, I'll have to test that and compare then open another PR for it. Also, how did you compare between auto & a value like 8 to find the magnitude of the difference? Would love to know. Also I know that `auto` checks for available system resource before assigning a process, so a fixed number like `8` makes a classical assumption that there will be always resource for 8 workers. Are there supposed downsides to this?

rasswanth-s: Sure @bitsofsteve , the runners in Github CI have two cores for ``ubuntu`` and three cores ``mac os``, when there was ``auto`` only two and three workers were set by default, My intution was that , increasing workers ,would make it more multi threaded,it worked sort of trial and error. 

One **downside : was that , when I set high number workers(more than 8) , it frequently crashed the workers, we should like kind of set an optimal number.
I remember windows runner having problem with 8 workers, maybe we could start with ``6`` and check?

madhavajay: Using more cores than there are CPUs would only be a good idea if there are a lot of non CPU bound tests. It might have been in the SyMPC repo that you had a lot of waiting going on between tests. I think for now `auto` is fine since we are running a lot of different tests across the Syft repo most of which are doing work, but I would be interested to see with some benchmarks if we can get things faster by tweaking stuff like this.

madhavajay (approved):
This is awesome work @bitsofsteve! I have just been playing around and I wanted to see if we can isolate different tests based on their specific sub packages. I think since we are touching all the test yaml files in this PR its hard to know if its working so lets merge and then make sure its filtering as we expect. I also renamed some tests so that they are different and split out the core / libs from syft to make things more parallel. Once this is merged ill change the Branch protection settings and then we can see if it works as expected.

I had to disable the SNYK tests coz they broke and I don't know why. I even regenerated the auth token and changed it in the settings, so its weird, I have no idea why they are broken but lets discuss in Heartbeat planning anyway.

bitsofsteve (commented):
@madhavajay , this looks great, but my assumption was that the directive `working-directory` in the workflow file sets the working directory per run to the specified directory, and then that is where the file-filter check is compared at that instance, or is this wrong? 

This way I didn't need to specify separately the file path for syft,hagrid and syft.

I'd want to know your thoughts around having different filters for the specific packages.
Thanks

PR #5999, opened 2021-09-22
Add get operations for SEPTs

fiza11:
Added support and tests for np get operations (copy, diagonal, take) to S.E.P.T.

### DIFF
--- a/packages/syft/src/syft/core/tensor/autodp/single_entity_phi.py
+++ b/packages/syft/src/syft/core/tensor/autodp/single_entity_phi.py
@@ -982,9 +982,9 @@ class SingleEntityPhiTensor(PassthroughTensor, AutogradTensorAncestor, ADPTensor
         mode: Optional[str] = "raise",
     ) -> SingleEntityPhiTensor:
         """Take elements from an array along an axis"""
-        data = self.child.take(indices=indices, mode=mode)
-        min_vals = self.min_vals.take(indices=indices, mode=mode)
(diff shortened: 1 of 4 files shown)


packages/syft/src/syft/core/tensor/autodp/single_entity_phi.py
```diff
@@ -959,6 +959,62 @@ def clip(
...
+        indices: np.ArrayLike,
+        axis: Optional[int] = None,
+        mode: Optional[str] = "raise",
+    ) -> SingleEntityPhiTensor:
+        """Take elements from an array along an axis"""
+        data = np.take(self.child, indices=indices, mode=mode)
+        return SingleEntityPhiTensor(
+            child=data,
+            entity=self.entity,
+            min_vals=self.min_vals,
```
iamtrask, line 989: Seems like this would lead to a tensor whose min_vals have a different shape than the data since min_vals wasn't transformed in line with take. Let's run self.min_vals.take() as well for min_vals and max_vals

packages/syft/src/syft/core/tensor/autodp/single_entity_phi.py
```diff
@@ -959,6 +959,62 @@ def clip(
             scalar_manager=self.scalar_manager,
         )
 
+    def copy(
```
iamtrask, line 962: this looks great!

iamtrask (changes requested):
Left comments line

PR #6003, opened 2021-09-26
Add support and tests for all transform operations for RowEntityPhiTensors

IshanMi:
## Description
Added support and tests for all transform operations for RowEntityPhiTensors:

- [x] Transpose
- [x] Reshape
- [x] Resize
- [x] Partition
- [x] Compress
- [x] Squeeze
- [x] Swapaxes
- [x] Ravel 

## Affected Dependencies
No affected dependencies.

## How has this been tested?

Added pytest tests to the `row_entity_phi_test.py` file, and used a `pre-commit run --all-files` to verify linting and other errors.

## Checklist

### DIFF
--- a/packages/syft/src/syft/core/tensor/autodp/row_entity_phi.py
+++ b/packages/syft/src/syft/core/tensor/autodp/row_entity_phi.py
@@ -267,7 +267,8 @@ class RowEntityPhiTensor(PassthroughTensor, ADPTensor):
             )
         new_list = list()
         for tensor in self.child:
-            new_list.append(tensor.swapaxes(axis1, axis2))
+            # Axis=0 for REPT.child is Axis=1 for REPT, so subtract 1
+            new_list.append(tensor.swapaxes(axis1 - 1, axis2 - 1))
         return RowEntityPhiTensor(rows=new_list, check_shape=False)
 
     def squeeze(
@@ -288,10 +289,6 @@ class RowEntityPhiTensor(PassthroughTensor, ADPTensor):
         return RowEntityPhiTensor(rows=new_list, check_shape=False)
 
     def reshape(self, *shape: List[int]) -> RowEntityPhiTensor:
-        print(shape, type(shape), self.shape, type(self.shape))
-        print(type(shape[0]))
-        print(type(self.shape[0]))
-
         # This is to fix the bug where shape = ([a, b, c], )
         if isinstance(shape[0], list) or isinstance(shape[0], tuple):
             shape = shape[0]
--- a/packages/syft/src/syft/core/tensor/autodp/row_entity_phi.py
+++ b/packages/syft/src/syft/core/tensor/autodp/row_entity_phi.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 # stdlib
+from collections.abc import Sequence
 from typing import Any
 from typing import List
 from typing import Optional
@@ -21,9 +22,9 @@ from ..broadcastable import is_broadcastable
 from ..passthrough import PassthroughTensor  # type: ignore
 from ..passthrough import implements  # type: ignore
 from ..passthrough import is_acceptable_simple_type  # type: ignore
-from ..types import AcceptableSimpleType  # type: ignore
+from ..types import AcceptableSimpleType
 from .adp_tensor import ADPTensor
-from .initial_gamma import InitialGammaTensor  # type: ignore
+from .initial_gamma import InitialGammaTensor
 from .single_entity_phi import SingleEntityPhiTensor
 
 
@@ -44,7 +45,7 @@ class RowEntityPhiTensor(PassthroughTensor, ADPTensor):
     # a list of attributes needed for serialization using RecursiveSerde
     __attr_allowlist__ = ["child"]
 
-    def __init__(self, rows: Any, check_shape: bool = True):
+    def __init__(self, rows: Sequence, check_shape: bool = True):
         """Initialize a RowEntityPhiTensor
 
         rows: the actual data organized as an iterable (can be any type of iterable)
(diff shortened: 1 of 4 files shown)


packages/syft/src/syft/core/tensor/autodp/row_entity_phi.py
```diff
@@ -112,21 +114,21 @@ def create_gamma(
 
     @property
     def shape(self) -> Tuple[Any, ...]:
-        return [len(self.child)] + list(self.child[0].shape)  # type: ignore
+        return tuple([len(self.child)] + list(self.child[0].shape))
```
madhavajay, line 117: We're promising a tuple here in the signature, so we should return one.

packages/syft/src/syft/core/tensor/autodp/row_entity_phi.py
```diff
@@ -44,15 +45,16 @@ class RowEntityPhiTensor(PassthroughTensor, ADPTensor):
...
         """Initialize a RowEntityPhiTensor
 
         rows: the actual data organized as an iterable (can be any type of iterable)
         check_shape: whether or not we are already confident that the objects in iterable
             'rows' all have the same dimension (check if we're not sure).
 
         """
-
+        # Container type heirachy: https://docs.python.org/3/library/collections.abc.html
+        self.child: Sequence
```
madhavajay, line 57: We can actually tell mypy that the instance variable .child will be a Sequence so it can chill out. The other ways are:
```python
# this looks ugly and exposes the parent implementation
def init...
    self.child = rows

# technically works but also allows a child on the Class which we dont want
class RowEntityPhiTensor:
    child: Sequence

    def init...
```

packages/syft/src/syft/core/tensor/autodp/row_entity_phi.py
```diff
@@ -187,7 +189,7 @@ def __mul__(  # type: ignore
                     )
                 else:
                     raise Exception(
-                        f"Tensor dims do not match for __sub__: {self.child.shape} != {other.shape}"  # type: ignore
+                        f"Tensor dims do not match for __sub__: {getattr(self.child, 'shape', None)} != {other.shape}"
```
madhavajay, line 192: Child probably shouldn't have a shape property but at least this can safely try to get it.

packages/syft/src/syft/core/tensor/autodp/row_entity_phi.py
```diff
@@ -248,35 +250,40 @@ def __and__(self, other: Any) -> RowEntityPhiTensor:
...
             return RowEntityPhiTensor(rows=new_list, check_shape=False)
         else:
             raise Exception(
-                f"Tensor dims do not match for __truediv__: {len(self.child)} != {len(other.child)}"  # type: ignore
+                f"Tensor dims do not match for __truediv__: {len(self.child)} != {len(other.child)}"
             )
 
     def repeat(
-        self, repeats: Union[int, Tuple[int, ...]], axis: Optional[int] = None
+        self, repeats: Union[int, List[int]], axis: Optional[int] = None
```
madhavajay, line 271: Numpy docs says repeats is a Union of int and List. I assume that we should mimic it but if im wrong we can change it back.

packages/syft/src/syft/core/tensor/autodp/row_entity_phi.py
```diff
@@ -248,35 +250,40 @@ def __and__(self, other: Any) -> RowEntityPhiTensor:
...
+
         if axis is None:
             raise Exception(
                 "Conservatively, RowEntityPhiTensor doesn't yet support repeat(axis=None)"
             )
 
         if axis == 0 or axis == -len(self.shape):
             new_list = list()
-            for r in range(repeats):  # type: ignore
+            for _ in range(repeats):
```
madhavajay, line 286: This was causing an error because one valid input is a Tuple for repeats, so we need to guard against that until its implemented with an Exception as above.

packages/syft/src/syft/core/tensor/autodp/row_entity_phi.py
```diff
@@ -299,12 +306,69 @@ def repeat(
...
+            int,
+            Union[Sequence[int], Sequence[Sequence[int]]],
+        ],
+    ) -> RowEntityPhiTensor:
+        if isinstance(shape, int):
+            raise Exception(
+                f"{type(self)}.reshape for shape: int {shape} is not implemented"
+            )
+        # This is to fix the bug where shape = ([a, b, c], )
+        if isinstance(shape[0], Sequence):
```
madhavajay, line 362: This was complaining because the shape signature was promising Union int and Sequence not a possible nested sequence... 🙄 yup this stuff can be annoying.

packages/syft/tests/syft/core/tensor/autodp/row_entity_phi_test.py
```diff
@@ -386,6 +386,144 @@ def test_sub_result_gamma(row_data_ishan: List, row_data_trask: List) -> None:
...
+    assert reference_tensor.shape != original_tensor.shape, "Resize didn't work"
+    assert reference_tensor.shape == new_shape, "Resize shape doesn't check out"
+
+
+@pytest.mark.skipif(dims == 1, reason="Dims too low for this operation")
+def test_reshape(row_data_ishan: List) -> None:
+    """Test to see if Reshape works for the ideal case"""
+    reference_tensor = REPT(rows=row_data_ishan)
+    original_shape = reference_tensor.shape
+    new_shape = tuple(
```
madhavajay, line 484: fixed to match the output for `shape`

madhavajay (approved):
Awesome work! I noticed there were a lot of `# type: ignore` so I thought it might be worth investigating. The problem with ignore is it usually lives forever and then potentially hides other errors in that line. In this case most of the errors could be removed after satisfying the compiler about that the type of `child` was in this subclass. In addition some of the other warnings were valid concerns where code passed in could lead to an exception so in those cases it might be better to filter them out with our own exceptions straight away so we at least know that type of input is not supported yet.

packages/syft/src/syft/core/tensor/autodp/row_entity_phi.py
IshanMi: Oh that's super helpful, I didn't know about the Sequence data type but reading about it now, I think it should work!

PR #6010, opened 2021-09-28
Fixed Inplace SQL Alchemy issue.

rasswanth-s:
## Description
For class functions, when a storeable object is manipulated , it is not reflected in the domain's store.
We reassign it after functionality is manipulated.

## How has this been tested?
Add a test for the same using `List.append`

## Checklist


packages/syft/src/syft/core/node/common/action/run_class_method_action.py
```diff
@@ -243,7 +243,8 @@ def execute_action(self, node: AbstractNode, verify_key: VerifyKey) -> None:
             args=tag_args,
             kwargs=tag_kwargs,
         )
-
+        # TODO :Make it work for inplace ops in sql alchemy.
+        node.store[self._self.id_at_location] = resolved_self  # type: ignore
```
gmuraru, line 247: Q: This seems wrong (that we have 2 `id_at_location`) one in `self` and one in `self._self`?

rasswanth-s: Great question george, self._self points to ``current pointer being manipulated``, self.id_at_location points to the result, while doing inplace operations , when we manipulate store object of that pointer , the current pointer is not automatically mutated in sql alchemy.

rasswanth-s: So we reassign back to the store the  ``Storeableobject`` after manipulation.

packages/syft/src/syft/core/node/common/action/run_class_method_action.py
```diff
@@ -226,6 +230,17 @@ def execute_action(self, node: AbstractNode, verify_key: VerifyKey) -> None:
...
+        # check if resolved_self has changed and if so mark as mutating_internal
+        # this prevents someone from mutating an object they own with something they
+        # do not own and the read_permissions not flowing backwards
+        if (
+            resolved_self_previous_bytes is not None
+            and resolved_self is not None
+            and resolved_self_previous_bytes
+            != sy.serialize(resolved_self.data, to_bytes=True)
+        ):
+            mutating_internal = True
```
rasswanth-s, line 242: Awesome

packages/syft/src/syft/core/node/common/action/run_class_method_action.py
madhavajay: Yes this is correct and the tests verify so.

iamtrask (approved):
Well done. Fantastic work on an incredibly nuanced and challenging problem. Now that I see it, we have had to implement fix/paradigms like this in previous versions of PySyft and so I have high confidence your approach will be successful long term. Great work.

PR #6013, opened 2021-09-29
Add qualifier and arithmetic operations for SEPTs

fiza11:
Added support and tests for np qualifier operations (any, all) and arithmetic operations (abs, pow, sum) to S.E.P.T.

### DIFF
--- a/packages/syft/tests/syft/core/tensor/autodp/single_entity_phi_test.py
+++ b/packages/syft/tests/syft/core/tensor/autodp/single_entity_phi_test.py
@@ -1181,6 +1181,68 @@ def test_clip() -> None:
     assert (clipped_tensor3 >= rand1).all()
 
 
+# tensor filled with zeros
+zeros_tensor = SEPT(
+    child=np.zeros(dims, dtype=np.int32),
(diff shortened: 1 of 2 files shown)


packages/syft/tests/syft/core/tensor/autodp/single_entity_phi_test.py
```diff
@@ -1181,6 +1181,68 @@ def test_clip() -> None:
     assert (clipped_tensor3 >= rand1).all()
 
 
+# tensor filled with zeros
+zeros_tensor = SEPT(
```
IshanMi, line 1185: Could we replace these new global variables and implement them as pytest fixtures instead? (I think we discussed this during the meeting today)
You could merge from the `dev_fixed_adp_tests` branch and make use of some of the ones already implemented :)

iamtrask: +1

iamtrask (changes requested):
See @IshanMi 's comment about global vars.

PR #6032, opened 2021-10-05
[WIP] SMPC - Private Multiplication and Parties communication.

rasswanth-s:
## Description
Implementation Private Multiplication for SMPC along with making nodes communicate with each other.

- [ ] Add multiplication for private and private values
- [ ] Setup a communication framework for parties to communicate with each other.
- [ ] Construct a _reveal method, such that parties reveal a secret among themselves
- [ ] Fixed point truncation after private multiplication
- [ ] Test the functionality

## Affected Dependencies
```GridHTTPConnection```

## How has this been tested?
Tests to be added.

## Checklist

### CODE REVIEW

File: packages/syft/src/syft/grid/client/grid_connection.py [hunk 1]
```diff
@@ -39,7 +38,8 @@ class GridHTTPConnection(HTTPConnection):
```
madhavajay, line 38: Could we use `@serializable(recursive_serde=True)` and remove the proto file and manual serde steps?

packages/syft/src/syft/core/node/common/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+        "syft",
+        "syft.core",
+        "syft.core.node",
+        "syft.core.node.common",
+        "syft.core.node.common.client",
+    ]
+
+    classes = [
+        (
+            "syft.core.node.common.client.Client",
```
madhavajay, line 30: Do we need an AST entry for Client? Do we need to remotely construct a Client on a remote system from a client itself? Also if this goes into the store, currently while we restrict `.get` we do not restrict arbitrary computation on objects in a store if you know the UID so could a remotely stored client with session token be used by someone to do something bad?

packages/syft/src/syft/grid/client/grid_connection.py
rasswanth-s: Sure , I will change it

PR #6109, opened 2021-10-30
Add version tests path back in workflow file

rasswanth-s:
## Description
Added the version tests folder back in workflow file, to prevent it from being run in each PR as it consumes lot of CI runners,and  we have a nightly test for the same.


.github/workflows/syft-version_tests.yml
```diff
@@ -12,6 +12,8 @@ on:
   pull_request:
     branches:
       - dev
+    paths:
+      - "packages/syft/tests/trigger/version_tests"
```
gmuraru, line 16: This is to run the tests for each python version?

rasswanth-s: This test run all the versions of torch, python in a combinatorial factor,
We have run it in a nightly test, If you would like to run it in a PR, you have to modify that particular or you could run it manually.
I modified it like  this way  before,Due to lot of PR yesterday, this got deleted by mistake.

gmuraru (approved):
LGTM!

PR #6150, opened 2021-11-04
Hagrid land networks

iamtrask:
Hagrid leaves up newtorks when running "hagrid land all" which eventually means docker runs out of ports if you're constantly spinning up and tearing down domain nodes.

Known Issue: "hagrid land all" is too powerful and this makes it more powerful


shubham3121:
Should we refer to it as `hagrid avada kedavra` 😂 as it can now destroy all containers on your system.

PR #6177, opened 2021-11-12
[WIP] feat: configure traefik self-signed TLS

tenevdev:
⚠️ 🚧 This is work in progress!

## Description

Exposed on port 443 and routes to front and back-ends.

## Affected Dependencies
- Self-signed certificates (i.e. `openssl`)

## How has this been tested?

You can either generate a self-signed cert with SAN extension or Traefik can generate a default self-signed certificate.
The current Traefik config assumes you generate one.

```shell
# generating self-signed certs with SAN extension for 127.0.0.1 in the right place
mkdir packages/grid/traefik/certs
cd packages/grid/traefik/certs
curl -sS https://raw.githubusercontent.com/antelle/generate-ip-cert/master/generate-ip-cert.sh |
    bash -s 127.0.0.1
```

```shell
# deploy with docker compose
cd packages/grid
source .env && docker compose -p eu --profile frontend up -d
```

After the frontend loads, open `https://127.0.0.1` in a browser. There will be a warning or error about the certificate not being trusted. Add the certificate to your trusted certificates - either from the browser settings or  for the whole system.

## Checklist


iamtrask:
Woot woot!!!

tenevdev:
Hey @madhavajay, just pushed my progress so far on the TLS story.

On the docker compose side of things, you said that it's ok to break things from the main `docker-compose.yml` file as it's only ever used with the override file and never on its own. Just to confirm, since the main file is the one with options for running with Docker Swarm, does it therefore mean Docker Swarm is never used as a deployment option? If so, I guess I'll remove all mentions of it and further simplify things.

When it comes to `hagrid`, I feel like things are starting to get a lot messier in there than I'd like to work with but I also don't feel comfortable refactoring everything because it's changing pretty quickly without me. What options do you think I should expose and what should happend automagically? For example, for switching the security features on and off, a flag such as `--insecure` (note that deployment would become secure by default). Then, should it ask the user to generate self-signed certificates for them, should it just auto-generate with no prompt, should it tell the user to generate the certs manually and re-run? If users will provide their own certificates, should there be new flags `--tls-certificate`, `--tls-key` with paths to their files or should it tell the user to put their files in the expected location with the expected names? If generating certificates, should it automatically import them to a store, should it prompt the user for options and import, should it tell the user about different ways to import? Should two ports be exposed or only the secure one? Should there be options to turn redirection on and off?

PR #6189, opened 2021-11-17
[WIP]windows Ci test runner setup

bitsofsteve:
## Description
This PR is to setup CI runners for windows testing.

### DIFF
--- a/.github/workflows/stack-integration_tests.yml
+++ b/.github/workflows/stack-integration_tests.yml
@@ -140,6 +140,13 @@ jobs:
           powershell.exe -command "docker rm $(docker ps -aq) --force"
         shell: cmd
 
+      - name: Prune all volumes
+        if: steps.changes.outputs.stack == 'true'
+        continue-on-error: true
+        run: |
+          powershell.exe -command "docker volume prune --force"
+        shell: cmd
+
       - name: Run integration tests
         if: steps.changes.outputs.stack == 'true'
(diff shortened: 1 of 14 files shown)


.github/workflows/stack-integration_tests.yml
```diff
@@ -118,7 +118,102 @@ jobs:
...
+        if: steps.changes.outputs.stack == 'true'
+        run: |
+          pip install tox --upgrade
+
+      - name: Install Docker Compose
+        if: runner.os == 'Windows'
+        run: |
+          curl -L https://github.com/docker/compose/releases/download/v2.1.1/docker-compose-windows-x86_64.exe -o "C:\Program Files\Docker\cli-plugins\docker-compose.exe"
+
+      - name: Setup WSL
```
madhavajay, line 177: this wont work, we need to pre-provision WSL2 and desktop docker on these machines. Once its working we can always create an image so we can destroy and re-create the machine without the setup work. This is also super handy for quickly getting it setup:
https://raw.githubusercontent.com/madhavajay/PySyft/madhava/win_bootstrap/packages/hagrid/hagrid/win_bootstrap.py

bitsofsteve: Thanks @madhavajay ! I wasn't going to depend on it actually, I just wanted to split out the windows CI test so it could fail separately.

PR #6194, opened 2021-11-23
[WIP] Add DP tensor pointer

fiza11:


rasswanth-s:
@fiza11  Shall we close this, as we had discussed for single tensor pointer for all DP operations

PR #6201, opened 2021-11-27
Comparison tests disable

rasswanth-s:
## Description
Temporary comparison tests disable


iamtrask:
Why do we want to disable these tests? I'm very inclined to keep them given tests are what prevent us from regressing on functionality.

iamtrask:
I don't think we should disable these tests - we need comparison to keep working

PR #6202, opened 2021-11-28
Warns user that they might have run out of privacy budget

fiza11:


packages/syft/src/syft/core/node/common/client.py
```diff
@@ -238,6 +237,16 @@ def send_immediate_msg_with_reply(
                 error(str(exception))
                 traceback_and_raise(exception)
             else:
+                # response.message.budget is the privacy budget. We put a check here and raise a warning?
+                try:
+                    # third party
+                    from scipy.stats import shapiro
+
+                    shapiro_result = shapiro(response.message.data)  # type: ignore
```
iamtrask, line 245: This is far too low-level for this kind of test. All messages of all kinds pass through here. Move this to .get() on the differential privacy compatible pointer.

iamtrask (changes requested):

PR #6203, opened 2021-11-28
Network node improvements

iamtrask:
- [x] syft.networks only prints networks which are online
- [x] check to make sure Azure CLI is installed when running hagrid - pointer to installation instructions if not
- [x] check to make sure ansible is installed when running hagrid - pointer to installation instructions if not
- [x] network nodes auto-join their own VPN
- [x] https://app.shortcut.com/openmined/story/4077/core-content-connecting-to-a-network-is-an-example-but-the-user-cannot-actually-do-it-in-the-notebook-or-ui
- [x] https://app.shortcut.com/openmined/story/4190/create-prototype-of-network-node-endpoint-for-final-exam-grading-process
- [x] https://app.shortcut.com/openmined/story/4203/network-nodes-don-t-display-domain-nodes-which-are-inactive-by-default
- [x] https://app.shortcut.com/openmined/story/4202/modify-network-node-with-parameter-to-automatically-let-domain-nodes-join


packages/grid/backend/grid/core/node.py
```diff
@@ -15,10 +18,74 @@
...
 
+
+def thread_function(*args, **kwargs) -> None:  # type: ignore
+    # TODO: call this after the Network node is deployed instead of using a timer.
+    time.sleep(30)
+
+    # third party
+    from requests import get
+
+    ip = get("https://api.ipify.org").content.decode("utf8")
```
rasswanth-s, line 29: MAYBE: Andrew should we avoid depending on external services in our current ecosystem?

packages/grid/backend/grid/core/node.py
```diff
@@ -15,10 +18,74 @@
...
+
 if settings.NODE_TYPE.lower() == "domain":
     node = Domain("Domain", db_engine=get_db_engine(), settings=settings)
 elif settings.NODE_TYPE.lower() == "network":
     node = Network("Network", db_engine=get_db_engine(), settings=settings)
+    format = "%(asctime)s: %(message)s"
+    logging.basicConfig(format=format, level=logging.INFO, datefmt="%H:%M:%S")
+
+    logging.info("Main    : before creating thread")
+    x = threading.Thread(target=thread_function)
```
rasswanth-s, line 82: MAYBE: I think we could use green threads here as gevent monkey patching could be done at level, which does implicit context switches?

packages/grid/backend/grid/core/node.py
iamtrask: As a general policy, definitely. In this case it's an attempt to get a public IP as one of multiple ways the system tries to establish a connection, and I'm not aware of a way to get a public IP reliably that doesn't involve connecting to an external service somehow. Thoughts?

packages/syft/src/syft/__init__.py
```diff
@@ -96,7 +96,7 @@
 
 sys.path.append(str(Path(__file__)))
 
-logger.add(sink=sys.stderr, level="CRITICAL")
+logger.start()
```
rasswanth-s, line 99: Q: Why did change to start() ?

rasswanth-s: https://github.com/OpenMined/PySyft/blob/dev/packages/syft/src/syft/logger.py
Our custom  logger class does not have start method I think

rasswanth-s: Ignore this, The changes you made to logger file were down :sweat_smile:

rasswanth-s: did not see it before

iamtrask: i wanted to be able to pause logging temporarily - so i created start() and stop() methods

PR #6242, opened 2022-01-11
Ionesio's codebase cleanup 1

IonesioJunior:
## Description
Removing duplicated code from `packages/syft/src/syft/core/node/common/node_service/get_all_requests/get_all_requests_service.py`


packages/syft/src/syft/core/node/common/node_service/object_request/object_request_service.py
```diff
@@ -49,6 +47,8 @@
 from .object_request_messages import CreateRequestResponse
 from .object_request_messages import DeleteRequestMessage
 from .object_request_messages import DeleteRequestResponse
+from .object_request_messages import GetAllRequestsMessage
+from .object_request_messages import GetAllRequestsResponseMessage
 from .object_request_messages import GetBudgetRequestsMessage
```
rasswanth-s, line 52: MAYBE:
in Line 407 keep only a single if
`if not _can_triage_request`
as _requests variable is already assigned to `node.data_requests.all()` in L405

packages/syft/src/syft/core/node/domain/client.py
```diff
@@ -72,7 +72,7 @@ def __init__(self, client: Client) -> None:
     def requests(self) -> List[RequestMessage]:
 
         # relative
-        from ..common.node_service.get_all_requests.get_all_requests_messages import (
+        from ..common.node_service.object_request.object_request_messages import (
             GetAllRequestsMessage,
         )
```
rasswanth-s, line 78: Q: In L83,84 why do we serialize and de serialize  the same message?
can't we send it directly?
<pre>
<code>
blob = serialize(msg, to_bytes=True)
msg = deserialize(blob, from_bytes=True)
</pre>
</code>

rasswanth-s (approved):
LGTM

packages/syft/src/syft/core/node/common/node_service/object_request/object_request_service.py
IonesioJunior: Solved by  ace647c

packages/syft/src/syft/core/node/domain/client.py
IonesioJunior: Solved by 639ed60

IonesioJunior:
Thank you for the feedback and valuable observations @rasswanth-s !

PR #6334, opened 2022-02-28
Added script to purge celery worker queue 

rasswanth-s:
Added warning filter for numpy


madhavajay (approved):
Looks good.
Can we just rename the script to something like, "flush_queue.sh" or something?

PR #6348, opened 2022-03-09
Update Readme with latest documentation.

rasswanth-s:
## Description
Updated Readme.


shubham3121:
@rasswanth-s I guess we need to raise the pull request against `0.6.0` branch instead of dev.

rasswanth-s:
@shubham3121 , should I do for both (dev and 0.6.0) as dev Readme.md is in the front page of the rep.

shubham3121:
@rasswanth-s sure let's do it. Should we update the version to 0.7.0 beta

PR #6350, opened 2022-03-09
Stephen/benchmark_task

bitsofsteve:
## Description
Setting benchmark to run Rept on 1 million rows.


benchmarks/executor.py
```diff
@@ -1,11 +1,21 @@
...
+from typing import Optional
+from typing import Tuple
 
 # third party
-import numpy as np
 import pyperf
-from syft_benchmarks import run_rept_suite
-from syft_benchmarks import run_sept_suite
+from syft_benchmarks import run_ndept_suite
+from syft_benchmarks import run_rept_suite  # noqa: F401
```
rasswanth-s, line 13: I think we could remove sept/rept benchmark suite
as Ishan has a PR for removing sept,rept
https://github.com/OpenMined/PySyft/pull/6436
which would be merged in few and would break the benchmarks

bitsofsteve: Thanks @rasswanth-s , I've removed it from the executor, for now, I'll phase out rept/sept suite entirely later. I'm making them a legacy code now.🙂

madhavajay (approved):
Great work @bitsofsteve !!! ❤️

PR #6407, opened 2022-04-15
add hagrid version command

bitsofsteve:
## Description
This PR is to add a `hagrid version` command to Hagrid which prints the hagrid version to the cli.


packages/hagrid/hagrid/cli.py
```diff
@@ -21,6 +22,9 @@
 import click
 import rich
 
+# syft absolute
+import syft
```
madhavajay, line 26: I'm not sure this is a good idea to have syft imported directly into HAGrid. Even when we add the ability for HAGrid to manage syft environments for us we probably won't want to import syft itself.

packages/hagrid/hagrid/cli.py
```diff
@@ -1890,3 +1896,19 @@ def generate_sec_random_password(length: int, alphabet: str = DEFAULT_ALPHABET)
 
     # Python 3 (urandom returns bytes)
     return "".join(alphabet[c % len(alphabet)] for c in urandom(length))
+
+
+# add Hagrid info to the cli
```
madhavajay, line 1901: I think users expect to type something like `version` and in that case perhaps we can just print the hagrid version since all the other information is already available under `hagrid debug`.

packages/hagrid/hagrid/cli.py
```diff
@@ -67,6 +71,8 @@ def get_azure_image(short_name: str) -> str:
 
 
 @click.group(cls=RichGroup)
+# experimenting with click.option
```
madhavajay, line 74: Remove unused commented lines to keep our PRs to only include changes that are being used.

bitsofsteve: Okay, thanks Madhava!

packages/hagrid/hagrid/cli.py
bitsofsteve: @madhavajay I was actually thinking the same since we have `hagrid debug` happening, we could add anything there and keep `--version` to just print hagrid version. I'll do that.

packages/hagrid/hagrid/cli.py
bitsofsteve: @madhavajay, right also since we are dumping all other stuff on the `debug` command, I'll strip it off.

madhavajay (approved):
👍

madhavajay:
@bitsofsteve can you resolve this conflict and ill hit merge.

PR #6412, opened 2022-04-19
CH-5242

IonesioJunior:
## Description

## Affected Dependencies

## How has this been tested?

## Checklist


madhavajay (approved):
Great work figuring out how to get the VPN to connect to itself. In this case I think it might be worth hiding the internal details like port and host from the outside system and in a way that will work for other deployment types like kubernetes, so I think the best way is to create a specific service which knows what to do internally. To avoid the confusion of the same code running in two different contexts of a network or a domain we can just split it out. The Celery Beat that you and @shubham3121 works great and I think we should explore this more with internal healthchecks. The next obvious thing to me is to keep the headscale key and the url to the headscale server somewhere in the database and then use it re-connect without having to generate a new one. We can then make this service also work for domains, so if a domain is disconnected this service can reconnect it automatically.

PR #6430, opened 2022-04-28
added integration test for public hagrid docker

bitsofsteve:
## Description
This PR adds a CI test for the public deployment of Hagrid and Docker.

### DIFF
--- a/.github/workflows/stack-integration_tests.yml
+++ b/.github/workflows/stack-integration_tests.yml
@@ -97,14 +97,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Check for file changes
-        uses: dorny/paths-filter@v2
-        id: changes
-        with:
-          base: ${{ github.ref }}
-          token: ${{ github.token }}
-          filters: .github/file-filters.yml
-
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/stack-integration_tests.yml
```diff
@@ -85,6 +85,72 @@ jobs:
...
+      matrix:
+        os: [ubuntu-latest]
+        python-version: ["3.10"]
+
+    runs-on: ${{matrix.os}}
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Check for file changes
```
madhavajay, line 100: Lets not trigger it based on file changes, lets always run it and move it to nightlies.

.github/workflows/stack-integration_tests.yml
```diff
@@ -85,6 +85,72 @@ jobs:
         run: |
           tox -e stack.test.integration
 
+  stack-integration-tests-public:
+    strategy:
+      max-parallel: 3
+      matrix:
+        os: [ubuntu-latest]
```
madhavajay, line 92: Once this works can we enable it on all OSes?

madhavajay: Porting the bash stuff to windows might take some time we can possibly add this as a task for later.

madhavajay (approved):
Great work! This is going to be awesome for our users!

PR #6446, opened 2022-05-04
making public test run on nightly alone

bitsofsteve:
This PR aims to separate Stack Integration tests for public deployments to be in its own workflow and run only in nightlies as it makes sense since we do not deploy new syft & hagrid per PR or merge.


madhavajay (approved):
Perfect!

PR #6460, opened 2022-05-06
[W.I.P.] lazyrepeatarray refactor

IshanMi:
## Description
Currently we lose a lot of the benefits of the `lazyrepeatarray` with a single `matmul` or `dot` operation, because we're forced to convert them to their full sized NumPy arrays to perform those operations.

This PR refactors the LazyRepeatArray class, and gives it true lazy evaluation. The lazyrepeatarray shape gets modified after every new operation, and errors related to improper inputs/shapes are still detected, but evaluation of all the functions will be delayed until the user calls `.evaluate()`- which will happen during/just before the `publish` method.

## Affected Dependencies
This will primarily affect PhiTensor and GammaTensor.

## How has this been tested?
Working prototype is in a Jupyter Notebook
@fiza11 is working on benchmarking
unit tests to be added before merging 

## Checklist


packages/syft/src/syft/core/tensor/lazy_repeat_array.py
```diff
@@ -50,87 +60,255 @@ def __init__(self, data: np.ndarray, shape: Tuple[int, ...]) -> None:
...
+                self.add_op(function="np.multiply", args=other)
+            return self
 
-        if self.data.shape == other.data.shape:
-            return self.__class__(data=self.data * other.data, shape=self.shape)
+        else:
+            raise Exception(f"not sure how to do this yet: {type(other)}")
 
-        raise Exception("not sure how to do this yet")
+    def __matmul__(self, other: Any):  # type: ignore
```
IshanMi, line 175: Could we fix the type annotations for `other`, since we know it'll only be a numpy/jnp array, lazyrepeatarray, or integer/float/etc

packages/syft/src/syft/core/tensor/lazy_repeat_array.py
```diff
@@ -231,3 +373,18 @@ def transpose(self, *args: List[Any], **kwargs: Dict[str, Any]) -> lazyrepeatarr
         return lazyrepeatarray(
             data=self.data.transpose(*args, **kwargs), shape=dummy_res.shape
         )
+
+    def add_op(self, function: str, selection=slice(None), args=None) -> None:  # type: ignore
```
IshanMi, line 377: Could we fix the type annotations here by specifying args is Optional, etc?

packages/syft/src/syft/core/tensor/lazy_repeat_array.py
```diff
@@ -231,3 +373,18 @@ def transpose(self, *args: List[Any], **kwargs: Dict[str, Any]) -> lazyrepeatarr
...
+        self.transforms.append((function, selection, args))
+
+    def evaluate(self):  # type: ignore
```
IshanMi, line 380: Could we add a return type hint here so we can remove the `type: ignore`?

packages/syft/tests/syft/core/tensor/adp/gamma_tensor_test.py
```diff
@@ -10,6 +10,7 @@
 from syft.core.adp.data_subject_ledger import DataSubjectLedger
 from syft.core.adp.ledger_store import DictLedgerStore
 from syft.core.tensor.autodp.gamma_tensor import GammaTensor
+from syft.core.tensor.autodp.gamma_tensor import jax2numpy
 from syft.core.tensor.autodp.phi_tensor import PhiTensor as PT
```
IshanMi, line 15: Could we add more tests here to make sure that lazyrepeatarray minvals/maxvals were initialized properly when going from PhiTensor -> GammaTensor, and that evaluate still works when you're adding GammaTensors?

IshanMi (commented):
@fiza11 Great work! This is looking pretty good- just a few more changes and it should be good to go :+1:

packages/syft/src/syft/core/tensor/lazy_repeat_array.py
```diff
@@ -231,3 +373,18 @@ def transpose(self, *args: List[Any], **kwargs: Dict[str, Any]) -> lazyrepeatarr
...
+        result = np.ones(self._shape) * self.data
+        for func, selection, args in self.transforms:
+            if func == "np.matmul":
+                if isinstance(args, lazyrepeatarray):
+                    args = args.to_numpy().evaluate()
+                result = eval(func)(result[selection], args)
```
madhavajay, line 386: We should be really careful about using `eval` I think we should discuss another way.

packages/syft/src/syft/core/tensor/lazy_repeat_array.py
fiza11: Should we use Union everywhere or declare it inside a class and then import it?

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
```diff
@@ -1167,8 +1167,8 @@ def sum(
         return GammaTensor(
             value=np.array(self.child.child.sum()),
             data_subjects=self.data_subjects.sum(),
-            min_val=float(self.min_vals.sum(axis=None)),
-            max_val=float(self.max_vals.sum(axis=None)),
+            min_val=self.min_vals.sum(axis=None),
+            max_val=self.max_vals.sum(axis=None),
```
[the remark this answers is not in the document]
rasswanth-s: This is sufficient for now

packages/syft/src/syft/core/adp/vectorized_publish.py
```diff
@@ -17,6 +17,7 @@
 from tqdm import tqdm
```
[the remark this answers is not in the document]
rasswanth-s: Yep , we could remove it as this code is mostly executed inside the celery worker, and we would not have a visual interface for it from the DS side.

PR #6464, opened 2022-05-08
local user group and hagrid launch message

gigaray:
Support for user account management in Windows 10 Home Edition is not well explained on Microsoft pages too. So, adding those two lines of text (as a Note) will save time for those who are trying hagrid for the first time. 

Also, the hagrid launch error message is cryptic - just changing to 'domain' works.

## Description
Those notes will help save time for those Windows 10 users with Home edition, who are eager to get hagrid launch to happen successfully.

## Affected Dependencies
No dependencies

## How has this been tested?
I verified the steps -


madhavajay:
@gigaray thanks for the PR. We need to actually update those docs because on you can now install HAGrid into windows python and use it there for docker deployments without any wsl2. However due to `ansible` not being available on windows if you want to do deployments to the cloud you either need to use the hagrid docker container or wsl2.

PR #6546, opened 2022-05-27
Patch dep torchcsprng for PySyft 0.5.0 on Apple M1 chips 

leriomaggio:
## Description
The setup configuration file has been modified to allow the installation of `syft 0.5.0` on Apple with M1 chips.

While testing the change (i.e. _within an environment without neither `torchsnprng` nor `sympc` installed_), I also ended up adjusting the rules for a few packages in the env that caused some clashing leading to failures in all the `syft.test` env in `tox`  (more details below).

#### Dropping `torchsnprng` dependency

The change affects the `setup.cfg` file now including an extra configuration group for `torchcsprng`.
This group has been added to the `torch_ecosystem` requirements, and ignores the installation of `torchcsprng` if `platform_machine=='arm64'` (i.e. Apple M1)

#### Fixing Package versions

Some packages in  in `syft` and `dev` configuration led to version clashing when bluntly installing the latest versions of some packages (e.g. `nbconvert 6.5`, `black 22.0.3`).
Those clashing affected all the corresponding test env in `tox`, ultimately leading to errors in all the (`syft`) tests.

FYI, `tox` `syft.test.*` have been used to test the removal of `torchcsprng` from the deps. 

## Affected Dependencies

None, really. 
In fact, `torchcsprng` is currently used only in `SyMPC` which is only listed as optional requirement for `syft_0.5.0` (via `requirements.unstable.txt`), and not even included when installing `syft==0.5.0` directly from PyPi.

## How has this been tested?

Running and checking that all the `syft.test.*` in `tox` complete successfully. 

_Note_: I don't have Apple M1 chip computer, so I tested it the other way around, that is negating the condition in `setup.cfg` to have environments **without** `torchcsprng` installed.
 
_Take away_: With my surprise this was true also for `syft.test.libs`. 
Therefore, in this PR I also changed `tox.ini` getting rid of the `requirements.unstable.txt` dependency whenever not needed.

## Checklist
- [X ] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [X ] My changes are covered by tests


madhavajay:
@leriomaggio awesome work thank you.
In the end I just stripped a bunch of stuff out and managed to get some of the tests to still pass. The core of Syft is probably fine but some of the lib support or grid might be flakey.

PR #6573, opened 2022-06-03
Disable checking all domain health by default

iamtrask:
normally when we run network_client.domains it checks the health of all the domains before doing anything. as the number of domains increases (and we also discover some bugs/issues) this seems to be causing problems by default - disabling for now

### DIFF
--- a/packages/syft/src/syft/core/node/common/client_manager/dataset_api.py
+++ b/packages/syft/src/syft/core/node/common/client_manager/dataset_api.py
@@ -133,7 +133,7 @@ class DatasetRequestAPI(RequestAPI):
         result = [
             content
             for content in self.perform_api_request(
-                syft_msg=self._get_all_message
+                syft_msg=self._get_all_message, timeout=1
             ).metadatas
         ]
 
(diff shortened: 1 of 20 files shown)


packages/syft/src/syft/core/node/common/client_manager/domain_api.py
```diff
@@ -41,49 +40,41 @@ def all(self, pandas: bool = True) -> List[Any]:
...
+            #     # n = len(_data)
+            #     # data = list()
+            #     # args = [
+            #     #     (self, i, n, domain_metadata["id"])
+            #     #     for i, domain_metadata in enumerate(_data)
+            #     # ]
+
+            #     # # Check domain status sequentially
+            #     # for i, arg in enumerate(args):
+            #     #     if check_domain_status(*arg):
```
shubham3121, line 60: @madhavajay should we re-enable this, since all messages are sent with timeouts or at least have some sort of default timeout?

shubham3121 (approved):
Great stuff !!! 💯 🎉

packages/syft/src/syft/core/node/common/client_manager/dataset_api.py
```diff
@@ -133,7 +133,7 @@ def all(self) -> List[Any]:
         result = [
             content
             for content in self.perform_api_request(
-                syft_msg=self._get_all_message
+                syft_msg=self._get_all_message, timeout=1
```
shubham3121, line 136: Should we have a larger timeout? since it fetches all datasets so deserialization can take time if there are a large number of datasets uploaded.

PR #6595, opened 2022-06-13
[WIP]Fix Incorrect privacy budget spent

rasswanth-s:
## Description
Since , we use FPT , which increases the range of the input tensor, this inturn causes large rdp constants, leading to high privacy budget spent, the PR aims to decode FPT values right before publish.

## Affected Dependencies
`Vectorized Publish`

## How has this been tested?
Existing Tests


iamtrask:
I'm not 100% sure that this can be done without causing privacy leakage. Can you explain a bit more about why FPT "increases the range of the input tensor". Why should it increase the range? Do you mean it reduces the precision?

It's actually quite important that we don't use floating point precision for differential privacy. Let's not merge this change until we are sure we understand it. https://www.microsoft.com/en-us/research/wp-content/uploads/2012/10/lsbs.pdf

PR #6632, opened 2022-07-08
Importing syft is very slow at the moment - trying to make it faster

iamtrask:
## Description
Main goal of this PR is to start deleting or tuning things to try to make the syft import faster.


iamtrask:
closing because this is a fool's errand atm

PR #6634, opened 2022-07-12
Mean op DP tensors

curt-mitch:
## Description
This PR adds a DP-friendly mean method to Syft based on the [NumPy mean method](https://numpy.org/doc/stable/reference/generated/numpy.mean.html). In addition to adding a mean method to the Phi and Gamma tensor classes, this PR adds a parameter to the `lazyrepeatarray` constructor called `data_type` that allows us to declare whether the array is known to consist entirely of minimum or maximum values, which can save us from performing some expensive calculations.

## Affected Dependencies
N/A

## How has this been tested?
Tested locally in notebooks, added unit and E2E tests.

## Checklist


packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
```diff
@@ -578,6 +578,71 @@ def sum(
...
+            result=result,
+            self_obj=self,
+            args=[],
+            kwargs={},
+        )
+        dummy_res = np.empty(self.public_shape).mean(*args, **kwargs)
+        result.public_shape = dummy_res.shape
+        result.public_dtype = self.public_dtype
+
+        return result
```
IshanMi, line 644: Let's add an e2e test in `PySyft/tests/integration/e2e` to double check this works as intended :)

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
```diff
@@ -522,6 +522,23 @@ def sum(
...
         return self.gamma.sum(*args, **kwargs)
 
+    def mean(
+        self,
+        *args: Tuple[Any, ...],
+        **kwargs: Any,
+    ) -> Union[
+        TensorWrappedPhiTensorPointer, MPCTensor, TensorWrappedGammaTensorPointer
+    ]:
+        """Apply the "truediv" operation between "self" and "other"
```
IshanMi, line 532: Let's update this docstring to correctly reflect the operation name and args and such :)

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
```diff
@@ -2022,6 +2025,13 @@ def sum(
         #     max_vals=max_val,
         # )
 
+    def mean(
+        self,
+        axis: Optional[Union[int, Tuple[int, ...]]] = None,
+        **kwargs: Any,
+    ) -> Union[PhiTensor, GammaTensor]:
+        return self.gamma.mean(axis, **kwargs)
```
IshanMi, line 2033: The result should still be a PhiTensor

packages/syft/src/syft/core/tensor/lazy_repeat_array.py
```diff
@@ -262,6 +262,10 @@ def sum(self, *args: Tuple[Any, ...], **kwargs: Any) -> lazyrepeatarray:
         res = np.array(self.to_numpy().sum(*args, **kwargs))
         return lazyrepeatarray(data=res, shape=res.shape)
 
+    def mean(self, *args: Tuple[Any, ...], **kwargs: Any) -> lazyrepeatarray:
+        res = np.array(self.to_numpy().mean(*args, **kwargs))
+        return lazyrepeatarray(data=res, shape=res.shape)
```
IshanMi, line 267: We discussed this in our 1:1- but let's take advantage of the fact that we can predict ahead of time what the min_val and max_val of the mean would be :)

packages/syft/tests/syft/core/tensor/adp/phi_tensor_test.py
```diff
@@ -344,6 +344,36 @@ def test_sum(
...
+        max_vals=upper_bound,
+        min_vals=lower_bound,
+    )
+
+    tensor_mean = tensor.mean()
+    tensor_mean_axis_1 = tensor.mean(**kwargs)
+
+    assert (tensor_mean.child == reference_data.mean()).all()
+    assert (tensor_mean_axis_1.child == reference_data.mean(**kwargs)).all()
+    assert zeros_tensor.mean().child == 0
```
IshanMi, line 374: Nicely done! Let's also add a few assert statements to check that the metadata and shape is what we expect it to be, so that we can be sure that DP noise is added correctly :)

packages/syft/tests/syft/core/tensor/lazy_repeat_array_test.py
```diff
@@ -94,3 +94,10 @@ def test_sum() -> None:
     lazyarray = lazyrepeatarray(data=np.array([1]), shape=array.shape)
 
     assert lazyarray.sum(axis=None).data == array.sum(axis=None)
+
+
+def test_mean() -> None:
+    array = np.array([1, 2, 3])
+    lazyarray = lazyrepeatarray(data=np.array([1, 2, 3]), shape=array.shape)
```
IshanMi, line 101: lazyrepeatarray.data is usually just passed as a single value (like an integer)

packages/syft/tests/syft/core/tensor/adp/phi_tensor_test.py
```diff
@@ -344,6 +344,36 @@ def test_sum(
     assert zeros_tensor.sum().child == 0
 
 
+@pytest.mark.parametrize("kwargs", [{"axis": (1)}])
```
IshanMi, line 347: Brownie points for good use of pytest! :tada:

IshanMi (changes requested):
Great work @curt-mitch !! There are a few changes to make, as we discussed in our 1:1, but this was great effort :clap: 

Let's make these changes, and then add a unit test testing this op for gamma_tensor (similar to the one you made for phi_tensor in `phi_tensor_test.py`), as well as an end-to-end test that can test the TensorPointers and publish.

Terrific job, and great progress!

packages/syft/src/syft/core/adp/data_subject_list.py
```diff
@@ -157,6 +157,12 @@ def sum(self, target_shape: tuple) -> DataSubjectList:
             self.data_subjects_indexed.reshape((-1, *target_shape)),
         )
 
+    def mean(self) -> DataSubjectList:
+        return DataSubjectList(
+            one_hot_lookup=self.one_hot_lookup.copy(),
+            data_subjects_indexed=self.data_subjects_indexed.mean(),
```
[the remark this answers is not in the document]
curt-mitch: Got it, I've taken out the method since the class is unlikely to stick around.

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
curt-mitch: Done!

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
curt-mitch: D'oh!

packages/syft/src/syft/core/tensor/lazy_repeat_array.py
curt-mitch: I _think_ I've got this at least started down the right path, where we'd start passing new min/max properties in the constructor method. Let's chat more because I think I'm still missing something in how this should be implemented.

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
curt-mitch: Fixed!

packages/syft/tests/syft/core/tensor/adp/phi_tensor_test.py
curt-mitch: I added a few more `assert` checks, but let me know if there are other props I should be testing for.

packages/syft/tests/syft/core/tensor/lazy_repeat_array_test.py
curt-mitch: Corrected!

packages/syft/src/syft/core/tensor/lazy_repeat_array.py
```diff
@@ -37,7 +37,9 @@ class lazyrepeatarray:
 
     __attr_allowlist__ = ["data", "shape"]
 
-    def __init__(self, data: np.ndarray, shape: Tuple[int, ...]) -> None:
+    def __init__(
+        self, data: np.ndarray, shape: Tuple[int, ...], data_type: str = "none"
```
madhavajay, line 41: Perhaps we should extract these stringly typed values to some file level CONSTANTS or an Enum and if there is a use case without min / max, make it Optional where `None` represents no min or max flatten operation. If more operations vary in their treatment this could be moved further up the chain by creating a Min and Max subclass which implement the correct handling leaving the general case to the base class?

IshanMi: (Great catch!) I'm not sure it's worth the effort to make min and max subclasses- I don't see lazyrepeatarray being used for anything else besides storing min and max values in DP tensors, so it might be best to keep things simple for now?

I actually didn't know about Enums in Python- looks cool and I definitely think they can fit in here :D (based on my understanding from https://stackoverflow.com/questions/37601644/python-whats-the-enum-type-good-for)

curt-mitch: Good suggestion, @madhavajay. I've used enums in TypeScript (https://www.typescriptlang.org/docs/handbook/enums.html) but didn't even know they were available in Python!

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
```diff
@@ -1730,6 +1795,29 @@ def _sum(state: dict) -> jax.numpy.DeviceArray:
...
+    ) -> GammaTensor:
+        def _mean(state: dict) -> jax.numpy.DeviceArray:
+            return jnp.mean(self.run(state).child)  # type: ignore
+
+        output_state = dict()
+        output_state[self.id] = self
+
+        child = self.child.mean(axis=axis, **kwargs)
+
+        min_v = child.min()
```
IshanMi, line 1809: Avoid directly computing min or max on the result- since this leaks information without spending privacy budget
Make use of the fact that we know the min_vals and max_vals _before_ the mean method is applied, and those numbers are stored in `self.min_vals.data` and `self.max_vals.data`

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
```diff
@@ -2022,6 +2025,21 @@ def sum(
...
+        axis: Optional[Union[int, Tuple[int, ...]]] = None,
+        **kwargs: Any,
+    ) -> PhiTensor:
+        child = self.child.mean(axis=axis, **kwargs)
+        data_subjects = np.array(self.data_subjects.mean(axis=axis, **kwargs))
+
+        return PhiTensor(
+            child=child,
+            data_subjects=data_subjects,
+            min_vals=self.min_vals.mean(),
```
IshanMi, line 2039: Pass in args and kwargs so that the min_vals and max_vals have the same shape as child if the Data Scientist uses any args/kwargs

packages/syft/src/syft/core/tensor/lazy_repeat_array.py
```diff
@@ -66,6 +68,14 @@ def __init__(self, data: np.ndarray, shape: Tuple[int, ...]) -> None:
                 if val < 0:
                     raise ValueError(f"Invalid shape: {shape}")
 
+        if self.data.size > 1:
+            if data_type == "min_val":
+                self.data = data.min()
+            elif data_type == "max_val":
+                self.data = data.max()
+            # else:
+            #     raise NotImplementedError("What on earth are you using this class for?")
```
IshanMi, line 77: We should raise an exception here :)

packages/syft/tests/syft/core/tensor/adp/phi_tensor_test.py
```diff
@@ -344,6 +344,44 @@ def test_sum(
...
+def test_mean(
+    reference_data: np.ndarray,
+    upper_bound: np.ndarray,
+    lower_bound: np.ndarray,
+    ishan: DataSubjectArray,
+    kwargs: Dict,
+) -> None:
+    ishan = np.broadcast_to(ishan, reference_data.shape)
+    zeros_tensor = PT(
+        child=reference_data * 0,
```
IshanMi, line 357: Try using: np.zeros_like(reference_data)

packages/syft/tests/syft/core/tensor/adp/phi_tensor_test.py
IshanMi: I'd also check that the min_vals and max_vals have the same shape as the child- but that's all :)

tests/integration/e2e/smpc_dp_operation_test.py
```diff
@@ -315,6 +315,54 @@ def test_gt(
     print("Published Result ", published_result.get())
 
 
+@pytest.mark.e2e
+def test_mean(
```
IshanMi, line 319: Let's split this up and make two separate tests for PhiTensor mean and GammaTensor mean. The reason this is important is because publishing the result of PhiTensor.mean() is expected to take a LOT more privacy budget than GammaTensor.mean() (because there's only 1 data subject in PhiTensor- meaning their data has nowhere to hide). 
To accomodate for this, we could use a higher value of sigma for the PhiTensor test.

Lmk if any of this doesn't make sense- happy to talk it over :)

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
IshanMi: Everything else is spot on though- in particular, well done with the `_mean()` method!

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
```diff
@@ -522,6 +522,23 @@ def sum(
...
+    ]:
+        """Apply the "mean" operation to "self"
+
+        Args:
+            y Tuple[Any, ...] : second operand.
+
+        Returns:
+            Union[TensorWrappedPhiTensorPointer,MPCTensor] : Result of the operation.
+        """
+        return self.gamma.mean(*args, **kwargs)
```
IshanMi, line 540: When calling `mean()` on a PhiTensor, the output should still be a PhiTensor, since no new data subjects were introduced.

packages/syft/src/syft/core/tensor/lazy_repeat_array.py
```diff
@@ -262,6 +272,19 @@ def sum(self, *args: Tuple[Any, ...], **kwargs: Any) -> lazyrepeatarray:
...
+        data_type = "none"
+        if res.size > 0:
+            all_min_val = np.all(res == res.min())
+            all_max_val = np.all(res == res.max())
+
+        if all_min_val:
+            data_type = "min_val"
+        elif all_max_val:
+            data_type = "max_val"
+        return lazyrepeatarray(data=res, shape=res.shape, data_type=data_type)
```
IshanMi, line 286: I believe we discussed this during our 1:1- let's make use of the fact that we can predict the output ahead of time, so we can avoid the potentially expensive `self.to_numpy()` operation.

IshanMi (changes requested):
Great progress @curt-mitch !! You're almost done here, just small changes to make (probably less than 5-6 lines of code).
You should be proud- the DP system isn't the easiest to work with, and you've done amazingly. Kudos! :tada: :taco:

packages/syft/src/syft/core/adp/data_subject_list.py
```diff
@@ -584,6 +590,9 @@ def var(self, *args: List[Any], **kwargs: Dict[Any, Any]) -> DataSubjectArray:
     def sqrt(self, *args: List[Any], **kwargs: Dict[Any, Any]) -> DataSubjectArray:
         return DataSubjectArray(self.data_subjects)
 
+    def mean(self) -> DataSubjectArray:
+        return DataSubjectArray(self.data_subjects)
```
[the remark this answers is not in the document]
curt-mitch: @IshanMi It doesn't look like the `DataSubjectArray` class has a `sum` method yet. Should this be something more like the `var` method? https://github.com/OpenMined/PySyft/blob/106a5ba193f3fc39594d3ddba4c1fd8dfbb5a84c/packages/syft/src/syft/core/adp/data_subject_list.py#L581-L582

packages/syft/src/syft/core/tensor/lazy_repeat_array.py
curt-mitch: @madhavajay @IshanMi I've added a TODO to create an enum in a separate PR for now.

packages/syft/tests/syft/core/tensor/adp/phi_tensor_test.py
curt-mitch: Added!

tests/integration/e2e/smpc_dp_operation_test.py
curt-mitch: I've split the DP mean method off into its own test, but lets review to make sure I'm setting it up as expected.

packages/syft/src/syft/core/adp/data_subject_list.py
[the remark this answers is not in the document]
rasswanth-s: MAYBE: We could remove mean method,The mean method would work automatically, like  since we covered the basic operations of mean  (addition and division).calling np.mean would invoke add and div automatically

Ex:

<pre>
<code>
from syft.core.adp.data_subject_list import DataSubjectArray

natsu = DataSubjectArray(["natsu"])
zerek = DataSubjectArray(["zerek"])

data_subjects = np.array([natsu,zerek])

res = data_subject.mean() or np.mean(data_subjects)
print(res)
</code>
</pre>

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
```diff
@@ -580,6 +580,71 @@ def sum(
...
+
+        Returns:
+            Union[TensorWrappedGammaTensorPointer,MPCTensor] : Result of the operation.
+        """
+        attr_path_and_name = "syft.core.tensor.tensor.Tensor.mean"
+        min_vals = self.min_vals.mean(*args, **kwargs)
+        max_vals = self.max_vals.mean(*args, **kwargs)
+
+        result = TensorWrappedGammaTensorPointer(
+            data_subjects=self.data_subjects,
```
rasswanth-s, line 601: MAYBE: Could we call the mean operation on the datasubjects also?

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
```diff
@@ -1363,8 +1428,12 @@ def _sub(state: dict) -> jax.numpy.DeviceArray:
             max_max = self.max_vals.data - other.max_vals.data
             _min_val = np.minimum.reduce([min_min, min_max, max_min, max_max])
             _max_val = np.maximum.reduce([min_min, min_max, max_min, max_max])
-            min_val = lazyrepeatarray(data=_min_val, shape=self.shape)
-            max_val = lazyrepeatarray(data=_max_val, shape=self.shape)
+            min_val = lazyrepeatarray(
+                data=_min_val, shape=self.shape, data_type="min_val"
```
rasswanth-s, line 1432: Q: Interesting ,Any semantic reason behind the data_type attribute?

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
rasswanth-s: @IshanMi  maybe a scenario like if the individual data points belonged to different data subjects, I think at that point it would be become a gamma tensor?

Ex:
<pre><code>
data_subjects = np.array( [DSA("GrandMaster"),DSA("Ishan")])

data_subjects.mean() would be 
np.array(DSA({"GrandMaster","Ishan"}))
</pre></code>

I think we could do like
<pre><code>
if len(self.data_sujects.sum()) ==1:  # to check the uniqueness.
      #create phiTensor pointer
else:
      create gamma tensor pointer.


</pre></code>

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
```diff
@@ -2106,6 +2190,25 @@ def sum(
         #     max_vals=max_val,
         # )
 
+    def mean(
+        self,
+        axis: Optional[Union[int, Tuple[int, ...]]] = None,
+        **kwargs: Any,
+    ) -> PhiTensor:
+        child = self.child.mean(axis=axis, **kwargs)
+        data_subjects = np.array(self.data_subjects.mean(axis=axis, **kwargs))
```
rasswanth-s, line 2199: MAYBE: We could do 
`data_subjects=self.data_subjects.mean(axis, **kwargs)`
as data_subjects we visualize as a numpy array with basic operation needed to perform mean.

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
```diff
@@ -1308,8 +1345,12 @@ def max(self, axis: Optional[Union[int, Tuple[int, ...]]] = None) -> PhiTensor:
         return PhiTensor(
             child=result,
             data_subjects=self.data_subjects[indices],
-            min_vals=lazyrepeatarray(data=result.min(), shape=result.shape),
-            max_vals=lazyrepeatarray(data=result.max(), shape=result.shape),
+            min_vals=lazyrepeatarray(
+                data=self.min_vals.data, shape=result.shape, data_type="min_val"
```
rasswanth-s, line 1349: NITPICK: I think we should include .min() on the data from previous code , same comment for the below max.

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
```diff
@@ -1318,8 +1359,12 @@ def min(self, axis: Optional[Union[int, Tuple[int, ...]]] = None) -> PhiTensor:
...
```
rasswanth-s, line 1363: NITPICK:  I think we should include .min() on the data from previous code , same comment for the below max.

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
rasswanth-s: `min_vals=lazyrepeatarray(data=self.min_vals.data.min(), shape=result.shape, data_type="min_val")`

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
```diff
@@ -2106,6 +2190,25 @@ def sum(
...
+        **kwargs: Any,
+    ) -> PhiTensor:
+        child = self.child.mean(axis=axis, **kwargs)
+        data_subjects = np.array(self.data_subjects.mean(axis=axis, **kwargs))
+
+        return PhiTensor(
+            child=child,
+            data_subjects=data_subjects,
+            min_vals=lazyrepeatarray(
+                data=self.min_vals.data, shape=child.shape, data_type="min_val"
```
rasswanth-s, line 2205: NITPICK: We could do the mean operation on the lazyrepeatarray also.

rasswanth-s: SUGGESTION : Maybe also we could move all the lazyrepeatarray computations to that particular class itself, to maintain modularity like
min_vals = self.min_vals.mean(*args,**kwargs) maybe now or for a future PR.

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
IshanMi: Great catch Rasswanth!

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
IshanMi: The idea was that if `lazyrepeatarray.data` ever becomes a multidimensional array instead of a single value, we could check if it is a min_val or max_val lazyrepeatarray, and collapse the array down to a single value accordingly by calling max or min

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
IshanMi: We will definitely have a PR to improve the lazyrepeatarray soon- adding Enums, and using some of the ideas from [here](https://github.com/OpenMined/PySyft/pull/6460) like lazy evaluation (in a safe way).

IshanMi: However I do think we should add some comments here in order to make this clear to people in the future

IshanMi: Another benefit in this case of not implementing a mean function directly in LazyRepeatArray, but rather creating the object in Phi/GammaTensor is that if we tried to do the former, we would have to call the expensive `to_numpy()` function in order to determine the shape of the output lazyrepeatarray. However, when implemented in the DP Tensor method, we could directly pass on the shape of self.child.mean() as the shape of the lazyrepeatarray, and avoid calling `to_numpy()`.

IshanMi (approved):

IshanMi:
`test_any()` and `test_all()` were flopping intermittently because once in a blue moon, `condition = np.random.choice(a=[False, True], size=(reference_data.shape[0])` would return a list of only `False` and this would fail the assertion check. I added a fix for this, which was to manually replace the last value in the list with a `True` so that there's always at least 1 element being tested.

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
rasswanth-s: Nice Explanation @IshanMi

packages/syft/src/syft/core/adp/vectorized_publish.py
```diff
@@ -377,19 +376,26 @@ def publish(
...
-def can_reduce_further(value: ArrayLike, zeros_like: ArrayLike) -> bool:
+def can_reduce_further(value: ArrayLike) -> bool:
     try:
-        result = value != zeros_like
+        result = value != 0
         # check we can call any or iterate on this value otherwise exit loop
         # numpy scalar types like np.bool_ are Iterable
         if not hasattr(result, "any") and not isinstance(result, Iterable):
-            return False
+            if result is True:
```
rasswanth-s, line 385: NIT:
<pre><code> 
      if isinstance(result,bool):
             return result
      else:
             raise Exception
</code></pre>

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
rasswanth-s: Reminder before Merge :)

tests/integration/e2e/dp_operation_test.py
```diff
@@ -0,0 +1,89 @@
...
+    # Check that datasets are visible
+    assert len(domain.datasets) > 0
+
+    # Check PB is available
+    assert domain.privacy_budget > 100
+    prev_domain_budget = domain.privacy_budget
+
+    domain_data = domain.datasets[-1]["data"]
+
+    result = domain_data.mean()
```
rasswanth-s, line 81: MAYBE:  Could we also add a test for TensorWrappedGammaTensorPointer?
We could modify the upload process in conftest.py to upload gamma tensor which would give us a gamma tensor pointer during indexing.

https://github.com/OpenMined/PySyft/blob/6b0bf2efdbd2bbf31e6012423c7d2fedf410e80f/tests/integration/conftest.py#L87

rasswanth-s (approved):
Well Done @curt-mitch  and @IshanMi  🎸

IshanMi:
Closing this PR- I had copied your implementation of the mean op into the working branch we were using to round out the 0.7 DS API since it was easier than having to solve all of these merge conflicts- some of which involve older implementations of operations which were insecure.

Terrific work @curt-mitch !!

PR #6674, opened 2022-08-14
[WIP] Serde refactor

tudorcebere:
Current progress:
* most of syft object are backed by recursive serde - exceptions: External types (np, torch) and primitives(int, bool)

Next steps:
* create a type registry to know what types we can handle
* remove .serialize(). /_object2whatever methods, keep a functional API
* remove GenerateWrapper (use the type bank)
* Decouple the need of using our own custom types (SyBool/SyInt, etc)
* Put a nail in the heart of DataMessage

Don't even try to review it for now, DM on slack if we want to discuss different aspects.

Knows issues
* json(...) not supported (will break plenty of things)
* no automated testing


teo-milea:
The Romanian Reaper is on the move 💀🌶️

PR #6681, opened 2022-08-18
Serde refactoring

tudorcebere:
Current progress:

most of syft object are backed by recursive serde - exceptions: External types (np, torch) and primitives(int, bool)
Next steps:

create a type registry to know what types we can handle
remove .serialize(). /_object2whatever methods, keep a functional API
remove GenerateWrapper (use the type bank)
Decouple the need of using our own custom types (SyBool/SyInt, etc)
Put a nail in the heart of DataMessage
Don't even try to review it for now, DM on slack if we want to discuss different aspects.

Knows issues

json(...) not supported (will break plenty of things)
no automated testing


packages/syft/src/syft/core/common/serde/recursive_primitives.py
```diff
@@ -0,0 +1,131 @@
...
+)
+
+recursive_serde_register(
+    OrderedDict,
+    serialize=serialze_kv,
+    deserialize=functools.partial(deserialize_kv, OrderedDict),
+)
+
+recursive_serde_register(
+    type(None), serialize=lambda _: b"1", deserialize=lambda _: None
```
madhavajay, line 99: Would this not be safer to be a 0 "just in case"

madhavajay:
omg omg omg its green ✅

madhavajay (approved):
This is one of those historical PRs!

The High Council will be pleased! ⚔️

PR #6684, opened 2022-08-24
Simplify concatenate smpc

gmuraru:
## Description
Simplify the concatenate logic.
@rasswanth-s  @iamtrask  I am thinking that we should might remove concatenate as a method from our tensors because if we stick with the numpy API - concatenate is a "static" method from the "numpy" module level --> Probably we should have an analogue for "np.concatenate" something like "syft.concatenate" (or "syft.tensor.concatenate")

## Affected Dependencies

## How has this been tested?
- In a colab

## Checklist


packages/syft/src/syft/core/tensor/tensor.py
```diff
@@ -212,26 +211,19 @@ def _apply_op(
...
+            func = getattr(self_mpc, op_str)
+            return func(other_mpc)
         elif isinstance(other, MPCTensor):
-            if op_str != "concatenate":
-                op = getattr(operator, op_str)
-                return op(other, self)
-            else:
-                return other.concatenate(self)
+            # "self" should be secretly shared
+            other_mpc, self_mpc = MPCTensor.sanity_checks(other, self)
```
rasswanth-s, line 224: GOOD Catch...

rasswanth-s (approved):

PR #6718, opened 2022-09-11
Tensor ops logic functions

teo-milea:
## Description
This PR adds pos, neg, and, or, all and any for Phi tensors and Gamma tensors

## Affected Dependencies
-

## How has this been tested?
I have added custom tests for each function

## Checklist


iamtrask:
Haven't reviewed yet - but I appreciate how clear and organized you are in your commit messages

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
```diff
@@ -2381,6 +2382,188 @@ def _squeeze(state: dict) -> jax.numpy.DeviceArray:
...
 
+    def any(
+        self,
+        axis: Optional[Union[int, Tuple[int, ...]]] = None,
+        keepdims: Optional[bool] = False,
+        where: Optional[ArrayLike] = None,
+    ) -> GammaTensor:
+        func = "any"
+
+        def _any(state: dict) -> jax.numpy.DeviceArray:
```
iamtrask, line 2393: Mild performance improvement: this method doesn't appear to have state (aside from the argument called "state"). Perhaps we could move it outside of the any() function so that we don't pay the cost of creating this method whenever we'd like to call any()?

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
```diff
@@ -2381,6 +2382,188 @@ def _squeeze(state: dict) -> jax.numpy.DeviceArray:
...
+                *[
+                    i.reconstruct() if isinstance(i, GammaTensor) else i
+                    for i in state.values()
+                ]
+            )
+
+        output_state = dict()
+        output_state[self.id] = self
+
+        mapper[func] = _any
```
iamtrask, line 2404: If we move _any() out as well, then we don't have to set and re-set this lookup table. We could stead just create this when syft is imported.

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
```diff
@@ -2381,6 +2382,188 @@ def _squeeze(state: dict) -> jax.numpy.DeviceArray:
...
+            out_child = np.array(
+                self.child.any(axis=axis, keepdims=keepdims, where=where)
+            )
+
+        new_data_subjects = DataSubjectArray.from_objs(np.empty(out_child.shape))
+
+        return GammaTensor(
+            child=out_child,
+            data_subjects=new_data_subjects,
+            min_vals=lazyrepeatarray(data=False, shape=out_child.shape),
```
iamtrask, line 2418: It appears that we might be accidentally adding a new type of tensor in here, yes? Have we created a boolean tensor type? I suspect this will create type errors somewhere although python can do some smart casting). In any case, I suggest we instead set min_vals and max_vals to 0 and 1.

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
iamtrask: (ditto for other methods)

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
iamtrask: (ditto for other methods)

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
iamtrask: (ditto for other methods)

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
```diff
@@ -1087,11 +1088,70 @@ def copy(self, order: Optional[str] = "K") -> PhiTensor:
...
+        # TODO: properly define data subjects and
+        # figure out if it is not a privacy violation to return bool
+        if where is None:
+            out_child = np.array(self.child.any(axis=axis, keepdims=keepdims))
+        else:
+            out_child = np.array(
+                self.child.any(axis=axis, keepdims=keepdims, where=where)
+            )
+
+        new_data_subjects = DataSubjectArray.from_objs(np.empty(out_child.shape))
```
iamtrask, line 1106: I'm curious as to why we wouldn't be sourcing data subjects from "self" here in some way. Surely "self" could be referring to a particular data subject, in which case the new_data_subjects variable would need to fetch from it, yes?

iamtrask: (ditto for other methods)

packages/syft/tests/syft/core/tensor/adp/gamma_tensor_test.py
```diff
@@ -821,3 +821,159 @@ def test_squeeze(
...
+        max_vals=upper_bound,
+        min_vals=lower_bound,
+    ).gamma
+
+    neg_tensor = reference_tensor.__neg__()
+
+    assert neg_tensor.func_str == "neg"
+    assert reference_tensor == neg_tensor.sources[reference_tensor.id]
+    assert (neg_tensor.child == reference_tensor.child * -1).all()
+    assert (neg_tensor.min_vals == reference_tensor.max_vals * -1).all()
```
iamtrask, line 869: well done on these

iamtrask (changes requested):
Great work! And well done jumping into a scientifically advanced section of the codebase.

Not a comprehensive review necessarily - but noted a few things that popped out to me.

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
teo-milea: well, bool is a subclass of int so the only issue that might appear would be from using type() on the elements of the lazyrepeatarray, but I will change it to be consistent

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
teo-milea: the main issue was that I didn't know to whom to "assign" the new values from the data subjects as the shape of the result could be very different from the shape of the data subjects. I guess we could reduce over the right axis with the add operation on the data subjects, At the same time, I didn't know if we need to implement any and all like that or if we consider the release of the result themselves to be a privacy violation. The main problem is that there are other sections of the code where we use all and any without the privacy wrapper, I guess maybe we can change it only to use them on the child values

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
teo-milea: if we dont need different objects, yeah definitely sounds like an improvement

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
teo-milea: yep sounds good

packages/syft/src/syft/core/tensor/autodp/phi_tensor.py
teo-milea: fixed

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
teo-milea: fixed

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
teo-milea: fix

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
teo-milea: fix

packages/syft/tests/syft/core/tensor/adp/gamma_tensor_test.py
teo-milea: <3

teo-milea:
TODO: discuss with Ishan how to reduce the number of data subjects on certain operations.

packages/syft/src/syft/core/adp/data_subject.py
```diff
@@ -149,12 +159,21 @@ def from_string(blob: str) -> DataSubjectGroup:
         return DataSubjectGroup(entity_set)  # type: ignore
 
     def __add__(
-        self, other: Union[DataSubjectGroup, DataSubject, int, float]
+        self,
```
rasswanth-s, line 162: Q: I think we this DataSubject Class is deprecated or are we using it again?

madhavajay: Yes lets address in a seperate PR.

PR #6751, opened 2022-09-14
update hagrid install instruction 

### LINKED ISSUE #6724: Update hagrid install README instructions
We should change `pip install hagrid` to `pip install -U hagrid` to ensure users get an updated version at all times.

Check the repo for areas where this is written and replace them.

bitsofsteve:
resolves #6724


shubham3121 (approved):

iamtrask (changes requested):
I'm not sure it's a requirement but conventionally I believe the "-U" goes after the end of the command (so after the word "hagrid")

bitsofsteve:
> I'm not sure it's a requirement but conventionally I believe the "-U" goes after the end of the command (so after the word "hagrid")

@iamtrask oh good to know, but looking at the documentation from pip here: https://pip.pypa.io/en/stable/installation/#upgrading-pip and how it's used it seems `pip install -U SomePackage` is also pretty standard.

madhavajay (approved):
Great work! 🙌🏽

PR #6752, opened 2022-09-14
 fixes #6736

### LINKED ISSUE #6736: Change HAGrid check to support Gitpod
hagrid check needs to detect gitpod and output a gitpod friendly url. This works in the quickstart jupyter output so just copy the same code which tests for gitpod and calculates the correct url.

bitsofsteve:
## Description
fixes #6736


madhavajay:
@bitsofsteve Great start. This needs a few fixes and testing on gitpod.

shubham3121:
@bitsofsteve I have updated the PR to use gitpod url only for display at hagrid check, for checking we can still continue to use localhost.

madhavajay:
Final fix, to make the url not wrap. 🙌🏽

PR #6794, opened 2022-09-29
Fix Linux Docker instructions

IonesioJunior:
## Checklist
- [x] Check if the user is root and warns them that using hagrid in root might cause issues if they run it later without root.
- [x] Check if the user they are using is in the _"sudo"_ group in the dependency checker on Linux to make sure they have permission ahead of time.
- [x] Check if they are in the docker group and if not, ask them to add themselves. Here we will need to test/make sure that either hagrid (through subprocess shells) or if there is some way we can restart the session so that we can access docker from their user without logging out and logging back in again.
- [x] Improve the error message of the docker service check to suggest that the docker should be installed or started. If we can possibly detect if the binary exists then say, start docker, and if the binary does not exist, print the OS-specific install instructions from the deps system.


IonesioJunior:
**Error message when docker binary is not found**
docker_uninstalled


**Error message when docker binary is found but the current user is not allowed to execute docker commands (Linux)**
user_permission_error

packages/hagrid/hagrid/deps.py
```diff
@@ -479,17 +484,55 @@ def check_docker_version() -> Optional[str]:
     return version
 
 
-def docker_running() -> bool:
+def docker_running() -> Tuple[bool, str]:
     try:
-        cmd = "docker info"
+        cmd = "docker --version"
```
shubham3121, line 489: @IonesioJunior we want to use `docker info` as it checks if docker is running or not on windows and mac and starts docker incase of ubuntu. `docker --version` doesn't detect that. 

I think we need to fix the order of checking sudo and docker running status. First check docker has sudo status and then check docker running status.

packages/hagrid/hagrid/deps.py
```diff
@@ -500,16 +543,23 @@ def check_docker_service_status(animated: bool = True) -> None:
...
+        user_allowed, permission_msg = allowed_to_run_docker()
     else:
         console = Console()
         # putting \t at the end seems to prevent weird chars getting outputted
         # during animations in the juypter notebook
         with console.status("[bold blue]Checking for Docker Service[/bold blue]\t"):
-            if not docker_running():
-                raise MissingDependency("❌ Docker service is not running.")
+            docker_installed, msg = docker_running()
+            user_allowed, permission_msg = allowed_to_run_docker()
```
shubham3121, line 554: here we can swap the order of these method calls. 
So first call `allowed_to_run_docker()` and then call `docker_running` method.

shubham3121 (changes requested):

shubham3121 (approved):
This is fantastic 🙌🏾 
Great work @IonesioJunior !! 🎉

PR #6828, opened 2022-10-07
Tensor ops abs arg min max

teo-milea:
## Description
Added abs, argmax and argmin from numpy to our tensors, according to the numpy reference.

## Affected Dependencies

## How has this been tested?
I have also added a test for each function.

## Checklist


packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
```diff
@@ -1994,6 +1994,88 @@ def __le__(self, other: Any) -> GammaTensor:
             sources=output_state,
         )
 
+    def __abs__(self) -> GammaTensor:
+
+        output_state = dict()
+        output_state[self.id] = self
+
+        data = self.child
+        child = np.abs(data)
```
rasswanth-s, line 2003: I think we could call data.__abs__() itself, because during SMPC the child is a FixedPrecisionTensor instead of Numpy array.

rasswanth-s (approved):
Amazing, Well Done @teo-milea  🚀

PR #6831, opened 2022-10-10
Refactoring Publish to use Worst Case Bound

rasswanth-s:
## Description
During DP-SMPC Publish operation, since we tend to work with Random ShareValues, which are in high orders of magnitude, which in turn exceeds the cache size during publish returning erroneous results.

The PR aims to modify Publish Operation to use max_values during Publish as the worst case bound, which is closer to the actual raw values.

- [x]  Works with Sufficient budget Scenario.
- [x]  Works with Insufficient Budget Scenario.

## Affected Dependencies
None

## How has this been tested?
Existing SMPC-DP Trade Demo Tests.

## Checklist


IshanMi:
There's an easier way to do this private budget modification- on [this line](https://github.com/OpenMined/PySyft/blob/45827b78b517b1a0072c02355b829e152592a764/packages/syft/src/syft/core/adp/vectorized_publish.py#L159) of vectorized_publish, simply replace `private=True` with `private=False`

packages/syft/src/syft/core/adp/vectorized_publish.py
```diff
@@ -143,6 +143,7 @@ def publish(
             lipschitz_bounds = coeffs.copy()
         else:
             lipschitz_bounds = tensor.lipschitz_bound
+            # ISHAN: Do we need to modify lipschitz_bound to work with max_val_array?
```
IshanMi, line 146: cc: I'm not quite sure- @iamtrask ?

IshanMi:
@iamtrask does the lipschitz bound need to be modified if we're using worst case bounds instead of real private data?

shubham3121 (approved):
This is amazing 🎉 
Great work @rasswanth-s !!

IshanMi (approved):

Issue #6841, opened 2022-10-11
Improve DP Tensor UI/UX


iamtrask:
```
On the following page https://openmined.github.io/PySyft/guides/data-owner/01-upload-data.html

we describe how to annotate with DP metadata using code like this:

In:

# run this cell
data_subjects = sy.DataSubjectArray.from_objs(dataset["ID"])

age_data = sy.Tensor(dataset["Age"]).annotated_with_dp_metadata(
   min_val=0, max_val=100, data_subjects=data_subjects
)
```

Instead let's do something like:
```
age_data = sy.Tensor([40, 39, 9, 8]).annotate_with_dp_metadata(
   min_val=0, max_val=100, data_subjects=["Ishan", "Shubham", "Irina", "Osam"] 
)
```

This maximizes the intuitive understanding of what's going on. This has several changes:

- Instead of using sy.DataSubjectarray.from_objs() to annotate using a list of IDs, we pass in a simple, intuitive list of strings.
- "annotated_with_dp_metadata" ->"annotate_with_dp_metadata"
- Instead of first creating a DataFrame we just create a simple list of integers. Note that these could still be displayed in their own table in Step 3 just for intuition's sake.
- Instead of IDs being integers they're the names of people.

shubham3121:
Fixed by: https://github.com/OpenMined/PySyft/pull/6845

rasswanth-s:
I am closing long standing Github Issues related to older versions of PySyft (<=0.7) and are not aligned in our current roadmap.Kindly re-open the issue if the needs to be addressed or is still relevant.

PR #6845, opened 2022-10-12
Remove need for user to provide DataSubjectArray for Syft Tensor

IshanMi:
## Description
To fix an issue raised in #6841 where someone using PySyft currently needs to either import the DataSubjectArray class (how would they know to do this), or provide a list with dimensions matching their tensor (too much effort required on the user's end).

This PR allows the user to provide a List of strings instead, and infers whether they're providing one data subject for the entire Tensor, one per row, or one per every data point.

i.e. a workflow like this should work:

```
import syft as sy
import numpy as np

data = np.random.randint(low=0, high=10, size=(5, 5))
people = ["Joe", "Mary", "Susan", "Albus", "Bob"]
tensor = sy.Tensor(data).private(min_val=0, max_val=10, data_subjects=people)

``` 

Note: There are plans downstream to support one Data Subject per Tensor (akin to old SEPT)  or one per image/array (akin to REPT), and not needing a DataSubjectArray to be created in those cases; this PR doesn't attempt to start those.

## Affected Dependencies
No dependencies changed, but docs can be changed :)

## How has this been tested?

Tested locally in a Jupyter Notebook here: [https://github.com/OpenMined/PySyft/blob/data_subject_creation_ux/notebooks/Experimental/Ishan/Test_No_DSA.ipynb](url)

## Checklist


iamtrask:
Small thing - any chance we can also add the ability to pass in a single string? (which would be interpreted as passing in a list of length 1)

IshanMi:
> Small thing - any chance we can also add the ability to pass in a single string? (which would be interpreted as passing in a list of length 1)

I believe that exists already: https://github.com/OpenMined/PySyft/blob/6f381d92514e71108d48fa0db4dd03c4a54c6590/packages/syft/src/syft/core/tensor/ancestors.py#L416

packages/syft/src/syft/core/tensor/ancestors.py
```diff
@@ -443,9 +443,24 @@ def _private(
...
-            )
+            if data_subjects.size == 1:
+                # 1 data subject for the entire tensor.
+                data_subjects = np.broadcast_to(data_subjects, self.shape)
+            if (
+                data_subjects.shape[0] == self.shape[0]
+                and len(data_subjects.shape) == 1
+            ):
+                # e.g. 1 data subject per image for 10 imgs: child = (10, 25, 25) and data_subjects = (10,)
+                data_subjects = np.broadcast_to(data_subjects, self.shape)
```
shubham3121, line 454: @IshanMi this broadcast operation will fail because data_subjects is a 1-D array. We will have to expand the data_subjects to correct the dimensions before performing the broadcast operation. Something like this should work:
```
# no of axes in the data
no_of_axes = len(self.shape)
axes_to_expand_along = list(range(no_of_axes))[1:]

# expand the data subject array to have the same dimensions as the data
data_subjects = np.expand_dims(data_subjects, axis=axes_to_expand_along)

# now we can perform the broadcast operation
data_subjects = np.broadcast_to(data_subjects, self.shape)
```

shubham3121 (changes requested):

packages/syft/src/syft/core/tensor/ancestors.py
```diff
@@ -443,9 +443,24 @@ def _private(
...
+            else:
+                raise Exception(
+                    "Data Subject shape doesn't match the shape of your data. Please provide either 1 data subject per "
+                    "data point, or 1 data subject per "
```
shubham3121, line 458: Incomplete sentence: `... or 1 data subject per`

rasswanth-s:
@IshanMi , would this PR functionality be part of the 0.7 release or for future releases?

iamtrask:
> @IshanMi , would this PR functionality be part of the 0.7 release or for future releases?

Let's plan to include it in 0.7 if possible. It's a significant improvement in UX for noobs which will likely be the primary users of 0.7.

IshanMi:
Sorry everyone- I was preoccupied with the Data Science API and forgot about this. I've made the changes requested and added tests to make sure calling `.private` does indeed return the correct things when receiving a string, list, tuple, array, etc.

shubham3121 (approved):
Superb work @IshanMi !!! 🎉🚀
Awesome work with intensive tests !!! 🔥🔥

PR #6849, opened 2022-10-12
Setup log collector for CI on workflow/jobs failure

bitsofsteve:
## Description
Considering the flakiness and a bit of how black-boxed the CI can sometimes be, this PR aims to provide some insights, especially during workflow fail, by collecting the logs of the running container per job and uploading it as a tarball to the GitHub artifact. This way should a workflow fail and needs to be debugged, one could easily look at the properly named and dated log files to see what happened.


tudorcebere:
I am hyped about this PR!!

PR #6790, opened 2022-09-27
[W.I.P.] Add more granular Publish tests

IshanMi:
## Description
At the moment we only test the output of our DP system- i.e. the output of calling `.publish()` on a DP tensor. This results in  the `publish` method being a black box where things may break and go undetected. Examples of this include `jnp.take` changing behaviour and not raising Out of Bounds/IndexErrors, or privacy budgets not spending properly.

This PR with @curt-mitch aims to test the internals of the publish method, and ensure that it fails gracefully and with some trace of the associated error. This will hopefully give us more confidence about the reliability of the DP system and the publish method.

## Affected Dependencies
List any dependencies that are required for this change.

## How has this been tested?
- Describe the tests that you ran to verify your changes.
- Provide instructions so we can reproduce.
- List any relevant details for your test configuration.

## Checklist
- [ ] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [ ] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [ ] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [ ] My changes are covered by tests

PR #6853, opened 2022-10-13
Fix publish breaking when not enough PB

shubham3121:
## Description
Publish breaks when there is not sufficient budget cause the comparison `value != zero_like` in `can_reduce_further` method return NotImplemented as output.

This is because zero_like is a GammaTensor tensor and JAX doesn't know how to interact with it.

## Affected Dependencies

## How has this been tested?
- Tested locally.

## Checklist


iamtrask:
Can we add a test for this?

IshanMi:
> Can we add a test for this?

We can add a test for this to the granular publish PR #6790

IshanMi (approved):

PR #6857, opened 2022-10-13
fix skill_test.py

mikaelapisani:
First padawan assesment


rasswanth-s (approved):
Well done @mikaelapisani  on first padawan PR 🎸

PR #6866, opened 2022-10-17
FIX Network Node version compatibility check

IonesioJunior:
## Description
- [ ] Trigger a warning asking if the user wants to continue with the application.
- [ ] domain.apply_to_network check Syft version compatibility
    - [ ] ADD Domain Client Syft version verification
    - [ ] ADD Environment Syft version verification
- [ ] Improve apply_to_network logs

### DIFF
--- a/packages/syft/src/syft/core/node/domain_client.py
+++ b/packages/syft/src/syft/core/node/domain_client.py
@@ -1,11 +1,15 @@
 # stdlib
 import logging
+import multiprocessing
+from multiprocessing import Event
+from multiprocessing.synchronize import Lock as LockBase
 import sys
 import time
 from typing import Any
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import Tuple
 from typing import Type
 from typing import Union
 
@@ -16,6 +20,7 @@ import names
 import pandas as pd
 
 # relative
+from ... import __version__
 from ...logger import traceback_and_raise
 from ...util import validate_field
 from ..common.message import SyftMessage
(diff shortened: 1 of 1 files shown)


IonesioJunior:
**No mismatch during connection**
network_success

IonesioJunior:
Cancel option due to a mismatch during connection
mismatch_version_n

IonesioJunior:
Ignore mismatch and keep applying to the network
mismatch_version_y

iamtrask:
Love where this is going! Especially love the GIFS!

packages/syft/src/syft/registry.py
```diff
@@ -54,13 +54,25 @@ def check_network(network: Dict) -> Optional[Dict[Any, Any]]:
...
-                    res = requests.get(url, timeout=0.5)
+                    ping_url = url + "ping"
+                    res = requests.get(ping_url, timeout=0.5)
                     online = res.status_code == 200
                 except Exception:
                     online = False
 
             if online:
+                # Check if syft version was described in NetworkRegistry
+                if not network.get("version", None):
```
shubham3121, line 65: @IonesioJunior should we have a check for `unknown` also here ? I'm thinking of the situation when a network node is online but unhealthy and therefore version is set to unknown, but when the node becomes healthy we should update the version. Thoughts?

IonesioJunior: Oh, good catch! Thank you for noticing it! ❤️

IonesioJunior: Is that what you meant? [ecc0cf6](https://github.com/OpenMined/PySyft/pull/6866/commits/ecc0cf6d71fe6e023641586cb9cc628d01495473)

shubham3121: Yes, this looks good 🙌🏾

shubham3121 (approved):
Wonderful work @IonesioJunior !! 💯 💪🏾

PR #6872, opened 2022-10-18
Fix account duplication during PyGrid initialization

IonesioJunior:
## Description
The `backend`, `backend-stream`, and `celery-worker` containers create the same account in the database during their initialization. This PR aims to avoid this duplication by making the email column unique and verifying if someone else with the same email is already registered in the database before registering a new account.

repeated_accounts

Fixes: https://github.com/OpenMined/PySyft/issues/6879


iamtrask:
YYEEESS!

rasswanth-s:
Wohoo!!!

packages/syft/tests/syft/core/node/common/service/user_manager_service_benchmark_test.py
```diff
@@ -53,18 +54,27 @@
...
+    faker: Faker, domain: DomainInterface, is_admin: bool = False
+) -> Tuple[dict, VerifyKey]:
 
     # Create dummy user
+    private_key = SigningKey.generate()
+    verify_key = private_key.verify_key
+
+    # Check if generated email is already been used for another account previously registered.
+    email = faker.free_email()
+    while domain.users.contain(email=email):
```
rasswanth-s, line 67: Q: should it be an if statement?

rasswanth-s (dismissed):
Amazing Work @IonesioJunior , 🥁  🤟

IonesioJunior:
Ohh, thank you for testing it and catching it out @rasswanth!

Yeah, lock seems a good way to solve it. I'll take a look at Madhava's solution and try to adapt it to this use case.

packages/syft/tests/syft/core/node/common/service/user_manager_service_benchmark_test.py
IonesioJunior: The `while ` statement is to ensure that the randomly generated email has not yet been used by any previously registered accounts.

IonesioJunior: While I was tidying these tests I came across situations where the same email was generated at two different times causing failures in the tests.

To ensure that this does not happen, the `while` loop will only allow the exit when we have an email that has never been used in any previously registered account.

IonesioJunior:
Can you help me to reproduce and test it out? For some reason in my environment, It just creates a unique account now. 

Maybe it is something related to the amount of concurrency we choose.

packages/syft/tests/syft/core/node/common/service/user_manager_service_benchmark_test.py
rasswanth-s: Ahh ok, it creates a new email on each function call, I initially thought faker was global and would provide same email each time 👍

rasswanth-s:
I just do a normal hagrid launch on my machine, 

Screenshot 2022-10-20 at 10 32 46 AM
They are created with a millisecond different in the created_at column.

rasswanth-s:
I think one idea would be to make the user creation run only at the backend container, we could separate the user creation at pre-start.sh (initial_data.py) and put in a separate shell script and we could make that script run in backend container only as an entry point. Thoughts?

packages/grid/backend/grid/db/init_db.py
```diff
@@ -16,28 +18,36 @@ def init_db(db: Session, signing_key: Optional[SigningKey] = None) -> None:
...
-            first_superuser_budget=5.55,
-            domain_name=settings.DOMAIN_NAME,
-        )
-    else:  # Uses the Owner root key to update node keys
-        setup = node.setup.first()
-        if setup.signing_key:
-            # the system has a key so use it
-            signing_key = SigningKey(
-                setup.signing_key.encode("utf-8"), encoder=HexEncoder
+        service_name = os.getenv("SERVICE_NAME", "")
```
rasswanth-s, line 21: MAYBE : Since we create the initial user in `prestart.sh` through `initial_data.py` , we could removed this line execution from prestart.sh, and in the docker compose in the backend service, we could add it to the command like
`command: /bin/sh -c " ./start.sh && python /app/grid/initial_data.py"`

rasswanth-s: I think it might be `/bin/sh or /bin/ash`

rasswanth-s: Thoughts?

IonesioJunior: I imagine this would be the ideal design, wrapping and isolating the launch functionality into the backend build.

However, this method does more than just initialize a new configuration; it also reloads updated database information such as node credentials (root signature keys). So we also need to run it in backend-stream and celery worker containers, so that the keys between the three images are the same.

IonesioJunior: What I did though, was split these two features (create and load ) into two different methods (`init_db` and `load_db`) and then delegated the responsibility of knowing which one of these we should execute to the `init_db.py` file.

rasswanth-s: Great find, I initially thought , it was like a common node object , since the containers are isolated, it makes sense that node objects are new instances.

packages/grid/backend/grid/db/init_db.py
```diff
@@ -11,7 +12,43 @@
 from grid.core.node import node
 
 
+def load_db(db: Session) -> None:
```
rasswanth-s, line 15: Nice work, on splitting it @IonesioJunior , the logic is solid.

rasswanth-s (approved):
Well done @IonesioJunior,Amazing 🎸  🥁

Issue #6890, opened 2022-10-24
wizard.check_grid_docker threw a TypeError


Kariten:
## Description
I met the exception below when running the notebook `01-install-wizard.ipynb` at in[7]: `wizard.check_grid_docker`.
How should I solve it?

## Screenshots
An exception occured: '>' not supported between instances of 'NoneType' and 'Version'.
Please file a bug report on GitHub Issues or in Slack #support

```json
{
    "datetime": "24/10/2022 13:40:56 PDT",
    "dependencies": {
        "ansible-playbook": "/root/.hagrid/quickstart/.venv/bin/ansible-playbook",
        "docker": "/usr/bin/docker",
        "git": "/usr/bin/git"
    },
[...]

PR #6901, opened 2022-10-25
Fixed TypeError while binary.version is None

Kariten:
#6890 
## Description
If users installed Docker without docker compose cli-plugin, `wizard.check_grid_docker` will threw a TypeError at
```python
MINIMUM_DOCKER_COMPOSE_VERSION = "2.0.0"
if binary_info.path and binary_info.version > version.parse(
    MINIMUM_DOCKER_COMPOSE_VERSION
):
    self.display = "✅ Docker Compose " + str(binary_info.version)
```
The reason is that docker can be found (binary_info.path is not None) and docker compose can not be found (binary_info.version is None), that causes the comparison between instances of 'NoneType' and 'Version'. So I added a check of binary_info.version.

## How has this been tested?
I ran the modified `wizard.check_grid_docker` in the previous environment, the wizard now can properly prompt users to manually install docker compose.


shubham3121:
Thank you @Kariten for the quick fix 🙌🏾 . 
Can you fix the lining errors?

rasswanth-s (approved):
Thank you for this important fix for our upcoming release.

Issue #6913, opened 2022-10-28
Calling .get() prematurely gives un-intuitive stack trace.


iamtrask:
Upon a freshly launched domain ("hagrid launch --tag=latest"), the following generates a very unintuitive stack trace.

```import syft as sy

me = sy.login(email="info@openmined.org", password="changethis", port=8081)

import numpy as np

n = 10000
_x = (np.random.randn(n) * 10).astype(np.int64)
x = sy.Tensor(_x).annotate_with_dp_metadata(min_val=0, 
                                                   max_val=20, 
                                                   data_subjects=list(map(lambda x:str(x), range(0,n))))

me.load_dataset(assets={"x":x}, 
                name="Test Dataset", 
                description="Short Desc")

me.create_user(name='Bob',
               email="andrew2@openmined.org",
               password="abc123",
               budget=10000000)

bob = sy.login(email="andrew2@openmined.org",
               password="abc123",
               port=8081)

initial_budget = bob.privacy_budget

x = bob.datasets[-1]['x']

out_ptr = x.mean().publish(sigma=10)
out = out_ptr.get()                                   #  <<<<<------   error occurs here

final_budget = bob.privacy_budget
```

![Image](https://user-images.githubusercontent.com/4328594/198748383-5174498c-26a0-4228-bd5c-a98890eaf11a.png)

iamtrask:
Because I know how PySyft works - I know that this is because the user called .get() before the object existed and so the user needs to call .block.get() in order to be able to run this code without triggering a trace. 

However, even though I know this it's still painful from a UX experience. In a perfect world we'd be able to say "Hey - it looks like you're asking for an object that won't exist for another 4 seconds"... or at maybe .get() should be a blocking function by default?

In any rate - the error of "object with id can't be found" is a really opaque way to fail and is quite misleading. Would love to hear better ideas for how this UX could be.

kirkdemeroukas:
Was having exactly the same issue and I was wondering why. Sometimes it worked but others it didn't even while changing absolutely nothing. I also tried calling .block.get() and while this doesn't throw an error anymore it just keeps loading for ever.

madhavajay:
@iamtrask we are discussing providing more robust feedback during 0.8 specifically relating to all of the `async` actions that have been queued which would allow us to provide information about this since we would know that it was submitted but not completed as opposed to something which never existed or was deleted. I think the correct solution here is to expose better feedback about the state of the action graph to users.

[...]

PR #6918, opened 2022-10-30
ADD Auto-block .get() for processing pointers and its childs

IonesioJunior:
## Description
Calling .get() prematurely gives an `object not found error`  due to the fact that the object wasn't registered into the node yet.
The changes in this PR aim to add some context to the client side, making it trace recent pointers that were sent or generated by some previous computation using previous pointers. This PR aims to achieve 3 goals:
- [x] Make the client API calls smooth.
- [x] Add some level of logic on the client side, making it decide which pointers to wait for and which to discard.
- [x] Still provide support manual blocking approach.

Closes: #6913

These changes are applied for:
- Syft Tensors and its operations
- Syft Klass object pointers and its operations
- Gamma Tensors and its operations
- Phi Tensors and its operations
- AST Callable Pointers
## How has this been tested?
-  **Removed all block statements from integration tests (smpc, dp, etc).**

## Checklist


packages/syft/src/syft/core/pointer/pointer.py
```diff
@@ -226,9 +227,30 @@ def _get(
...
         )
+
+        while is_processing_pointer:
+            try:
+                obj = self.client.send_immediate_msg_with_reply(
+                    msg=obj_msg, timeout=GET_OBJECT_TIMEOUT, verbose=True
+                )
+                self.client.processing_pointers.pop(self.id_at_location, None)
+                is_processing_pointer = False
+            except UnknownPrivateException:
```
iamtrask, line 242: This feels like it runs the risk of creating infinite loops which may be difficult to debug at a later date. I feel we should pass in a timeout which defaults to 90 seconds (looking at the tests commented out in this PR) and give an intuitive message if the timeout occurs "Object not found - it may be the that this object doesn't yet exist or it may be that this object has been deleted. If you believe that the object doesn't yet exist, please retry your query momentarily (or set a longer timeout .get(timeout=<insert longer time here>)   ). However, if you worry the object have have been pre-maturely deleted or that an earlier failure may have prevented the object from being created, more advanced interrogation of the system is needed. Start by attempting to re-run your code, checking that each preceeding object exists (you can call my_pointer.exists to check). Hopefully this will show you where an error appears to be occurring. If this does not resolve the issue and you are not an admin, you may need to contact the Domain's administrator for help."

tests/integration/e2e/spicy_swan_performance_test.py
```diff
@@ -211,7 +211,6 @@ def test_benchmark_datasets() -> None:
 
         start_time = time.time()
         publish_ptr = sum_ptr.publish(sigma=500_000)
-        publish_ptr.block_with_timeout(timeout)
```
iamtrask, line 214: For all of these commented out "block_with_timeout" methods, we should move the timeout parameter to the .get() method instead. Infinite blocking could cause all sorts of issues later on. Additionally, if something does timeout we would do well to make a really obvious stdout so that we can see why CI would be running so slowly.

iamtrask (changes requested):
While the functionality satisfies the requirements of the project, I worry that a small aspect of this might make future work difficult to debug. The change I recommend isn't particularly large, but I think it could have big impacts on how effectively we can debug in the future.

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
```diff
@@ -454,12 +456,22 @@ def reconstruct(self, delete_obj: bool = True) -> np.ndarray:
...
         for share in self.child:
-            if not share.exists:
-                raise Exception(
-                    "One of the shares doesn't exist. This probably means the SMPC "
-                    "computation isn't yet complete. Try again in a moment or call .block.reconstruct()"
-                    "instead to block until the SMPC operation is complete which creates this variable."
-                )
+            # Check if share pointer is in processing stage
+            # If so, wait until it exists.
+            if share.client.processing_pointers.get(share.id_at_location, False):
```
madhavajay, line 461: @rasswanth-s Can you check this? It could be that the old code was relying on the exception and a retry mechanism elsewhere in a green thread? Do you know if doing this change and using normal sleep here will cause any issues?

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
```diff
@@ -732,6 +744,10 @@ def __apply_op(
         shape = utils.get_shape(op_str, self.shape, y_shape)
         ring_size = utils.get_ring_size(self.ring_size, y_ring_size)
 
+        if utils.ispointer(result[0]):
```
madhavajay, line 747: Does this code just get the result of the dict but assign it to nothing?

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
rasswanth-s: Since the MPCTensor is run on the server side, this should work, but I think we could remove the unbounded wait using time. sleep(), as sometimes the smpc computation is not complete we would go in an infinite loop.

madhavajay: Thanks for checking, yes that makes sense. I have added a timeout just like the above pointer code as well.

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
madhavajay: @IonesioJunior was this supposed to set the boolean to True for these?

madhavajay:
@IonesioJunior Great work. I had a play and made some small changes and added a test.

Since there was already a default timeout in the http code this now flows through and does the while loop if its in the pointer registry as well but there is no concept of No timeout because 0 timeout is not allowed in the http code.

I think this means currently in my changes the default retry is the GET_OBJECT_TIMEOUT.
The code is becoming a little busy because we had a timeout function in `get` already which was relating to request block.

The other thing we could do is have some concept of retry=True/False but right now the code is already getting a little overloaded so i think we should circle back to this later rather than have it hold up our existing bugs in 0.7.

As long as this doesn't break anything for now I think we can merge it but we should keep an eye on it because .get is used everywhere and there could be unexpected side effects of changing the default behavior. I noticed some of the network tests take longer because they rely on the .get failing and getting caught in a try catch which doesnt happen as fast now so we might need to change those tests.

packages/syft/src/syft/core/tensor/smpc/mpc_tensor.py
IonesioJunior: Yup, you're right! Just fixed it now, thank you for reviewing and noticing it.

IonesioJunior: Fixed by: [e47e21d](https://github.com/OpenMined/PySyft/pull/6918/commits/e47e21d3ed28b150676f4c5884b6677cf32a10c0)

tests/integration/e2e/spicy_swan_performance_test.py
IonesioJunior: Fixed by: [a8c5abf](https://github.com/OpenMined/PySyft/pull/6918/commits/a8c5abf1494356f854a81631b814e5928bc0eb8b)

IonesioJunior:
>The other thing we could do is have some concept of retry=True/False but right now the code is already getting a little overloaded so I think we should circle back to this later rather than have it hold up our existing bugs in 0.7.

+100

IonesioJunior:
>  I noticed some of the network tests take longer because they rely on the .get failing and getting caught in a try-catch which doesn't happen as fast now so we might need to change those tests.

Actually, this shouldn't happen since if we're looking for a pointer ID that isn't there, then it's not in the processing_pointers register as well. So it should skip the while and raise the `Exception` instantly.

packages/syft/src/syft/core/pointer/pointer.py
```diff
@@ -242,17 +242,24 @@ def _get(
...
             try:
                 obj = self.client.send_immediate_msg_with_reply(
                     msg=obj_msg, timeout=timeout_secs, verbose=True
                 )
+
+                # If we reached here it's because we didn't have any failure,
+                # so we were able to retrieve the pointer successfully.
+                # So it isn't a processing pointer anymore and we can exit the while loop.
+                # without wait the timeout
+                is_processing_pointer = False
```
madhavajay, line 256: good catch

madhavajay:
@IonesioJunior awesome work. 🙌🏽

iamtrask:
Really nice work - also @IonesioJunior i love that design decision you mentioned about calling the exception immediately if the pointer isn't there. Well done!

PR #6994, opened 2022-11-07
Add validation for budget

### LINKED ISSUE #6985: Able to assign negative PB during create user
## Description
Currently, we are able to assign a negative Privacy Budget to a user.

## How to Reproduce
1. Launch domain: `hagrid launch canada domain to docker:8081 --tag=latest --tail=false --dev`
2. Code to reproduce:
```
import syft as sy

domain_client = sy.login(email="info@openmined.org", password="changethis", port=8081)

domain_client.create_user(name="Joker", email="joker@test.com", password="test", budget=-1)

domain_client.users
```

Screenshot from 2022-11-07 12-49-19

## Expected Behavior
It should throw an error conveying that a user cannot be allocated a negative PB

gmuraru:
## Description
Fixes #6985 
* Add validation for the budget when creating a user.
* Add test for user creation validation


packages/syft/src/syft/core/node/domain_client.py
```diff
@@ -759,7 +759,9 @@ def load_dataset(
             "\n\nRun `<your client variable>.datasets` to see your new dataset loaded into your machine!"
         )
 
-    def create_user(self, name: str, email: str, password: str, budget: int) -> dict:
+    def create_user(self, name: str, email: str, password: str, budget: float) -> dict:
+        if budget <= 0:
+            raise ValueError(f"Budget should be a positive number, but got {budget}")
         try:
             self.users.create(name=name, email=email, password=password, budget=budget)
```
rasswanth-s, line 766: MAYBE: we could raise the exception in self.users.create (UserRequestsAPI)  function,
if they by  mistake call 
domain.users.create(**Data_Scientist_details)
Some of our old notebooks, create Data Scientist in this manner.

rasswanth-s: Yep makes sense george 👍

packages/syft/tests/syft/core/node/domain/domain_test.py
```diff
@@ -144,3 +144,14 @@ def test_object_exists_on_domain(domain: sy.Domain) -> None:
     assert not ptr.exists
     ptr = x.send(domain_client, id_at_location_override=uid)
     assert ptr.exists
+
+
+@pytest.mark.asyncio
+def test_create_user(domain: sy.Domain) -> None:
+    domain.create_user(email="test@test.com", password="changethis", budget=1)
```
rasswanth-s, line 151: Could we replace the domain with `root_client` fixture as the create_user is present only on Client Object.

rasswanth-s (approved):
Well Done @gmuraru

packages/syft/src/syft/core/node/domain_client.py
shubham3121: I still think we should have a check for the budget being negative in `def create` in `syft/core/node/common/client_manager/user_api.py`, as it is still the API being exposed to the client. Ideally, I think these kinds of checks would also lie on the server side in the user message service, but since we don't propagate those exceptions to the client currently, throwing those errors won't be communicated to the client anyway.

gmuraru: From what I see it is something like:
- `create_user` is the API exposed to the user from the `client_side`.
- `create` is the method called by the `create_user` - the `user` should not directly call `create`, right?

I think we could do the followings:
* raise the exception in `create_user`
* raise the exception in `create`
* raise the exception in both places - to have a safeguard on the client-side and also server-side

If we raise the exception only in `create_user` the error is caught earlier in the calling chain and it does not forward the call to the server (by calling `create`).

The downside is that a `user` could change their part of the `create_user` implementation to remove that validation and in this case it could call `create` with a negative budget - is there any problem if there is a user created with a negative budget?

If we raise the exception in both places - we have a safeguard in both places until we figure out how to route exception from server to client side (now I have a preference for this method :smile:).
Also, if we choose to implement it also `server side` I would prefer to add a new parameter to the `create` method - `budget` (and also keep `**kwargs`). Is there any scenario when a user is created without explicitly specifying a budget?
(I prefer doing `if budget <= 0` raise error than `if "budget" in kwargs and kwargs["budget"] <= 0" raise error - if budget is always mandatory we could make it explicitly required -- but I am not sure if that is the case)

What do you think?

shubham3121: Yeah, I like the idea of raising the exception in both places.
`is there any problem if there is a user created with a negative budget?` -> Ideally, nothing bad would happen but since the budget will always be >=0, so I feel it doesn't feel right to be able to assign a negative budget 😅 .
We can have `if budget < 0`, as we assign zero by default for new users when they register using `sy.register` method.

gmuraru:
@madhavajay @rasswanth-s  I think the `k8` tests are failing here. Should I continue to restart them until they pass?

rasswanth-s:
@gmuraru  the k8s issues are solved now.

packages/syft/src/syft/core/node/domain_client.py
```diff
@@ -759,7 +759,9 @@ def load_dataset(
             "\n\nRun `<your client variable>.datasets` to see your new dataset loaded into your machine!"
         )
 
-    def create_user(self, name: str, email: str, password: str, budget: int) -> dict:
+    def create_user(self, name: str, email: str, password: str, budget: float) -> dict:
```
rasswanth-s, line 762: I think we should also include a budget  sanity  check in the service side code also  in the service part of 
```py
 def create_user_msg
```
 of user_manager_service.py
https://github.com/OpenMined/PySyft/blob/8f550d8fe6a249d0e956335b033c1870b9870b1f/packages/syft/src/syft/core/node/common/node_service/user_manager/user_manager_service.py#L36
As the  code would also allow data scientist to create custom message with their custom values.

gmuraru: I will look a little bit at this! I will come back with an update

packages/syft/src/syft/core/node/common/client_manager/user_api.py
```diff
@@ -49,6 +49,11 @@ def __delitem__(self, key: int) -> None:
         self.delete(user_id=key)
 
     def create(self, **kwargs: Any) -> None:
+        budget = kwargs.get("budget")
```
gmuraru, line 52: Q: @rasswanth-s  probably this should be deleted? (since this is still on the client side and they were already verified in `domain_client`?)

rasswanth-s: I think we could have it here also george, as we expose the client side API , of domain_client.users.create in addition to domin_client.create_user.

rasswanth-s (approved):
LGTM with updated changes.

packages/syft/tests/syft/core/node/domain/domain_test.py
rasswanth-s: @gmuraru  Ping, I think we get this test to wok, we are good for merge on the PR.

packages/syft/src/syft/core/node/common/client_manager/user_api.py
gmuraru: Done! I made the changes locally, but I forgot to add the `test` directory to the commit :facepalm:

PR #6997, opened 2022-11-08
Fix dataset dtype reflecting when viewing assets from .dataset API

IonesioJunior:
## Description
The error happened due to the fact that dtype was a string `< class ...Tensor >`. The `<>` were mixed with HTML syntax, creating an error while trying to render it in jupyter notebook.

## Affected Dependencies

## How has this been tested?

## Checklist


packages/syft/src/syft/core/node/common/util.py
```diff
@@ -115,7 +115,7 @@ def upload_result_to_s3(
         asset_name=asset_name,
         dataset_name=dataset_name,
         node_id=domain_id,
-        dtype=data_dtype,
```
gmuraru, line 118: Q: Could we delete the `data_type`? (it is used anywhere else?)

packages/syft/src/syft/core/node/common/util.py
```diff
@@ -262,7 +262,7 @@ def upload_to_s3_using_presigned(
             asset_name=asset_name,
             dataset_name=dataset_name,
             node_id=client.id,
-            dtype=data_dtype,
```
gmuraru, line 265: Also here

packages/syft/src/syft/core/node/common/util.py
IonesioJunior: Good catch! Yup, we must delete it.

IonesioJunior: Solved by: [54a6c1e](https://github.com/OpenMined/PySyft/pull/6997/commits/54a6c1e92d38de1bc27952bf332c902ec0a53a67)

packages/syft/src/syft/core/node/common/util.py
IonesioJunior: Solved by: [54a6c1e](https://github.com/OpenMined/PySyft/pull/6997/commits/54a6c1e92d38de1bc27952bf332c902ec0a53a67)

rasswanth-s:
@IonesioJunior ,I think when module name is too large , it overlaps with ID , when we view the datasets, 
Maybe we could print just the class Name without the path or reformat the pandas table?
Screenshot 2022-11-08 at 4 51 58 PM

IonesioJunior:
Ohh, yeah! Thank you for noticing it @rasswanth-s! ❤️

IonesioJunior:
Looks better now, but maybe in the future we can change this table to change its size dynamically according to the size of the output
datasets_vis

rasswanth-s (approved):
Well done @IonesioJunior 🤟

Issue #6950, opened 2022-11-03
Min vals division by zero in lazyrepeatarrays may be causing overflow errors


IshanMi:
When we compute the final min_vals and max_vals, we often do something like: min_min, min_max, max_max, max_min and compare all the values to see which is the smallest.

For an op like truediv this might involuntarily involve dividing by zero and raising an error.

co-authored by: @teo-milea

IshanMi:
![Image](https://user-images.githubusercontent.com/32711264/199743372-c9bb595e-8510-426f-84b8-3ef7b82c6186.png)

IshanMi:
This led to this in the logs:



![Image](https://user-images.githubusercontent.com/32711264/199743504-8118bf83-be22-424d-ad1b-c6dcdc56d976.png)

iamtrask:
As a bug this is more complex than a simple overflow issue. It is genuinely an infinite privacy spend risk WRT the public privacy API. If the denominator has a min_val of -1 and a max_val of 1, then it's possible that the private value is indeed zero and the computation is invalid. The permutation of min_min, min_mmax, etc. is correctly capturing this. That is to say - it's correct to determine that it's possible this is an invalid computation.

We will need to spend more time thinking on this. For 0.7 - please fix by raising an exception on the pointer when a denominator could be zero (isn't positive-definite or negative-definite). The exception should read something like, "I'm sorry, but our privacy accountant can't yet divide by a number which might be zero. Please find a way to approximate your computation without dividing by a private value which might be zero. This can usually be done by computing the function piecewise - by performing three computations which are merged through masking and summing: one which addresses what happens if the denominator is positive-definite, another which addresses what happens if the denominator is negative-definite, and another which addresses what happens if the denominator is 0 exactly. Use comparison operators, masking, and summing to accomplish this (not if statements - which don't work on private values). "

teo-milea:
In order to be more general, during the publishing system I raised an exception when we find any `NaN` or `Inf` in the L2 norms in this PR: https://github.com/OpenMined/PySyft/pull/7008. Indeed, I created an exploit that allowed a DS to publish entire arrays while only consuming the default budget for publishing noise. It seems like these checks were enough. On the overflow issue, however, I could reverse the min and max values, but couldn't abuse this in any way. I will add some warnings on the DS side for this issue https://github.com/OpenMined/PySyft/issues/6952, but at the moment in doesn't seem we need any check on the server. Let me know if you have any recommendations!

[...]

PR #7008, opened 2022-11-09
Add check for `NaN` values in the publish mechanism

### LINKED ISSUE #6956: Test if we can handle `NaN` values in the publish system without revealing private data
Currently, in our test suite, we don't test with `NaN` values which can be a problem as we expect to spend `NaN` privacy budget when publishing arrays with at least one unknown value. 

Co-authored by @IshanMi

teo-milea:
## Description
This fixes #6956 

## How has this been tested?
There is a nb that replicates the problem.

## Checklist


packages/syft/src/syft/core/adp/vectorized_publish.py
```diff
@@ -155,7 +157,19 @@ def publish(
         # Step 2: Calculate the epsilon spend for this query
 
         # rdp_constant = all terms in Theorem. 2.7 or 2.8 of https://arxiv.org/abs/2008.11193 EXCEPT alpha
-        rdp_constants = compute_rdp_constant(rdp_params, private=private)
+        if any(np.isnan(l2_norms)):
```
madhavajay, line 160: @teo-milea nice! Could we extract this into a function which takes the input and then raises a custom exception so theres no need for indenting and we can catch that specific exception in the outer part of the function?

packages/syft/src/syft/core/adp/vectorized_publish.py
```diff
@@ -177,10 +195,16 @@ def publish(
...
                 "Please contact the OpenMined support team for help."
+                "For that you can either:"
+                " * describe your issue on our Slack #support channel. To join: https://openmined.slack.com/"
+                " * send us an email describing your problem at support@openmined.org"
+                " * leave us an issue here: https://github.com/OpenMined/PySyft/issues"
             )
 
         # Step 3: Check if the user has enough privacy budget for this query
         privacy_budget = get_budget_for_user(verify_key=ledger.user_key)
+        print(privacy_budget)
```
madhavajay, line 206: I think we could add these to the log because we want them most of the time and the data owners probably would too.

packages/syft/src/syft/core/adp/vectorized_publish.py
```diff
@@ -155,7 +157,20 @@ def publish(
         # Step 2: Calculate the epsilon spend for this query
 
         # rdp_constant = all terms in Theorem. 2.7 or 2.8 of https://arxiv.org/abs/2008.11193 EXCEPT alpha
-        rdp_constants = compute_rdp_constant(rdp_params, private=private)
+        if any(np.isnan(l2_norms)):
+            if any(np.isnan(l2_norm_bounds)) or any(np.isinf(l2_norm_bounds)):
```
rasswanth-s, line 161: MAYBE: We could also raise the error when we extract min/max values if the min/max values hold nan values.

packages/syft/src/syft/core/adp/vectorized_publish.py
```diff
@@ -155,7 +157,20 @@ def publish(
         # Step 2: Calculate the epsilon spend for this query
```
rasswanth-s, line 158: Q: Would it be possible to catch these(the nan value errors) on the DS side and intimate them(ex: based on the min, max values, and the computation performed), or are we planning for client-side checks in `0.8`?

IshanMi:
I added a check inside the `compute_min_max` function of the lazyrepeatarrays which will help us detect NaN and Infs on the data scientist's side and raise Exceptions early, on the TensorPointers themselves :+1: 

This will also help #6950

packages/syft/src/syft/core/adp/vectorized_publish.py
IshanMi: Great idea Rasswanth, added in [3e18454](https://github.com/OpenMined/PySyft/pull/7008/commits/3e184546b926986b6a293cf90beb8cb1d541d050)

packages/syft/src/syft/core/adp/vectorized_publish.py
IshanMi: Unrelated to Rasswanth's comment- but would the use of `any()` fail here if `l2_norms` is a singleton, like the result of `sum()`?

IshanMi (approved):

PR #7011, opened 2022-11-09
Add support for Hagrid logs & set --tail=false to default

### LINKED ISSUE #6971: "hagrid launch --tail=False" should be default + a table that lists containers and the commands to see their live logs
The new user experience when --tail=False is incredible and by far the best experience for most users (everyone except devs?), especially since there's a definitive "you're fully launched' moment, as opposed to just watching logs until it stops. --tail=false should be the default experience except when you're in dev mode.

That said - finding logs is hard and scary to docker noobs and they're going to experience bugs. I think it would be great if we added a table to the final output of hagrid launch --tail-False which listed every container with a single command for how to see the logs. (and a highlight for where "if you get errors they'll be here or here" pointing somehow towards the streaming and backend containers. 

STRETCH: link to Rasswanth's video for finding logs in case people need it.

bitsofsteve:
## Description
fixes #6971 

This PR adds a new command to hagrid, which is `hagrid logs`, and generates an output like so. 

Also, this PR sets --tail=false to the default behavior in hagrid when launching domain nodes.

Screenshot 2022-11-10 at 02 18 33

### DIFF
--- a/packages/hagrid/hagrid/cli.py
+++ b/packages/hagrid/hagrid/cli.py
@@ -3522,7 +3522,7 @@ cli.add_command(ssh)
 
 
 # Add hagrid logs command to the CLI
-@click.command(help="Get the logs of the HAGrid service")
+@click.command(help="Get the logs of the HAGrid node")
 @click.argument("domain_name", type=str)
 def logs(domain_name: str) -> None:  # nosec
 


packages/hagrid/hagrid/cli.py
```diff
@@ -3511,3 +3519,81 @@ def ssh(ip_address: str, cmd: str) -> None:
 
 
 cli.add_command(ssh)
+
+
+# Add hagrid logs command to the CLI
+@click.command(help="Get the logs of the HAGrid service")
```
rasswanth-s, line 3525: NITPICK: maybe replace by
Get the logs of the HAGrid `node`

packages/hagrid/hagrid/cli.py
```diff
@@ -3511,3 +3519,81 @@ def ssh(ip_address: str, cmd: str) -> None:
...
+@click.argument("domain_name", type=str)
+def logs(domain_name: str) -> None:  # nosec
+
+    """
+    host_name = get_host_name(domain_name)
```
rasswanth-s, line 3530: Q: should this comment be deleted?

packages/hagrid/hagrid/cli.py
```diff
@@ -3511,3 +3519,81 @@ def ssh(ip_address: str, cmd: str) -> None:
...
+
+
+long_string = (
+    "ℹ [bold green]To view the live logs of a container,copy the log command and paste it into your terminal.[/bold green]\n"  # noqa: E501
+    + "\n"
+    + "ℹ [bold green]The logs will be streamed to your terminal until you exit the command.[/bold green]\n"
+    + "\n"
+    + "ℹ [bold green]To exit the logs, press CTRL+C.[/bold green]\n"
+    + "\n"
+    + "🚨 The [bold white]backend & backend_stream[/bold white] [bold green]containers are the most important to monitor for debugging.[/bold green]\n"  # noqa: E501
```
rasswanth-s, line 3584: MAYBE: we could specifiy celery worker also.

rasswanth-s:
For example, when people, tend to use hagrid logs for  `not` launched domain, should we raise an intimation that the domain is not launched yet?
Screenshot 2022-11-22 at 12 33 53 PM

rasswanth-s (approved):
Fantastic Work @bitsofsteve  , the UI is really cool

bitsofsteve:
> For example, when people, tend to use hagrid logs for `not` launched domain, should we raise an intimation that the domain is not launched yet? Screenshot 2022-11-22 at 12 33 53 PM

@rasswanth-s I wonder if, for sake of scope, we can move this improvement to a new ticket so we can limit the "ifs" and work on it as a separate issue?

rasswanth-s:
Sure @bitsofsteve we could do that 👍

PR #7024, opened 2022-11-11
Wrapped data_subjects check with an numpy array to work for singular values

### LINKED ISSUE #6933: Data subject checks fail when using the result of an operation like sum
In several places in the codebase, we do the following check:
`if (self.data_subjects == other.data_subjects).all():`

If `self` and `other` is the result of an operation like sum, then this `self.data_subjects == other.data_subjects` returns a single Bool (since we're comparing two singletons) instead of an array of booleans (which has the attribute `all()`). This causes the check above to fail: 

Image

teo-milea:
## Description
Fixes #6933.

## How has this been tested?
Using the Twitter demo notebook.

## Checklist


shubham3121 (approved):
Great work @teo-milea !!! This was a neat fix. 🎉 💯

PR #7028, opened 2022-11-11
[scripts] dev_tools improve win support & performance

yashgorana:
## Description
Improve `dev_tools.sh` support on Windows for launching specific ports in using the `start.exe` command. Script's dependency on `python3` was also causing it to fail on systems that did not have `python3` command available. Instead of fixing the python/python3 conflict, a more robust fix is to use the faster [`docker inspect --format`](https://docs.docker.com/engine/reference/commandline/inspect/#list-all-port-bindings) that uses [Go Templates](https://pkg.go.dev/text/template). The script has also been refactored in a little to support any future docker-specific additions.

image

## Affected Dependencies
- Padawan notebooks may need to be updated to reflect the new output.

## How has this been tested?
- Verified by running the dev tools script on Windows, WSL2 & Linux.

## Checklist

### DIFF
--- a/packages/grid/docker-compose.dev.yml
+++ b/packages/grid/docker-compose.dev.yml
@@ -30,8 +30,8 @@ services:
   queue:
     image: rabbitmq:3-management
     ports:
-      - "5672"
-      - "15672"
+      - "15672" # admin web port
+      # - "5672" # AMQP port
 
   backend:
     volumes:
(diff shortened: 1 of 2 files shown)


yashgorana:
@shubham3121 @madhavajay can you please take a look at this.

I can also directly print out `http://localhost:[port]` for ease-of-use, but trying to keep parity with the previous script's output.

madhavajay:
Awesome work @yashgorana. This works great!
Screen Shot 2022-11-17 at 4 29 57 pm

I removed the two unused ports for seaweedfs and rabbitmq (s3 and amqp) so that users dont try to open them with the dev tools and get nothing useful.

madhavajay (approved):
Awesome work! 🙌🏽

Issue #6744, opened 2022-09-14
First syft import is slow


madhavajay:
Is there any way we can speed up the first import of syft because its horribly slow.
Are we able to somehow profile it to know why?

I believe doing lazy execution on the exported modules is one way to do this.

kiendang:
Profiling `import syft` here https://kien.ai/syft-import-profile/

```
python -X importtime -c "import syft"
```

Looks like importing external libraries (`torch`, `jax`, `flax`, `pandas`) accounts the majority of total import time. We probably should lazy load these. No point loading `jax` if user only uses `torch`.

With [PEP 690](https://peps.python.org/pep-0690/) lazy loading should become the default behavior but I'll look into ways to do this manually for now.

madhavajay:
That refactor was epic. Great work. I will be discussing the `torch` aspect shortly because its not actually used in any of our current functionality so i suspect we could actually just cut it. I think until we get an answer on priority lets just pause this.

PR #7030, opened 2022-11-13
Replace syft absolute imports inside of the codebase

IonesioJunior:
## Description
In order to provide a better code structure, detect future circular imports, implement lazy import we should avoid using absolute syft import inside of the codebase. This PR aims to replace the syft absolute import occurency all over the codebase by its relative import.

## Affected Dependencies

## How has this been tested?

## Checklist


rasswanth-s:
Code Cleanuppp!!!! 🤟

kiendang:
+1 for this PR and the circular import one. One problem with `syft` right now is slow import #6744. One solution is lazy import and for that to work we need to get rid of circular imports.

packages/syft/src/syft/core/node/domain_client.py
```diff
@@ -408,15 +403,6 @@ def reset(self) -> None:
         if response == "y":
             response = self.routes[0].connection.reset()  # type: ignore
 
-    def search(self, query: List, pandas: bool = False) -> Any:
```
rasswanth-s, line 411: Q: Do we not use network search message anymore?

packages/syft/src/syft/core/node/common/node_manager/redis_store.py
```diff
@@ -16,6 +16,8 @@
 import syft
```
rasswanth-s, line 16: maybe we could remove this import absolute import also?

rasswanth-s:
I found absolute syft imports in file:
`packages/syft/src/syft/util.py`
`packages/syft/src/syft/ast/klass.py`
`packages/syft/src/syft/core/common/serde/serializable.py`
`packages/syft/src/syft/core/common/serde/serialize.py`
`packages/syft/src/syft/core/node/common/node_manager/redis_store.py`
maybe we could replace them also.

packages/syft/src/syft/core/node/domain_client.py
IonesioJunior: We removed only from the Domain Client. But it's still in NetworkClient API
network_client_search

packages/syft/src/syft/util.py
```diff
@@ -55,6 +53,10 @@ def validate_type(_object: object, _type: type, optional: bool = False) -> Any:
     )
 
 
+def get_syft_built() -> ModuleType:
+    return sys.modules[__name__.split(".")[:-1][0]]
```
kiendang, line 57: Hmm why not just `__name__.split(".")[0]`? The only case when there's a difference is when `__name__.split(".")` has length 1 but then it would throw an error?

packages/syft/src/syft/core/node/domain_client.py
rasswanth-s: Ahh ok

packages/syft/src/syft/util.py
rasswanth-s: Q: Why did we replace from "syft" string to using `__name__`?

rasswanth-s (approved):
LGTM, Fantastic Start  on revamping syft 🏖️

packages/syft/src/syft/util.py
kiendang: also NIT but should it be `get_syft_build()` instead of `built`?

IonesioJunior: > Hmm why not just `__name__.split(".")[0]`? The only case when there's a difference is when `__name__.split(".")` has length 1 but then it would throw an error.

Yup, you're right, not only that but the current solution isn't reliable if we change this source file to another place. Thanks for noticing it.

IonesioJunior: > Q: Why did we replace from "syft" string to using `__name__`?

Not a cybersecurity expert here. 
But I thought in a scenario where `syft` key is overwritten in `sys.modules` making this package load a malicious package inside of it. Not completely sure if it's feasible but better safe than sorry

IonesioJunior: > also NIT but should it be `get_syft_build()` instead of `built`?

The main purpose of this method is to get syft package when its completely loaded, that's why I named it with "built". but now, thinking a bit about that, I guess a better name would be `get_loaded_syft`.

IonesioJunior: Solved by: [5194633](https://github.com/OpenMined/PySyft/pull/7030/commits/51946335e520361eeda3824d64a3796e77e9c511) and [91c73f1](https://github.com/OpenMined/PySyft/pull/7030/commits/91c73f19ebf8a7d552168b69c6b56fc3ad5a6d7d)

rasswanth-s: > > Q: Why did we replace from "syft" string to using `__name__`?
> 
> I'm not a cybersecurity expert. But I thought in a scenario where `syft` key is overwritten in `sys.modules` making this package load a malicious package inside of it. Not completely sure if it's feasible but better safe than sorry

I see, good point 👍

Issue #6940, opened 2022-11-03
`repeat` method on data pointer does not accept list arg


kiendang:
## Description

The `repeat` method on data pointer errors if passed in a list.

```py
d_ptr.repeat([1] * 10000)
``` 

but works fine if passed in an np array

```py
d_ptr.repeat(np.array([1] * 10000))
``` 

## How to Reproduce

Use the same setup as @shubham3121 in #6928

```python
import syft as sy
import numpy as np

domain_client = sy.login(email="info@openmined.org", password="changethis", port=8081)
arr = np.random.randint(0, 10000, (10000))
data = sy.Tensor(arr).annotate_with_dp_metadata(-1, 11000, data_subjects="Bob")
data_ptr = data.send(domain_client, tags=["my_data"])

domain_client.create_user(name="DS", email="test@email.com", password="password", budget=99999)
ds_client = sy.login(email="test@email.com", password="password", port=8081)

ds_client.store

d_ptr = ds_client.store[data_ptr.id_at_location]
```

run `repeat`

```python
d_ptr.repeat([1] * 10000)
```

Currently this returns `AttributeError: 'Reply' object has no attribute 'upload_id'` caused by a bug in the exception handling logic. #6979 fixes this.

<details>

<summary>details</summary>

```py
Uploading `6b0502f29a1b4152b8e74e6e72886856`: 100%|██████████████████| 1/1 [00:00<00:00, 157.75it/s]

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
File ~/pysyft/packages/syft/src/syft/core/node/common/util.py:267, in upload_to_s3_using_presigned(client, data, chunk_size, asset_name, dataset_name)
    259         obj_public_kwargs.update(child_kwargs)
    261     proxy_data = ProxyDataset(
    262         asset_name=asset_name,
    263         dataset_name=dataset_name,
    264         node_id=client.id,
    265         dtype=data_dtype,
    266         fqn=data_fqn,
--> 267         shape=data.shape,
    268         obj_public_kwargs=obj_public_kwargs,
    269     )
    270 except (Exception, KeyboardInterrupt) as e:

AttributeError: 'List' object has no attribute 'shape'

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
Cell In [27], line 1
----> 1 repeat = d_ptr.repeat([1] * 10000)

[...]

PR #7080, opened 2022-11-17
fix data pointer accepting list arg

ivyclare:
PR for [issue](https://github.com/OpenMined/PySyft/issues/6940)


packages/syft/src/syft/core/node/common/util.py
```diff
@@ -256,13 +257,20 @@ def upload_to_s3_using_presigned(
             child_kwargs = getattr(data.child, "proxy_public_kwargs", {})
             obj_public_kwargs.update(child_kwargs)
 
+        if isinstance(data, list):
```
madhavajay, line 260: I think we could go a step further with:

```python
data_shape = (0,)
if hasattr(data, "shape"):
    data_shape = data.shape
elif hasattr(data, "__len__"):
    data_shape = (len(data), )
```

This would ensure that any type with .shape including our own types will work and then as a backup any collection type with a length can be used.

Also I would say its best to preserve the `tuple` of 1 dimension expectation for `.shape` otherwise all the downstream code will get confused if its sometimes an int.

kiendang:
Thanks @ivyclare.

We could merge this but as @madhavajay said `ProxyDataset` was originally only meant for tensors/arrays. This PR makes it work with other things with `__len__` but also open to all kind of variables and might cause other issues down the line?

Should we merge this and create an issue to remind us to implement `ProxyObject` for things other than tensors/arrays?

kiendang:
Oh apologies. Looks like that has always been the plan. I only just now saw the empty `proxy_dataset.py` file.

kiendang (changes requested):
Nice job @ivyclare 👍. Just one thing. There's an empty file named `Python:` that got included in your PR. Could you exlude that? Also could you create a new issue to remind us to implement `ProxyObject` based on [this comment](https://github.com/OpenMined/PySyft/issues/6940#issuecomment-1305085449) by @madhavajay?

kiendang (approved):
Great work @ivyclare! I just added a test. Ready for merging now.

PR #7086, opened 2022-11-20
[WIP]SMPC Notebook

### LINKED ISSUE #7082: DS could get the result of MPC computations without request/response.
## Description
Data Scientist could get the result of SMPC computations without a request/response
```python
import syft as sy
import numpy as np
alice = sy.login(email="info@openmined.org",password="changethis", port=8081)
bob = sy.login(email="info@openmined.org",password="changethis",port=8082)
# create a data scientist account using alice and bob client
skywalker_alice = sy.login(port=8081, email="rey@skywalker.net", password="jakku")
skywalker_bob = sy.login(port=8082, email="rey@skywalker.net", password="jakku")
data = sy.Tensor(np.array([1,2,3]))
data.send(alice)
data.send(bob)
alice_ptr = skywalker_alice.store[-1]
bob_ptr = skywalker_bob.store[-1]
res = alice_ptr+bob_ptr # SMPC computation.
res.get() 
```
.get() should fail, as the Data Scientist should not get the result of intermediate computation in SMPC without a request/response.

Solution:
`remove DS verify_key in przs_action.py`

### LINKED ISSUE #7089: SMPC: Insecure Share Generation
## Description
During SMPC computations , the creation of shares of a secret is insecure, as we allow Data Owners ,to view the shares generated by all the parties in the computation.
https://github.com/OpenMined/PySyft/blob/6972b46ee759c132d15da12e6152331f2fbcb722/packages/syft/src/syft/core/tensor/smpc/share_tensor.py#L335

Even though the share creation is done as PseudoRandomZeroSharing.
This would allow malicious Data owners to retrieve the `private  data` of other data owners, if the initial sharing phase is followed by multiplication.

Even more  vulnerability is that the seed for generating the shares is generated at the Data scientist side

## Expected Behavior
Should distribute shares in pairwise circle fashion for generation PRZS.
The seed generation should be done at the server side.

More context on PRZS could be found on this notebook:
https://github.com/OpenMined/PySyft/blob/608871f0481a70556422aebb4e419e1a068fa1de/notebooks/padawan/07-SMPC.ipynb

rasswanth-s:
## Description
The PR contains the SMPC notebooks for Padawan sessions.

Fixes #7082 
Fixes #7089


gmuraru (commented):
For the notebook.
I will try to leave them on a per section.

#### Motivations
Some questions might arise:
- Why not each DO add noise, DS fetches the results from each DO node and then does whatever they want with it without involving SMPC.

#### Example:
Scenario 1: DS fetches results from DOs (with noise) then does addition (no SMPC involved)
Scenario 2: DS does addition (SMPC) and then fetches the results.
I think that in _Scenario 1_ case there might be a need to be added more noise (or the DS might need to spend more budget) than in _Scenario 2_.

"Consider the weights of a Neural Network which are both data and a model and just like personal data the possibility of being able to utilize a model while also protecting it from leaking creates a lot of value and usecases."
I think here we can specify that the weights is 'data about the model'.

#### Introduction to SMPC
NITPICK: "Secure Mutli-party Computation(SMPC)" -> "Secure Mutli-party Computation (SMPC)" (leave one empty space between "Computation" and "SMPC")
MAYBE: It makes sense to say that the "parties" in this scenario are the "DOs"?
MAYBE: Does it make sense in the "Motivations" section to add a clarification that "Data Owner" means "data" (actual dataset) or "model" (weights) owner? -- I am not sure if people would think that `DO` means only a person that has a dataset.
Q: Privacy? (should we also specify the intermediary values from the protocol?) - it might not be that relevant.

### Secret Sharing
OUGHTTO: I would remove the atomic bomb use case and add the hospital use case
MAYBE:  At the two phases I would add:
MAYBE:  "Sharing" - when the party that holds the secrets decides to share the governance over it with others -- (this does not imply that the data owner loses their secret) -- it is more a preparatory step to do further computation between more data.

### SMPC Boolean Circuit Example
MAYBE: Specify how the private bits are constructed.
Alice shares bit `s` such that `s = a ^ b` (ring size 2) 
Bob shares bit "s'" such that `s1 = a' ^ b'`
Here you could also introduce that "+" in ring size two is actually the "^" (xor) or it might be too much?
TYPO?: "Let's break down the problem see how we could compute the basic gates." --> "Let's break down the problem and see how we could compute the basic gates.'"

### XOR Gate
TYPO? "Assume there are two stakeholder S1 and S2 having private inputs would like to compute XOR by secure computation." -"Assume there are two stakeholders S1 and S2 having private inputs and they would like to compute XOR by secure computation." 
MAYBE: I am not sure why we add the stakeholders S1 and S2 - can't we simply say that Alice and Bob have the data (they are the DOs) and DS could find something by doing private computation on their data, but without revealing their data?
NITPICK: "values,hence " --> "values, hence" (leave one empty space after the comma)
NITPICK: "Sharing Phase , The Stakeholders " --> "Sharing Phase, The Stakeholders" (remove one space before the comma)
OUGHTTO: "They generate random bit r independently and construct the Shares as" --> I am not sure if it is understood from the context who are they (specify that the shareholders?)
NITPICK: "S1 constructs shares α ,β" --> Same stuff with white space
Q: "bit r independently", but then there are used "r1" and "r2". Maybe mention "r1" and "r2"?
MAYBE: "The stakeholders give the shares α,α' to Alice and β,β' to Bob" -- specify here that it it does not matter which part of the share (from a secret) each party receives - because no share bit would reveal anything about the secret bit.
TYPO: bob --> Bob
MAYBE: In the picture Ana should have α and α' to be in sync with "The stakeholders give the shares α,α' to Alice and β,β' to Bob" (even if α == a ^ r1)
MAYBE: Sometimes Stakeholders is written with uppercase "S" and sometimes not. Maybe stick to one and if you want to emphasize the stakeholders, then use "bold"?
NITPICK: "γ(gamma)" --> leave empty space before opening parentheses. The same for delta.
NITPICK: "Using Associativity property of xor and rearranging the equations." --> "Using the associativity property..."
MAYBE: Also the equations could use less/more spaces in some places.
NITPICK: "Let's take look at the XOR table" --> "at the XOR table"
MAYBE: There are more places in the text with the "space" problem before/after commas (I will skip those) - also there are some punctuation marks missing in some phrases.
TYPO: "its's"
MAYBE: "Does the former protocol described above does not compute the xor function securely?" Rephrase this.
Q: "Since the XOR function itself ,reveals the private input, it is not a security Leakage" - I do not understand this. Reveals the private input of the other party if the output is revealed.
OUGHTTO: Secure Function Evaluation  --> this concept is not defined.
TYPO: "exchaning"  -- "exchanging"

#### AND Gate
MAYBE: Maybe only specify "Beaver Triple" without random "Beaver Triple"?
MAYBE: Add the "&" sign for AND -- `ab` will become `a & b`
MAYBE: I am not sure what value is the code adding for the explanation -- the text explanation is straightforward.
MAYBE: The sharing x, y and z could be added together -- same rows and different columns -- this also for some others before this. I feel like there is a lot of "empty space" in the notebook.
MAYBE: Remove the pre-processing phase since it is not explained. I think maybe we should focus on the concepts of SMPC and not how we should "improve" the process.
MAYBE: "Alice computes", "Bob Computes" -- change "Computes" to "computes" - it is good to have consistency.

OUGHTTO: 
```
=> γ ⊕ δ
=> (εα' ⊕ ηα ⊕ za ⊕ εη) ⊕ (εβ' ⊕ ηβ ⊕ zb)
Using Associativity of XOR grouping common terms
=> ε(ya ⊕ yb) ⊕ η(xa ⊕ xb) ⊕ (za ⊕ zb) ⊕ εη
=> εy ⊕ ηx ⊕ z ⊕ εη (Expading constant terms)
=> (x ⊕ a)y ⊕ (y ⊕ b)x ⊕ z ⊕ (x⊕a)(y⊕b)
=> xy ⊕ ay ⊕ xy ⊕ bx ⊕ z ⊕ xy ⊕ bx ⊕ ay ⊕ ab (Grouping Common terms)
=> (xy ⊕ xy) ⊕ (bx ⊕ bx) ⊕ (z ⊕ xy)⊕ (ay ⊕ ay) ⊕ ab
=> ab ⊕ (z ⊕ xy) (since z= xy)
=> ab ⊕ (xy ⊕ xy)
```
I am not sure if there is an easier way to showcase this? (maybe underline some of the terms when reducing them?)

#### Dating problem
Q: "a date with other." --> "date with each other"?

Good work Rasswanth! I still have to review from the :point_up_2: section.

rasswanth-s:
Keeping this a placeholder for answering the review comments

Motivations
TO Answer


Introduction to SMPC
✅ NITPICK: spelling SMPC
✅MAYBE  
[PENDING] MAYBE: Does it make sense in the "Motivations" section to add a clarification that "Data Owner" means "data" (actual dataset) or "model" (weights) owner? -- I am not sure if people would think that DO means only a person that has a dataset.
✅  Q: Privacy : Since we mention any abrtirary information apart from the public output, it would encompass the intermediate, but we solidify this intuition in the later part of the notebook

gmuraru (commented):
Left more comments.

Some sentences are ending with a period and some are not (they do not have any marks at the end).

#### AND Gate
TYPO: "Simple" --> "simple"
MAYBE: Use send here rather than "share" -- use "shares" only as a noun in the SMPC context.
"creates shares of each value of the tuple and **shares** it with the parties"
TYPO: " sends it" --> "sends them"
TYPO: "Since Alice and Bob have the shares required to reconstruct " the second reconstruction should use x_b and y_b.
MAYBE: "Observe that the constant term is only added to Alice's share as this prevent duplication of constant term in final output." --> you are not talking about the constant term anywhere. Also there are some "the"s missing.
OUGHTTO: The image is not synchronized with the terms computed.
Q: Why not use all the row to write "Alice and Bob either send each of their shares to the stakeholder for them to reconstruct individually or they send it between themselves and make it a public output." (it seems like there is added a character limit per row)
MAYBE: "Similar to xor gate, we will try exploring the AND gate" -> "XOR gate" (consistency).

#### Explanation 1
MAYBE: I would stick to using "they"/"them" both for Bob and/or Alice

#### Explanation 2 (there is a typo in the title)
Q: "Since , this is revealead by the function output itself, it is not a security leakage" (some typo and the sentence might need more words after "Since")

#### Solving the Original Circuit Example
MAYBE: The image could contain 2 wires coming out from each gate (to show the shares)

#### Security Models
MAYBE: "Semi-honest" I feel like the "language"/words could be simplified. For example "However, they try to learn information from the transcript of the messages communicated in the protocol." --> something around the lines "they could look at intermediary values computed by the protocol, but they should not be able to find stuff about the secret from those"

TBD (I will continue tomorrow)

rasswanth-s:
🟠[PENDING] Placeholder for the second set of comments:

Some sentences are ending with a period and some are not (they do not have any marks at the end).

AND Gate
✅ TYPO
✅ MAYBE
✅ TYPO
TYPO: "Since Alice and Bob have the shares required to reconstruct " the second reconstruction should use x_b and y_b.
MAYBE: "Observe that the constant term is only added to Alice's share as this prevent duplication of constant term in final output." --> you are not talking about the constant term anywhere. Also there are some "the"s missing.
OUGHTTO: The image is not synchronized with the terms computed.
Q: Why not use all the row to write "Alice and Bob either send each of their shares to the stakeholder for them to reconstruct individually or they send it between themselves and make it a public output." (it seems like there is added a character limit per row)
MAYBE: "Similar to xor gate, we will try exploring the AND gate" -> "XOR gate" (consistency).

Explanation 1
MAYBE: I would stick to using "they"/"them" both for Bob and/or Alice

Explanation 2 (there is a typo in the title)
Q: "Since , this is revealead by the function output itself, it is not a security leakage" (some typo and the sentence might need more words after "Since")

Solving the Original Circuit Example
MAYBE: The image could contain 2 wires coming out from each gate (to show the shares)

Security Models
MAYBE: "Semi-honest" I feel like the "language"/words could be simplified. For example "However, they try to learn information from the transcript of the messages communicated in the protocol." --> something around the lines "they could look at intermediary values computed by the protocol, but they should not be able to find stuff about the secret from those"

SMPC in PySyft
MAYBE: "Crypten Project by Facebook" --> leave a link with their repo?

Async Execution Model
Q: "The server side doesn't need to execute async messages , asynchronously, but by doing so, we could have a lot more functionality" -- async messages + asynchronously does not sound too good (I am thinking about the formulation).
MAYBE: Explain that there are tradeoffs ^^ doing stuff async does not come without negative parts.
MAYBE: " by a resource or an output" after that it is "have a pending I/O request,' -- I would rather leave only one of the two sentences (probably only the I/O part -- it feels like both sentences are saying the same thing - that we switch in case we are doing I/O)
MAYBE: Add links to RabbitMQ and Celery
MAYBE: "Syft uses , RabbitMQ as the messaging queue" -- maybe it is better to talk about the concepts in the SMPC in PySyft Chapter or another section with "Remote execution in PYSyft". And in this part present "RabbitMQ", "Celery worker".
OUGHTTO: There are a lot of concepts introduced and I would rather stick to more generic terms. I feel like a new reader will need a lot of time to digest all of those terms.
Some terms that are used with no context/definition/explanation (I am not sure what level it is expected from a Padawan):

gmuraru (commented):
One more round! 😄 

#### Gevent
* MAYBE: Explain libev and eventloop?
#### SMPC in Action
* MAYBE: The output when creating a Tensor looks a little bit unaligned:
  * `Tensor created! You can activate Differential Privacy protection by calling .private() or             .annotate_with_dp_metadata().` (Probably a fix is required in the SyftTensor?)
* MAYBE: When starting hagrid some of the ascii art table does not seem properly aligned? 
* MAYBE: I would skip the "Request for Budget" and introduce it at the end -- how to combine SMPC And DP.
* Q: You wanted to call `get`?
```
# ⚔️ Runnable Code
c.get(delete_obj=False)
c
```
#### MPCTensor
* MAYBE: If the MPCTensor, ShareTensor did not change in PySyft. "The child values of the MPC tensor represents the parties involved in the computation" --> "The child values of the MPCTensor are Pointers to the underlying data (in our case the Shares)".
* Q: Strange that this code does not show "PointerTensor" (in the notebook are the names alice, bob and the ids)
```
# ⚔️ Runnable Code
print(c.child[0])
print(c.child[1])
```
* OUGHTTO: "Let's get the actual data of the TensorPointer of alice."
Both Alice and Bob shares are fetched.
* NITPICK: "Now, let's explore the ShareTensor class and the construction of them" add "full stop" (.) at the end.

#### ShareTensor
NITPICK: "Again , in simple terms , Sharetensor are the troops that execute the command from the MPC Tensor" --> there are a lot of spaces + consistency for naming stuff --> "Again, in simple terms, ShareTensor are the troops that execute the command from the MPC Tensor"
* MAYBE: In the image "SharedTensor" -> "ShareTensor"
* MAYBE: The rectangle from the MPCTensor needs some alignment.
* NITPICK: There are some spaces before/after commas that need to be removed/added.
* OUGHTTO: Remove the "They know to compute the Turing Complete operations" --> There is introduced a new term "Turing Complete" and it should be defined. I think it is better to use a simpler language

#### Pseudo Random Generators (PRGs)
* NITPICK: "revealed(or are predictable)" space between "revealed" and "("
* MAYBE: Try to build a little bit on this "If the random bits are revealed(or are predictable) the entire structure collapses". I was thinking at rephrasing it as the following.
   * "Consider a two party scenario from above. If one of the parties know what random number the other party would use or even predict with a good chance what that number might be then the entire structure collapses"
* NITPICK: "Obtaining, true randomness is expensive" --> no comma here.
* MAYBE: That PRG image could be smaller I think?
* Q: It is expected that they know what a seed is?
* MAYBE: "Let's use the secrets library for generating random numbers, as it is cryptographically secure" --> "We will use the secrets library" (The phrase "Let's" is also on the previous row).
* MAYBE: "Let's generate a small seed to be used to by the generator" (I would remove that "small")

Really great work @rasswanth-s  - Sorry that it is taking me so long to review it.

You can go ahead and merge it.
I will try to review more later in the night!

rasswanth-s:
## Gevent
🟠[PENDING] MAYBE: Explain libev and eventloop?

SMPC in Action
🟠[PENDING] MAYBE: The output when creating a Tensor looks a little bit unaligned:
Tensor created! You can activate Differential Privacy protection by calling .private() or .annotate_with_dp_metadata(). (Probably a fix is required in the SyftTensor?)
🟠[PENDING] MAYBE: When starting hagrid some of the ascii art table does not seem properly aligned?
🟠[PENDING] MAYBE: I would skip the "Request for Budget" and introduce it at the end -- how to combine SMPC And DP.
✅ Q: You wanted to call get? modified to show result

## MPCTensor
✅ MAYBE: Shares name
✅ Q: Strange that this code does not show "PointerTensor" Currently it shows
✅ OUGHTTO: "Let's get the actual data of the TensorPointer of alice."
✅ NITPICK: ShareTensor explanation modified

## ShareTensor

✅ NITPICK: explanation changed and fixed

MAYBE: In the image "SharedTensor" -> "ShareTensor"
MAYBE: The rectangle from the MPCTensor needs some alignment.
NITPICK: There are some spaces before/after commas that need to be removed/added.
OUGHTTO: Remove the "They know to compute the Turing Complete operations" --> There is introduced a new term "Turing Complete" and it should be defined. I think it is better to use a simpler language
Pseudo Random Generators (PRGs)
NITPICK: "revealed(or are predictable)" space between "revealed" and "("
MAYBE: Try to build a little bit on this "If the random bits are revealed(or are predictable) the entire structure collapses". I was thinking at rephrasing it as the following.
"Consider a two party scenario from above. If one of the parties know what random number the other party would use or even predict with a good chance what that number might be then the entire structure collapses"
NITPICK: "Obtaining, true randomness is expensive" --> no comma here.
MAYBE: That PRG image could be smaller I think?
Q: It is expected that they know what a seed is?
MAYBE: "Let's use the secrets library for generating random numbers, as it is cryptographically secure" --> "We will use the secrets library" (The phrase "Let's" is also on the previous row).
MAYBE: "Let's generate a small seed to be used to by the generator" (I would remove that "small")

PR #7095, opened 2022-11-21
ignoring py safety warning

bitsofsteve:
## Description
Fixing the Hagrid PR failing due to safety warnings about `py`

### DIFF
--- a/.github/workflows/hagrid-pr_tests.yml
+++ b/.github/workflows/hagrid-pr_tests.yml
@@ -73,14 +73,14 @@ jobs:
         run: |
           pip install importlib-metadata==4.13.0
           bandit -r hagrid
-          safety check -i 42923
+          safety check -i 42923 -i 51457
 
       - name: Scan for security issues python 3.7
         if: steps.changes.outputs.hagrid == 'true' && matrix.python-version == '3.7'
         run: |
           pip install importlib-metadata==4.13.0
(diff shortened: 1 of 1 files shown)


.github/workflows/hagrid-pr_tests.yml
```diff
@@ -73,14 +73,14 @@ jobs:
...
-          safety check -i 42923
+          safety check -i 42923 -i 51457
 
       - name: Scan for security issues python 3.7
         if: steps.changes.outputs.hagrid == 'true' && matrix.python-version == '3.7'
         run: |
           pip install importlib-metadata==4.13.0
           bandit -r hagrid
-          safety check -i 42923
+          safety check -i 42923 -i 51457
```
rasswanth-s, line 83: Q:  This risk is critical, by the webpage https://pyup.io/v/51457/f17/
Maybe we could upgrade the package which depends on py or some alternatives?

bitsofsteve: @rasswanth-s yeah, the only problem is that packages like Pytest and the rest(...that I don't know yet) install `py`, and since py is the culprit across all versions of it, installing the update of the package that uses py as dependency won't help, since no fix i.e a new version of `py` have not been released. So safety will still flag it.

rasswanth-s (approved):

Issue #6953, opened 2022-11-03
Data Scientist can crash a domain node for all users by forcing the RDP Cache to extend infinitely


IshanMi:

IshanMi:
sigma was set at 0.001 and this created an RDP constant of 41M which would take a LONG time to generate

![Image](https://user-images.githubusercontent.com/32711264/199752273-f16bfef2-4d31-42b8-971a-115d42f90f14.png)

shubham3121:
I had timed it for values in the range 12e6 it was taking like 2.5 hours 😂 for the whole process. I was thinking of solving this in a few ways:
- increase the size of the cache file
- multithreaded the `_get_optimal_alpha_for_constant` computations, as each computation is independent of the other.
- have a background task (via celery) running that increases cache size every few intervals in a separate thread. Although, this may cause memory overload at some time.

madhavajay:
This doesn't seem like something we can quickly fix for 0.7. My suggestion is we revisit this as part of the work to provide the  User access to the Action Graph in 0.8. In the mean time I guess we have to raise an Exception. Forcing a cache rebuild and there by overloading the domain is not really a solution. Also, do we need to extend the entire cache just to solve for a value which is out of range? Can't we just force calculate things outside of the cache but not cache the results?

iamtrask:
There are two quick fixes for this:
1) pre-generate a larger RDP cache (possibly in a thread)
2) the client can deny a .publish() event if it's likely to generate a privacy budget spend higher than the upper bound that the client already knows about.

madhavajay:
I think rather than extending the cache, anything which is outside the cache range should just be passed to `find_alpha_and_eps` sure its going to be slow but its predictably slow where as if you jump really high you generate everything from 1.2 million RDP constants to where ever you jumped inclusive when often you will only want a few values. We can then also introduce a timeout so that if it fails to solve all the RDP Constants for a query by a certain time we raise an Exception. Then at least we arent crashing the server.

Thoughts?

Then for a future task:
- Making that run in parallel in the background will then be easier.
- We can also later change the cache structure so the ranges <50 >50 etc can be seperate arrays. If a user jumps into the distance we can just create a new array at that position, solve the answer and then optionally kick off a multithreaded cache building search around that value range to close the gaps.

iamtrask:
Madhava's solution is the best one. Great thinking sir.

iamtrask:
[...]

PR #7102, opened 2022-11-22
Manually calculate epsilon if the RDP constant is significantly larger than the cache size

IshanMi:
## Description
Fixing issue #6953 - instead of forcing the cache to extend, if it's beyond a certain threshold, we directly calculate the epsilon values for every non-cacheable value in the array.

We determined the cutoff by seeing how long it would take to extend the cache (without docker) and found it took:
- Time taken for 10,000 more values = 4.212839603424072s
- Time taken for 50,000 more values = 18.544428825378418s
- Time taken for 100,000 more values = 33.84507870674133s
- Time taken for 150,000 more values = 53.744325399398804 s

(These numbers might vary based on the hardware, the amount of RAM given to docker, etc)

co-authored by: @ivyclare 

## Affected Dependencies

## How has this been tested?
Largely in notebooks for now

## Checklist


packages/syft/src/syft/core/adp/data_subject_ledger.py
```diff
@@ -375,29 +401,31 @@ def _get_batch_rdp_constants(
...
     def _get_epsilon_spend(self, rdp_constants: np.ndarray) -> np.ndarray:
-        # rdp_constants_lookup = (rdp_constants - 1).astype(np.int64)
         rdp_constants_lookup = convert_constants_to_indices(rdp_constants)
-        try:
-            # needed as np.int64 to use take
-            eps_spend = jax.jit(jnp.take)(
-                self._cache_constant2epsilon, rdp_constants_lookup
-            )
+        if rdp_constants_lookup.max() - len(self._cache_constant2epsilon) >= 150_000:
+            eps_spend = self._fetch_eps_spend_for_big_rdp(rdp_constants)
```
rasswanth-s, line 406: MAYBE: We could pass rdp_constants lookup, to the `_fetch_eps_spend_for_big_rdp` function to save on recalculating the `cache_indices` variable.

rasswanth-s (approved):
Awesome work, Well Done @ivyclare, @IshanMi

Issue #6966, opened 2022-11-04
Publish fails for insufficient budget scenario.


rasswanth-s:
## Description
Publish Operation fails for insufficient budget scenario.
When we try to publish a DP annotated tensor for insufficient budget scenario.we receive a stack trace at the celery work side.
The screenshots contains details to reproduce the error.

## How to Reproduce
![Screenshot 2022-11-04 at 6 02 27 PM](https://user-images.githubusercontent.com/43314053/199973416-64cdbbe3-6f2e-4931-86b6-e8c9fd3dd980.png)

Error Stack Trace:
![Screenshot 2022-11-04 at 6 03 35 PM](https://user-images.githubusercontent.com/43314053/199973646-46f4aa5b-d3c2-457b-bc93-9eecf30c636d.png)

Expected Behaviour:
To be able to publish on insufficient budget scenario.

madhavajay:
The code is supposed to exit the while loop where it tries to reduce the accuracy and remain within budget if it has not successfully reduced the budget spend from the previous iteration to prevent infinite looping. In the fail case it should return noise.

That exception should be caught and the end of the functions generate noise in the shape of the tensor code should still return.

madhavajay:
Fixed by: https://github.com/OpenMined/PySyft/pull/7103

PR #7103, opened 2022-11-22
Refactoring publish function 

### LINKED ISSUE #7050: Publish While Loop Variable Broken
## Description
This PR has changed the behavior of the publish while loop and prevents it from working: https://github.com/OpenMined/PySyft/commit/02d6e83c98d3fd7bc61757c2ab4a15b5c7934f57#diff-3a545178990897379af8b2d29a78b4573397be7f1e84853b7e9f4fa4e57500e4

All references of `tensor` changed to `value` need to change back so that the line:
`tensor = tensor.swap_state(filtered_sourcetree)` can effect the while loop and 
`original_output = tensor.child` can exit with filtered data.

teo-milea:
## Description
Fixes:
- https://github.com/OpenMined/PySyft/issues/7148
- https://github.com/OpenMined/PySyft/issues/7145
- https://github.com/OpenMined/PySyft/issues/6966

The significant changes we've made in this PR are:
- changing the `data_subjects` from `PhiTensor` to `data_subject` as we expect only to have one DataSubject per PhiTensor
- removing `bounds` and `data_subjects` from GammaTensors
- change the `sources` dictionary to include only the `PhiTensors` from which the `GammaTensor` originated
- change `func_str` to `func` which can be reconstructed recursively using a state dictionary which is a mapping between `PhiTensors` ids and the arrays we want to use.
- fixed lipschitz bound search but it is too slow for more than 10 numbers (in `0.8` we need to use more pessimistic bounds that are easier to compute for scale)
- fixed the publishing mechanism
- modified some ledger functions to reflect the new publishing
- added tests for publishing and lipschitz bound search
- *NEW* Pickle 🥒 free JaxOp graph for reconstruct / filter

Left TODOs:
- add some more tests (Ishan is working on that)
- fix the broken parts of the codebase due to our changes (I think 'annotate_with_metadata' and the `Pointer` tensors are the main parts)
- tests that this PR fixed the previous issues


tests/integration/e2e/publish_test.py
```diff
@@ -0,0 +1,67 @@
+# stdlib
+from datetime import datetime
```
IshanMi, line 2: Can we remove this import, since it doesn't seem we're using it in the test file?

tests/integration/e2e/publish_test.py
```diff
@@ -0,0 +1,67 @@
...
+import time
+import uuid
```
IshanMi, line 4: Ditto- is this an unused import?

tests/integration/e2e/publish_test.py
```diff
@@ -0,0 +1,67 @@
...
+    velma_nums = np.array([8, 11, 10, 50, 44, 32, 55, 29, 6, 1])
+    assert len(fred_nums) == len(velma_nums)
+    fred_tensor = sy.Tensor(fred_nums).private(min_val=0, max_val=122, data_subjects="fred")
+    velma_tensor = sy.Tensor(velma_nums).private(min_val=0, max_val=122, data_subjects="velma")
+    
+    domain_client.create_user(
+        name="Scooby",
+        email="scooby@doo.org",
+        password="snacks",
+        budget=0,
```
IshanMi, line 43: Is this 0 privacy budget intentional?

IshanMi (changes requested):
Terrific job making this part of the DP codebase a lot more readable and user friendly!

tests/integration/e2e/publish_test.py
IonesioJunior: Solved by: [3c58940](https://github.com/OpenMined/PySyft/pull/7103/commits/3c5894036a961a55432c12e0b550db843ddef8fe)

tests/integration/e2e/publish_test.py
IonesioJunior: Solved by: [3c58940](https://github.com/OpenMined/PySyft/pull/7103/commits/3c5894036a961a55432c12e0b550db843ddef8fe)

tests/integration/e2e/publish_test.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+
+    ds_client = sy.login(email="scooby@doo.org", password="snacks", port=DOMAIN1_PORT)
+
+    fred_nums_ptr = ds_client.datasets[0]["fred_nums"]
+    velma_nums_ptr = ds_client.datasets[1]["velma_nums"]
+    comb_ptr = fred_nums_ptr + velma_nums_ptr
+    op1 = comb_ptr * 2
+    op2 = op1.sum()
+    op2.publish(sigma=1.5)
```
rasswanth-s, line 62: MAYBE: we could call `op2.get()` to check if the result is available?
we could also add op2.block_with_timeout() as publish might not be instantaneously available before `.get()`

tests/integration/e2e/publish_test.py
teo-milea: Yep, I wanted the publishing to fail for the first instance

tests/integration/e2e/publish_test.py
IonesioJunior: @rasswanth-s , @teo-milea @IshanMi What should we expect to receive since there's no budget for .get()?

IonesioJunior:
Is there anything missing from this PR?

teo-milea:
@IonesioJunior yes, I will add the fix for the filtering system here as well (by tomorrow it should be fixed)

tests/integration/e2e/publish_test.py
rasswanth-s: MAYBE: We could make sure that the computation was completed successfully by `.get()` and that there was no error on the publishing side,  I am also not sure, how we verify the correctness of a noisy result, maybe like checking the max noise bound on the values.

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
```diff
@@ -4806,26 +4225,15 @@ def _bytes2object(buf: bytes) -> GammaTensor:
...
-                )
-            )
-
-            min_val = deserialize(gamma_msg.minVal, from_bytes=True)
-            max_val = deserialize(gamma_msg.maxVal, from_bytes=True)
             is_linear = gamma_msg.isLinear
-            id_str = gamma_msg.id
-            func_str = gamma_msg.funcStr
+            id_str = UID.from_string(gamma_msg.id)
+            import dill
```
madhavajay, line 4230: Can we discuss adding `dill`.

madhavajay (changes requested):
We need to fix all the linting issues and remove dill. I can help but I need to be clear on what code is truely dead.

packages/syft/tests/syft/core/tensor/adp/data_subject_array_test.py
```diff
@@ -1,9 +1,19 @@
...
 import syft as sy
+from syft.core.adp.data_subject import DataSubject
+from syft.core.adp.data_subject import dstonumpyutf8
+from syft.core.adp.data_subject import numpyutf8tods
 from syft.core.adp.data_subject_list import DataSubjectArray
 from syft.core.adp.data_subject_list import dslarraytonumpyutf8
 from syft.core.adp.data_subject_list import numpyutf8todslarray
 
+def test_data_subject_serde() -> None:
+    data_subject = DataSubject("🥒pickles")
```
madhavajay, line 12: 😭

IshanMi (approved):

PR #7107, opened 2022-11-24
Fixed Any Pointer result in publish operation

### LINKED ISSUE #6967: Modifying the Pointer result of Publish Operation.
## Description
The result of Publish operation is an Any pointer, which is  programmatically a correct result, but it is unintuitive for a beginner, 
we should re-enable the `resolve pointer type` service, which resolves the the underlying data in Any Pointer to corresponding result.

Screenshot of output:
Screenshot 2022-11-04 at 6 07 17 PM

rasswanth-s:
## Description
Fixes #6967 
## Affected Dependencies

## How has this been tested?
added a test in trade demo test.

## Checklist


kiendang (approved):
👍 @rasswanth-s. I'm thinking about `PUBLISH_POINTER_TYPE` vs `publish_pointer_type`? Right now it's hard coded as a constant so I do see why `PUBLISH_POINTER_TYPE` is appropriate but in the future if we support more publish types it would not be a constant anymore?

rasswanth-s:
@kiendang , it would be constant also in the future , 
This constant fixing would be same like us fixing the return types of each operation in the AST.
In this case, since the DP pointers are custom subclasses, we fix the return type in the class itself.

PR #7115, opened 2022-11-25
Fix mutation check at server side

### LINKED ISSUE #6976: Incorrect Mutation check by Capnp
## Description
For the `inplace` operations in the `RunClassMethodAction.py` , after the function has completed execution, we do a mutation check by serializing the original data before and after the computation, to check if an `inplace` operation has happened by which we modify the original data in the database.
Code reference:
https://github.com/OpenMined/PySyft/blob/6f92240c336ad233cdef3e4e739cb1d15420b6fa/packages/syft/src/syft/core/node/common/action/run_class_method_action.py#L238

`Protobuf` seemed to generate the same bytes value, when we serialize the same object   several times, but `pycapnp` serialized objects tends to have different bytes when we serialize the same data.

This inturn causes it to think , it is mutated and we store the same data again to the Database.

The screenshot contains the serialization example with `pycapnp`

Screenshot 2022-11-05 at 2 31 43 PM

## Expected Behavior
To be able to detect correctly mutated and non mutated data and modify the database data accordingly

## Bonus Performance Point
Serialization happens at four stages at RunClassMethod.

data = Original data to operate on .
- Deserialization of `data` to operate on from the DB
- Serialization of the data as resolved_self_previous_bytes( for mutation check)
- Serialization of the data after the computation has completed(for mutation check)
- Serialization of the result of the method of RunClassMethod.

Serialization is costly when we work with large datasets, migrating from serialization to check for mutation would give a `major performance hit`.

rasswanth-s:
## Description
Fixes #6976 
For all the Py-primitives(int,float,slice) we associate a  random UID, during object creation.
This causes incorrect mutation checks during Function execution(RunClassMethodAction,FunctionorConstructorAction...)
Storing the same object back to redis.

## How has this been tested?
Added tests for the same

## Checklist


rasswanth-s:
Working example of the bug reported!
Screenshot 2022-11-25 at 10 44 50 AM

tudorcebere:
Great work @rasswanth-s ! I think this is a nice solution until we merge 0.8.0. You killed plenty of dead code. One remark - do you think there is any way to catch this behaviour reliably? I am pretty sure it only happened in the context of upcast.

Another line of work that would significantly improve this yet again would be to drop the double serialization to check for mutability. I think that at some point, all the serializable objects should be hashable (this feature would be cool for plenty of reasons + we can derive this quite easily I believe - we know the fields we care about, we just add a `__hash__` function that recursively hashes the underlying selected objects) - one would be to check for mutability without the time/performance hit. I am sure we should definitely explore that for 0.8.0. What do you think?

rasswanth-s:
To catch this reliably we might have to add serde bytes check to all our custom classes, but it depends explicitly on the custom class if they tend to have random UID, like the one with Py Primitives.

I have wanted for a long time to remove the double serde in RunClassMethodAction, totally sede happens `four` times for a single operation 😅. Yep totally, we could definitely explore  that, for 0.8

tudorcebere (approved):
LGTM!

tudorcebere:
I will create a ticket to explore the removal of serialization when we check for immutability, I think that making our objects hashable would helps us in plenty of places.

IonesioJunior (approved):
Remarkable and high-quality work @rasswanth! I loved that during the bug resolution you also managed to reduce our overall tech debt by removing code duplications and dead code. Amazing work!

PR #7135, opened 2022-11-30
ADD basic Network frontend page

IonesioJunior:
## Description
This PR adds a simple HTTP frontend web page so that when users go to it they know it's not broken.

## How has this been tested?
- Hagrid API check
hagrid_check
- Trying to access it via browser.
browser_check

## Checklist


madhavajay:
@IonesioJunior I tweaked this a bit to use an nginx static config for the network node so we have no code now but can re-use the same frontend build and container as domains for future.

The tricky part was getting it to respond to the `NODE_TYPE` env variable but it seems as though nginx now supports this automatically on startup if you use their `.template` files.

Until we have a frontend for the network, network dev mode will simply use the production container which itself has an nginx network config.

madhavajay (approved):
🙌🏽

Issue #6921, opened 2022-10-31
UX Improvement: More hand-holding. When creating a Syft Tensor, tell DS they can annotate. After annotate, tell DS they can load_dataset


iamtrask:
It isn't obvious how a new PySyft user would know they had to call `.private` or `.annotate` on their Syft Tensor. Perhaps we could do some signposting and when a new Syft Tensor is created, we could print something like "Syft Tensor created! To activate Differential Privacy protection, please call `.annotate` and load into a domain node!"

iamtrask:
I like this idea. I think this and other queries might be best turned on/off by a syft.helper=True flag (which could default to True)

iamtrask:
This was actually opened by @IshanMi

madhavajay:
We could also persist this `syft settings` to their `~` home directory the same way we do with `hagrid`.

iamtrask:
> We could also persist this `syft settings` to their `~` home directory the same way we do with `hagrid`.

Stretch

madhavajay:
I have had to disable this for now because every time a tensor gets created with internal code it prints. For SMPC this can result in hundreds of the same message getting printed.

https://github.com/OpenMined/PySyft/pull/7086

Let's create a new PR which enables this but does not trigger. Perhaps we could have a global timeout or something similar to prevent it from flooding prints?

iamtrask:
Relying on a syft.helper = True flag seems like a reasonable middle ground. Default it to False and set it to True in all our tutorials.

PR #7136, opened 2022-12-01
[W.I.P] Add syft helper for handholding/messages

IshanMi:
## Description
Adding a syft.helper based on issue #6921  

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/notebooks/Experimental/Ishan/HelperMessages.ipynb
+++ b/notebooks/Experimental/Ishan/HelperMessages.ipynb
@@ -1,0 +1,142 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "a796c05c-dbbc-4883-b6a7-6da5af6c901c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import syft as sy"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "0443ee0c-03b1-4efb-8fd7-5cec073b05d4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "id": "4ee5a936-4554-4cba-ab58-3e90985f1fa9",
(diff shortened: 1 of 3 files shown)


packages/syft/src/syft/user_settings.py
```diff
@@ -0,0 +1,41 @@
...
+root_path = os.path.dirname(os.path.abspath(__file__))
+
+# get the .env file path
+env_path = os.path.join(root_path, ".env")
+
+# load the .env file
+load_dotenv(dotenv_path=env_path)
+
+# TODO: Add this to a file in ~/.syft and add some ENV overrides
+tutorial_mode = bool(os.environ.get("SYFT_TUTORIAL_MODE", False))
```
madhavajay, line 23: I think the default was True?

packages/syft/src/syft/user_settings.py
```diff
@@ -0,0 +1,41 @@
...
+# stdlib
+import os
+
+# third party
+from dotenv import load_dotenv
+
+# relative
+from .core.common.decorators import singleton
+
+# Get root path of project
```
madhavajay, line 13: Let's remove the .env loading stuff for now this looks like something that might cause some issues. I think we want something like ~/.syft/settings.json but lets look into it as a seperate task. For now, lets put SYFT_TUTORIAL_MODE=False in all the docker container environment variables and later when we notice this in notebooks we can add something like sy.settings.helper = False or something.

packages/syft/setup.cfg
```diff
@@ -55,6 +55,7 @@ syft =
     tqdm==4.64.0
     typing_extensions==4.1.1 # backport to older python 3
     pyzmq==23.2.1 # https://github.com/zeromq/pyzmq/issues/1764
+    python-dotenv==0.21.0
```
madhavajay, line 58: I think we want to remove this for now.

packages/syft/src/syft/user_settings.py
madhavajay: @bitsofsteve

bitsofsteve: Okay! @madhavajay

Issue #7140, opened 2022-12-02
Hagrid command line argument names should use all hyphens instead of snakecase


kiendang:
## Description
Things like `--auth_type`, `--ansible_extras` should be `--auth-type` and `--ansible-extras`. 3 reasons for this

1. This is how options in most cli tools are defined.
2. Extra key stroke, typing `-` is much faster than `_`
3. Most importantly, there could some options that Hagrid passes to the underlying cli tools. For example I was working on implementing k3d and k8s support for hagrid and there are some options that are passed to `k3d` or `kubectl`, `devspace` cli, for example `--registry-port`, `--registry-volume`. Setting them as `--registry_port`, `--registry_volume` would be a source of confusion to users who were familiar with those tools.

## Are you interested in working on this improvement yourself?
Yes. Please see # for planning.

## Additional Context
Add any other context or screenshots.

PR #7169, opened 2022-12-11
Change hagrid ArgCache api to dict-like item instead of attr

kiendang:
- `arg_cache[key] = value` instead of `arg_cache.key = value`
- use `self.__cache` for internal cache instead of `self.__dict__` to avoid potential conflicts.

~This would make #7140 easier.~ Not required anymore. See #7170. There might still be advantages of using a dict like interface over attribute:

- can use any string as key. attr cannot support keys such as `hyphen-separated-string`
- attributes are more suitable for pre-defined, not dynamic keys. key-value pairs stored as attributes might clash with other attributes of the object. using attr when users `argcache.key = ...` we need to have some heuristics to determine if `key` belongs to the cache or is it an exising attribute of `argcache`. in the code currently we do `if not key.startswith("__")`. with dict we don't have to worry about this at all.

## Description

## Affected Dependencies

## How has this been tested?

## Checklist


madhavajay:
Awesome work!

PR #7213, opened 2023-01-12
Added JS SDK prototype

madhavajay:
## Description
- Added flask server
- Added capnp-ts library
- Added capnpc-ts compiled address.capnp js file
- Requires libcapnp, capnp and nodejs


madhavajay:
Moved to here:
https://github.com/OpenMined/PySyft/pull/7297

PR #7214, opened 2023-01-12
add click upgrade to hagrid ansible task

bitsofsteve:
## Description
This PR tends to fix an issue with Hagrid deploying to azure. During hagrid deploy the package click seems to be missing , even after the Hagrid setup ansible task already ran. This can cause the hagrid deploy to azure to fail. 

To reproduce. 

run `hagrid launch domain to azure --node-count 7` (the node count is totally optional, and go grab a beer or something)

Output (Don't bother with the IP's I already nuked them :))
Screenshot 2023-01-12 at 09 48 04

A bit of a digging(Thanks to @shubham3121 's support) shows that click seems to not have been installed. 
Screenshot from 2023-01-12 16-00-30


packages/grid/ansible/roles/containers/tasks/hagrid.yml
```diff
@@ -8,3 +8,8 @@
   shell: "runuser -l {{ om_user }} -c 'pip install -e {{ syft_dir }}/packages/hagrid'"
   become: yes
   when: vagrant is defined and install == "true"
+
+- name: Install HAGrid
```
madhavajay, line 12: This is identical to the rule above and its also the wrong rule as this one will only run if `vagrant` is defined which we use in local Vagrant vm mode.

shubham3121:
@bitsofsteve, found that the Click version on the azure machines is 7.0.0 which does not support `context_settings={"show_default": True}` as arguments on the `click.command` method. Versions > 7.1.0 support this argument. Upgrading the version of click should fix this.

packages/grid/ansible/roles/containers/tasks/hagrid.yml
bitsofsteve: @madhavajay that was a weird text override.. was meant to be a click upgrade task.

bitsofsteve:
> @bitsofsteve, found that the Click version on the azure machines is 7.0.0 which does not support `context_settings={"show_default": True}` as arguments on the `click.command` method. Versions > 7.1.0 support this argument. Upgrading the version of click should fix this. Screenshot from 2023-01-13 14-18-55

Yeah @shubham3121 , I tried pinning it, but it still wouldn't install a click version > 7.0.0

shubham3121 (approved):

packages/grid/ansible/roles/containers/tasks/hagrid.yml
```diff
@@ -8,3 +8,7 @@
   shell: "runuser -l {{ om_user }} -c 'pip install -e {{ syft_dir }}/packages/hagrid'"
   become: yes
   when: vagrant is defined and install == "true"
+
```
madhavajay, line 11: Can we delete this now that its not needed.

packages/hagrid/setup.py
```diff
@@ -11,7 +11,7 @@
 
 packages = [
     "ascii_magic",
-    "click",
+    "click>7.1",
```
madhavajay, line 14: We probably want >=7.1

madhavajay (changes requested):
Two small changes and were good to go.

PR #7259, opened 2023-01-27
[k8s] bootstrap secrets for k8 pods

yashgorana:
## Description

Closes https://github.com/OpenMined/Heartbeat/issues/8

- Provision secrets through `kubectl create secret generic --from-file"
- Update bootstrap.py to create a credentials file
- Fix tox new update error for "passenv"
- Enable K8s integration test

## Affected Dependencies
N/A

## How has this been tested?
- `tox -e stack.test.integration.k8s`
 
## Checklist


yashgorana:
@madhavajay this is up for review - please take a look at it. Will push another commit to fix the linting issue

packages/grid/devspace.yaml
```diff
@@ -364,10 +364,19 @@ deployments:
...
+              secretName: backend-credentials
         containers:
           - image: ${CONTAINER_REGISTRY}${DOCKER_IMAGE_BACKEND}
             command:
               - /start.sh
+            volumeMounts:
+              - containerPath: /credentials
+                volume:
+                    name: credentials-data
+                    readOnly: true
```
robert-wagner, line 379: Is there a reason this is readonly but the others are not?

iamtrask: Why hello there.

yashgorana: Good catch, ideally this should be `true` everywhere as Node IDs & Pvt key should be read-only. The bootstrap script already prevents overrides.

[Devspace docs](https://www.devspace.sh/component-chart/docs/guides/persistent-volumes#secret-volumes) doesn't really mention anything about flagging secrets as read only, but [Kubernetes docs](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod) explicitly states that we should set `.spec.containers[].volumeMounts[].readOnly = true`. So given what we are already doing + what's recommended this should be `true`

robert-wagner: Caveat being that I had not heard of devspace before a minute ago but is there a reason we do this as a volume mount?

yashgorana: This was discussed between Madhava, Shubham & I. Sharing a credentials volume was one of them: this required bringing all backend containers into a single pod, but that would create conflict between backend & backend-stream containers and was not recommended by Madhava. The method I implemented is simple and one of the ways [suggested by Kubernetes](https://kubernetes.io/docs/concepts/configuration/secret/#uses-for-secrets), while also keeping the internal deployment arch unchanged & pretty straightforward to deploy.

PS - the readOnly true/false mismatch that you pointed out is actually a leftover when I was exploring the first approach :)

robert-wagner: I think I'm a bit confused. It seems to me that you are using a shared volume for the credentials (which you mentioned is not what you want to do). I see in the tox.ini that you create a k8s secret but I don't see where you reference that secret

yashgorana: The volumes aren't shared across pods, they just have the same name. Volume sharing can only happen for multiple _containers_ in a pod (which is not what we want to do)

```
volumes:
  - name: credentials-data
    secret:
      secretName: backend-credentials
```

Here , Kubernetes/Devspace will look for `backend-credentials` secret, which is provisioned through the `kubectl create secret ...` command, and mount them. Hope this clears things up :)

robert-wagner: I totally missed the secret line. This makes a ton more sense now. Thanks for the clarification!

robert-wagner (approved):
Overall lgtm, that being said I haven't had write access to this repo for a while now (so someone else should review this)

madhavajay:
@yashgorana I moved the PR here: https://github.com/OpenMined/PySyft/pull/7295

PR #7261, opened 2023-01-27
Fix check for empty function parameter annotation

kiendang:
## Description
The current check for empty function parameter annotation doesn't work since `param.annotation` is not a class/type and passing it to `issubclass` will throw `TypeError`. The idiomatic way to do this is to check if it is [`inspect.Parameter.empty`](https://docs.python.org/3/library/inspect.html#inspect.Parameter.empty), which is also the same as `inspect._empty` but actually documented. See https://stackoverflow.com/questions/62010367/python-check-if-signature-annotation-is-of-specific-class.

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/packages/syft/src/syft/core/node/new/api.py
+++ b/packages/syft/src/syft/core/node/new/api.py
@@ -152,7 +152,7 @@ def generate_remote_function(signature: Signature, path: str, make_call: Callabl
                 t = param.annotation
             msg = None
             try:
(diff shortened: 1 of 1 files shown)


teo-milea (approved):

madhavajay:
@kiendang great catch and interesting to learn more about the signature api.

PR #7285, opened 2023-02-02
Custom Policies

teo-milea:
## Description
This is the work for the first Policy Demo. Tasks left to do:
- [x] Create the Policy objects needed for working with policies
- [x] Implementing messaging API for allowing Policy transmission
- [x] Create and store policy on task approval
- [ ] Implement a .public_state message
- [x] Added policy evaluation on task execution
- [x] Develop a new mock notebook for a long-term vision 

## Checklist


packages/syft/setup.cfg
```diff
@@ -61,6 +61,7 @@ syft =
     tqdm==4.64.0
     typeguard==2.13.3
     typing_extensions==4.4.0
+    ipython==8.5.0
```
madhavajay, line 64: Do we need to make this mandatory with the syft package?

packages/syft/src/syft/core/common/serde/deserialize.py
```diff
@@ -9,6 +10,7 @@ def _deserialize(
     blob: Any,
     from_proto: bool = True,
     from_bytes: bool = False,
+    class_type: Type = type(None),
```
madhavajay, line 13: I can't seem to find where this is being used. What is it for?

packages/syft/src/syft/core/node/common/node_service/task_submission/task_submission.py
```diff
@@ -217,25 +226,126 @@ def run(  # type: ignore
...
+
         task = node.tasks.find(search_params={"uid": self.payload.task_uid})
         if not task:
             raise Exception(f"The given task_id:{self.payload.task_uid} does not exist")
         task = task[0]
 
+        init_state = init_policy_state(task.policy_code, task.policy_name)
+        # TODO use our better serialization for the state variable
+
+        node.tasks.update(
```
madhavajay, line 238: Is this the deprecated API?

packages/syft/setup.cfg
teo-milea: I tried to eliminate it and I had an issue with running kernels in vscode. We can eliminate it and keep in mind if it comes up as an issue for the users

packages/syft/src/syft/core/node/common/node_service/task_submission/task_submission.py
teo-milea: yes, I will eliminate it, I was not sure if there was anyone using that file, but I guess this is not he case anymore

packages/syft/src/syft/core/common/serde/deserialize.py
teo-milea: so in order to store the state of an instantiated policy inside the database without too much effort, I am inserting the serializable decorator in the code string to use that. However, whenever I try to deserialize on code executed with `exec`, it seems that the new type cannot be found in the type bank. I didn't invest too much time ensuring that the type bank is correctly updated in this niche situation, as I already know what type I need. I use this optional argument to directly pass the type for user submitted policies at https://github.com/OpenMined/PySyft/blob/01ee892a0693180a0fa97de2ed649c8af7f06091/packages/syft/src/syft/core/node/new/policy.py#L202

madhavajay (approved):
Excellent work. That was a mountain but we climbed it. 🙌 There is no doubt that we solved some of the harder work that is needed in future for custom classes, so I think it was worth doing it now.

PR #7286, opened 2023-02-03
Added pretty printing to SyftAPI and SyftObject

madhavajay:
## Description
- Enabled args and kwargs directly to Services endpoints
- Added cache flag to login
- Added initial SyftError type for Service Union return types


shubham3121 (approved):
🎉

PR #7289, opened 2023-02-06
Code Execution Requests

rasswanth-s:
## Description
Integration of Cross Domain Manual  Code Execution Requests in Enclaves

## How has this been tested?
Tests to be Added

## Checklist


packages/syft/src/syft/core/node/new/task/task.py
```diff
@@ -0,0 +1,78 @@
...
+    node_uid: UID
+
+    @staticmethod
+    def from_client(client: AbstractNodeClient):
+        return NodeView(name=client.name, node_uid=client.node_uid)
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, NodeView):
+            return False
+        elif self.name == other.name and self.node_uid == other.node_uid:
```
gmuraru, line 33: MAYBE: Remove the `elif` and simply do return

packages/syft/src/syft/core/node/new/task/task_service.py
```diff
@@ -0,0 +1,400 @@
...
+        code: str,
+        outputs: List[str],
+        owners: List[NodeView],
+        task_id: UID,
+        oblv_metadata: Optional[dict] = None,
+    ) -> Result[Ok, Err]:
+        """Enclave Submit task"""
+        # TODO 🟣 Check for permission after it is fully integrated
+
+        if owners and not isinstance(owners, list):
```
gmuraru, line 86: Q: In what case we might not have a list of owners

packages/syft/src/syft/core/node/new/task/task_service.py
```diff
@@ -0,0 +1,400 @@
...
+                skip_permission=True,
+            )
+            if domain_input.is_ok():
+                inputs.update(domain_input.ok().base_dict[enclave_task.id])
+            else:
+                return domain_input.err()
+        return inputs
+
+    def execute_task(self, enclave_task: Task, context: AuthedServiceContext) -> None:
+        global stdout_
```
gmuraru, line 320: OUGHTTO: There is no way to remove those?

Maybe keeping another variable where we collect the stdout and stderr and then collect them in another module?

packages/syft/src/syft/core/node/new/task/task_service.py
```diff
@@ -0,0 +1,400 @@
...
+
+        code = enclave_task.code
+        inputs = self.fetch_private_inputs(enclave_task=enclave_task, context=context)
+        outputs = enclave_task.outputs
+
+        try:
+            logger.info(f"inital outputs: {outputs}")
+
+            # Check overlap between inputs and vars
+            global_input_inter = set(globals().keys()).intersection(set(inputs.keys()))
```
gmuraru, line 331: Q: I don't think you need to use `set` on the `inputs.key()`/`vars.keys()` . I think it should already be a set.
(also for the next line).

packages/syft/src/syft/core/node/new/task/task_service.py
```diff
@@ -0,0 +1,400 @@
...
+            # byte_code = compile_restricted(code, "<string>", "exec")
+            # exec(byte_code, restricted_globals)
+            exec(code)  # nosec
+
+            for output in outputs:
+                logger.info(f"variable: {output} result: {vars()[output]}")
+                outputs[output] = vars()[output]
+
+            # restore stdout and stderr
+            sys.stdout = stdout_
```
gmuraru, line 361: Q: Where is `sys.stdout` and `sys.stderr` overwritten?

packages/syft/src/syft/core/node/new/task/task_service.py
```diff
@@ -0,0 +1,400 @@
...
+            enclave_task.outputs = {"output_id": new_id}
+            enclave_task.execution = "Done"
+            self.task_stash.update(enclave_task)
+        except Exception as e:
+            sys.stdout = stdout_
+            sys.stderr = stderr_
+            raise e
+            print("Task Failed with Exception", e)
+        finally:
+            sys.stdout = stdout_
```
gmuraru, line 389: MAYBE?: I think that having `finally` should also cover the `Exception` case (I think you can remove the `sys.stdout = stdout_` and `sys.stderr = stderr` from the `except Exception as e` case.

packages/syft/src/syft/oblv/deployment_client.py
```diff
@@ -221,6 +228,65 @@ def sanity_check_oblv_response(self, req: requests.Response):
...
+        if inspect.isfunction(code):
+            code = inspect.getsource(code)
+
+        # Step1 : Restructing inputs
+        node_input_map = {}
+        clients = inputs.keys()
+        # TODO: 🟣 Simplify code structure for inputs fields
+        # TODO: 🟣 Additional sanity checks for inputs fields
+        for client, client_input_map in inputs.items():
+            if not isinstance(client, AbstractNodeClient):
```
gmuraru, line 249: MAYBE: Could we move the validation to another function? (also the one from 255)?

In this way we keep separate the validation logic and the client mapping logic.

gmuraru:
MAYBE: We should also add some tests?

packages/syft/src/syft/core/node/new/task/task.py
rasswanth-s: Done in [b1c9af5](https://github.com/OpenMined/PySyft/pull/7289/commits/b1c9af5f30c99117d59aa92dc15e8c9698d49b96)

packages/syft/src/syft/core/node/new/task/task_service.py
rasswanth-s: Good catch, with the new services architecture we would do the type checking before sending a request on the client side itself, I mainly wanted to sanity check that they are passing at least one owner when doing computation on an enclaves
I have modified it in [b1c9af5](https://github.com/OpenMined/PySyft/pull/7289/commits/b1c9af5f30c99117d59aa92dc15e8c9698d49b96)

packages/syft/src/syft/core/node/new/task/task_service.py
```diff
@@ -0,0 +1,428 @@
...
+            )
+            if domain_input.is_ok():
+                inputs.update(domain_input.ok().base_dict[enclave_task.id])
+            else:
+                return domain_input.err()
+        return inputs
+
+    def execute_task(self, enclave_task: Task, context: AuthedServiceContext) -> None:
+        global stdout_
+        global stderr_
```
kiendang, line 336: `stdout_` and `stderr_` are only used in this function. We can just put

```py
stdout_ = sys.stdout
stderr_ = sys.etderr
```

here instead of using `global`? Similar to 43746afcb1498edb175d31a6a014a904f6874b20

packages/syft/src/syft/core/node/new/task/task_service.py
kiendang: One problem is the print statement would still use the replaced `sys.stdout/err`, but we can do `print("Task Failed with Exception", e, file=stderr_) # or stdout_` in addition to removing the duplicated lines.

packages/syft/src/syft/core/node/new/task/task_service.py
rasswanth-s: @kiendang , Initially , we implemented the old task service system.
Madhava suggested on porting to the new user code service..
This code would soon be deleted
Since the user code service was being developed parallely, 
We thought we could merge this PR and work on  porting to the new user code service in a new task/PR...

kiendang: I see. That explains why I see so much duplicated code with the code submission api.

PR #7305, opened 2023-02-10
HB30 - Fixed error with serde when trying to serialize signature with typing.Dict

the-elancier:
## Description
Closes https://github.com/OpenMined/Heartbeat/issues/30

When we tried to serialize a signature with typing.Dict (as in the code snippet below), it was failing. It threw the following error: `Exception: typing._SpecialGenericAlias not in TYPE_BANK`

```
import syft as sy
from typing import Dict, Union
import result
import inspect

def func(nested_dict: Dict[int, Dict]) -> Union[result.result.Ok[result.result.Ok]]:
    return ""
                                                
sig = inspect.signature(func)

sy.serialize(sig, to_bytes=True)

```

Fixed it with @shubham3121 's help.

## Checklist


packages/syft/src/syft/core/common/serde/recursive_primitives.py
```diff
@@ -14,20 +15,17 @@
...
-
-# stdlib
 from typing import cast
 
 # relative
 from .capnp import get_capnp_schema
 from .recursive import recursive_serde_register
 
+# import types unsupported on python 3.8
+_UnionGenericAlias = getattr(typing, "_UnionGenericAlias", None)
```
madhavajay, line 25: Should we also attempt to add Serde for _UnionGenericAlias below?

madhavajay: add `if sys.version_info < (3, 9):`

PR #7352, opened 2023-02-22
Refactored LinkedDocument to be LinkedObject

madhavajay:
## Description
- Added service types and lookups

### DIFF
--- a/packages/syft/src/syft/core/node/new/linked_obj.py
+++ b/packages/syft/src/syft/core/node/new/linked_obj.py
@@ -1,0 +1,78 @@
+# stdlib
+from typing import Any
+from typing import Optional
+from typing import Type
+
+# third party
+from typing_extensions import Self
+
+# relative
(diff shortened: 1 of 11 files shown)


packages/syft/src/syft/core/node/new/linked_obj.py
```diff
@@ -0,0 +1,78 @@
...
+    __canonical_name__ = "LinkedObject"
+    __version__ = SYFT_OBJECT_VERSION_1
+
+    node_uid: UID
+    service_type: Type[Any]
+    object_type: Type[SyftObject]
+    object_uid: UID
+
+    def __str__(self) -> str:
+        return f"<{self.service_type}: {self.object_uid}@<Node: {self.node_uid}>"
```
shubham3121, line 27: I think this should be {self.object_type} instead of {self.service_type}

shubham3121 (approved):
LGTM 💯 !!

PR #7385, opened 2023-03-01
added __mul__ operation

letv3:
This pull request contains the implementation of the __mul__ operation for PhiTensor and GammaTensors

### DIFF
--- a/packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
+++ b/packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
@@ -2161,6 +2161,36 @@ class GammaTensor:
     def divmod(self, other: Any) -> Tuple[GammaTensor, GammaTensor]:
         return self.__divmod__(other)
 
+    def __mul__(self, other: SupportedChainType) -> GammaTensor:
+        from .phi_tensor import PhiTensor
+        output_state = dict()
+        output_state[self.id] = self
+        if isinstance(other, PhiTensor):
+            other = other.gamma
+
+        if isinstance(other, GammaTensor):
+            output_state[other.id] = other
+
+            child = self.child * other.child
+            min_val = self.min_vals * other.min_vals
+            max_val = self.max_vals * other.max_vals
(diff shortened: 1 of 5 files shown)


packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
```diff
@@ -2161,6 +2161,36 @@ def __divmod__(self, other: Any) -> Tuple[GammaTensor, GammaTensor]:
...
+        output_state = dict()
+        output_state[self.id] = self
+        if isinstance(other, PhiTensor):
+            other = other.gamma
+
+        if isinstance(other, GammaTensor):
+            output_state[other.id] = other
+
+            child = self.child * other.child
+            min_val = self.min_vals * other.min_vals
```
iamtrask, line 2175: One possibly non-intuitive aspect here is signs. If min_vals and max_vals are both positive, both negative, or mix between positive and negative then it can flip whether it's min or max.

iamtrask (changes requested):
Exciting contribution to an important part of the codebase. I'm not the most up-to-date on the state of this part of the code ATM but I did have one meaningful piece of feedback. Great work!

packages/syft/src/syft/core/tensor/autodp/gamma_tensor.py
letv3: Hey @iamtrask  this PR is the `Trial of Spirit` from the padawan program (no new contributions).
However, it is interesting how we should deal with the combinations min_val and max_val
Here is the table of options.


|Firts Val | Seconda Val| Outcome
|---|---|---|
|positive| positive | positive |
|positive|negative|negative|
|negative|positive|negative|
|negative|negative|positive|

In cas of positive its obvious. In other cases, Simple `AND` operation, will change the outcome.
Strainght forward approach is to remember what value was negative and than preform `__mull__` on absolute values (but not sure if the result will be right).

madhavajay:
@letv3 great work! 🙌

PR #7393, opened 2023-03-03
Fix login 

IonesioJunior:
## Description
Due to PySyft API changes, we also had to update the way the frontend handles messages. This PR aims to fix and improve JSClient and the `/login` route.


madhavajay:
Dev frontend wont build at the moment so im merging this so we can get CI up and running again.

PR #7400, opened 2023-03-07
Fix UI User Setttings

IonesioJunior:
## Description
This PR aims to fix Account settings properties enabling the ability to change User information using PySyft UI using the new PySyft API and message structure.


tcp (approved):
👍

PR #7408, opened 2023-03-08
Fix config page

IonesioJunior:
## Description
This PR aims to update the /config route by updating the javascript client requests and its way to interact with PyGrid Node via services. This PR includes:

- New Metadata Service (including MetadataStash, MetadataService, and new NodeMedataUpdate struct)
- JSClient method to update Node metadata Information
- /config fixes to work with this new API
- small lint fixes

This PR uses `integrate_signup` as the base branch, it should wait until integrate_signup is reviewed and merged.


packages/syft/src/syft/core/node/new/metadata_service.py
```diff
@@ -0,0 +1,70 @@
...
+                new_metadata = current_metadata[0].copy(
+                    update=metadata.dict(exclude_unset=True)
+                )
+                update_result = self.stash.update(new_metadata)
+                if update_result.is_ok():
+                    return Ok(result)
+                else:
+                    return SyftError(message=update_result.err())
+        else:
+            return SyftError(message=result.err())
```
IonesioJunior, line 70: @madhavajay, could you take a look at this class? The purpose of creating this service is to modify the properties and state of the metadata. Currently, we cannot change it. Although I remember that we were not supposed to add any new features, this service was already developed and was lost during the API changes. I thought we could reintroduce it now. Please let me know your thoughts.

packages/syft/src/syft/core/node/worker.py
```diff
@@ -533,6 +538,32 @@ def queue_task(
...
+                deployed_on=datetime.now().date().strftime("%m/%d/%Y"),
+            )
+            result = metadata_stash.set(metadata=new_metadata)
+            if result.is_ok():
+                return result.ok()
+            return None
+    except Exception as e:
+        print("create_worker_metadata failed", e)
+
+
```
IonesioJunior, line 566: @madhavajay  I created this method to persist the metadata for the first time. I chose to declare it here since we also declared the `create_admin_new` method in this same location. Do you think we should move it to another location instead? Let me know your thoughts.

tcp (approved):
👍

PR #7411, opened 2023-03-09
Partition key bug

teo-milea:
## Description
When we try to create two objects that have the same user_verify_key but different other keys, the store will return an error because it will only match that key. I added a quick fix, but it seems the problem is deeper. 

## How has this been tested?
I have added a notebook to replicate the issue under notebooks/bugs

## Checklist

### DIFF
--- a/notebooks/bugs/partition_key_duplication_error.ipynb
+++ b/notebooks/bugs/partition_key_duplication_error.ipynb
@@ -1,0 +1,214 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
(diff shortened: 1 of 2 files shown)


madhavajay:
@kiendang yes im pretty sure this just needs to be moved to `__attr_searchable__` since if it was unique then a user could only upload 1 UserCode object? @teo-milea is that the issue you were having?

madhavajay:
@kiendang yes im pretty sure this just needs to be moved to `__attr_searchable__` since if it was unique then a user could only upload 1 UserCode object? @teo-milea is that the issue you were having?

teo-milea:
@madhavajay yep that was the problem

kiendang:
@teo-milea A change similar to #7399 would fix this.

PR #7427, opened 2023-03-16
service access control

koenvanderveen:
## Description
The goal of this task is to create unit tests which will test the @service_method decorator and its endpoints and adding some kind of simple role gate such that service methods can have some RBAC controls applied.

```
@service_method(path="user.create", name="create", roles=[GUEST, DATA_OWNER])
def create(
    self, context: AuthedServiceContext, user_create: UserCreate
) -> Union[UserView, SyftError]:
```
 

- [x] Add roles GUEST, DATA_SCIENTIST, DATA_OWNER
- [x]  Add roles to service method decorator
- [x]  Add Tests to reject access endpoints which aren't allowed
- [x]  Update SyftAPI.for_user() to filter available endpoints based on users ROLE
- [ ]  Add some kind of cache invalidation / update of this API after someone has logged in or changed roles
- [ ]  Follow the User CRUD operations process through to ensure users can't elevate their roles unless they are DATA_OWNERS

## How has this been tested?
Tests for accessing apis with different permissions

## Checklist


koenvanderveen:
@madhavajay @shubham3121 with this PR some parts of the API become a bit more closed off. I expect that when writing test something like this would be super common

```
node = Worker()
guest_client = node.root_client.guest()
guest_client.register("x@y.com", "mypw")
user_id = #get user id
guest_client.login("x@y.com", "mypw")
node.root_client.services.api.users.update(user_id, UserUpdate(role=ServiceRole.DATASCIENTIST))

# do stuff with guest client, accept stuff with root_client
```  

Earlier, we didnt need this because every endpoint was accessible to everyone. Now we do, but it feels like it would be nice to make this somehow a bit shorter. 

- we could give register a parameter `login` that does both in one step. Which also prevents retyping your credentials
- we could add a convenience method `node.root_client.update_user_role("myaccount@openmined.org", ServiceRole.DATASCIENTIST)`

```
node = Worker()
ds_client = node.root_client.guest()
ds_client.register("x@y.com", "mypw")
node.root_client.update_user_role("x@y.com", ServiceRole.DATASCIENTIST)
```

Something like a login code would be even better to build a reliable api, but also a bit more work.

shubham3121:
Great points @koenvanderveen. Auto-login post register is currently implemented, so this should work.
```
node = Worker()
ds_client = node.root_client.guest()
ds_client.register("x@y.com", "mypw")
```
I think we will definitely look at implementing a method like this `node.root_client.update_user_role`, it's just that the client-side UX for user profile-related views hasn't been designed yet.

PR #7475, opened 2023-03-30
Use `TypeVar` for `serializable` to preserve typing

kiendang:
## Description
Currently the `serializable` decorator has return type `Any` so when we apply it to any class the class will lose all typing. For example, currently with this

```py
@serializable
class PartitionKey(BaseModel):
    ...

PTKey = PartitionKey(...)
```

we won't get any typing, completion, ... when using PTKey.

This PR fixex that by using type parameter (`TypeVar`) to type `serializable`.

## Affected Dependencies
List any dependencies that are required for this change.

## How has this been tested?
- Describe the tests that you ran to verify your changes.
- Provide instructions so we can reproduce.
- List any relevant details for your test configuration.

## Checklist
- [x] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [x] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [x] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [x] My changes are covered by tests

PR #7462, opened 2023-03-28
message_stash.py tests - 100% coverage.

PeterChung241:
## Description

Needs review, happy for some feedback. 

## Affected Dependencies

N/A

## How has this been tested?

Locally

## Checklist
- [ X] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [ X] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [ X] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [ X] My changes are covered by tests

PR #7476, opened 2023-03-30
Misc fixes (`BaseStash` tests' flakiness, `@seriablizable` and `@instrument` typings)

kiendang:
## Description
- Remove unused `**kwargs` from `@serializable` plus remove the now deprecated `recursive_serde` in tests for `BaseStash`.
- Fix `BaseStash` tests' flakiness by avoiding collision from random generators.
- Fix typing for `@serializable` and `@instrument` so that `@serializable`- and `@instrument`-ed classes and functions now offer correct typing and code completion (originally from #7475).

> Currently the `serializable` decorator has return type `Any` so when we apply it to any class the class will lose all typing. For example, currently with this
>
> ```py
> @serializable
> class PartitionKey(BaseModel):
>    ...
>
> PTKey = PartitionKey(...)
> ```
>
> we won't get any typing, completion, ... when using PTKey.
>
> This PR fixex that by using type parameter (`TypeVar`) to type `serializable`.

- Fix `test_messagestash_get_all_by_verify_key_for_status` added in #7462.

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/packages/syft/src/syft/telemetry.py
+++ b/packages/syft/src/syft/telemetry.py
@@ -1,6 +1,5 @@
 # stdlib
 import os
-from typing import Any
 from typing import Optional
 
 
@@ -15,10 +14,10 @@ def str_to_bool(bool_str: Optional[str]) -> bool:
 TRACE_MODE = str_to_bool(os.environ.get("TRACE", "False"))
 
 
-def setup_tracer() -> Any:
+def setup_tracer():
     if not TRACE_MODE:
 
-        def noop(func: Any) -> Any:
+        def noop(func):
             return func
 
         return noop
--- a/packages/syft/src/syft/core/node/new/serializable.py
+++ b/packages/syft/src/syft/core/node/new/serializable.py
@@ -20,7 +20,6 @@ def serializable(
     without: Sequence[str] = [],
     inherit: Optional[bool] = True,
     inheritable: Optional[bool] = True,
-    **kwargs,
 ) -> T:
     """
     Recursively serialize attributes of the class.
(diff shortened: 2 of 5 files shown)


packages/syft/src/syft/telemetry.py
```diff
@@ -15,10 +14,10 @@ def str_to_bool(bool_str: Optional[str]) -> bool:
 TRACE_MODE = str_to_bool(os.environ.get("TRACE", "False"))
 
 
-def setup_tracer() -> Any:
+def setup_tracer():
     if not TRACE_MODE:
 
-        def noop(func: Any) -> Any:
+        def noop(func):
```
shubham3121, line 20: Any reason to remove typing ?

shubham3121 (approved):
Superb work @kiendang !!! 🎉💯

rasswanth-s:
Well Done @kiendang

packages/syft/src/syft/telemetry.py
kiendang: This typing is not accurate. The correct one would be

```py
T = TypeVar("T")

def noop(func: T) -> T:
    ...
```

If you don't type it as `Any`, type inference would figure out that `noop` returns the same type as `func`. Earlier, because `serializable` and `instrument` decorators as `Any` were given `Any` as the return type, anything with these decorator would lose its type and become `Any`. This is the reason why we didn't have type or code completion for `BaseStash`, `PartitionKey`, `QueryKey`, ... Either use type parameters to type decorators correctly, or just don't give them types and let type inference figures it out.

kiendang: Removing `Any` here and `setup_tracer` already makes things work so I just left it at that. Type inference for decorators is not implemented perfectly either so I was just going with the simplest solution that works here.

kiendang: Explicit typing re-added in #7483.

PR #7481, opened 2023-04-02
Replace user_verify_key by admin_verify_key during request.submit

IonesioJunior:
## Description
This PR aims to fix a small bug/vulnerability by replacing the usage of `user_verify_key` which uses a hardcoded email to retrieve the root key with the `admin_verify_key` which uses the actual admin service role. This could be a problem if the admin changed his email during the onboarding process.


madhavajay (approved):
Good spot!

PR #7483, opened 2023-04-03
Re-add explicit typing for decorators (`@instrument`, `@serializable`)

kiendang:
## Description
See https://github.com/OpenMined/PySyft/pull/7476#discussion_r1154245386 for context.

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/packages/syft/src/syft/core/node/new/serializable.py
+++ b/packages/syft/src/syft/core/node/new/serializable.py
@@ -1,6 +1,6 @@
 # stdlib
+from typing import List
 from typing import Optional
-from typing import Sequence
 from typing import TypeVar
 
 # syft absolute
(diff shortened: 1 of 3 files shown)


packages/syft/src/syft/core/node/new/serializable.py
```diff
@@ -12,12 +12,12 @@
 module_type = type(syft)
 
 
-T = TypeVar("T")
+T = TypeVar("T", bound=type)
 
 
 def serializable(
-    attrs: Sequence[str] = [],
```
madhavajay, line 19: good spot, these should be picked up by bugbear, i swapped flake8 for ruff but maybe it doesnt have these checks... hrm

packages/syft/src/syft/telemetry.py
```diff
@@ -1,6 +1,19 @@
 # stdlib
 import os
+import sys
+from typing import Callable
 from typing import Optional
+from typing import TypeVar
+from typing import Union
+
+if sys.version_info >= (3, 10):
+    # stdlib
```
madhavajay, line 10: I imagine this is the same as the other PR, since we want typing extensions for older python we should just use it until we deprecate a particular version such as 3.8.

kiendang: Yup I also replied under the other comment. This is more like self documentation to remind us to move from `typing_extensions` to `typing` when the minimum required version Python supports the feature. It's just a convention I saw widely used for using standard library backports.

madhavajay (approved):
Nice work!

PR #7486, opened 2023-04-03
`Worker` fixes

kiendang:
## Description
See commits

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/packages/syft/src/syft/core/node/worker.py
+++ b/packages/syft/src/syft/core/node/worker.py
@@ -7,19 +7,12 @@ from datetime import datetime
 from functools import partial
 import hashlib
 import os
-import sys
 from typing import Any
 from typing import Callable
 from typing import List
 from typing import Optional
 from typing import Type
 from typing import Union
-
-if sys.version_info >= (3, 11):
-    # stdlib
-    from typing import Self
-else:
-    from typing_extensions import Self
 
 # third party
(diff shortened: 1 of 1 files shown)


packages/syft/src/syft/core/node/worker.py
```diff
@@ -7,13 +7,20 @@
...
 import os
+import sys
 from typing import Any
 from typing import Callable
 from typing import List
 from typing import Optional
 from typing import Type
 from typing import Union
 
+if sys.version_info >= (3, 11):
```
madhavajay, line 18: we cant support python 3.11 yet but even if we can, typing_extensions defaults to native implementations where available:
Screenshot 2023-04-05 at 3 52 08 pm

https://github.com/python/typing_extensions/blob/37909ec67adf56070fa7218efa475e87812834e5/src/typing_extensions.py#L1658

kiendang: Yup that's always the case with `typing_extensions`. I agree that leaving it as just `from typing_extensions import ...` no matter the Python version would be no difference technically. The benefit is that it reminds us to move it to `from typing import ...` later on when the minimum Python version supported becomes 3.11. I agree the benefit for doing that in this case is extremely small, more about just following a recommended pattern.

PR #7503, opened 2023-04-11
[WIP] Jax Level 2

teo-milea:
## Description
Open Questions:
- how will we use pointers? There was a previous version of action object pointers, but right now that seems to be outdated, but we couldn't find any information on the new pointers
- ~How will we send a compiled function? Does jax support reverse compiling? This looks more like a code execution request, so the benefit of a Level 2 work is not clear, as long as we try to offer this feature. It sounds like it would be better if we just implement an already available API like jnp~ We can get the raw code but the problem of approving this like a code execution request still stands
- Do we need to implement all the device array types? We have managed to create a `DeviceArrayObject` but are not sure if we need `SharedDeviceArray` or `GlobalDeviceArray` at the moment.
- Can jax compile unsafe functions? (For example, can we get access to the file system in a compiled function?)

## Checklist


teo-milea:
Moved to #7600

PR #7550, opened 2023-04-25
Rpc 2

koenvanderveen:
Includes:

- implements eager execution
  - methods (pointer.method())
  - getattribute (pointer.x)
  - setattribute (pointer.x = new_val)
  - functions (numpy.func())
  - .request() for pointers
  - .get() for pointers
  - action for setting objects (necessary when passing arguments as non ActionObjects in a Plan)
- plans
  - creation of plans using @planify
  - tracing of actions via hooks
  - tracing of lib functions in the designated wrapper (this switched to the tracing client)
  - execution of plans: reconnect inputs to the action graph, separate action logic
- simple logic for read permission flow for eagerly executed results
- refactored action.execute
- a bunch of tests for new functionalities

@madhavajay I added a notebook for playing around with the new code


madhavajay (approved):
Excellent work. Can you remove the _experiemental notebook and make sure the api/0.9 one works / asserts correctly (it would be good to verify without probing the internal store).

When running with a web server (port=8080) Im seeing this exception but it seems to not matter? Are you able to figure out whats wrong there?
Screenshot 2023-05-10 at 11 32 08 am

PR #7566, opened 2023-04-27
Fixes network.domains and network.api.services.network.get_all_peers not working

### LINKED ISSUE #7565: network.domains and network.api.services.network.get_all_peers not working
## Description
network.domains and network.api.services.network.get_all_peers not working

## How to Reproduce
1. Launch a node using `hagrid launch "network" network to docker:8000 --tag=beta --build-src=dev`
2. Try listing domains with `network.domains` 
3. It fails with the following error
```
<class 'syft.service.response.SyftError'>: Exception calling network.get_all_peers. BaseStash.get_all() missing 1 required positional argument: 'credentials'
```

## Expected Behavior
network.domains must return list of domains

KamalGalrani:
fixes #7565


madhavajay:
@KamalGalrani great spot, we need some test coverage on this as it got missed before, can you add something quickly? We also need to deploy an updated server for this.

madhavajay:
@KamalGalrani it looks like this got resolved but thanks for the PR and the other ones I already merged. ❤️

PR #7568, opened 2023-04-27
Fixes serialization for pydantic.networks.EmailStr

### LINKED ISSUE #7567: User Search: pydantic.networks.EmailStr not in TYPE_BANK
## Description
image

## How to Reproduce
`network_client.api.services.user.search(email=EmailStr('kamal.galrani@tvarit.com'))`

## Expected Behavior
Return user with email address

KamalGalrani:
fixes #7567


madhavajay:
@KamalGalrani I guess the problem with this solution is the result being deserialized will be a `str`. 

Can you add something here:
https://github.com/OpenMined/PySyft/blob/dev/packages/syft/src/syft/serde/third_party.py

And a test to verify it does serialize and deserialize correctly on both sides?

KamalGalrani:
Done

PR #7591, opened 2023-05-01
ADD Login error message in case login fails

IonesioJunior:
## Description
This PR enhances the application's front end by adding the ability to change the border color of the login form and displaying a small error message in case a login attempt fails. These improvements will provide users with a better user experience and more helpful feedback in case of errors.

## Changes Made

- Added a border color change to the login form when the user clicks the submit button.
- Added a hidden `<p>` with the error message to be displayed below the login form when the login attempt fails.

## Screenshots

Before:
before

After:
after

## Testing
- I have manually tested the login functionality to ensure that the error message is displayed when a login fails.


madhavajay (approved):
nice! no tests are breaking too so looks good.

PR #7616, opened 2023-05-09
Refactoring CMP to work with jax

teo-milea:
## Description
Fixed some issues with missing shortcut imports and circular imports. Did some refactoring and changed the DFS exploration of a package with a BFS approach.

## Checklist


packages/syft/src/syft/serde/lib_service_registry.py
```diff
@@ -79,10 +84,13 @@ def __init__(
     def set_signature(self) -> None:
         pass
 
-    def build(self) -> None:
+    def build(self, stack: List, root: Optional[CMPBase] = None) -> None:
```
koenvanderveen, line 87: isnt the root always self? the only place where we call .build() is with c.build(stack=self.stack, root=c)

packages/syft/src/syft/serde/lib_service_registry.py
```diff
@@ -309,9 +341,13 @@ def __init__(self, children: List[CMPModule]):
         self.children = {c.path: c for c in children}
 
     def build(self) -> Self:
+        self.stack = []
```
koenvanderveen, line 344: what is the benefit of doing this with a global stack and not recursive?

packages/syft/src/syft/serde/lib_service_registry.py
```diff
@@ -129,65 +137,76 @@ def init_child(
...
-                else:
-                    return False
-        except Exception:  # nosec
-            pass
-        return False
 
-    @staticmethod
-    def parent_is_parent_module(parent_obj: Any, child_obj: Any) -> Optional[str]:
-        try:
+            # if the name of the parent can be found at the start of the name of the child we are good
```
koenvanderveen, line 192: does this assume both child and parent are modules? I could see how this would easily give wrong results if they wouldnt

packages/syft/src/syft/serde/lib_service_registry.py
```diff
@@ -129,65 +137,76 @@ def init_child(
...
+            if child_obj.__name__.startswith(parent_obj.__name__):
+                return True
+
+            # if the child has a module, then we should just make sure it has the same start
+            # as the parent, but not the entire name, as there might be relative imports
+            # in other parts of the codebase
             if hasattr(child_obj, "__module__"):
-                return child_obj.__module__ == parent_obj.__name__
+                return child_obj.__module__.startswith(
```
koenvanderveen, line 200: Do we need to check whether they are modules? Just thinking out loud

packages/syft/src/syft/serde/lib_service_registry.py
teo-milea: yeah, we can change that, I missed it when I moved to bfs

packages/syft/src/syft/serde/lib_service_registry.py
```diff
@@ -98,13 +106,13 @@ def build(self) -> None:
                     except Exception:  # nosec
                         continue
                     child = self.init_child(  # type: ignore
-                        self.obj,
+                        root.obj,
```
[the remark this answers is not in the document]
teo-milea: with your earlier observation, they are not, so we can keep `self`

packages/syft/src/syft/serde/lib_service_registry.py
teo-milea: I suspected that was the case for why initially we wrapped that function in a try expect. Of course, if the parent is not a submodule it is fine to return False (maybe there should be problems with static methods of some classes, but at the same time we need some rigorous testing as there are still corner cases that we miss, I assumed we would do that at the end of the quarter

packages/syft/src/syft/serde/lib_service_registry.py
teo-milea: same idea as earlier, without in-depth testing I couldn't say what we are missing and what we should avoid. For my immediate goals this worked, but I am also sure there are corner cases we are missing

packages/syft/src/syft/serde/lib_service_registry.py
teo-milea: in my opinion, it is cleaner to check if we have already inspected submodules in lower-level files if we have a bfs approach, rather than a recursive one, as we need to keep that information as a global variable anyway, so it gets messier if we do both

koenvanderveen:
closing because we dont have capacity to merge right now, will reopen when relevant again

PR #7643, opened 2023-05-17
Pandas cookbook with pointers

koenvanderveen:
chapters 1-8 with pointers


packages/syft/src/syft/service/action/action_service.py
```diff
@@ -55,22 +135,73 @@ def np_array(self, context: AuthedServiceContext, data: Any) -> Any:
         np_pointer = self.set(context, np_obj)
         return np_pointer
 
+    def preprocess_action_to_store(
+        self,
+        context: AuthedServiceContext,
+        action_object: Union[ActionObject, TwinObject],
+    ):
+        if isinstance(action_object, ActionObject) and is_collection(
```
koenvanderveen, line 143: the idea is here that when we do for instance client.lib.pandas.concat([df1,df2]), that this list that is sent before we execute the action does not contain a nested structure of ActionObject, but one ActionObject that contains the underlying [df1.syft_action_data, df2.syft_action_data ]

koenvanderveen: This collection (list/dict/tuple/set) structure could be nested though, so we are searching through it. While we are searching we are also computing a few other things, like do we have read permission on the result

.github/workflows/pr-tests-syft.yml
```diff
@@ -152,7 +152,15 @@ jobs:
         os: [ubuntu-latest]
         python-version: ["3.9", "3.10"]
         node-type: ["enclave"]
-        notebook-paths: ["tutorials"]
+        notebook-paths:
```
koenvanderveen, line 155: makes sure that we have a separate runner for all tutorial tests

koenvanderveen:
closing because we dont have capacity to merge right now, will reopen when relevant again

PR #7660, opened 2023-05-23
Update skill_test.py

d-m-silva:
added my name to the file

## Description
The task asked to submit the name of the Padawan right after "skywalker", so I did that


teo-milea:
LGTM! Great work @d-m-silva !!

PR #7665, opened 2023-05-26
Request Improvements

shubham3121:
## Description

## Affected Dependencies

- [ ] Capture history of changes at Request Level
     - [x] Create a class to capture Request history
     - [x] Track request change as history as part of the Request Object when changes are applied or rejected
     - [x] Dynamically calculate `status` of Request from Request change history
     - [x] Add a service/stash method to save/update Request object.
     - [x] Update Project to have request as a Linked Object.

- [ ] Make SyftObjects NodeAware
    - [x] Add a function to set an attribute on a SyftBaseObject instance recursively.
    - [x] Add `syft_node_location` to object when it is dispatched by a service
        - set node_uid to Services during initialisation so that they are node aware.
        - remove __init__ from pydantic classes and replace them with pydantic.validators decorators
    - [x] Add `syft_client_verify_key` to object when object is received on the client side.
- [x] Update APIRegistry to use both `node_uid` and `credentilas.verify_key` to cache API
- [x] get api from APIRegistry based on `syft_node_location` and `syft_client_verify_key`

## How has this been tested?

## Checklist


packages/syft/src/syft/service/service.py
```diff
@@ -311,6 +311,30 @@ def service_method(
         # TODO: this is dangerous, we probably want to be more conservative
         roles = DATA_OWNER_ROLE_LEVEL
 
+    def _attach_node_location(self, result) -> Any:
```
madhavajay, line 314: 🦾

packages/syft/src/syft/types/twin_object.py
```diff
@@ -30,35 +34,27 @@ class TwinObject(SyftObject):
...
-        if id is None:
-            id = UID()
-        super().__init__(
-            private_obj=private_obj,
-            private_obj_id=private_obj_id,
-            mock_obj=mock_obj,
-            mock_obj_id=mock_obj_id,
-            id=id,
-        )
+    mock_obj_id: Optional[UID]
```
kiendang, line 41: I feel like here it's better to leave `private_obj_id: UID. # type: ignore` and `mock_obj_id: UID. # type: ignore`. Because of the pydantic validator these 2 are never gonna be None so the type is more accurate this way. Plus it's used in [pydantic doc](https://docs.pydantic.dev/latest/usage/validators/#validate-always) and there's an open [PEP](https://peps.python.org/pep-0712/) for this so I think this is ok we're just doing `# type: ignore` to work around the current type checker imperfection.

kiendang (commented):
Should we prefix all `@pydantic.validator` functions with `__`? These functions should not be exposed to users.

packages/syft/src/syft/node/credentials.py
```diff
@@ -22,9 +22,10 @@
 class SyftVerifyKey(SyftBaseModel):
     verify_key: VerifyKey
 
-    @pydantic.validator("verify_key", pre=True, always=True)
-    def make_verify_key(cls, v: Union[str, VerifyKey]) -> VerifyKey:
-        return VerifyKey(bytes.fromhex(v)) if isinstance(v, str) else v
+    def __init__(self, verify_key: Union[str, VerifyKey]):
+        if isinstance(verify_key, str):
+            verify_key = VerifyKey(bytes.fromhex(verify_key))
+        super().__init__(verify_key=verify_key)
```
kiendang: Oh I thought these 2 do exactly the same thing? Genuine question coz I've been using validator for this. Am I missing anything?

kiendang: (I saw your commit reverting from validator back to this)

packages/syft/src/syft/service/request/request.py
```diff
@@ -61,6 +61,20 @@ def is_type(self, type_: type) -> bool:
...
+    __canonical_name__ = "ChangeStatus"
+    __version__ = SYFT_OBJECT_VERSION_1
+
+    id: Optional[UID]
+    change_id: UID
+    applied: bool = False
+
+    @staticmethod
+    def from_change(change: Change, applied: bool) -> Self:
+        return ChangeStatus(change_id=change.id, applied=applied)
```
kiendang, line 75:
```suggestion
    @classmethod
    def from_change(cls, change: Change, applied: bool) -> Self:
        return cls(change_id=change.id, applied=applied)
```

Not a big deal, just that if using classmethod then if someone subclass this it still works correctly.

packages/syft/src/syft/node/credentials.py
shubham3121: I reverted it back for now because of its use in `convert_types` function. e.g. `convert_types(["id", "verify_key"], [UID, SyftVerifyKey])`. I will investigate if we can revert back to use validators

packages/syft/src/syft/types/syft_object.py
```diff
@@ -44,14 +45,37 @@
...
+class SyftBaseObject(BaseModel, SyftHashableObject):
     class Config:
         arbitrary_types_allowed = True
 
     __canonical_name__: str  # the name which doesn't change even when there are multiple classes
     __version__: int  # data is always versioned
 
+    syft_node_location: Optional[UID]
+    syft_client_verify_key: Optional[SyftVerifyKey]
+
```
yashgorana, line 78: As a headsup once we make `SyftBaseObject` as hashable and have node & client info, the following classes will be have them as well:
- APIEndpoint
- LibEndpoint
- SyftAPIData
- PreHookContext
- BaseConfig
- StoreConfig
- MongoDict
- ChangeContext
- Context

If this is specific to node/request then would suggest we don't add them to SyftBaseObject, and instead create a new sub-class and use it for the concrete ones that need these attrs.

shubham3121: This is not specific to request, but most of the SyftObjects, Error and Permission objects. Yes I will be creating a sub-class, to prevent all pydantic classes have these attributes.

packages/syft/src/syft/types/syft_object.py
```diff
@@ -44,14 +45,37 @@
...
-class SyftBaseObject(BaseModel):
+
+class SyftHashableObject:
+    __hash_exclude_attrs__ = []
+
+    def __hash__(self) -> int:
+        return int.from_bytes(self.__sha256__(), byteorder="big")
+
+    def __sha256__(self) -> bytes:
+        self.__hash_exclude_attrs__.extend(DEFAULT_HASH_EXCLUDE_ATTRS)
```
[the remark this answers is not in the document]
shubham3121: But the problem is that any subclass defining __hash_exclude_attrs__ will overwrite this. So we want a behaviour where we can always exclude `syft_node_location` and `syft_client_verify_key` in the hash comparison. Otherwise, I need to define both of these attributes in all subclasses where __hash_exclude_attrs__ is overwritten.

packages/syft/src/syft/types/syft_object.py
```diff
@@ -241,6 +251,8 @@ def _repr_debug_(self) -> str:
         _repr_str = f"class {class_name}:\n"
         fields = getattr(self, "__fields__", {})
         for attr in fields.keys():
+            if attr in DEFAULT_HASH_EXCLUDE_ATTRS:
```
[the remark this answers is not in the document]
shubham3121: Again, this exclusion is specific to `syft_node_location` and `syft_client_verify_key`, probably I will rename the variable to reflect the same and therefore not using `self.__hash_exclude_attrs__`

packages/syft/src/syft/types/syft_object.py
```diff
@@ -4,16 +4,19 @@
 import inspect
 from inspect import Signature
 import types
+from typing import AbstractSet
```
kiendang, line 7:
```suggestion
from collections.abc import Set
```

AbstractSet is [deprecated](https://docs.python.org/3/library/typing.html#typing.AbstractSet) since Python 3.9

packages/syft/src/syft/types/syft_object.py
```diff
@@ -37,6 +40,11 @@
 from .syft_metaclass import PartialModelMetaclass
 from .uid import UID
 
+IntStr = Union[int, str]
+AbstractSetIntStr = AbstractSet[IntStr]
```
kiendang, line 44:
```suggestion
AbstractSetIntStr = Set[IntStr]
```

packages/syft/src/syft/types/syft_object.py
```diff
@@ -4,16 +4,19 @@
...
 from inspect import Signature
 import types
+from typing import AbstractSet
 from typing import Any
 from typing import Callable
 from typing import ClassVar
 from typing import Dict
 from typing import KeysView
 from typing import List
+from typing import Mapping
```
kiendang, line 14: `typing.Mapping` is [deprecated](https://docs.python.org/3/library/typing.html#typing.Mapping) in favor of `collections.abc.Mapping` as well, but I think we use `typing.Mapping` a lot in our codebase we might want to do 1 separate PR that replaces all of them.

packages/syft/src/syft/service/action/action_object.py
```diff
@@ -977,7 +986,12 @@ def _syft_wrap_attribute_for_bool_on_nonbools(self, name: str) -> Any:
             )
 
         debug("[__getattribute__] Handling bool on nonbools")
-        context = PreHookContext(obj=self, op_name=name)
+        context = PreHookContext(
```
rasswanth-s, line 989: Q: What is PreHookContext?

rasswanth-s: Why do customly add  node_uid and verify key here?

packages/syft/src/syft/service/action/action_object.py
```diff
@@ -536,7 +539,7 @@ def request(self, client):
         return client.api.services.request.submit(submit_request)
 
     def _syft_try_to_save_to_store(self, obj) -> None:
-        if self.syft_node_uid is None:
+        if self.syft_node_uid is None or self.syft_client_verify_key is None:
```
rasswanth-s, line 542: Q: Do we not save the object to store , if either uid or verify key is None ?

packages/syft/src/syft/node/credentials.py
```diff
@@ -53,10 +54,9 @@ def __hash__(self) -> int:
 class SyftSigningKey(SyftBaseModel):
     signing_key: SigningKey
 
-    def __init__(self, signing_key: Union[str, SigningKey]) -> None:
-        if isinstance(signing_key, str):
-            signing_key = SigningKey(bytes.fromhex(signing_key))
-        super().__init__(signing_key=signing_key)
+    @pydantic.validator("signing_key", pre=True, always=True)
+    def make_verify_key(cls, v: Union[str, SigningKey]) -> SigningKey:
```
rasswanth-s, line 58: NITPICK: the method name is make_signing_key?

packages/syft/src/syft/node/credentials.py
```diff
@@ -53,10 +54,9 @@ def __hash__(self) -> int:
...
+        return SigningKey(bytes.fromhex(v)) if isinstance(v, str) else v
```
rasswanth-s, line 59: Q: Should we also throw an error, if any other value other than SigningKey or str is passed?

packages/syft/src/syft/client/client.py
```diff
@@ -514,6 +512,7 @@ def register(
             return SyftError(message=str(e))
         response = self.connection.register(new_user=new_user)
         if isinstance(response, tuple):
+            self.credentials = response[1].signing_key
```
rasswanth-s, line 515: I think there was a feature request,to prevent register method from modifying the user.
I think Koen made a PR earlier to remove the same.
Do we want to have the same user?

packages/syft/src/syft/client/api.py
```diff
@@ -49,17 +50,29 @@
...
+
+        key = (node_uid, user_verify_key)
+
+        cls.__api_registry__[key] = api
 
     @classmethod
-    def api_for(cls, node_uid: UID) -> SyftAPI:
-        return cls.__api_registry__.get(node_uid, None)
+    def api_for(cls, node_uid: UID, user_verify_key: SyftVerifyKey) -> SyftAPI:
+        key = (node_uid, user_verify_key)
```
rasswanth-s, line 74: ❤️

packages/syft/src/syft/service/metadata/node_metadata.py
```diff
@@ -108,7 +108,7 @@ def check_version(self, client_version: str) -> bool:
 @transform(NodeMetadata, NodeMetadataJSON)
 def metadata_to_json() -> List[Callable]:
     return [
-        drop("__canonical_name__"),
+        drop(["__canonical_name__"]),
```
rasswanth-s, line 111: Good Catch!

packages/syft/src/syft/service/request/request.py
```diff
@@ -127,21 +141,43 @@ class Request(SyftObject):
...
+        if len(self.history) == 0:
+            return RequestStatus.PENDING
+
+        change_applied_map = {}
+        for change_status in self.history:
+            # only store the last change
+            change_applied_map[change_status.id] = change_status.applied
+
+        all_changes_applied = all(change_applied_map.values()) and (
+            len(change_applied_map) == len(self.changes)
```
rasswanth-s, line 167: 👌

rasswanth-s (approved):
Stellar Work 🏅  @shubham3121 , and also kudos on the attention to details., I have some minor questions , left them as comment on the PR.

packages/syft/src/syft/service/action/action_object.py
shubham3121: PreHookContext is a type of context class used when we processing pre_hooks in action_object. e.g. make_action_side_effect. Need to pass node_uid and verify_key as the ActionObjects are updated based on this context.

packages/syft/src/syft/node/credentials.py
shubham3121: Pydantic automatically throws an ValueError if value is not SigningKey.

packages/syft/src/syft/client/client.py
shubham3121: 👍🏾

packages/syft/src/syft/service/action/action_object.py
shubham3121: Yeah, we don't save to store this case, cause it means someone has created this object locally.

packages/syft/src/syft/node/credentials.py
shubham3121: Fixed by: https://github.com/OpenMined/PySyft/pull/7665/commits/d22ffe63abbfc2d76c8dcdfa162c474e2af0ca8d

PR #7668, opened 2023-05-26
Implementation of Paginated Search for User Names via Frontend API

IonesioJunior:
## Description
This pull request introduces the capability to perform search requests and filter users based on their names through the frontend API, interacting directly with our search user service. The modifications are primarily focused on presenting the results in a paginated format to enhance usability and performance.

### The specific changes include:

1. Fix the `searchUserByName` function in the `frontend.lib.api.user` source file. This function is designed to communicate with the search user service, initiating search requests and filtering based on user names.

2. The addition of `chunk_size` and `chunk_index` as optional parameters to the `user.search` function. This enhancement facilitates a paginated response, allowing the user to specify the size of data chunks (i.e., pages) and the index of the chunk they wish to view.

3. Small fix in `services.py` to allow Optional service parameters to have their default values if they're not defined by the request.
These enhancements will provide users with a more efficient, user-friendly interface when searching and filtering users
through our frontend API.


packages/syft/src/syft/service/service.py
```diff
@@ -259,11 +259,13 @@ def reconstruct_args_kwargs(
         autosplat_objs[autosplat_key] = autosplat_type(**init_kwargs)
 
     final_kwargs = {}
-    for param_key, _ in signature.parameters.items():
+    for param_key, param in signature.parameters.items():
         if param_key in kwargs:
             final_kwargs[param_key] = kwargs[param_key]
         elif param_key in autosplat_objs:
             final_kwargs[param_key] = autosplat_objs[param_key]
+        elif not isinstance(param.default, type(Parameter.empty)):
```
madhavajay, line 267: awesome

PR #7669, opened 2023-05-26
remove changing to new user when registering new user

koenvanderveen:


rasswanth-s (approved):
One small question, otherwise looks good to me.

PR #7674, opened 2023-05-28
Minor fixes to Syft Projects for tutorials

IshanMi:
## Description
- Add default output policy to the syft_function decorator so tutorials don't have to be as verbose
- fix pydantic validator bug arising when no email was specified while creating a Syft Project, causing a KeyError

## Affected Dependencies
No new dependencies, only cosmetic changes.

## How has this been tested?

## Checklist


madhavajay:
@IshanMi Looking good, just need to fix the error the linter found:
```
packages/syft/src/syft/service/code/user_code.py:399:47: B008 Do not perform function call `SingleExecutionExactOutput` in argument defaults
```

Otherwise every code object will have the same OutputPolicy instance.

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -395,13 +396,16 @@ def debox_asset(arg: Any) -> Any:
 
 def syft_function(
     input_policy: Union[InputPolicy, UID],
-    output_policy: Union[OutputPolicy, UID],
+    output_policy: Union[OutputPolicy, UID] = None,
```
rasswanth-s, line 399: Q: Maybe we could enforce the users to always provide a output policy?,
As it would enforce them to also understand the restrictions that the code would be processed.
It would also make the users be context aware while choosing policies.

rasswanth-s (approved):
One small question, otherwise  looks good to me.

packages/syft/src/syft/client/client.py
```diff
@@ -360,6 +361,9 @@ def api(self) -> SyftAPI:
 
         return self._api
 
+    def view_projects(self) -> List:
```
shubham3121, line 364: We can have it has a property. Similar to users or code, etc.
 ```
    @property
    def users(self) -> Optional[APIModule]:
        if self.api is not None and hasattr(self.api.services, "user"):
            return self.api.services.user
        return None
```

File: packages/syft/src/syft/service/code/user_code.py (hunk 1 above)
Comment (shubham3121) [inline packages/syft/src/syft/service/code/user_code.py:399]: Also can we update the typing for `output_policy` to be Optional if we want to set it None by default.

File: packages/syft/src/syft/service/project/project.py [hunk 3]
```diff
@@ -1051,29 +1054,30 @@ def make_shareholders_and_leader_route(cls, values) -> Dict:
 
     @root_validator(pre=True)
     def check_user_email_and_users(cls, values) -> Dict:
-        if values["user_email_address"] is not None:
-            users = values["users"]
-            if len(users) == 0:
-                raise SyftException(
-                    "Users cannot be empty if user email address is provided"
-                )
+        if "user_email_address" in values:  # avoids KeyError below if no email provided
```
shubham3121, line 1057: 👍🏾

packages/syft/src/syft/service/project/project.py
```diff
@@ -1105,10 +1109,29 @@ def exchange_routes(
 
         return SyftSuccess(message="Successfully Exchaged Routes")
 
+    def create_request(self, obj: SubmitUserCode, client: SyftClient):
```
shubham3121, line 1112: should we rename it to something related to user code maybe `create_code_request` ?

shubham3121 (commented):
Nice work guys !! Just left a few minor comments.

packages/syft/src/syft/client/client.py
IshanMi: Great catch- added in https://github.com/OpenMined/PySyft/pull/7674/commits/e5868c32ab358ef37cf645f9a9213ceb46bedc46

PR #7677, opened 2023-05-29
Re-enable flake8-bugbear checks in Ruff

kiendang:
## Description
Re-enable flake8-bugbear checks in Ruff to catch potential buggy code.

One of those is B006 `Do not use mutable data structures for argument defaults.`

for functions, instead of

```python
def f(ls=[]):
   ...
```

do

```python
def f(ls=None):
    if ls is None:
        ls = []
    ...
```

This is an [anti-pattern](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html) and might lead to subtle bugs.
## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,9 +72,9 @@ repos:
         # exclude: ^(packages/syft/src/syft/proto)
         files: "^packages/syft(?!(.*_pb2\\.py$)).*\\.py"
 
-  - repo: https://github.com/charliermarsh/ruff-pre-commit
+  - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: "v0.0.246"
+    rev: "v0.0.247"
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
--- a/packages/hagrid/hagrid/cli.py
+++ b/packages/hagrid/hagrid/cli.py
@@ -3370,14 +3370,13 @@ def get_docker_status(
         # If there are worker containers with an internal port
         # fetch the worker container with the launched worker name
         worker_containers = worker_containers_output.split("\n")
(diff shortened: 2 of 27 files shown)


kiendang:
Ah looks like pydantic handles this differently so things like
```python
class CreateAsset(SyftObject): 
    ...
    contributors: List[Contributor] = []
```

is valid.

kiendang:
Reopen to fix some stuff that is not pydantic.

madhavajay:
@kiendang yes, great catch. We used to use bugbear which had a check for this but we switched to `ruff`: https://github.com/charliermarsh/ruff Can you see if we can add back the right bugbear check to ruff for this.

https://github.com/PyCQA/flake8-bugbear
```
B006: Do not use mutable data structures for argument defaults. They are created during function definition time. All calls to the function reuse this one instance of that data structure, persisting changes between them.
```

kiendang:
@madhavajay I've re-added the flake8-bugbear checks and attempted to fix the failed checks. There are only a couple of places where fixing the checks makes the code doesn't work anymore. Those will be followed up separately.

madhavajay (approved):
Dude, stellar work! ❤️💫

I am keen to also get the mypy checks back in but im worried about the 500+ errors we need to fix first.

```
- repo: https://github.com/pre-commit/mirrors-mypy
    rev: v1.0.0
    hooks:
      - id: mypy
        name: mypy-syft
        always_run: true
        # files: "^packages/syft/src/syft/serde|^packages/syft/src/syft/util"
        files: "^packages/syft/src/syft/serde"
```

I think we just need to progressively open this up as we go.

kiendang:
Yup I actually also tried reenabling mypy but it's close to 1000 errors now. I think we can gradually resolve them by reenabling individual parts of the code base

PR #7708, opened 2023-06-06
use client.code instead of client.api.services.code

koenvanderveen:


koenvanderveen:
I hardcoded the ports in many of the notebook tests as it is a source of flakyness when running parallel notebook tests that all use port=auto. They can still end up using the same port

PR #7709, opened 2023-06-06
Add project helper attr/functions to SyftClient

yashgorana:
## Description

Closes https://github.com/OpenMined/Heartbeat/issues/332

Adds new helper methods/attributes for projects...
```python
client.get_project(name="Project Name")   # Returns project by name
client.get_project(uid=sy.UID(hex))       # Returns project by its UID
client.get_project()                      # returns list of all projects
```

... which use these new project APIs
```python
client.api.services.project.get_by_uid(uid: UID)
client.api.services.project.get_by_name(name: str)
```

## Affected Dependencies
- `ProjectSubmit` and `Project` now have unique names.

## How has this been tested?
- Notebook based tests
- TODO: Unit tests

## Checklist

### DIFF
--- a/notebooks/api/0.8/01-submit-code.ipynb
+++ b/notebooks/api/0.8/01-submit-code.ipynb
@@ -305,6 +305,7 @@
    },
    "outputs": [],
    "source": [
+    "# Creating a new project\n",
     "project_create = sy.Project(\n",
     "    name=\"My Cool UN Project\",\n",
     "    description=\"\"\"Hi, I want to calculate the trade volume in million's with my cool code.\"\"\",\n",
@@ -326,6 +327,16 @@
     "projects = project_create.start()\n",
     "assert len(projects) == 1\n",
     "project = projects[0]"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Or when working on a project that already exists\n",
+    "project = guest_domain_client.get_project(name=\"My Cool UN Project\")"
    ]
   },
   {
--- a/notebooks/api/0.8/01-submit-code.ipynb
+++ b/notebooks/api/0.8/01-submit-code.ipynb
@@ -336,7 +336,8 @@
    "outputs": [],
    "source": [
     "# Or when working on a project that already exists\n",
-    "project = guest_domain_client.get_project(name=\"My Cool UN Project\")"
+    "project = guest_domain_client.get_project(name=\"My Cool UN Project\")\n",
+    "assert project"
    ]
   },
   {
@@ -443,7 +444,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.16"
+   "version": "3.11.2"
   },
   "toc": {
    "base_numbering": 1,
--- a/packages/syft/src/syft/client/client.py
+++ b/packages/syft/src/syft/client/client.py
@@ -1,11 +1,14 @@
+# future
+from __future__ import annotations
+
 # stdlib
 from enum import Enum
 import hashlib
 import json
 from typing import Any
 from typing import Dict
-from typing import List
 from typing import Optional
+from typing import TYPE_CHECKING
 from typing import Type
 from typing import Union
 from typing import cast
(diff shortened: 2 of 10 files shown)


packages/syft/src/syft/client/client.py
```diff
@@ -415,45 +414,69 @@ def apply_to_gateway(self, client: Self) -> None:
 
     @property
     def data_subject_registry(self) -> Optional[APIModule]:
-        if self.api is not None and hasattr(self.api.services, "data_subject"):
+        if self.api is not None and self.api.has_service("data_subject"):
```
yashgorana, line 417: Added a helper function `has_service(name)` to make this check a bit cleaner.

Aditionally, By looking at the code `self.api is not None` doesn't seem to be None. Because invoking `self.api` always invokes `_fetch_api()`. Let me know if this is redundant, I'll remove it.

packages/syft/src/syft/service/project/project_service.py
```diff
@@ -49,19 +49,6 @@ def create_project(
...
-            )
-
-            if project_id_check.is_err():
-                return SyftError(message=f"{project_id_check.err()}")
-
-            if project_id_check.ok() is not None:
-                return SyftError(
-                    message=f"Project with id {project_id} already exists. Kindly use a different id"
-                )
-
```
yashgorana, line 64: `UID` is anyway unique at stash level, so this will already be handled. Let me know if there's a case that I may be unaware of.

packages/syft/src/syft/service/project/project_stash.py
```diff
@@ -40,11 +40,17 @@ def get_all_for_verify_key(
...
         self, credentials: SyftVerifyKey, uid: UID
     ) -> Result[Optional[Project], str]:
         qks = QueryKeys(qks=[UIDPartitionKey.with_obj(uid)])
         return self.query_one(credentials=credentials, qks=qks)
+
+    def get_by_name(
+        self, credentials: SyftVerifyKey, project_name: str
+    ) -> Result[Optional[Project], str]:
+        qks = QueryKeys(qks=[NamePartitionKey.with_obj(project_name)])
+        return self.query_one(credentials=credentials, qks=qks)
```
yashgorana, line 56: New API call to get directly by name instead of iterating-and-filter over all projects. `"name"` column is now unique, and hence makes sure that it will return only one item

notebooks/api/0.8/01-submit-code.ipynb
```diff
@@ -328,6 +329,16 @@
...
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Or when working on a project that already exists\n",
+    "project = guest_domain_client.get_project(name=\"My Cool UN Project\")"
```
rasswanth-s, line 339: It would good also to add assert statements, as pytest stops the execution only with exception(Link in the case where a project does not exist).on that particular line itself
I have added it in my upcoming commit.

Pinging here, as it would be helpful in the upcoming PR's

packages/syft/src/syft/client/client.py
rasswanth-s: ❤️ , I think we can remove it

shubham3121: We don't always fetch, we cache it in self._api where self.api acts a property pointing to self._api

packages/syft/src/syft/node/node.py
```diff
@@ -621,7 +620,7 @@ def handle_api_call_with_unsigned_result(
             # Publisher system which pushes to a Queue
             worker_settings = WorkerSettings.from_node(node=self)
 
-            message_bytes = serialize._serialize(
+            message_bytes = _serialize(
```
rasswanth-s, line 623: 🙌

packages/syft/src/syft/service/project/project.py
```diff
@@ -648,6 +648,8 @@ class Project(SyftObject):
     users: List[UserIdentity] = []
 
     __attr_repr_cols__ = ["name", "shareholders", "state_sync_leader"]
+    __attr_unique__ = ["name"]
```
rasswanth-s, line 651: 🚀

packages/syft/src/syft/service/project/project_stash.py
rasswanth-s: Wonderful

rasswanth-s (approved):
Well Done @yashgorana  🎸

rasswanth-s:
I think there was a test missing for get_by_uid, I think we could add it in the second PR.

PR #7715, opened 2023-06-07
Domain register toggling

khoaguin:
## Description
Current anyone can register on a domain. We would want a functionality where we can toggle the register functionality. This might be required by some organisations where only the DO can create new users.
More detailed can be seen in [this ticket](https://github.com/OpenMined/Heartbeat/issues/333)


packages/syft/src/syft/service/user/user_service.py
```diff
@@ -325,11 +328,43 @@ def admin_verify_key(self) -> Union[SyftVerifyKey, SyftError]:
...
+        return SyftSuccess(message=f"Registration feature successfully {message}")
+
     def register(
         self, context: NodeServiceContext, new_user: UserCreate
     ) -> Union[Tuple[SyftSuccess, UserPrivateKey], SyftError]:
         """Register new user"""
 
+        request_user_role = (
+            ServiceRole.GUEST
+            if new_user.created_by is None
```
koenvanderveen, line 355: Cant we not just use context.credentials here? Does that remove the need to have this new_user.created_by field?

notebooks/api/0.8/06-domain-register-control-flow.ipynb
```diff
@@ -0,0 +1,359 @@
+{
```
rasswanth-s, line 1: NITPICK: Could we rename the notebook to `07-domain-register-control-flow` as we already have `06`

packages/syft/src/syft/client/client.py
```diff
@@ -656,6 +659,9 @@ def login(
     if email and password:
         login_credentials = UserLoginCredentials(email=email, password=password)
 
+    if login_credentials is None:
```
koenvanderveen, line 662: I think logging in without credentials is a bit confusing. I think being explicit about creating a guest account isnt a bad thing. If we want to keep this we should at least inform the user about what is happening

packages/syft/src/syft/service/user/user_service.py
```diff
@@ -325,11 +328,43 @@ def admin_verify_key(self) -> Union[SyftVerifyKey, SyftError]:
         except Exception as e:
             return SyftError(message=str(e))
 
+    @service_method(path="user.toggle_registration", name="toggle_signup")
```
koenvanderveen, line 331: can we rename this? when I read the word toggle as a user I would expect changing from false->true when false and true->false when true. But instead it seems like we are setting the value

packages/syft/tests/syft/users/user_service_test.py
```diff
@@ -604,3 +625,37 @@ def mock_get_by_email(credentials: SyftVerifyKey, email: str) -> Err:
...
+    assert isinstance(response_1, SyftSuccess)
+    # by default, the guest client can't register new user
+    response_2 = guest_domain_client.register(
+        email=email2, password="harley123", name="Harley"
+    )
+    assert isinstance(response_2, SyftError)
+
+    assert any([user.email == email1 for user in root_domain_client.users])
+
+    # only after the root client enable other users to signup, they can
```
koenvanderveen, line 647: a DO client would still be able to register users right? Can we add a test for that?

notebooks/api/0.8/06-domain-register-control-flow.ipynb
koenvanderveen, line 1: currently all of the api/0.8 notebooks contain more or less full user stories, while the tutorials are more about "how do I do X". I think toggling this setting would be a better fit in the tutorials. If there is anything missing there we can move it there. Would you agree?

koenvanderveen:
overall great PR ❤️

packages/hagrid/hagrid/cli.py
```diff
@@ -215,6 +215,11 @@ def clean(location: str) -> None:
     is_flag=True,
     help="Enable Jupyter Notebooks",
 )
+@click.option(
+    "--enable-signup",
```
rasswanth-s, line 219: 🙌

packages/syft/src/syft/client/client.py
```diff
@@ -370,9 +370,11 @@ def api(self) -> SyftAPI:
         return self._api
 
     def guest(self) -> Self:
-        self.credentials = SyftSigningKey.generate()
-        self.__logged_in_user = ""
-        return self
+        return SyftClient(
```
rasswanth-s, line 373: Good Catch on this Bug 🚀

packages/syft/src/syft/client/client.py
rasswanth-s: +1 : MAYBE: a print statement: 
Logging into {DOMAIN NAME} ... as GUEST

packages/syft/src/syft/service/settings/settings.py
```diff
@@ -1,23 +1,24 @@
...
 from ...serde.serializable import serializable
+from ...types.syft_object import PartialSyftObject
 from ...types.syft_object import SYFT_OBJECT_VERSION_1
 from ...types.syft_object import SyftObject
 from ...types.uid import UID
 
 
 @serializable()
-class NodeSettingsUpdate(SyftObject):
+class NodeSettingsUpdate(PartialSyftObject):
```
rasswanth-s, line 12: Q: Interesting , what is PartialSyftObject, how does it help this context ?

packages/syft/src/syft/service/settings/settings_service.py
```diff
@@ -62,7 +62,7 @@ def update(
             current_settings = result.ok()
             if len(current_settings) > 0:
                 new_settings = current_settings[0].copy(
-                    update=settings.dict(exclude_unset=True)
+                    update=settings.to_dict(exclude_empty=True)
```
rasswanth-s, line 65: Q: What is difference between  exclude_unset and exclude_empty?

packages/syft/tests/conftest.py
```diff
@@ -28,6 +29,7 @@ def faker():
 
 @pytest.fixture(autouse=True)
 def worker(faker):
+    flags.CAN_REGISTER = False
```
rasswanth-s, line 32: By default it is false,maybe we could remove it from the worker fixture here?

packages/syft/src/syft/client/client.py
koenvanderveen: what would even happen when we would login as a guest while guest registration is disabled?

notebooks/api/0.8/06-domain-register-control-flow.ipynb
khoaguin: done

packages/syft/src/syft/service/user/user_service.py
shubham3121: Yes you're right, we can just use context.credentials. I will update it.

packages/syft/src/syft/client/client.py
shubham3121: @koenvanderveen as a guest user (without registration) we can still explore datasets on the domain.

packages/syft/tests/syft/users/user_service_test.py
shubham3121: Yes, they can. Will add a test for it.

notebooks/api/0.8/06-domain-register-control-flow.ipynb
shubham3121: Yes, I agree. Currently we have added it both places. 0.8/ has a notebook with asserts and then we have added the same to the account settings tutorial (scroll to the end).
https://github.com/OpenMined/PySyft/blob/98f8eae95223c650ed12ae6b486c7d2fcb143376/notebooks/tutorials/data-owner/02-account-management.ipynb
Do you think we need a separate notebook instead ?

packages/syft/src/syft/service/settings/settings.py
shubham3121: So, PartialSyftObject class, allows us to declare attributes without marking them Optional all the time. This allows us to havev a distinction between values when they are None and when the user has not set them.
This means, if I don't provide description in a Project object, then it is value is marked as `Empty` instead of None. 

Refer to this issue for more detail:
https://github.com/OpenMined/Heartbeat/issues/117

packages/syft/src/syft/service/settings/settings_service.py
shubham3121: `exclude_unset` is provided by Pydantic, and it differentiates between set and unset values by a value is None or not. But this causes confusion when actually the value being set is None. Therefore we needed the notion of SyftPartialObjects and values being marked as `Empty` if they are not None.

image

packages/syft/src/syft/service/user/user_service.py
shubham3121: @koenvanderveen I rechecked, we don't have `context.credentials` at this point. This is because, the `context` is a NodeServiceContext object and not AuthedServiceContext
Refer to: https://github.com/OpenMined/PySyft/blob/8b98a07e66d193a7233cd4120681c1da016f4dde/packages/syft/src/syft/node/routes.py#L144

rasswanth-s: I think it might be security issue shubham, without authedservicecontext.
Since the verify key of the domain is public.

I could create UserCreate Object Customly, and include the verify key of the domain node and post it to the register endpoint.

I think it would good to have signature verification in register endpoint

rasswanth-s: I am thinking, on whether to sign the UserCreate Message or pass the signing key itself.

For for the first option:
When we sign the UserCreate Message , 

we could extract the verify key , from it directly.

For the second option:
Even passing the signing key, directly, should be not be  a problem..
Since signing key is a very important primitive in our system.
I  was thinking , maybe we could avoid passing signing key, during exchange.
I am also not fully sure.

Thinking more on it.

rasswanth-s: First I thought of removing /register endpoint, and we could directly call

`domain_client.api.services.user.register()`

Which would directly give us the authedservicecontext.

But I think ,the /register is used by the frontend team for registration in UI.
Not sure , if it would be a good idea to remove that.

notebooks/api/0.8/06-domain-register-control-flow.ipynb
koenvanderveen: I meant to propose to remove the notebook from 0.8/api, as this is just a minor setting and not really a full user flow

koenvanderveen: but if you feel like it should stay here thats also fine with me

packages/syft/src/syft/service/user/user_service.py
koenvanderveen: the current approach is still not secure I believe. Normally we ensure that the signing key passed into the request is paired with the signing key that is used to sign the message on the client side, and therefore getting by verify key is equivalent to getting by signing key (please correct me if I am wrong here). For .register we are not doing this whole verification dance (as we may not have the keys for the user yet). Therefore we should get the user by signing key in order for this to be secure. In the current scenario you could pass a verify key, which is public, of another user

packages/syft/tests/conftest.py
shubham3121: Fixed by: https://github.com/OpenMined/PySyft/pull/7715/commits/b76fd05f983d0368a46dc630c29e759307ed42e2

packages/syft/src/syft/service/user/user_service.py
shubham3121: Fixed by: https://github.com/OpenMined/PySyft/pull/7715/commits/b0f3e791af7392c7e670d5232637b1211465a2e8

koenvanderveen (approved):

PR #7745, opened 2023-06-13
Showing name and shape of asset before uploading (`CreateAsset`)

khoaguin:
## Description
Fixing the bug that [when creating dataset + asset, before upload, ids of assets show None](https://github.com/OpenMined/Heartbeat/issues/398)


khoaguin:
Instead of showing the `id`, we will show the name and shape of `CreateAsset` since we only want to create the `id` for an asset on the server side when it's uploaded

khoaguin:
Now we only show the name of the `CreateAsset List` since the `id` and `shape` will be `None` before uploading 
image

PR #7789, opened 2023-06-20
refactor get

koenvanderveen:


packages/syft/src/syft/service/action/action_object.py
```diff
@@ -729,6 +730,22 @@ def get_from(self, client: SyftClient) -> Any:
...
+        from ...client.api import APIRegistry
+
+        api = APIRegistry.api_for(
+            node_uid=self.syft_node_location,
+            user_verify_key=self.syft_client_verify_key,
+        )
+        res = api.services.action.get(self.id)
+
+        if not isinstance(res, ActionObject):
+            return Err(res)
```
shubham3121, line 745: Should we wrap this into SyftError instead ?

shubham3121 (approved):
Great work @koenvanderveen !!! 
Some tests are failing, otherwise PR looks good.

PR #7820, opened 2023-06-21
add ability to pass asset into user func

koenvanderveen:


packages/syft/src/syft/service/code/user_code.py
```diff
@@ -373,6 +375,16 @@ def wrapper(*args: Any, **kwargs: Any) -> Callable:
...
         return wrapper
 
+    def _repr_markdown_(self):
+        md = f"""class UserCode
+    id: str = {self.id}
+    status.approved: str = {self.status.approved}
+    service_func_name: str = {self.service_func_name}
+    code:
+
+{self.raw_code}"""
```
rasswanth-s, line 385: Just Tested, looks Awesome, 
Well Done @koenvanderveen

PR #7822, opened 2023-06-21
CI cancel previous run

bitsofsteve:
This will help cancel previous runs when a new commit is made, hence preventing having to wait for a workflow even though in progress to finish.


rasswanth-s (approved):
Amazing work @bitsofsteve , Kudos for using github actions sytax itself

bitsofsteve:
> Amazing work @bitsofsteve , Kudos for using github actions syxtax itself
> 
> Maybe after the PR merge, we could test this workflow also on the PR's from forked branches to see if it works.

@rasswanth-s I'm sure it will work, since the concurrency grouping we used makes it unique to the PR. Great work on switching to `github.event.pull_request.number`.

PR #7910, opened 2023-07-09
[WIP] Init experimental wallet

iamtrask:
## Description
I got tired of logging in all the time and trying to remember what URLs corresponded to what domains. This feature lets me save and load all of that from my 1password wallet.


madhavajay:
Very cool. I do think we can probably use something like passkeys now that they are generally available on all major browsers, phones, operating systems etc.

I will close this for now and we can revisit it in the future when it comes back up.

PR #7912, opened 2023-07-10
User attribute setter methods

khoaguin:
According to [this issue](https://github.com/OpenMined/Heartbeat/issues/521)
- [x] when logging in to a domain node as the administrator, if the default password is still set, giving a warning and tell you how to change the password
- [x] setter method for user password
- [x] setter method for user email
- [x] setter method for user name
- [x] setter method for user institute
- [x] setter method for user website
- [x] setter method for user role
- [x] make a property `client.me` that returns the same thing with `client.users.get_current_user()`
- [x] after setting a new password, require repeat password. Hint: Use `getpass()`
- [x] other than password and email, change the `set_<>` to `update` for other attributes. Now we have `set_password`, `set_email`, `set_role` and `update` for name, institution and website.
- [x] writing tests for these setter methods (in `user_test.py`)


khoaguin:
Currently I am checking the default password like this (with ` if password == "changethis":`). Is there a better way to sync it with `defaultRootPassword = "changethis"` as in `values.yaml`?
image


Fixed using ` get_default_root_password()` from `node.py`

khoaguin:
Also, after resetting the password, shall the old `domain_client` be logged out?
image

khoaguin:
There is still an issue: In the test and notebook, we have different results when setting a role of a user to `admin` or `data_owner`. For example, if we have 3 users in the test function, the new admin can see all 3 of them. However, in the notebook, the new admin can only see 2 of them (not the original admin user) 
```
def test_user_view_set_role(worker, guest_client) -> None:
    admin_client = get_mock_client(worker.root_client, ServiceRole.ADMIN)
    assert admin_client.me.role == ServiceRole.ADMIN
    admin_client.register(
        name="Sheldon Cooper",
        email="sheldon@caltech.edu",
        password="changethis",
        institution="Caltech",
        website="https://www.caltech.edu/",
    )
    sheldon = admin_client.users[-1]
    sheldon.set_role("admin")
    ds_client = guest_client.login(email="sheldon@caltech.edu", password="changethis")
    assert len(ds_client.users.get_all()) == len(admin_client.users.get_all())
```
Then the `assert len(ds_client.users.get_all()) == len(admin_client.users.get_all())` will pass.

However, if we try to do the same thing in jupyter notebooks, we will have something different like below:
The original admin
image

The 2nd admin
image

khoaguin:
The bug "Role Permission not being propagated on query or get based calls" is fixed like below

image

shubham3121 (approved):
Amazing work @khoaguin !!! 
This will be super useful from a UX perspective.

PR #7914, opened 2023-07-10
Enclave single notebook high low

koenvanderveen:


packages/syft/src/syft/service/code/user_code.py
```diff
@@ -643,12 +644,15 @@ def add_custom_status(context: TransformContext) -> TransformContext:
             node_id=context.node.id,
             verify_key=context.node.signing_key.verify_key,
         )
-        if node_view in input_keys or len(input_keys) == 0:
```
koenvanderveen, line 646: We need to think about how we handle this. Not sure if it would be a security issue to not check this at all

shubham3121 (approved):
💯 🎉

packages/syft/src/syft/service/code/user_code.py
shubham3121: @rasswanth-s can you review this once.

notebooks/tutorials/enclaves/Enclave-single-notebook-DO-DS.ipynb
```diff
@@ -13,6 +13,7 @@
    ]
   },
   {
+   "attachments": {},
```
rasswanth-s, line 16: Q: Are both notebooks different in tutorials/enclaves?
I noticed similar testing.
We could keep one and delete the other?

koenvanderveen: one contains the high side and low side, the other does not

koenvanderveen: From a testing perspective they have the same coverage. From a tutorial perspective I think they are different things

packages/syft/src/syft/client/domain_client.py
```diff
@@ -62,11 +62,24 @@ def upload_dataset(self, dataset: CreateDataset) -> Union[SyftSuccess, SyftError
                 return tuple(valid.err())
             return valid.err()
 
-    def apply_to_gateway(self, client: Self) -> None:
+    def connect_to_gateway(
+        self,
+        url: Optional[str] = None,
+        port: Optional[int] = None,
+        handle: Optional["NodeHandle"] = None,  # noqa: F821
```
rasswanth-s, line 69: MAYBE: Instead of handle , we could have client directly ?

I think handle , is a specific , to when deployed by orchestra.

For example ,considering a domain node wanting to associate with a gateway.

1. ) They would initially explore all the gateways by

```python3
sy.gateways
```
2.) They would choose one gateway , they would like to explore

by 
```python3
gateway_client = sy.gateways[0]
```

since indexing sy.gateways automatically returns a client.

They could do directly
```python3
domain_client.connect_to_gateway(gateway_client)
```
or they would have to do
```python3
domain_client.connect_to_gateway(url = gateway_client.connection.url, port=gateway_client.connection.port)
```

packages/syft/src/syft/service/code/user_code.py
rasswanth-s: I think for now, we could proceed with the change.

One thing which is unclear, to me is the UX 
on how a data owner would identify , if this particular request is a request for some object on my machine, or it is request for data belonging to other people.

That is something, we could discuss.

packages/syft/src/syft/client/domain_client.py
koenvanderveen: To me, a UX where you are connecting a client to a client is very confusing. If I would be a user without context, I would not know what the result would be of that call. Assuming we wouldnt do `client1.connect(client2)`, I would for simplicity not create `client2` at all, but just pass in the url and port directly. But thats basically what we already have, which I would prefer.

koenvanderveen: cc @IrinaMBejan

packages/syft/src/syft/service/code/user_code.py
koenvanderveen: They wouldnt, as they wouldnt even be aware of the request in our use case

packages/syft/src/syft/client/domain_client.py
koenvanderveen: I am going to merge this, we can address this question in a separate discussion/PR

IrinaMBejan: I see Rasswanth's point - there's no way (that I am aware of) to get a handle for a gateway when your node is deployed using hagrid as a domain server. The only way to get access to it is via client = sy.login(url="", .., ..). So when sy.orchestra is not used, how does this call look like: 

gateway_node = ???
do_client.connect_to_gateway(handle=gateway_node)

koenvanderveen: so instead of 
```
gateway_client = sy.login((url="http://192.168.0.1", port=8000)
do_client.connect_to_gateway(gateway_client)
```
You would have
```
do_client.connect_to_gateway(url="http://192.168.0.1", port=8000)
```

shubham3121: How about we change on how its being read by the user:

`do_client.connect_to_gateway(via_client=gateway_client)`

This way we're saying that the clients are not connected but connection is made through the client.
We can still have other Options as well.

koenvanderveen: This would also work for me, but I do still think its kind of wasteful and potentially confusing dor users to be forced to create a client they only use for one thing (that doesnt require a client)

shubham3121: So I meant to propose, may be we have all three options 😅 
cause all of them are valid in different scenarios.

```python
do_client.connect_to_gateway(
    via_client: Optional[SyftClient], 
    handle: Optional[NodeHandle], 
    url: Optional[str], 
    port: Optional[int]
)
```

Passing client is easy when it is a OM hosted public gateway. Since we have access to client directly.
```python
gateway_client = sy.gateways[0]
do_client.connect_to_gateway(via_client=gateway_client)
```

Using node if we have access to one
```python
gateway_node = sy.orchestra(....)
do_client.connect_to_gateway(handle=gateway_node)
```

Passing url alternatively
```
url = "..''
do_client.connect_to_gateway(url=url)
```

koenvanderveen: oh, I like that

PR #7921, opened 2023-07-11
Syft function versioning

ivyclare:
## Description

This PR addresses [the issue of overwriting user-defined functions in PySyft](https://github.com/OpenMined/Heartbeat/issues/354). Currently, submitting a function with the same name as an existing one results in an error. With the changes introduced in this PR, users should be able to submit different versions of the function without any issues. Also, users should be able to access previous versions of their function.

To achieve this functionality, we have introduced a CodeInterface that maintains a mapping of function names to their respective versions. A CodeInterfaceStash is used to store these interfaces, and a CodeInterfaceService to manage them. 

This PR is a work-in-progress.

## Affected Dependencies

## How has this been tested?

## Checklist
- [] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)


packages/syft/src/syft/service/code_history/code_history_service.py
```diff
@@ -0,0 +1,261 @@
...
+        self, context: AuthedServiceContext
+    ) -> Union[List[CodeHistory], SyftError]:
+        """Get a Dataset"""
+        result = self.stash.get_all(context.credentials)
+        if result.is_ok():
+            return result.ok()
+        return SyftError(message=result.err())
+
+    @service_method(
+        path="code_history.get_by_id", name="get_by_id", roles=DATA_SCIENTIST_ROLE_LEVEL
```
koenvanderveen: maybe just .get?

packages/syft/src/syft/service/code_history/code_history_service.py
```diff
@@ -0,0 +1,261 @@
...
+    ) -> Union[SyftSuccess, SyftError]:
+        """Get a User Code Item"""
+        result = self.stash.get_by_uid(context.credentials, uid=uid)
+        if result.is_ok():
+            code_history = result.ok()
+            return code_history
+        return SyftError(message=result.err())
+
+    @service_method(path="code_history.delete_by_id", name="delete_by_id")
+    def delete(self, context: AuthedServiceContext, uid: UID):
```
koenvanderveen, line 121: Lets do .delete for consistency with other services

packages/syft/src/syft/service/code_history/code_history_service.py
```diff
@@ -0,0 +1,261 @@
...
+
+        final_result = None
+        if isinstance(code, SubmitUserCode):
+            if add_request:
+                result = user_code_service.request_code_execution(
+                    context=context, code=code, reason=reason
+                )
+                if isinstance(result, SyftError):
+                    return result
+                code = (
```
koenvanderveen: lets add a type here and on 72, 78

packages/syft/src/syft/service/code_history/code_history_service.py
```diff
@@ -0,0 +1,261 @@
...
+                    .changes[0]
+                    .linked_obj.resolve_with_context(context=context)
+                    .ok()
+                )
+                final_result = result
+            else:
+                result = user_code_service.submit(context=context, code=code)
+                if isinstance(result, SyftError):
+                    return result
+                uid = UID.from_string(result.message.split(" ")[-1])
```
koenvanderveen: maybe we can access the stash directly or create a intermediate method that does this and returns the object so we dont have to parse the id from the message

packages/syft/src/syft/service/code_history/code_history_service.py
```diff
@@ -0,0 +1,261 @@
...
+            credentials=context.credentials, service_func_name=code.service_func_name
+        )
+
+        if result_code_history_list.is_err():
+            return SyftError(message=result_code_history_list.err())
+
+        code_history = None
+        code_history_list = result_code_history_list.ok()
+
+        for elem in code_history_list:
```
koenvanderveen: this logic should be handled by the stash, ideally it would also just use both keys (service_func_name and user verify key) to search instead of loop

packages/syft/src/syft/service/code_history/code_history_service.py
```diff
@@ -0,0 +1,261 @@
...
+
+    @service_method(path="code_history.delete_by_id", name="delete_by_id")
+    def delete(self, context: AuthedServiceContext, uid: UID):
+        result = self.stash.delete_by_uid(context.credentials, uid)
+        if result.is_ok():
+            return result.ok()
+        else:
+            return SyftError(message=result.err())
+
+    def fetch_history_for_user(
```
koenvanderveen: the singular naming confuses me a bit here, because it still gets all the CodeHistorie**s** but not for all users. It looks like we have an inconsistency between what we call a history in the frontend (all versions of all functions for one user), and in the backend (all versions of one function for one user)

packages/syft/src/syft/service/code_history/code_history.py
```diff
@@ -0,0 +1,110 @@
...
+
+@serializable()
+class UserHistoryDict(SyftObject):
+    # version
+    __canonical_name__ = "UserHistoryDict"
+    __version__ = SYFT_OBJECT_VERSION_1
+
+    id: UID
+    node_uid: UID
+    user_dict: Optional[Dict[str, List[str]]] = {}
```
koenvanderveen: why optional?

packages/syft/src/syft/service/code_history/code_history_service.py
```diff
@@ -0,0 +1,261 @@
...
+    @service_method(
+        path="code_history.get_histories",
+        name="get_histories",
+        roles=DATA_OWNER_ROLE_LEVEL,
+    )
+    def get_histories_group_by_user(self, context: AuthedServiceContext):
+        result = self.stash.get_all(credentials=context.credentials)
+        if result.is_err():
+            return SyftError(message=result.err())
+        code_histories = result.ok()
```
koenvanderveen: lets add a type, makes it much easier to read the code

packages/syft/src/syft/service/code_history/code_history_service.py
koenvanderveen: maybe similar to how we do it in get_by_func_name_and_user_email?

packages/syft/src/syft/service/code_history/code_history.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+@serializable()
+class CodeVersions(SyftObject):
+    # version
+    __canonical_name__ = "CodeVersions"
+    __version__ = SYFT_OBJECT_VERSION_1
+
+    id: UID
+    user_code_history: Optional[List[UserCode]] = []
```
koenvanderveen: naming here is also slightly confusing, we call it history, but its value is a list of UserCode, while we have a separate object called history

koenvanderveen: Also it shouldnt be optional I believe, but possibly an empty list

packages/syft/src/syft/service/code_history/code_history.py
```diff
@@ -0,0 +1,110 @@
...
+class CodeHistory(SyftObject):
+    # version
+    __canonical_name__ = "CodeHistory"
+    __version__ = SYFT_OBJECT_VERSION_1
+
+    id: UID
+    node_uid: Optional[UID]
+    user_verify_key: SyftVerifyKey
+    enclave_metadata: Optional[EnclaveMetadata] = None
+    user_code_history: Optional[List[UID]] = []
```
koenvanderveen: remove optional

packages/syft/src/syft/service/code_history/code_history.py
```diff
@@ -0,0 +1,110 @@
...
+    service_func_name: str
+    comment_history: Optional[List[str]] = []
```
koenvanderveen: remove optional

packages/syft/src/syft/service/code_history/code_history.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+@serializable()
+class CodeHistoryDict(SyftObject):
+    # version
+    __canonical_name__ = "CodeHistoryDict"
+    __version__ = SYFT_OBJECT_VERSION_1
+
+    id: UID
+    code_versions: Optional[Dict[str, CodeVersions]] = {}
```
koenvanderveen: remove optional

koenvanderveen:
few notes on reprs that I communicated earlier:
- Would be nice with repr for a list of codeversion to see how many versions each CodeVersioon has (jane_client.code_history)
- Would be nice for a each version of a UserCode to see which version it is and when it was submitted time (jane_client.code_history.test_func)
- admin_client.code_histories, needs a repr

koenvanderveen:
apart from all the comments, I want to mention again that this is an amazing PR :heart:

koenvanderveen:
Thought about the negative indexing thing and whether its necessary, I think its still something we dont want for the DO because we want to prevent
```
code = client.code_histories["my_user"].my_func[-1]
func = code.unsafe_function
res = func()
request.accept_by_depositing_result(res)
```

packages/syft/src/syft/client/domain_client.py
```diff
@@ -148,6 +156,20 @@ def projects(self) -> Optional[APIModule]:
...
         return None
 
+    @property
+    def code_history_service(self) -> Optional[APIModule]:
+        if self.api is not None and self.api.has_service("code_history"):
+            return self.api.services.code_history
+        return None
+
+    @property
+    def code_history(self) -> Dict[str, CodeHistory]:
```
koenvanderveen: returntype is not accurate, returns a CodeHistoryDict

packages/syft/src/syft/client/domain_client.py
```diff
@@ -148,6 +156,20 @@ def projects(self) -> Optional[APIModule]:
...
+        return self.api.services.code_history.get_history()
+
+    @property
+    def code_histories(self) -> Dict[str, CodeHistory]:
```
koenvanderveen: returns as UserHistoryDict

koenvanderveen:
I think in general we need more consistency in the naming in the code_history_service, we have

code_history.get -> return CodeHistory
code_history.get_all -> returns [CodeHistory]
code_history.get_history -> returns CodeHistoryDict[CodeVersions]
code_history.get_history_for_user -> CodeHistoryDict[CodeVersions]
code_history.get_by_name_and_user_email -> returns [CodeHistory]
code_history.get_histories_group_by_user -> returns UserHistoryDict

there are a 2 main inconsistencies

1. CodeHistoryDict does not contain Codehistory object but CodeVersions
2. get_history give you versions instead of histories

the solution could be
CodeVersion and CodeHistory are almost the same object. Maybe we should call them CodeHistory and CodeHistoryView to communicate that they are similar. This would also make more sense for 2), because you would actually have the right type of object in your dict (view instead of real, but thats fine I suppose)

packages/syft/src/syft/service/code_history/code_history_service.py
teo-milea: sure!

packages/syft/src/syft/service/code_history/code_history_service.py
teo-milea: not sure if I understand this one. What we need here is the code object to validate de UID and the service_func_name, and we try to get that from the server to avoid any attacks. The result of user_code.submit was not returning anything useful which is way we added the id and parsed it in the function. Not sure how to do that without bloating the message result

packages/syft/src/syft/service/code_history/code_history_service.py
teo-milea: yep, it was an oversight on our part

packages/syft/src/syft/service/code_history/code_history.py
teo-milea: will remove

packages/syft/src/syft/service/code_history/code_history_service.py
teo-milea: for code_histories?

koenvanderveen: yes

koenvanderveen (approved):

PR #7926, opened 2023-07-12
make code.request_code_exeuction and request_code_execution the same …

koenvanderveen:


packages/syft/src/syft/client/enclave_client.py
```diff
@@ -35,10 +35,19 @@ class EnclaveMetadata(SyftObject):
...
+    __api_patched = False
+
     @property
     def code(self) -> Optional[APIModule]:
         if self.api is not None and self.api.has_service("code"):
-            return self.api.services.code
+            res = self.api.services.code
+            # the order is important here
+            # its also important that patching only happens once
+            if not self.__api_patched:
```
rasswanth-s, line 46: Superb!

PR #7955, opened 2023-07-18
propagate outputs to shareholders on enclaves

koenvanderveen:


packages/syft/src/syft/client/enclave_client.py
```diff
@@ -42,6 +42,7 @@ class EnclaveClient(SyftClient):
 
     @property
     def code(self) -> Optional[APIModule]:
+        self
```
shubham3121: this needs to be removed ?

packages/syft/src/syft/service/action/action_service.py
```diff
@@ -211,8 +213,18 @@ def _user_code_execute(
             syft_object=result_action_object,
             has_result_read_permission=True,
         )
+
         if set_result.is_err():
             return set_result.err()
+
+        if len(code_item.output_policy.output_readers) > 0:
```
shubham3121, line 220: 👍🏾

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -507,16 +558,20 @@ def debox_asset(arg: Any) -> Any:
...
         input_policy=ExactMatch(*args, **kwargs),
         output_policy=SingleExecutionExactOutput(),
+        share_results_with_owners=share_results_with_owners,
     )
 
 
 def syft_function(
     input_policy: Union[InputPolicy, UID],
     output_policy: Optional[Union[OutputPolicy, UID]] = None,
+    share_results_with_owners=False,
```
shubham3121, line 574: Should we make it True by default because this is something that DS controls ?

shubham3121 (approved):
Excellent work @koenvanderveen !! 
This makes the code much more clean and readable 💯 🎉

packages/syft/src/syft/service/code/user_code.py
koenvanderveen: irina suggested False, I agree. In theory the output still could be data that is sensitive for the data owners so I would not have it on by default

packages/syft/src/syft/client/enclave_client.py
koenvanderveen: yes, done

PR #7970, opened 2023-07-27
Comment out pip install to speed up CI

kiendang:
## Description
Comment out all the `%pip install` in notebooks to speed up CI since each of these takes seconds (up to 5 or 6s on my local machine) even when `syft` is already installed.

All the notebooks under `notebooks/tutorials/pandas-cookbook` are alr doing this.

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/notebooks/api/0.8/00-load-data.ipynb
+++ b/notebooks/api/0.8/00-load-data.ipynb
@@ -32,7 +32,7 @@
    "source": [
     "SYFT_VERSION = \">=0.8.2.b0,<0.9\"\n",
     "package_string = f'\"syft{SYFT_VERSION}\"'\n",
-    "%pip install {package_string} -f https://whls.blob.core.windows.net/unstable/index.html -q"
+    "# %pip install {package_string} -f https://whls.blob.core.windows.net/unstable/index.html -q"
    ]
   },
   {
@@ -680,7 +680,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.16"
+   "version": "3.11.3"
   },
   "toc": {
    "base_numbering": 1,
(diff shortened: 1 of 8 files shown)


notebooks/api/0.8/00-load-data.ipynb
```diff
@@ -32,7 +32,7 @@
    "source": [
     "SYFT_VERSION = \">=0.8.2.b0,<0.9\"\n",
     "package_string = f'\"syft{SYFT_VERSION}\"'\n",
-    "%pip install {package_string} -f https://whls.blob.core.windows.net/unstable/index.html -q"
+    "# %pip install {package_string} -f https://whls.blob.core.windows.net/unstable/index.html -q"
```
rasswanth-s, line 35: I think we introduced this method, as when people run the notebook as it is , without syft, installed, it would help them with the right installation of syft.

But is the CI , getting really slow due to this ?

Probably , we could comment them.

@madhavajay  Ping

kiendang: > I think we introduced this method, as when people run the notebook as it is , without syft, installed, it would help them with the right installation of syft.

If they don't have syft installed they would know to uncomment that. Plus we already commented this out in the notebooks/tutorials/pandas-cookbook notebooks.

> But is the CI , getting really slow due to this ?

It's up to 5, 6s on my machine. Since time taken is probably mostly IO (contacting the server), I would expect it's the same for everyone. It also adds up fast. We have a few other `api/0.8` notebooks in development right now. This would add a couple minutes to each CI job.

There is a way to check if they have the right `syft` version (SYFT_VERSION) with just vanilla python. However the code is harder to read. I'd prefer people uncommenting the line themselves other than blindly running an obscure code cell.

rasswanth-s: Agree 👍

rasswanth-s (approved):
Well Done @kiendang  🚀

PR #7975, opened 2023-08-02
Fix the issue "Use of a broken or weak cryptographic hashing algorithm on sensitive data"

khoaguin:
Fixing this issue: https://github.com/OpenMined/PySyft/security/code-scanning/1
Although SHA-256 is a strong cryptographic hash function, it is not suitable for password hashing since it is not computationally expensive. Trying to switch to `Argon2` (through the `argon2-cffi` PyPI package), which is a strong password hashing algorithm (and includes a per-password salt by default) instead.


shubham3121 (approved):
🎉 ❤️

PR #7978, opened 2023-08-03
[WIP] Add initial API bridge login and auth demo

akalliokoski:
## Description

* NOTE: This implementation relies on manual fix in `openapi3` package
   * fixes applying session headers to the request
   * changed a line in `paths.py`, line 346
     * from: `result = session.send(self._request.prepare(), verify=verify)`
     * to: `result = session.send(session.prepare_request(self._request), verify=verify)`
* instead of setting the headers, an alternative solution could be to use `HTTPBearer`
  * and injecting `callable_op.security = { 'HTTPBearer': '<token>'  }`
  * didn't get it working yet, though
* content type `application/x-www-form-urlencoded` not supported by openapi3
* [aiopenapi3](https://github.com/commonism/aiopenapi3) could also be investigated 
     * fork of `openapi3`
     * e.g. has [manual access to request object](https://aiopenapi3.readthedocs.io/en/latest/advanced.html#manual-requests), [plugins](https://aiopenapi3.readthedocs.io/en/latest/plugin.html)

### TODO
 * Notify User if not logged in on Authed endpoints
 * Store information in a User Session in the Syft server
 * Customize login name method (either infer it from the API or allow AOs to provide it)
 

## Affected Dependencies
- manually modified openapi3 package to fix applying session headers

## How has this been tested?
- ran the `08-api-bridge.ipynb` notebook

## Checklist

### DIFF
--- a/.github/workflows/pr-tests-syft.yml
+++ b/.github/workflows/pr-tests-syft.yml
@@ -8,6 +8,7 @@ on:
       - dev
       - main
       - "0.8"
+      - "23_8_bridge"
 
   workflow_dispatch:
     inputs:
--- a/notebooks/api/0.8/08-api-bridge.ipynb
+++ b/notebooks/api/0.8/08-api-bridge.ipynb
@@ -4,7 +4,9 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "70877e23-bd5d-4d6d-bcdd-9dfa671ce12f",
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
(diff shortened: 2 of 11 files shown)


madhavajay (approved):
@akalliokoski excellent work! I dug around the openapi3 library and it does seem like they support HTTPBearer so I switched to that so we can at least test having a session. I think in reality openapiv3 will allow us to detect these auth mechanisms and automatically support them but I agree the library we are using might not be the best, its quite out of date now, but for now I think it demonstrates our desired flow.

PR #7981, opened 2023-08-04
Obsolete Workflow command "set-output"

bashbang:
## Description
The Github Action command set-output is being obsoleted in favour of using env $GITHUB_OUTPUT. See details here: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
This PR migrates to the new method.
Some of the workflows what run windows environments needed to the shell explicitly set to bash, so I conformed them all for consistency.
Note: There was one action that had the command commented out. This was also updated to the new method in case it's uncommented in the future.

## How has this been tested?
- A run of the pr-tests-* all passed except pr-tests-stack which was unable to run due to unavailability of om-ci-16vcpu-ubuntu2204. However the same method was implemented in all workflows so expect that the other workflows should work as expected.

## Checklist


.github/workflows/pr-tests-stack.yml
```diff
@@ -261,8 +264,9 @@ jobs:
       - name: Get pip cache dir
         if: steps.changes.outputs.stack == 'true'
         id: pip-cache
+        shell: bash
         run: |
-          echo "::set-output name=dir::$(pip cache dir)"
+          echo "dir=$(pip cache dir)" >> %GITHUB_OUTPUT
```
kiendang:
```suggestion
          echo "dir=$(pip cache dir)" >> $GITHUB_OUTPUT
```

kiendang: Great work. Thank you! Just 1 small fix.

rasswanth-s:
@bitsofsteve  PTAL

.github/workflows/pr-tests-stack.yml
bashbang: Good catch!

kiendang (approved):

bitsofsteve (approved):
@bashbang LGTM 👍🏾, Great work spotting these! I'll merge this now.

PR #7983, opened 2023-08-04
Specify if result was obtained from mock or real data with `SyftInfo`. Show `SyftWarning` for some WARNING messages

khoaguin:
According to this ticket: [Specify if result was obtained from mock or real data](https://github.com/OpenMined/Heartbeat/issues/608)

Also show some warning messages with `SyftWarning` instead of just printing out


khoaguin:
image

khoaguin:
Before:
image

Now:
image

khoaguin:
If we only put `Asset` class as an argument to the `unsafe_function`, we get the message telling it's PRIVATE data if the user has the permission, otherwise MOCK data
image

khoaguin:
Otherwise if we input something specific like `data=asset.mock` into the `unsafe_function`, we will get back the message like below
image

khoaguin:
image

tcp (approved):

PR #7987, opened 2023-08-07
SeaweedFS blob storage

kiendang:
## Description

## Affected Dependencies

## How has this been tested?

## Checklist


packages/syft/src/syft/service/action/action_object.py
```diff
@@ -881,8 +942,8 @@ def __post_init__(self) -> None:
             if side_effect not in self._syft_post_hooks__[HOOK_ALWAYS]:
                 self._syft_post_hooks__[HOOK_ALWAYS].append(side_effect)
 
-        if isinstance(self.syft_action_data, ActionObject):
-            raise Exception("Nested ActionObjects", self.syft_action_data)
+        if isinstance(self.syft_action_data_type, ActionObject):
```
kiendang, line 1082: should this be `issubclass` instead of `isinstance` since `syft_action_data_type` is a type?

packages/syft/src/syft/service/user/user_service.py
```diff
@@ -210,7 +210,12 @@ def get_current_user(
                 SyftError(message="User not found!")
         return SyftError(message=str(result.err()))
 
-    @service_method(path="user.update", name="update", roles=GUEST_ROLE_LEVEL)
+    @service_method(
+        path="user.update",
+        name="update",
+        roles=GUEST_ROLE_LEVEL,
+        autosplat=["user_update"],
```
kiendang: @koenvanderveen I reverted this for now. This breaks `user_test.py` even after updating the syntax. `UserUpdate` has a couple of validators which don't get applied if using `autosplat`.

https://github.com/OpenMined/PySyft/blob/687aefd64076f312e26ebd855221baf200d418f2/packages/syft/src/syft/service/user/user.py#L114-L122

koenvanderveen: 👍

packages/syft/src/syft/service/action/action_object.py
```diff
@@ -436,29 +468,97 @@ class ActionObject(SyftObject):
...
+                if isinstance(blob_retrieval_object, SyftError):
+                    print(
+                        "Detached action object, object exists but is not linked to data in the blob storage",
+                        blob_retrieval_object,
+                    )
+                    return blob_retrieval_object
+                self.syft_action_data_cache = blob_retrieval_object.read()
+                self.syft_action_data_type = type(self.syft_action_data)
+
+    def _set_syft_action_data(self, data: Any) -> None:
```
koenvanderveen: I think if we could rename this, it basically just uploads to the blob store. The fact that it then sets the data is maybe less important, especially because its only used in _save_to_blob_store, which just takes the existing data from the object and then passes it into this method.

packages/syft/src/syft/service/action/action_object.py
```diff
@@ -474,6 +574,40 @@ def make_id(cls, v: Optional[UID]) -> UID:
...
+                values["syft_action_data_repr_"] = (
+                    v._repr_markdown_()
+                    if hasattr(v, "_repr_markdown_")
+                    else v.__repr__()
+                )
+            values["syft_action_data_str_"] = str(v)
+            values["syft_has_bool_attr"] = hasattr(v, "__bool__")
+        return values
+
+    def _save_to_blob_store(self) -> Optional[SyftError]:
```
koenvanderveen: I think this always coupled with client.api.services.action.set, maybe we should have method that does both? I am still thinking about the coupling of these two, because if you only call client.api.services.action.set OR \_save_to_blob_store, and not the other, you end up in pretty weird state. One solution could be to have an endpoint that does both, and make the two individual ones "private" (prefix with '_')

packages/syft/src/syft/store/blob_storage/seaweedfs.py
```diff
@@ -1,43 +1,229 @@
...
+                Key=file_name,
+            )
+            upload_id = result["UploadId"]
+            return SeaweedSecureFilePathLocation(upload_id=upload_id, path=file_name)
+        except BotoClientError as e:
+            return SyftError(
+                message=f"Failed to allocate space for {obj} with error: {e}"
+            )
 
+    def write(self, obj: BlobStorageEntry) -> BlobDeposit:
```
koenvanderveen, line 178: maybe rename this method? I think "write" gives a different expectation, especially since we write the result again, resulting in this pattern

```
obj = ...
res = client.write(obj)
res.write()
```

packages/syft/src/syft/service/action/action_object.py
```diff
@@ -436,29 +468,97 @@ class ActionObject(SyftObject):
     _syft_post_hooks__: Dict[str, List] = {}
     syft_twin_type: TwinMode = TwinMode.NONE
     syft_passthrough_attrs = BASE_PASSTHROUGH_ATTRS
+    syft_action_data_type: Optional[Type]
```
koenvanderveen, line 472: what is the difference between syft_internal_type and syft_action_data_type

packages/syft/src/syft/service/action/action_object.py
```diff
@@ -579,6 +713,19 @@ def _syft_try_to_save_to_store(self, obj) -> None:
         # relative
         from ...client.api import APIRegistry
 
+        if obj.syft_node_location is None:
+            obj.syft_node_location = obj.syft_node_uid
```
koenvanderveen, line 715: what is the difference between syft_node_location and syft_node_uid?

packages/syft/src/syft/store/blob_storage/seaweedfs.py
kiendang: Ya these method names are not precise and definitely bikeshed-able but not a huge deal to me tbh.
The `read` and `write` methods under Connection create an object to be sent to the client which they can use to retrieve/deposit a file or `SyftObject` from/to blob storage. Could be `(make/prepare/to)_read/write` but these are not precise either.

packages/syft/src/syft/service/action/action_object.py
kiendang: Doesn't look like there's a difference to me either. Will discuss with Shubham on this and other comments. Thanks for the review!

packages/syft/src/syft/service/action/action_object.py
shubham3121: As of now `syft_action_data_type` always stores the type of the data being stored and is calculated on the fly. While what I can see `syft_internal_type` is set explicitly during different Action Object subclasses like `NumpyBoolObject`, `NumpyScalarObject`, etc.

packages/syft/src/syft/service/action/action_object.py
shubham3121: There is not much difference, `syft_node_location` is set everytime an object is passed through the service_method decorator. While `syft_node_uid` is used to store the id of the node when its saved. Since both of them were developed in parallel, there is a slight redundancy in their use. I agree we should combine them, but would prefer to do that in a seperate PR.

packages/syft/src/syft/service/action/action_object.py
```diff
@@ -474,6 +575,40 @@ def make_id(cls, v: Optional[UID]) -> UID:
...
+            return data
+        if isinstance(data, ActionDataEmpty):
+            print(f"cannot store empty object {self.id}")
+            return data
+        result = self._set_syft_action_data(data)
+        if isinstance(result, SyftError):
+            return result
+        if not TraceResult.is_tracing:
+            self.syft_action_data_cache = self.as_empty_data()
+
```
teo-milea: I don't understand the purpose of emptying the cache here. What is TraceResult?

packages/syft/src/syft/service/action/action_object.py
```diff
@@ -474,6 +575,40 @@ def make_id(cls, v: Optional[UID]) -> UID:
...
+            values["syft_has_bool_attr"] = hasattr(v, "__bool__")
+        return values
+
+    def _save_to_blob_store(self) -> Optional[SyftError]:
+        data = self.syft_action_data
+        if isinstance(data, SyftError):
+            return data
+        if isinstance(data, ActionDataEmpty):
+            print(f"cannot store empty object {self.id}")
+            return data
```
teo-milea: We should return an error here as well, so we don't have to check everytime if the result is an error or an ActionDataEmpty

kiendang:
@shubham3121 I did the same `accept_by_depositing_result` fix but got another error in notebook 05 after that

```python
res_ptr = domain_client.code.func(x=x_pointer)
res = res_ptr.get()
res

AttributeError                            Traceback (most recent call last)
Cell In[21], line 2
      1 res_ptr = domain_client.code.func(x=x_pointer)
----> 2 res = res_ptr.get()
      3 res

AttributeError: 'SyftError' object has no attribute 'get'
```

Do you encounter the same thing locally?

packages/syft/src/syft/service/action/action_object.py
kiendang: Agree. We could change the name of this. Initially this was created to be the setter for `syft_action_data` as you can see the line right after this function `syft_action_data = syft_action_data.setter(_set_syft_action_data)`. As we don't set `syft_action_data` directly now but through `syft_action_data_cache` we should delete this and rename the function.

There is `def delete_data(self)` which sets `self.syft_action_data = empty`. I don't see it being used anywhere. Should we remove it or modify it (so that it sets `syft_action_data_cache` and removes from blob storage as well)?

teo-milea:
I got the same issue as @kiendang locally with the same fix, how did the tests pass?

shubham3121:
yes, getting the same error, that's weird the tests passed. The problem seems to be in `apply_output` in `RepeatedCallPolicy`: outputs[output_arg] is throwing the error.

kiendang:
Hmm I thought `outputs[output_arg]` raised the previous error which got fixed with `accept_by_depositing_result`?
The error from `domain_client.code.func(x=x_pointer)` looks like a separate issue.

```python
res_ptr = domain_client.code.func(x=x_pointer)
res_ptr

SyftError: Failed to run. 'NoneType' object has no attribute 'services'
```

teo-milea:
I have changed the example policy from notebook 5 as this feature is not really used and we could offer support to anyone trying it out. Given the changes from the last few months, we need to refactor and rethink the UX for custom policies, but it does not feel like a priority right now.

packages/syft/src/syft/service/action/action_service.py
```diff
@@ -71,10 +81,22 @@ def set(
     ) -> Result[ActionObject, str]:
         """Save an object to the action store"""
         # 🟡 TODO 9: Create some kind of type checking / protocol for SyftSerializable
+
+        if isinstance(action_object, ActionObject):
+            action_object.syft_created_at = DateTime.now()
+        else:
+            action_object.private_obj.syft_created_at = DateTime.now()
+            action_object.mock_obj.syft_created_at = DateTime.now()
+
```
teo-milea, line 90: Nitpick: Could we be specific here that we check if the object is a TwinObject, and let the default case to be the ActionObject?

packages/syft/src/syft/service/request/request.py
```diff
@@ -97,29 +99,63 @@ def _run(
...
             obj_uid = self.linked_obj.object_uid
             obj_uid = obj_uid.id if isinstance(obj_uid, LineageID) else obj_uid
 
+            action_obj = action_store.get(
+                uid=obj_uid,
+                credentials=context.approving_user_credentials,
+            )
+
+            if action_obj.is_err():
+                return Err(SyftError(message=f"{action_obj.err()}"))
```
teo-milea, line 116: Nitpick: do we need both Err and SyftErr?

packages/syft/src/syft/service/request/request.py
```diff
@@ -792,7 +841,7 @@ def _run(
                 return Err(valid)
             obj = self.linked_obj.resolve_with_context(context)
             if obj.is_err():
-                return SyftError(message=obj.err())
+                return Err(SyftError(message=obj.err()))
```
teo-milea, line 844: NP: same problem with Err and SyftError

packages/syft/src/syft/service/request/request.py
```diff
@@ -603,7 +652,7 @@ def _run(
         try:
             obj = self.linked_obj.resolve_with_context(context)
             if obj.is_err():
-                return SyftError(message=obj.err())
+                return Err(SyftError(message=obj.err()))
```
teo-milea, line 655: NP: same problem with Err and SyftError

kiendang:
Notebook 06 also fails for me locally.

teo-milea:
Works for me, can you share the error?

kiendang:
Screenshot 2023-09-01 at 5 43 04 PM

shubham3121:
@kiendang notebook 6 is working fine. Can you take the latest pull ?

teo-milea (approved):

kiendang:
> @kiendang notebook 6 is working fine. Can you take the latest pull ?

Yup it works now after clearing the db from previous runs.

packages/syft/src/syft/service/action/action_object.py
kiendang: @koenvanderveen

packages/syft/src/syft/service/request/request.py
shubham3121: @teo-milea there is a bit inconsistent use of Err and SyftError at request and Action store. Resorting to still use Result library over SyftError to prevent breaking existing functionality. We need to refactor this a bit and make their use consistent all across the codebase.

packages/syft/src/syft/service/action/action_object.py
shubham3121: TraceResult is used in Plans (check planify method). TraceResult is a global state and when it is enabled, then we're executing a Plan over local variables. At that point we don't want to clear the cache. 

Reason for Setting cache to empty:
This is because, now saving Action Objects include two steps:
- save actual data (in syft_action_data_cache) to blob store
- post that, save the Action Object itself to the Action Store. Now since we don't want to save the actual data present in `syft_action_data_cache` to the ActionStore (as that data is already stored to blob store) we clear it or set it as empty post saving to blob store.

packages/syft/src/syft/service/action/action_object.py
shubham3121: I think we should remove this method as it creates confusion and doesn't perform the functionality as it suggests.

madhavajay:
🚂 Choo Choo

PR #7996, opened 2023-08-10
ContainerCommand Hooks

madhavajay:
## Description
- moved logic for skypilot resources to pre hooks with session data
- added a100 to fastapi application
- added pre / post hook code to container command endpoints
- added extra kwargs to allow custom overrides
- added debug to keep containers running
- added training with full dataset


rasswanth-s:
GPU!!!!
PVC_AEPerformance_Brian-5976060

PR #7997, opened 2023-08-11
Syft CLI

yashgorana:
## Description
Closes https://github.com/OpenMined/Heartbeat/issues/634

## Affected Dependencies
* Adds `typer` as a new dep for the project

## How has this been tested?
- Unit tests

## Checklist


packages/syft-cli/README.md
```diff
@@ -0,0 +1,10 @@
+# Syft CLI
+
+## Development
+
+    $ pip install -e packages/syft-cli[dev]
+    $ syftcli hello
```
kiendang: A bit nitpick but I prefer putting shell commands in Github markdown like this

    ```sh
    pip install -e packages/syft-cli[dev]
    syftcli hello
    ```

instead of prefixing with `$ `. The reason is Github added the copy button to code blocks so without the `$ ` I can very quickly hit copy and paste into terminal and run immediately without having to delete the `$ `.

Though I do agree putting `$ ` makes it more clear it's a shell command so it's ok either way.

madhavajay (commented):
Great work. Can you add the same linting checks we have on github and in pre-commit that we normally run for hagrid but for this new project as well?

packages/syft-cli/README.md
rasswanth-s: +1 , good suggestion Kien

packages/syft-cli/cli.py
```diff
@@ -0,0 +1,29 @@
...
+
+@app.command()
+def hello():
+    typer.echo("Syft CLI 👋")
+
+
+@app.command()
+def version():
+    # syft absolute
+    import syft
```
rasswanth-s: MAYBE: We could put importing syft in a try catch, for the usecases, in which use the CLI independently without installing syft to prevent the errors

packages/syft-cli/setup.py
```diff
@@ -0,0 +1,34 @@
+# third party
+from setuptools import find_packages
+from setuptools import setup
+from version import __version__
+
+packages = [
+    "typer[all]==0.9.0",
+    "typing_extensions==4.6.3",
```
rasswanth-s, line 8: Q: I noticed we are not using typing_extensions currently or do we plan to use it later ?

packages/syft-cli/tests/version_test.py
```diff
@@ -0,0 +1,12 @@
+# third party
+from cli import app
+from typer.testing import CliRunner
+
+runner = CliRunner()
+
+
+def test_version():
```
rasswanth-s: Flagging this in the Cool Down Period:

Currently to be able to move fast, we could shift adding github tests for the new CLI in the cool down period.

We would need to add
- A new workflow file for testing the new CLI
- Adding a tox test which automatically test the CLI


I will create a separate Github Issue for the same.


Currently, I have tested the tests manually.

rasswanth-s: Maybe  we could add the only tox test in this PR, itself, (as it is a short change)

We could create one with name `syftcli.test.unit` (or any other name also works) like we have for syft.test.unit

Kindly have a look at tox.ini, we could replicate syft.test.unit for syftcli.

rasswanth-s: We could keep the task of adding it to Github CI in the cool down period.

rasswanth-s:
Well Done @yashgorana , on the initial start.

I think once ,we address all the comments above.

I think, we should be good to merge the PR.

packages/syft-cli/cli.py
rasswanth-s: I am also thinking, on should we even display the syft version , when printing cli version ,as we are thinking of cli as an independent tool.

rasswanth-s: Printing syft version also is okay , it should be not be an issue.

packages/syft-cli/tests/version_test.py
```diff
@@ -0,0 +1,12 @@
...
+    result = runner.invoke(app, ["version"])
+    assert result.exit_code == 0
+    assert "CLI" in result.stdout
+    assert "Library" in result.stdout
```
rasswanth-s, line 12: MAYBE: When we modify syft version to be optional (during syft version), I think we would modify this line also (to optionally check Library string), when people run syft cli tests without installing syft .

packages/syft-cli/cli.py
yashgorana: > We could put importing syft in a try catch

Ah great idea - implemented this

packages/syft-cli/setup.py
yashgorana: Yes for using `Annotated` type hint - https://typer.tiangolo.com/tutorial/options/help/

.pre-commit-config.yaml
```diff
@@ -93,7 +93,7 @@ repos:
     rev: v1.0.0
```
rasswanth-s, line 91: Could we include , syft-cli also in `black` pre-commit hook, line no: 73 in .pre-commit-config.yaml
under `files`

rasswanth-s (approved):
Wonderful Work @yashgorana , left one minor comment with black, other than that, we could merge once the tests pass.

yashgorana:
> Wonderful Work @yashgorana , left one minor comment with black, other than that, we could merge once the tests pass.

The regex `files: ^packages/syft(?!(.*_pb2\\.py$)).*\\.py` already picks up syft-cli. 
image

.pre-commit-config.yaml
yashgorana: Please see this comment - https://github.com/OpenMined/PySyft/pull/7997#issuecomment-1678621326

PR #8017, opened 2023-08-17
Permission Model for MongoDB

khoaguin:
Related issues:
[[Security] Enable Permissions Model in MongoDB Store#642](https://github.com/OpenMined/Heartbeat/issues/642)
[Inconsistency in Display Order with different Databases#645](https://github.com/OpenMined/Heartbeat/issues/645)


khoaguin:
Now after `node = sy.orchestra.launch(name="test-domain", reset=True)` we have 2 permissions collections for `User` and `NodeSettings`
image

shubham3121 (approved):
Wonderful work @khoaguin !! 🎉 🔥

packages/syft/src/syft/store/mongo_document_store.py
```diff
@@ -211,23 +232,43 @@ def _set(
...
         can_write = self.has_permission(write_permission)
+
+        store_query_key = self.settings.store_key.with_obj(obj)
+        collection_status = self.collection
+        if collection_status.is_err():
+            return collection_status
+        collection = collection_status.ok()
+
+        store_key_exists = (
+            collection.find_one({"_id": store_query_key.value}) is not None
```
koenvanderveen: can we assume this is always an id?

packages/syft/src/syft/store/mongo_document_store.py
```diff
@@ -211,23 +232,43 @@ def _set(
...
-                return collection_status
-            collection = collection_status.ok()
-
             if ignore_duplicates:
                 collection = collection.with_options(write_concern=WriteConcern(w=0))
             try:
                 collection.insert_one(storage_obj)
             except DuplicateKeyError as e:
                 return Err(f"Duplicate Key Error for {obj}: {e}")
+
```
koenvanderveen, line 264: not sure if the logic here is completely the same as the logic in the kv_docstore, (althought that logic also seems a bit flawed). In the kv_doc store, if the item already exists and we do ignore duplicates, we would **not write**, but return the original object instead. I am not saying that that logic makes sense, but it looks like there is a discrepancy between that and this logic

packages/syft/src/syft/store/mongo_document_store.py
```diff
@@ -345,12 +386,117 @@ def _delete(
         return Err(f"Failed to delete object with qk: {qk}")
 
     def has_permission(self, permission: ActionObjectPermission) -> bool:
```
koenvanderveen, line 398: :+1:

koenvanderveen:
Overall this PR looks very good to me, great work @khoaguin! I am not completely sure about the test coverage of the mongo store, do we have any visibility of on which tests use this? Also, it would be quite good to have a notebook that shows how you can easily start a node with a mongo store using orchestra (maybe for a later PR as well, but we can make a ticket now).

packages/syft/src/syft/store/mongo_document_store.py
khoaguin: Actually we can't be sure since `store_query_key` is a `QueryKey` and its `value` has type `Any`
image

khoaguin:
Thank you @koenvanderveen . From what I know we have the tests in `mongo_document_store_test.py` but they are more about the functionality of the methods like `set`, `update`..., rather than the permissions

koenvanderveen:
> Thank you @koenvanderveen . From what I know we have the tests in `mongo_document_store_test.py` but they are more about the functionality of the methods like `set`, `update`..., rather than the permissions

Oke, I think its not necessary to do it now, but we can at least add a ticket for triaging

khoaguin:
Yes sure @koenvanderveen. Please start the ticket and assign it to me, I think I can and should start working on it right away since memory is still fresh. By the way, I am done with the notebook to test this feature. Where should I put it?

khoaguin:
Hi @shubham3121  @koenvanderveen, here is the notebook containing some code to test the feature
[mongodb-permission-model.zip](https://github.com/OpenMined/PySyft/files/12496379/mongodb-permission-model.zip)

packages/syft/src/syft/store/mongo_document_store.py
khoaguin: when not setting `WriteConcern(w=0)`, an error will be raised if the item already exists.
When we set  `WriteConcern(w=0))`, adding existed items will not raise error, and the database will not be changed (so it kept the existed item)

packages/syft/src/syft/store/mongo_document_store.py
shubham3121: `store_query_key.as_dict_mongo` handles that logic. Updated the logic there.
Resolved here: https://github.com/OpenMined/PySyft/pull/8017/commits/6b5cc8e70c46e6564e53b2d988a774b2030b7f21

shubham3121:
@koenvanderveen @khoaguin We're running testing the notebooks against mongo in the CI. Check pr-tests-notebook-stack which runs `tox -e stack.test.notebook` with `ORCHESTRA_DEPLOYMENT_TYPE: "container_stack"`, which launches docker containers with Mongo. Task runner: `om-ci-16vcpu-ubuntu2204`

packages/syft/src/syft/store/mongo_document_store.py
koenvanderveen: I resolved this in the latest commits, please have a look @khoaguin @shubham3121

packages/syft/src/syft/store/mongo_document_store.py
```diff
@@ -211,27 +231,60 @@ def _set(
...
+        store_key_exists = (
+            collection.find_one(store_query_key.as_dict_mongo) is not None
+        )
+        if (not store_key_exists) and (not self.item_keys_exist(obj, collection)):
+            # attempt to claim ownership for writing
+            ownership_result = self.take_ownership(uid=obj.id, credentials=credentials)
+            can_write: bool = ownership_result.is_ok()
+        elif not ignore_duplicates:
+            unique_query_keys: QueryKeys = self.settings.unique_keys.with_obj(obj)
+            keys = ", ".join(f"`{key.key}`" for key in unique_query_keys.all)
```
shubham3121, line 250: I'm fine with this for now, but this is a redundant check as we already create unique indexes in Mongo against unique keys (`_create_update_index`), which should already take care of this logic.

khoaguin:
After the commit [a914c87](https://github.com/OpenMined/PySyft/pull/8017/commits/a914c87fe7ba2ddf0eb43463fc95baf4b20144df), we had a bug like in the picture below where we got error if we add things with `ignore_duplicates = True` in `_set()` of `MongoStorePartition`, for example when using `data_subject_registry.add_data_subject`

image

I reverted back to the commit [6b5cc8e](https://github.com/OpenMined/PySyft/pull/8017/commits/6b5cc8e70c46e6564e53b2d988a774b2030b7f21) where we did not get this issue like below,
image,
however, we still got an issue: Although we only had 3 data subjects
image
but in the `DataSubject_permissions` collection we have 4 entries like below
image

Furthermore, if I tried to re-upload the same dataset to create duplicates, we will get an error when uploading the dataset
image
and will still only have 1 dataset. However, we added 1 more permission to the `Dataset_permissions` collection
image

khoaguin:
Recheck after this commit [a9d6290](https://github.com/OpenMined/PySyft/pull/8017/commits/a9d6290995a8273100a9267cfcb5725b7ab66749) and previous issues within [this comment](https://github.com/OpenMined/PySyft/pull/8017#issuecomment-1704720674) are fixed

PR #8019, opened 2023-08-17
Adding PyInstaller  for  building standalone binary executables of syft cli 

rasswanth-s:
## Description
Currently, our new syft cli tool, we would be  releasing it to PyPI.

It would be good to package it as a standalone binary , that could be used in major Operating Systems (MacOS, Linux, Windows)

## How has this been tested?
 - Yet to be added, during subsequent PR's

## Checklist


bitsofsteve (commented):
@rasswanth-s  Good work. I guess with binary compilations and all we might want to extend this PR to add  `pr-tests-syftcli.yml` or something, this way we can mimic what we have for hagrid, where we can do some setup and few syftcli commands and make sure things are in check.

rasswanth-s:
Yeah @bitsofsteve , Testing is something , which we need to figure out optimally.

I was mainly , waiting for some initial mature code, to choose our testing strategy

We have plethora of testing options
1. Testing python and  standalone binary options ( 2 options)
2. Testing Operating systems (windows, mac, linux) (3 options)
3. Testing python versions ( 3.9- 3.11)

We would need around 18 different runners, to test all the scenarios

I thought it would be best to do it in the `cool down period`,
Since yash is also working on the bundle part of syft cli currently

We could get a good idea, in choosing a testing strategy that is optimal with number of runners by then.

bitsofsteve:
> Yeah @bitsofsteve , Testing is something , which we need to figure out optimally.
> 
> I was mainly , waiting for some initial mature code, to choose our testing strategy
> 
> We have plethora of testing options
> 
> 1. Testing python and  standalone binary options ( 2 options)
> 2. Testing Operating systems (windows, mac, linux) (3 options)
> 3. Testing python versions ( 3.9- 3.11)
> 
> We would need around 18 different runners, to test all the scenarios
> 
> I thought it would be best to do it in the `cool down period`, Since yash is also working on the bundle part of syft cli currently
> 
> We could get a good idea, in choosing a testing strategy that is optimal with number of runners by then.

Sounds good then.

packages/syft-cli/syft_cli/cli.py
```diff
@@ -2,8 +2,8 @@
 from rich import box
 from rich.console import Console
 from rich.table import Table
+from syft_cli.version import __version__
```
yashgorana: Why do we need to create a module `syft_cli` for pyinstaller?

packages/syft-cli/setup.py
```diff
@@ -22,14 +23,14 @@
...
-    extras_require={"dev": dev_packages},
+    extras_require={"dev": dev_packages, "build": build_packages},
     include_package_data=True,
     entry_points={
         "console_scripts": [
-            "syft = cli:app",
-            "syft-cli = cli:app",
-            "syftcli = cli:app",
-            "syftctl = cli:app",
+            "syft = syft_cli.cli:app",
```
yashgorana: We now have two monikers `packages/syft-cli` `packages/syft-cli/syft_cli`. Can we make this consistent?

packages/syft-cli/setup.py
```diff
@@ -22,14 +23,14 @@
     author="OpenMined <info@openmined.org>",
     packages=find_packages(),
     install_requires=packages,
-    extras_require={"dev": dev_packages},
+    extras_require={"dev": dev_packages, "build": build_packages},
```
yashgorana, line 28: `pip install -e packages/syft[dev]` should include pyinstaller as well right? With this we will have to install both `[dev]` and `[build]`

yashgorana: Ah I see that it needs a fresh environment. So we can keep dev_packages i.e.
```
dev_packages = [
    "pytest",
] + build_packages
```

that way syft-cli[dev] will auto-include syft-cli[build]

packages/syft-cli/setup.py
rasswanth-s: That is a great question, 
This is something, which took me long enough to also decide

This is mainly from the problem that the python modules 

do not allow hypens `-`

Why a separate package : 
One advantage for us to go for a a new module (like syft_cli inside syft-cli) packages .
We could easily separate out, all the commands apart from other config files (setup.py, documentation and several other files) when our project goes large.

Why use hypen in the initial package:

I think hypens, are only mainly improving the readability,most of the cli when publish on pypi 
have <tool_name>-cli.

I have looked several cli tools , most implement like this

repo name: typer-cli
module name: typer_cli
link: https://github.com/tiangolo/typer-cli

But I found azure cli,has a good structure for the same.

https://github.com/Azure/azure-cli

packages/syft-cli/setup.py
rasswanth-s: Done

PR #8032, opened 2023-08-24
setup for syftcli binary release

bitsofsteve:
## Description
This PR enables us release Syftcli binary for different platform  in Github

## Checklist


.github/workflows/cd-syftcli.yml
```diff
@@ -94,3 +100,95 @@ jobs:
         run: |
           tox -e syftcli.publish
           twine upload packages/syftcli/dist/*
+
+  create-syftcli-binaries:
+    needs: [deploy-syft-cli]
+    if: always() && (needs.deploy-syft-cli.result == 'success' && ${{needs.deploy-syft-cli.outputs.current_hash != needs.deploy-syft-cli.outputs.previous_hash }}  || github.event.inputs.skip_tests == 'true')
```
rasswanth-s: NITPICK: we could remove github.events.inputs.skip_tests in this work, as it the skip tests affect only the deployment.

tox.ini
```diff
@@ -80,6 +81,22 @@ commands =
     python -c 'from shutil import rmtree; rmtree("build", True); rmtree("dist", True)'
     python -m build .
 
+[syftcli.build]
+basepython = python3
+changedir = {toxinidir}/packages/syftcli
+description = Build SyftCLI Binary for each platform
+commands =
+    python -m pip install --upgrade pip
+    pip install --upgrade setuptools wheel twine tox build
```
rasswanth-s: NITPICK: We could remove this pip statement, as we would probably not need these pacakges for pyinstaller building?

tox.ini
```diff
@@ -80,6 +81,22 @@ commands =
     python -c 'from shutil import rmtree; rmtree("build", True); rmtree("dist", True)'
     python -m build .
 
+[syftcli.build]
```
rasswanth-s: Smart  ❤️ , Having a tox build would be great

tox.ini
```diff
@@ -80,6 +81,22 @@ commands =
...
+changedir = {toxinidir}/packages/syftcli
+description = Build SyftCLI Binary for each platform
+commands =
+    python -m pip install --upgrade pip
+    pip install --upgrade setuptools wheel twine tox build
+    pip install -e ".[build]"
+    python -c 'from shutil import rmtree; rmtree("build", True); rmtree("dist", True)'
+    ;check if we are on macos add additional flags if not run as is
+    bash -c 'if [[ "$OSTYPE" == "darwin"* ]]; then \
+         pyinstaller --clean --onefile --name syftcli --distpath ./dist/cli src/syftcli/cli.py --target-arch universal2; \
```
rasswanth-s: MAYBE: By default, the distpath , takes ./dist, do we want to place it in ./dist/cli/src/syftcli/cli.py ?

We could keep dist itself?

rasswanth-s: Q: should  we not provide the path to cli.py ?

like

```python3
pyinstaller --clean --onefile --name syftcli  src/syftcli/cli.py --target-arch universal2;
```

?

.github/workflows/cd-syftcli.yml
```diff
@@ -94,3 +100,95 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.SYFT_BUMP_TOKEN  }}
+
+      - name: Check if version is released on Pypi
+        id: check-pypi
+        run: |
+          VERSION = ${{needs.deploy-syft-cli.outputs.deployed_version}}
```
rasswanth-s: Q: For storing environment variable, should there no spaces like (between equal to symbol)
or it would be interpreted as binary?

```python3
VERSION=${{needs.deploy-syft-cli.outputs.deployed_version}}
```

.github/workflows/cd-syftcli.yml
```diff
@@ -94,3 +100,95 @@ jobs:
...
+
+          RESPONSE=$(curl -s --head https://pypi.org/pypi/syftcli/$VERSION/json)
```
rasswanth-s, line 163: Really good sanity check 🙌

.github/workflows/cd-syftcli.yml
```diff
@@ -94,3 +100,95 @@ jobs:
...
+
+          if [[ $RESPONSE == *"200 OK"* ]]; then
```
rasswanth-s: Q: I did a sample, testing with it, by below snippet

```sh
curl -s --head https://pypi.org/pypi/syftcli/0.1.3/json | grep "200 OK"
```

OK keywork is not in the received text, 

but the HTTP/2 200 is present

```sh
curl -s --head https://pypi.org/pypi/syftcli/0.1.3/json | grep "HTTP/2 200"
```

.github/workflows/cd-syftcli.yml
```diff
@@ -30,6 +30,11 @@ jobs:
     if: always() && (needs.call-pr-tests-linting.result == 'success' || github.event.inputs.skip_tests == 'true')
     runs-on: ubuntu-latest
 
+    outputs:
+      deployed_version: ${{ steps.bump-version.outputs.deployed_version }}
```
rasswanth-s, line 34: Q: Should the outputs, step, be defined, after the bump-version, step, if we define it before, maybe it would be null?

.github/workflows/cd-syftcli.yml
```diff
@@ -94,3 +100,95 @@ jobs:
...
+          pip install -U tox
+
+      - name: Build syftcli
+        run: |
+          tox -e syftcli.build
+
+      - name: upload binaries
+        uses: actions/upload-artifact@v2
+        with:
+          name: syftcli-binaries
```
rasswanth-s: Q: Do we need to setup , unique name for each OS binary tagged with version?

So , it would be easy for downloading and attaching it to the github release

I know the PR is still being worked on , just a small reminder.

.github/workflows/cd-syftcli.yml
bitsofsteve: Good question, it won't actually be null, defining it at the start is pretty standard. See here: https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs

.github/workflows/cd-syftcli.yml
bitsofsteve: Oh yeah sure, this was just a stub. We will have upload and download for each I suppose

.github/workflows/cd-syftcli.yml
rasswanth-s: Oh I see, they are evaluated , at the end of each job.

tox.ini
bitsofsteve: The distpath is actually definitive. When that runs, it will still place it in ./dist/cli not ./dist/cli/src/syftcli/cli.py. 

Also for Q2, we specified the path to cli.py , after the distpath argument. so it's more like `pyinstaller --clean --onefile --name  --distpath src`

rasswanth-s (approved):
Marvelous Work @bitsofsteve

PR #8054, opened 2023-08-31
Added permission check to ActionStore get_pointer method.

PeterChung241:
## Description

Added permission check to ActionObject get_pointer method.

## Affected Dependencies
N/A

## How has this been tested?
Locally with JN

## Checklist


packages/syft/src/syft/service/action/action_store.py
```diff
@@ -83,23 +84,38 @@ def get(
...
                 obj = self.data[uid]
+                read_permission = ActionObjectREAD(uid=uid, credentials=credentials)
                 if isinstance(obj, TwinObject):
-                    obj = (
-                        obj.mock if not is_action_data_empty(obj.mock) else obj.private
-                    )
+                    if self.has_permission(read_permission):
+                        obj = (
+                            obj.mock
+                            if not is_action_data_empty(obj.mock)
```
koenvanderveen: I think this should return private right?

PeterChung241:
Sorry for the late reply, at work right now. Think I was trying to change a
few things around to test the return results in the notebooks before I had
to leave so I committed what I was doing at the time. I'll keep looking at
it when I get home in the morning 🙂

PeterChung241:
Also I think the notebooks seem to prioritize returning the mock which is
what was confusing me a bit. Because when I change it to private first the
notebooks start breaking. API/0.8/00

teo-milea:
I tested the permissions on the [seaweed integration](https://github.com/OpenMined/PySyft/pull/7987) branch, and the problem was fixed, code for reference: 
```
import syft as sy
node = sy.orchestra.launch("test-domain", reset=True)

root_client = node.login(email="info@openmined.org",password="changethis")
data = sy.ActionObject.from_obj([1, 2,3])
data_ptr = data.send(root_client)
data_ptr
guest_client = node.client.guest()
guest_client
res = guest_client.api.services.action.get_pointer(data_ptr.id)
res.syft_action_data # guest does not have access to the data
res.get()
res = root_client.api.services.action.get_pointer(data_ptr.id)
res.syft_action_data # root has access to the data
```

packages/syft/src/syft/service/action/action_store.py
PeterChung241: The thing is:
`domain_client.datasets[0].assets[0].data` calls the get() method
`domain_client.datasets[0].assets[0].mock` calls the get_pointer() method
If we the set this to private it will return the private data not the mock to the data owner which breaks the notebook. api/0.8/00-load-data.ipynb

madhavajay:
@shubham3121 thanks for the fix!

PR #8066, opened 2023-09-04
fix styling issue

koenvanderveen:
Screenshot 2023-09-04 at 12 46 22
Screenshot 2023-09-04 at 12 47 01


madhavajay:
Like a boss 🎩

PR #8069, opened 2023-09-05
Tests for MongoDB permissions

khoaguin:
Tests for MongoDB Permissions


khoaguin:
TODO: if we delete an object from a partition, delete its corresponding entry in the permissions collection (in the `_delete` method in `kv_document_store` and `mongo_document_store`)

shubham3121 (approved):
Great work @khoaguin !! Great catch and fix with the delete functionality.

PR #8085, opened 2023-09-14
Helm

koenvanderveen:


packages/grid/backend/grid/start.sh
```diff
@@ -1,6 +1,7 @@
 #! /usr/bin/env bash
 set -e
 
+pip install nltk
```
koenvanderveen, line 4: todo: this is not a great solution

packages/grid/seaweedfs/start.sh
```diff
@@ -1,6 +1,6 @@
 #! /usr/bin/env bash
 
 sleep 30 &&
-echo "s3.configure -access_key ${S3_ROOT_USER} -secret_key ${S3_ROOT_PWD} -user iam -actions Read,Write,List,Tagging,Admin -apply" \
```
koenvanderveen, line 4: todo: cleanup

packages/hagrid/hagrid/orchestra.py
```diff
@@ -259,6 +262,7 @@ def deploy_to_python(
                 host=host,
```
koenvanderveen, line 264: todo: also add queue options here

koenvanderveen: Not sure how we exactly want to organize default setup for docker/k8s? Do we want dynamic consumer creation on as default? Or do we want to have a default set of consumers? Currently async execution is dependent on consumers, so if we have none we cannot execute async work.

packages/syft/src/syft/client/client.py
```diff
@@ -333,6 +333,13 @@ def get_node_metadata(self, credentials: SyftSigningKey) -> NodeMetadataJSON:
         else:
             return self.node.metadata.to(NodeMetadataJSON)
 
+    def to_blob_route(self, path: str, host=None) -> GridURL:
+        # TODO: FIX!
+        if host is not None:
+            return GridURL(host_or_ip=host, port=8333).with_path(path)
+        else:
+            return GridURL(port=8333).with_path(path)
```
koenvanderveen, line 353: @shubham3121, I know we talked about this. I feel like I understand it well enough to patch the problem for situation but I dont know if I might be causing problems elsewhere, could we again have a look at this?

packages/syft/src/syft/gevent_patch.py
```diff
@@ -37,6 +34,6 @@ def is_notebook() -> bool:
 
 jupyter_notebook = is_notebook()
```
koenvanderveen: so I had to remove this to get rid of threading problems. There is a much longer story here, but I suspect we have to dedicate maybe 1 or 2 weeks to come up with a proper multithreading/processing plan

packages/syft/src/syft/node/node.py
```diff
@@ -252,7 +263,8 @@ def __init__(
             if id is None:
                 id = UID()
             self.id = id
-        self.packages = get_venv_packages()
+        self.packages = ""
```
koenvanderveen, line 276: this was breaking so I disabled i. it looks it wasnt doing anything

packages/syft/src/syft/node/node.py
```diff
@@ -748,17 +852,25 @@ def get_role_for_credentials(self, credentials: SyftVerifyKey) -> ServiceRole:
...
         signed_result = SyftAPIData(data=result).sign(self.signing_key)
 
         return signed_result
 
     def handle_api_call_with_unsigned_result(
-        self, api_call: Union[SyftAPICall, SignedSyftAPICall]
+        self,
+        api_call: Union[SyftAPICall, SignedSyftAPICall],
+        job_id: Optional[UID] = None,
+        check_call_location=True,
```
koenvanderveen, line 1002: I added `check_call_location` here because when a consumer reads a message the message node_uid does not match the nodes id as it was initially sent to a different node. I was not completely sure how to do this properly as the message is signed

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -675,11 +689,17 @@ def process_code(
...
 
-    keywords = [ast.keyword(arg=i, value=[ast.Name(id=i)]) for i in input_kwargs]
+    call_args = function_input_kwargs
+    # call_args = policy_input_kwargs
+    if "domain" in function_input_kwargs:
+        # print("adding domain")
+        # call_args = ["domain"] + call_args
+        context.output["uses_domain"] = True
+    call_stmt_keywords = [ast.keyword(arg=i, value=[ast.Name(id=i)]) for i in call_args]
     call_stmt = ast.Assign(
```
koenvanderveen, line 790: this slightly changed because we (optionally) added `domain` as an argument to usercode. In general I think this code could be more clear if we refactor it a bit

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -847,24 +869,177 @@ class UserCodeExecutionResult(SyftObject):
...
+        # stdlib
+        import builtins as __builtin__
+
+        original_print = __builtin__.print
+
+        class LocalDomainClient:
+            def __init__(self):
+                pass
+
+            def launch_job(self, func: UserCode, **kwargs):
```
koenvanderveen, line 1091: most of the juggling with new usercode and approval can be removed once @teo-milea finishes the UX for the new policy

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -847,24 +869,177 @@ class UserCodeExecutionResult(SyftObject):
...
+        # sys.stderr = stderr
 
         # statisfy lint checker
         result = None
 
         exec(code_item.byte_code)  # nosec
+        _locals = locals()
+
+        user_code_service = context.node.get_service("usercodeservice")
+        for user_code in user_code_service.stash.get_all(context.credentials).ok():
```
koenvanderveen: this is necessary because in the outer syft function function we pass in the the batch function name as the first argument to launch_job, e.g.
`domain.launch_job(run_batch, .....)` so this function needs to be present in the scope

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -847,24 +869,177 @@ class UserCodeExecutionResult(SyftObject):
...
+            def __init__(self):
+                pass
+
+            def launch_job(self, func: UserCode, **kwargs):
+                # original_print("launching job for ", func.service_func_name)
+                # relative
+                from ... import UID
+
+                # get reference to node (TODO)
+                node = context.node
```
koenvanderveen: this is obviously the dangerous thing here. I am not sure how easy it would be to get a reference to this node object from within the syft function, but once you do you can basically do anything server side

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -878,7 +1053,13 @@ def execute_byte_code(code_item: UserCode, kwargs: Dict[str, Any]) -> Any:
         )
 
     except Exception as e:
+        # stdlib
+        import traceback
+
```
koenvanderveen, line 1183: maybe check dev_mode here

packages/syft/src/syft/service/code/user_code_service.py
```diff
@@ -269,12 +289,22 @@ def call(
...
                 return result.mock
             else:
                 return result.as_empty()
         except Exception as e:
-            return SyftError(message=f"Failed to run. {e}")
+            # stdlib
+            import traceback
+
+            # import ipdb
+            # ipdb.set_trace()
```
koenvanderveen: check dev_mode

packages/syft/src/syft/service/queue/queue.py
```diff
@@ -78,31 +90,67 @@ def handle_message(message: bytes):
...
                 status = Status.ERRORED
+                job_status = JobStatus.ERRORED
         except Exception as e:  # nosec
             status = Status.ERRORED
+            job_status = JobStatus.ERRORED
             result = SyftError(message=f"Failed with exception: {e}")
-
-        item = QueueItem(
+            print("HAD AN ERROR WHILE HANDLING MESSAGE")
+        # print("result", type(result.message.data))
```
koenvanderveen: check dev mode

packages/grid/docker-compose.multinode.yml
```diff
@@ -0,0 +1,460 @@
+version: "3.8"
+services:
+  # docker-host:
```
koenvanderveen: we might get rid of this entire file eventually, I used it as a way to start testing the multinode setup

packages/grid/seaweedfs/new_seaweed.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+echo "got api key"
+echo ${STACK_API_KEY}
+export STACK_API_KEY=${STACK_API_KEY}
```
rasswanth-s: Q: Is the syft stack api key used in seaweedfs ?

packages/grid/seaweedfs/seaweedfs.dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY ./start.sh /start.sh
+COPY ./mount_command.sh /mount_command.sh
+COPY ./app.py /app.py
+
+RUN chmod +x /start.sh
+RUN chmod +x /mount_command.sh
+
+CMD ["bash", "/seaweed.sh"]
```
rasswanth-s: Q: I think seaweed.sh is not copied, for running the custom script?

packages/grid/backend/grid/start.sh
yashgorana: @khoaguin perhaps we can move this into data_science optional dependencies

packages/grid/docker-compose.multinode.yml
```diff
@@ -0,0 +1,461 @@
...
+      - "${HTTP_PORT}:81"
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    labels:
+      - "orgs.openmined.syft=this is a syft proxy container"
+
+    # depends_on:
+    #   - "docker-host"
+
+  tailscale:
```
yashgorana: @shubham3121 @koenvanderveen  we have removed tailscale on "dev". perhaps we have to sync this file against dev as well. Please let me know if we still need tailscale for multinode setup.

packages/grid/docker-compose.multinode.yml
```diff
@@ -0,0 +1,461 @@
...
+      - credentials-data-worker2:/storage
+      - ${RELATIVE_PATH}./backend/grid:/app/grid
+      - ${RELATIVE_PATH}../syft:/app/syft
+      - ${RELATIVE_PATH}./data/package-cache:/root/.cache
+    stdin_open: true
+    tty: true
+    labels:
+      - "orgs.openmined.syft=this is a syft backend container"
+
+  backendworker3:
```
yashgorana: We might want to use replicas over here instead of different instances of backendworkers

packages/grid/default.env
```diff
@@ -50,6 +50,9 @@ DOMAIN_CHECK_INTERVAL=60
 ASSOCIATION_TIMEOUT=10
 USERS_OPEN_REGISTRATION=False
 DEV_MODE=False
+QUEUE_PORT=5556
+CREATE_PRODUCER=False
+N_CONSUMERS=0
```
yashgorana, line 57: Env vars present in dockerfile but not here:
* ~`HOST_GRID_PATH`~ (set as pwd from dockerfile, and run only in dev mode, so perhapse set it to empty instead?)
* `BACKEND_STORAGE_PATH`
* `SEAWEED_MOUNT_PORT`
* `STACK_API_KEY`

packages/grid/seaweedfs/app.py
```diff
@@ -0,0 +1,34 @@
...
+# third party
+from flask import Flask
+from flask import request
+
+# Flask application instance
+app = Flask(__name__)
+
+
+@app.route("/configure_azure", methods=["POST"])
+def test() -> str:
```
yashgorana: rename this function to configure_azure?

packages/grid/seaweedfs/app.py
```diff
@@ -0,0 +1,34 @@
...
+    first_res = json.loads(request.data.decode("utf-8").replace("'", '"'))
```
yashgorana, line 16: we need to manager errors here, because we may pass incorrect JSON payload, and the error needs to be returned accordingly.

packages/grid/seaweedfs/app.py
```diff
@@ -0,0 +1,34 @@
...
+    account_name = first_res["account_name"]
+    account_key = first_res["account_key"]
+    container_name = first_res["container_name"]
+    remote_name = first_res["remote_name"]
+    bucket_name = first_res["bucket_name"]
+
+    res = subprocess.run(
```
yashgorana, line 23: There might be a subprocess error here
* There was en error starting the command
* or, There was an error in mount_command.sh
* or, there was a system error resulting in failed mount.
These must  be returned.

packages/grid/seaweedfs/mount_command.sh
```diff
@@ -0,0 +1,15 @@
...
+
+echo 's3.bucket.create -name $3 ' | weed shell
+
+echo "remote.mount -dir=/buckets/$3 -remote=$1/$4" | weed shell
+
+weed filer.remote.sync
+
+
+
+echo "remote.configure -name=$1 -type=azure -azure.account_name=$2 -azure.account_key=$5" | weed shell
```
yashgorana: duplicate of line#3?

packages/grid/seaweedfs/new_seaweed.sh
yashgorana: This looks the same as start.sh

packages/grid/seaweedfs/seaweedfs.dockerfile
yashgorana: It should either be start.sh or new_seaweed.sh

packages/grid/seaweedfs/app.py
yashgorana: Good To have: Rate limit this api at an HTTP level or at the subprocess invocation to avoid spawning too many `bash mount_command.sh`

packages/grid/seaweedfs/mount_command.sh
```diff
@@ -0,0 +1,15 @@
+echo "$1 $2 $3 $4 $5"
```
yashgorana: Perhaps we can make this a bit more descriptive like

```
echo remote_name=$1 account_name=$2...
```

packages/grid/seaweedfs/seaweedfs.dockerfile
yashgorana: We must expose all environment variables with default values.

```
ENV SEAWEED_MOUNT_PORT=4001 \
    STACK_API_KEY=""
```

packages/grid/traefik/docker/dynamic.yml
```diff
@@ -40,6 +48,22 @@ http:
...
         - "blob-storage-host"
+    blob-storage-mount:
+      rule: "PathPrefix(`/mount`)"
+      entryPoints:
+        - web
+        - vpn
+      service: "seaweedfsmount"
+      middlewares:
+        - "blob-storage-mount-url"
+    vpn:
```
yashgorana, line 59: vpn has been removed. cc: @shubham3121

packages/hagrid/hagrid/orchestra.py
yashgorana: Yes based on what we had demo'd in our meeting - we will have a fixed set of default workers and spawn custom env workers once they are approved by the DO.

packages/hagrid/hagrid/orchestra.py
```diff
@@ -314,12 +322,17 @@ def deploy_to_python(
                 )
         else:
             raise NotImplementedError(f"node_type: {node_type_enum} is not supported")
+
+        def stop() -> None:
+            worker.stop()
```
yashgorana, line 325: check if worker is defined, because it is instantiated only when `node_type_enum in worker_classes`

packages/syft/src/syft/client/client.py
```diff
@@ -153,7 +153,7 @@ def get_cache_key(self) -> str:
     def api_url(self) -> GridURL:
         return self.url.with_path(self.routes.ROUTE_API_CALL.value)
 
-    def to_blob_route(self, path: str) -> GridURL:
+    def to_blob_route(self, path: str, **kwargs) -> GridURL:
```
yashgorana, line 156: do we need this?

packages/syft/src/syft/client/client.py
yashgorana: This should be `port=Settings.S3_PORT` instead of `port=8333`

packages/syft/src/syft/node/node.py
```diff
@@ -334,16 +352,48 @@ def __init__(
...
 
         NodeRegistry.set_node_for(self.id, self)
 
+    @property
+    def host_syft_location(self):
+        """e.g. /Users/<user>/workspace/pysyft"""
+        if not self.dev_mode:
+            raise ValueError("You can only get host location in dev mode")
+        # We set this variable in docker compose (from pwd)
+        env_val = os.environ.get("HOST_GRID_PATH", None)
```
yashgorana: PWD can be anything inside syft. For example, while developing grid or building dockerfiles we need to be in` packages/` or `packages/grid`... so perhaps we can just look for tox.ini and keep traversing to the parent, that way we eliminate the need for any env var altogether.

packages/syft/src/syft/node/node.py
```diff
@@ -794,13 +895,37 @@ def __eq__(self, other: Any) -> bool:
...
+        from ..service.queue.queue import Status
+
+        while True:
+            result = self.queue_stash.pop_on_complete(credentials, uid)
+            if not result.is_ok():
+                return result.err()
+            else:
+                res = result.ok()
+                if res.status == Status.COMPLETED:
+                    return res
```
yashgorana, line 903: need to check for errors if `res.status != Completed` then what?

yashgorana (commented):
just adding a couple of comments for the team to get this PR ready for merge.

packages/grid/seaweedfs/mount_command.sh
teo-milea: this was only for debugging, we could remove it all together

packages/grid/docker-compose.multinode.yml
koenvanderveen: this file can be removed probably

packages/grid/docker-compose.multinode.yml
koenvanderveen: see other comment

packages/syft/src/syft/service/queue/queue.py
```diff
@@ -38,8 +86,15 @@ def create_consumer(
         )
         return consumer
 
-    def create_producer(self, queue_name: str):
-        return self._client.add_producer(queue_name=queue_name)
+    def create_message_queue(self, queue_name: str):
+        return self._client.add_message_queue(queue_name)
+
```
yashgorana: Don't see `add_message_queue` implemented and `create_message_queue` used anywhere. Should we remove this?

packages/syft/src/syft/service/queue/queue.py
```diff
@@ -78,31 +224,41 @@ def handle_message(message: bytes):
...
+        job_item = worker.job_stash.get_by_uid(credentials, queue_item.job_id).ok()
+
+        queue_item.status = Status.PROCESSING
+        queue_item.node_uid = worker.id
+
+        job_item.status = JobStatus.PROCESSING
+        job_item.node_uid = worker.id
+
+        worker.queue_stash.set_result(credentials, queue_item)
+        worker.job_stash.set_result(credentials, job_item)
```
yashgorana: Instead of duplicating this we can instead just forward the message bytes to the process function `handle_message_multiprocessing` where we can run this routine followed by handling and setting results.

```
@staticmethod
def handle_message(message: bytes):
    p = multiprocessing.Process(
            target=handle_message_multiprocessing,
            args=(message),
        )
    p.start()
    p.join()
```

yashgorana: Aditionally, we can have fixed size process pool of size "n", so that we can process "n" messages without blocking the worker

packages/grid/seaweedfs/mount_command.sh
shubham3121: I think we can remove  this, might be for just debugging.

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -862,24 +977,195 @@ class UserCodeExecutionResult(SyftObject):
...
+                log_service.append(context=context, uid=log_id, new_str=new_str)
+                time = datetime.datetime.now().strftime("%d/%m/%y %H:%M:%S")
+                return __builtin__.print(
+                    f"{time} FUNCTION LOG ({job_id}):",
+                    *new_args,
+                    end=end,
+                    sep=sep,
+                    file=sys.stderr,
+                )
+
```
yashgorana, line 1125: A few key points:
- This is only capturing logs generated from print function & not logging std lib.
- For each print() we are triggering a Mongo API call - any failure exception in this API or downstream monog server outage results in direct failure of the function and the job even though the function itself has no issues.
- There's no rate limit on the add/apend API call itself hence a `while(True): print(...)` can probably overwhelm the single-node mongo server.

I would recommend that we setup logging & patch print to write to a file in /var/log instead. Once the function completes (or errors out) we simply upload the logs in a single Mongo API call.

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -862,24 +977,195 @@ class UserCodeExecutionResult(SyftObject):
...
+                for x in user_service.stash.partition.data.values()
+                if x.verify_key == context.credentials
+            ][0]
+            data_protcol = get_data_protocol()
+            user_api = node.get_api(context.credentials, data_protcol.latest_version)
+            user_api.signing_key = user_signing_key
+            # We hardcode a python connection here since we have access to the node
+            # TODO: this is not secure
+            user_api.connection = PythonConnection(node=node)
+
```
yashgorana, line 1014: Must discuss this

packages/syft/src/syft/service/code/user_code.py
koenvanderveen: yes those are definitely next steps. Not sure about writing the logs only after finishing. The whole point of logs is that you can see them as the program runs. This is especially useful if its a long running job

packages/syft/src/syft/store/linked_obj.py
```diff
@@ -65,6 +65,7 @@ def update_with_context(
         )
         if result.is_ok():
             return result
+        return result
```
kiendang: This could just be `return result`.

kiendang (commented):
nitpick: for the notebooks under `notebooks/helm` could we number them so that they're easier to navigate? Also could we use kebab case for all of them like the other notebooks? Right now there are `dataset_ux.ipynb`, `consumer node.ipynb`, `helm-syft-mongo.ipynb`. There's one `Untitled.ipynb` as well.

packages/syft/src/syft/node/node.py
shubham3121: Maybe we can check for status is Error/Failed and return

packages/syft/src/syft/service/action/action_service.py
```diff
@@ -104,6 +106,50 @@ def set(
...
+                    obj = obj.mock.as_empty()
+                    obj.syft_point_to(context.node.id)
+                else:
+                    obj.mock.syft_point_to(context.node.id)
+                    obj.private.syft_point_to(context.node.id)
+            else:
+                obj = obj.as_empty()
+            return Ok(obj)
+        return result
+
```
kiendang: Are these only for debugging purposes and should be removed?

packages/grid/seaweedfs/mount_command.sh
```diff
@@ -0,0 +1,15 @@
...
+
+echo "remote.configure -name=$1 -type=azure -azure.account_name=$2 \
+        -azure.account_key=$5" \
+        | weed shell
+
+echo 's3.bucket.create -name $3 ' | weed shell
```
kiendang: This should use double quotes? Otherwise `$3` will not be substituted. Try
```sh
HELLO="hello"
echo '$HELLO'
echo "$HELLO"
```

packages/hagrid/hagrid/orchestra.py
koenvanderveen: I think its always there (if you have no port)

packages/syft/src/syft/client/client.py
koenvanderveen: I think so, but we would have to look more deeply into doing this properly. We are using it in BlobRetrievalByURL._read_data and  SeaweedFSBlobDeposit.write. Depending on the deployment settings we need both but there is probably a better way of solving this

packages/syft/src/syft/client/client.py
koenvanderveen: grid is not a package though

PR #8090, opened 2023-09-18
Fix the bug "Guest user is able to change email to owner and essentially kills nodes"

khoaguin:
Fix the bug in https://github.com/OpenMined/Heartbeat/issues/747


IonesioJunior (approved):
LGTM! Awesome work @khoaguin !

packages/syft/src/syft/service/user/user_stash.py
```diff
@@ -76,6 +76,13 @@ def get_by_email(
         qks = QueryKeys(qks=[EmailPartitionKey.with_obj(email)])
         return self.query_one(credentials=credentials, qks=qks)
 
+    def email_exists(self, email="") -> bool:
```
shubham3121: Can we skip setting email to empty string. and add missing annotation for email field.

shubham3121 (approved):
Left a minor comment, otherwise looks good. Nice work !! 🎉

PR #8137, opened 2023-10-09
Demo improvements

teo-milea:
## Description
There are multiple changes meant to improve the flow of a future helm demo:
* parallelization of `file.iter_lines()` and processing
* ~new policy `ExecuteOncePerCombination`~ -> moved to #8160
* UX improvements for jobs 
* fixed problem with logs for subjobs

## How has this been tested?
* notebooks under the helm directory

## Checklist


packages/syft/src/syft/service/action/action_service.py
```diff
@@ -183,9 +183,11 @@ def _user_code_execute(
         code_item: UserCode,
         kwargs: Dict[str, Any],
     ) -> Result[ActionObjectPointer, Err]:
-        filtered_kwargs = code_item.input_policy.filter_kwargs(
+        input_policy = code_item.input_policy
+        filtered_kwargs = input_policy.filter_kwargs(
             kwargs=kwargs, context=context, code_item_id=code_item.id
         )
+        code_item.input_policy = input_policy
```
koenvanderveen, line 191: why is this needed?

teo-milea: if you check the UserCode class, you can see that the input policy is actually a property that also has a setter method, so in order to save the state we need to do this. This is exactly the same mechanism as the one for the output policy

koenvanderveen:
great job!

PR #8158, opened 2023-10-13
devspace quality-of-life commands

yashgorana:
## Description

Closes https://github.com/OpenMined/Heartbeat/issues/780

## Affected Dependencies

## How has this been tested?

## Checklist


madhavajay:
@yashgorana the test `pr-tests-stack` looks like its saying:
```
2023-10-13T05:05:59.5226266Z ##[error]No space left on device : '/home/runner/runners/2.309.0/_diag/blocks/f92775bc-6976-478a-9cee-6977f7b58277_8d2d333e-4dbd-5553-5489-587358cf17ab.1'
```

We previously did a work around to free up more space on the github runners. I think that it might be time to hunt down this problem.

I guess perhaps the images are being built on Docker and then pushed to k3d registry which means they are duplicate and possibly taking up too much space? But still thats a lot space, is there possibly too many copies of our DL libraries in layers being kept or some other large images during our tests which are causing this issue?

yashgorana:
> @yashgorana the test `pr-tests-stack` looks like its saying:
> 
> ```
> 2023-10-13T05:05:59.5226266Z ##[error]No space left on device : '/home/runner/runners/2.309.0/_diag/blocks/f92775bc-6976-478a-9cee-6977f7b58277_8d2d333e-4dbd-5553-5489-587358cf17ab.1'
> ```
> 
> We previously did a work around to free up more space on the github runners. I think that it might be time to hunt down this problem.
> 
> I guess perhaps the images are being built on Docker and then pushed to k3d registry which means they are duplicate and possibly taking up too much space? But still thats a lot space, is there possibly too many copies of our DL libraries in layers being kept or some other large images during our tests which are causing this issue?

yes this is something @rasswanth-s and I will look at

tox.ini
```diff
@@ -925,3 +925,54 @@ changedir = {toxinidir}/packages/syftcli
...
+passenv=HOME,USER
+allowlist_externals =
+    bash
+commands =
+    bash -c 'k3d registry create registry.localhost --port 12345 -v $HOME/.k3d-registry:/var/lib/registry'
+    bash -c '\
+        CYAN="\033[1;36m"; \
+        CLEAR="\033[0m"; \
+        printf "\n${CYAN}---------------------------------------------------------\n" ;\
+        printf "Make sure to add the following to your /etc/hosts file:${CLEAR}\n";\
```
rasswanth-s, line 940: I think for windows users, it might be different from /etc/hosts

rasswanth-s: We could add helper single line command like

for Linux/Mac
```sh
echo "127.0.01 k3d-registry.localhost" | sudo tee -a /etc/hosts
```

For windows:
```sh
Add-Content -Path "$env:systemroot\System32\drivers\etc\hosts" -Value "127.0.01 k3d-registry.localhost"
```

rasswanth-s (approved):
LGTM, left some comments

tox.ini
yashgorana: Yes I was actually thinking of setting it up automatically, but since this command would be run frequently, you would end up seeing multiple entries like this:

image

I'll create a script for this later but it would still need sudo (or run as admin on win) to actually patch the file. Moreover, WSL does not persist /etc/hosts/ patches because they are overwritten by host windows' hosts - so it will have a separate logic.

PR #8170, opened 2023-10-17
Fix folder removed in ghpages in cd-syft.yml

shubham3121:


.github/workflows/cd-syft.yml
```diff
@@ -335,7 +335,7 @@ jobs:
 
       - name: Copy helm repo files from Syft Repo
         run: |
-          rm -rf ghpages/helm/repo/
+          rm -rf ghpages/helm/
```
rasswanth-s: maybe we could  add
`mkdir -p ghpages/helm`

shubham3121: Updated the remove command to this:
 rm -rf ghpages/helm/*

This doesn't delete the directory, only the contents of the directory

rasswanth-s (approved):

PR #8189, opened 2023-10-23
switch all runners to github

bitsofsteve:
## Description
Switching all runners to Github's stock runners.


madhavajay:
If we do add back self hosted runners we should add the following:
1) a script which checks their uptime and aborts with an error if the runner has been up for longer than about a day or so
2) run the runners in duplicate with github actions for a while until we have confidence they aren't flakey

shubham3121:
Closing this. Covered by #8191

PR #8224, opened 2023-11-07
Fix list repr when something goes wrong one of the jobs

IonesioJunior:
## Description
This PR addresses and resolves two significant issues: the fragility of the job list when encountering unexpected job behavior, and the inadequate tracking of SyftErrors on the Consumer side. By implementing these fixes, we aim to enhance the robustness of job handling and improve error monitoring and logging.

## After Changes:
subjob_list


packages/syft/src/syft/service/code/user_code.py
```diff
@@ -1065,7 +1066,12 @@ def to_str(arg: Any) -> str:
         exec(code_item.parsed_code, _globals, locals())  # nosec
 
         evil_string = f"{code_item.unique_func_name}(**kwargs)"
-        result = eval(evil_string, _globals, _locals)  # nosec
+        try:
```
koenvanderveen, line 1086: I think we should add a couple of things here
- make sure that the error is printed to stdout such that we can see it in the logs of the server side
- add the stacktrace to the error using traceback.format_exc() as in line 1096
- make sure that we add the error to the job, but in such a way that admins can see the error and users cannot. Later we can add some kind of permission on top of that such that its possible for users to request the error message.

koenvanderveen:
I think another place where I have seen this break is when we return an ActionDataEmpty as result of the job. In that case it sometimes returns a None as the output of __str__() during repr. I am not sure if this is solved by that. Can we check that?

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -1065,7 +1066,24 @@ def to_str(arg: Any) -> str:
...
+            result = eval(evil_string, _globals, _locals)  # nosec
+        except Exception:
+            if context.job is not None:
+                # stdlib
+                import traceback
+
+                time = datetime.datetime.now().strftime("%d/%m/%y %H:%M:%S")
+                error_msg = traceback.format_exc()
```
IonesioJunior: Solves: `add the stacktrace to the error using traceback.format_exc()`

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -1065,7 +1066,24 @@ def to_str(arg: Any) -> str:
...
+                __builtin__.print(
+                    f"{time} EXCEPTION LOG ({job_id}):", error_msg, file=sys.stderr
+                )
```
IonesioJunior: Solves: `Error is printed to stdout such that we can see it in the logs of the server side`

packages/syft/src/syft/service/log/log_service.py
```diff
@@ -57,7 +70,18 @@ def get(
...
+
+    @service_method(path="log.get_error", name="get_error", roles=ADMIN_ROLE_LEVEL)
+    def get_error(
+        self, context: AuthedServiceContext, uid: UID
+    ) -> Union[SyftSuccess, SyftError]:
+        result = self.stash.get_by_uid(context.credentials, uid)
+        if result.is_err():
+            return SyftError(message=str(result.err()))
+
+        return Ok(result.ok().stderr)
```
IonesioJunior, line 84: Solves: `make sure that we add the error to the job, but in such a way that admins can see the error and users cannot.`

IonesioJunior: The `Job.logs()` method has been updated to include an optional `stderr` boolean parameter, which is set to false by default. When this parameter is set to true, the method will invoke a specialized logging service designed to fetch stderr messages. However, since this service requires  `ADMIN_ROLE_LEVEL` privileges, common users will not be able to retrieve it.
exception_logs

IonesioJunior:
> I think another place where I have seen this break is when we return an ActionDataEmpty as result of the job. In that case it sometimes returns a None as the output of **str**() during repr. I am not sure if this is solved by that. Can we check that?

I  couldn't reproduce this scenario. Maybe we can sync and check if it's still a problem.

koenvanderveen:
awesome PR ❤️

packages/syft/src/syft/client/client.py
```diff
@@ -477,6 +477,16 @@ def __init__(
 
         self.post_init()
 
+    def rrr(self):
```
madhavajay, line 480: @IonesioJunior Was this committed by accident?

koenvanderveen: ah that was me, will remove

koenvanderveen: added it while debugging the reconstruction of the stacktrace for syft functions, with a pointer to the line that throws the error

PR #8254, opened 2023-11-14
Build custom worker with config

shubham3121:
- create a class to parse a config yaml

## Description

## Affected Dependencies

```python

# syft absolute
from syft.custom_worker.builder import CustomWorkerBuilder
from syft.custom_worker.config import CustomWorkerConfig

if __name__ == "__main__":
    config = """
    build:
        gpu: false
        python_version: 3.11
        system_packages:
            - ffmpeg
        python_packages:
            - ffmpeg-python==0.2.0
            - llama-index==0.9.6.post2
    """
    config = CustomWorkerConfig.from_str(config)
    builder = CustomWorkerBuilder()

    print("building image...")
    builder.build_image(config)
    print("done")
```

## How has this been tested?

## Checklist


packages/grid/backend/grid/start.sh
yashgorana, line 1: undoing this change for simplicity at the moment - syft worker will just pull from openmined/grid-backend and install packages.

packages/grid/backend/backend.dockerfile
yashgorana, line 1: for backend to build custom worker, its dockerfile must be present in the backend container.

packages/syft/src/syft/custom_worker/builder.py
```diff
@@ -0,0 +1,98 @@
...
+                buildargs=build_args,
+            )
+
+            return
+        except docker.errors.BuildError as e:
+            raise e
+        except docker.errors.APIError as e:
+            raise e
+        except Exception as e:
+            raise e
```
kiendang, line 75: Doesn't 

```py
except Exception as e:
    raise e
```

already catches all of these?

packages/syft/src/syft/custom_worker/builder.py
```diff
@@ -0,0 +1,98 @@
+# stdlib
+import os.path
+from pathlib import Path
+from typing import Tuple
+
+# third party
+import docker
```
kiendang, line 7: Shouldn't `docker` be added to `syft` dependencies?

yashgorana: Ouh yes - this is in helm PR, but i missed out to add it here

packages/syft/src/syft/custom_worker/builder.py
yashgorana: i'm keeping them here so that we know what kind of errors we want to handle when integrating with worker service. This will eventually get streamlined...

test.py
```diff
@@ -0,0 +1,21 @@
+# syft absolute
```
shubham3121: I think this file got committed by mistake ?

tox.ini
```diff
@@ -1055,21 +1055,22 @@ setenv =
...
     bash -c '\
         docker build \
-            -f grid/backend/syft_base_cpu.dockerfile . \
+            -f grid/backend/worker_cpu.dockerfile . \
             -t cpu-worker:opendp \
             --build-arg PIP_PACKAGES="$PIP_PACKAGES" \
             --build-arg SYSTEM_PACKAGES="$SYSTEM_PACKAGES"'
     bash -c 'for pkg in $PIP_PACKAGES; do docker run --rm cpu-worker:opendp pip list | grep "$pkg"; done'
     bash -c 'for pkg in $SYSTEM_PACKAGES; do docker run --rm cpu-worker:opendp apk -e info "$pkg"; done'
+    bash -c 'docker rmi cpu-worker:opendp'
```
kiendang, line 1069: 👍

.github/workflows/cd-syft.yml
```diff
@@ -129,17 +129,17 @@ jobs:
...
+      #   id: syft-base-cpu-build
+      #   uses: docker/build-push-action@v5
+      #   with:
+      #     context: ./packages
+      #     file: ./packages/grid/backend/worker_cpu.dockerfile
+      #     platforms: ${{ steps.release_metadata.outputs.release_platform }}
+      #     target: backend
+      #     outputs: type=image,name=openmined/syft-base-cpu,push-by-digest=true,name-canonical=true,push=true
+      #     cache-from: type=registry,ref=openmined/syft-base-cpu:cache-${{ steps.release_metadata.outputs.short_release_platform }}
+      #     cache-to: type=registry,ref=openmined/syft-base-cpu:cache-${{ steps.release_metadata.outputs.short_release_platform }},mode=max
```
kiendang, line 142: If the `worker_cpu` image is always meant to be built inside a backend container in prod and not prebuilt and pushed to DockerHub then maybe we can just delete this instead of commenting it out?

shubham3121: I think we will shift to having a base image with just the syft code in it. And the workers can later build from that. But if I think in that case as well maybe backend builds from the base image and worker builds from backend. Anyways I think we are just keeping this commented for bookkeeping, but we can choose to delete this, @yashgorana thoughts ?

packages/syft/src/syft/custom_worker/config.py
```diff
@@ -0,0 +1,88 @@
+# stdlib
+from hashlib import sha256
+from pathlib import Path
+from typing import List
+
+# third party
+from packaging import version
+from pydantic import validator
+from typing_extensions import Self
+import yaml
```
kiendang, line 11: We should add PyYAML to syft dependencies as well? PyYAML is currently listed as dependency in hagrid and syft-cli, not syft.

packages/syft/src/syft/custom_worker/config.py
```diff
@@ -0,0 +1,88 @@
...
+    version: str = "1"
+
+    @classmethod
+    def from_dict(cls, config: dict) -> Self:
+        return cls(**config)
+
+    @classmethod
+    def from_str(cls, content: str) -> Self:
+        config = yaml.safe_load(content)
+        return cls.from_dict(config)
```
kiendang, line 86: I think we need a test for this class. I doubt it will work here since `CustomWorkerConfig` take a `CustomBuildConfig` argument, but `config = yaml.safe_load(content)` would be just a normal dictionary here so `cls(**config)` might not work. We probably need to use `cls.parse_obj(config)` instead.

kiendang: hmm `parse_obj` is not recursive either. This is something that is more convenient in Pydantic V2, we would have to do things a bit more manually here.

kiendang: ah my bad `parse_obj` actually works. there was a bug in my test.

kiendang: hmm from my test `cls(**config)` works too. that's nice though unexpected. will push the test.

yashgorana: Yeah `ExampleObj(**some_dict)` is a [valid API shared in pydantic examples](https://docs.pydantic.dev/latest/#pydantic-examples)

.github/workflows/cd-syft.yml
yashgorana: Yeah for now we can keep this. we can housekeep at the towards the end of the helm feature.

packages/syft/src/syft/custom_worker/config.py
shubham3121: yes agreed, lets add that to the syft setup.cfg

packages/syft/src/syft/custom_worker/config.py
```diff
@@ -85,7 +86,7 @@ def from_str(cls, content: str) -> Self:
         return cls.from_dict(config)
 
     @classmethod
-    def from_path(cls, path: Path | str) -> Self:
+    def from_path(cls, path: Union[Path, str]) -> Self:
```
shubham3121, line 89: 👍🏾

packages/syft/src/syft/custom_worker/config.py
kiendang: It's a bit different here. I'd expect `CustomWorkerConfig(**config)` to work where `config = {"version": version, "build": build_config}` if `build_config` is of type `CustomBuildConfig`. However here `config = yaml.safe_load(content)` so `build_config` would be a `dict`, not parsed into a `CustomBuildConfig` yet. It's a bit surprising to me that `CustomWorkerConfig(**config)` still works here.

shubham3121: CustomWorkerConfig(**config) and CustomWorkerConfig.from_dict(config) are both valid syntax.

shubham3121 (commented):
Wonderful work @yashgorana and @kiendang. !!!

packages/syft/src/syft/custom_worker/config.py
kiendang: I'm talking about the `from_str`

```py
def from_str(cls, content: str) -> Self:
    config = yaml.safe_load(content)
    return cls.from_dict(config)
```

which is the same as 


```py
def from_str(cls, content: str) -> Self:
    config = yaml.safe_load(content)
    return cls(**config)
```

If the yaml file looks like

```yaml
version: "1"
build:
  python_packages:
    - nltk==3.8.1
  ...
```

`config` would be `{"version":  "1", "build": {"python_packages": ["nltk"==3.8.1], ...}}` which surprises me that `cls(**config)` still works since I expected `cls(**config)` to work with `config={"version":  "1", "build": ConfigBuildConfig(python_packages=["python_packages==3.8.1"])}}` only.

kiendang: Anw I added more tests in #8288 for peace of mind.

PR #8336, opened 2023-12-07
fix UX for nested .get()

IonesioJunior:
## Description
Currently if we have a function that spawn a job and returns `job.result` (in the outer job), we get a double nested actionobject. We probably want to make that single nested. So, this PR aims to achieve this by adding small changes in the way the client checks if a job is resolved and checking its job result type, allwoing to work in a nested promises scenario.

## Affected Dependencies
Job Stash


IonesioJunior:
After some discussion, we've decided to discard the proposed solution. I'll be submitting a new pull request (PR) soon to address this issue more effectively.

PR #8337, opened 2023-12-08
Adding `__repr__` for container workload classes

khoaguin:
Solving these issues:
- [x] [Add __repr__ for SyftImage related classes#905](https://github.com/OpenMined/Heartbeat/issues/905)
- [x] [Add __repr__ to SyftWorker and SyftWorkerPool#906](https://github.com/OpenMined/Heartbeat/issues/906)

Summary of what this PR does:
-  `<client>.images` should show a table containing only the main ones: `id, image_tag, image_hash, created_at`
- Doing `client.worker_pools` would be preferred over client.worker_pool.get_all()
- If `tag = "openmined/test-nginx:0.7.8"`, we can index worker image with `client.images[tag]`
-  `client.worker_pools` show a list of worker pools with `id, pool_name, workers, healthy (healthy/all), running (running/all), image tag/name, created_at`
- we can also index a worker pool with `client.worker_pools[tag]`
- visualising a worker pool, should show first the inline fields like in `client.worker_pools` and after a table of workers show workers' information
- `client.worker_pools[0].workers` shows all the worker `id, image tag/name, resources pool name, healthcheck (healthy/unhealthy), status (created/running/stopped), job`
- `worker = client.worker_pools[0].workers[0]`, then worker will show `id/name, image tag, worker pool name, healthy, status`


khoaguin:
Before building image
image

SyftWorkerImage repr
image

After building image
image

khoaguin:
should we change something about this repr?
image

khoaguin:
image

khoaguin:
image

khoaguin:
image

khoaguin:
image

shubham3121:
@khoaguin can you resolve merge conflict and merge with dev

packages/syft/src/syft/protocol/protocol_version.json
```diff
@@ -1,5 +1,5 @@
 {
-  "1": {
+  "dev": {
```
shubham3121: I think the procotol_version.json seems incorrect, maybe to incorrect merge with dev. Can you reset this file with `dev` 
`git checkout dev -- packages/syft/src/syft/protocol/protocol_version.json` and then in a new terminal
`tox -e syft.protocol.check` to stage any protocol changes.

khoaguin: I noticed there are some differences in my branch's `protocol_version.json` and `dev`'s `protocol_version.json`. Will simply copying the content of `dev`'s `protocol_version.json` into mine work?

khoaguin:
image

khoaguin:
image

khoaguin:
image

khoaguin:
image
image
image
image

khoaguin:
Changes after mergingin with `dev`: we have a new column for `Job` in `WorkerPool`'s repr

image

khoaguin:
image

packages/syft/src/syft/service/worker/worker_image.py
```diff
@@ -91,6 +101,10 @@ class SyftWorkerImage(SyftObject):
     image_hash: Optional[str]
     created_at: DateTime = DateTime.now()
     created_by: SyftVerifyKey
+    source_file: Optional[str]
+    full_tag_str: Optional[str]
```
kiendang: nitpick: This could just be `full_tag`? The type is already `str`.

packages/syft/src/syft/service/worker/worker_pool.py
```diff
@@ -43,6 +64,7 @@ class SyftWorker(SyftObject):
     image_hash: Optional[str]
     healthcheck: Optional[WorkerHealth]
     status: WorkerStatus
+    full_image_tag_str: Optional[str]
```
kiendang: `full_image_tag`

packages/syft/src/syft/service/worker/worker_pool.py
```diff
@@ -79,11 +107,76 @@ class WorkerPool(SyftObject):
...
     max_count: int
     workers: List[SyftWorker]
+    created_at: DateTime = DateTime.now()
+
+    @property
+    def running_workers(self) -> List[SyftWorker]:
+        return [
+            worker
+            for worker in self.workers
+            if get_worker_container_status(worker) == WorkerStatus.RUNNING
```
kiendang: This would create multiple `DockerClient`s, one for each worker, but don't worry about this I'll push a fix for this.

packages/syft/src/syft/service/worker/worker_pool.py
```diff
@@ -99,3 +194,57 @@ class ContainerSpawnStatus(SyftBaseModel):
...
+
+    if isinstance(container, SyftError):
+        return container
+
+    container_status = container.status
+
+    return _CONTAINER_STATUS_TO_WORKER_STATUS.get(
+        container_status,
+        SyftError(message=f"Unknown container status: {container_status}"),
+    )
```
kiendang, line 284: @rasswanth-s

kiendang:
@khoaguin Just resolved conflicts with dev. Will take a look at the `DictTuple` thing.

packages/syft/src/syft/client/api.py
```diff
@@ -384,6 +384,9 @@ def _repr_html_(self) -> Any:
         results = self.get_all()
         return results._repr_html_()
 
+    def __len__(self) -> int:
+        return len(self.get_all())
+
```
kiendang: I think we should not add this here. This is an expensive call to get `__len__` and it's not obvious to users. They could just do `len(client...get_all())` instead.

khoaguin: I think this depends on the `API.get_all()`, but we already use it `_repr_html` and `__getitem__` so it gets called all the time
image

khoaguin: And I add this as it makes a lot of sense to do something like `len(client.images)` or `len(client.worker_pools)` to see how many are there

kiendang: > I think this depends on the `API.get_all()`, but we already use it `_repr_html` and `__getitem__` so it gets called all the time

It's true. These are also expensive, and I plan to discuss with the product team about this. Doesn't mean we should just keep adding to it in the meantime though. Plus this is beyond the scope of this PR. You're not just adding to `client.images` or `client.worker_pools` but all other endpoints (e.g. `client.datasets`) as well.

packages/syft/src/syft/client/domain_client.py
```diff
@@ -197,6 +198,35 @@ def code_history(self) -> CodeHistoriesDict:
...
+        if isinstance(result, DictTuple):
+            worker_names = ""
+            image_tag_workers = {}
+            for worker_pool in result:
+                for worker in worker_pool.workers:
+                    worker_names += f"{worker.name}\n"
+                image_tag_workers[worker_pool.syft_worker_image_name_tag] = worker_names
+            return DictTuple(image_tag_workers)
+        else:
+            return SyftError(message=f"{result.err()}")
```
kiendang: It makes sense that `client.workers()` does not print out the `_repr_html_`. Plain `DictTuple` objects do not have the `_repr_html_` method.

Also if we're not using this right now I think it's better to remove it from this PR. We can add it back when we actually use it.

khoaguin: Done

packages/syft/src/syft/client/api.py
khoaguin: Thanks @kiendang. This makes sense. I removed it

packages/syft/src/syft/service/worker/worker_image.py
khoaguin: Done

packages/syft/src/syft/service/worker/worker_pool.py
khoaguin: Done

packages/syft/src/syft/service/worker/worker_image.py
```diff
@@ -91,6 +101,10 @@ class SyftWorkerImage(SyftObject):
     image_hash: Optional[str]
     created_at: DateTime = DateTime.now()
     created_by: SyftVerifyKey
+    source_file: Optional[str]
+    full_tag: Optional[str]
```
rasswanth-s: NIT: since already have a SyftWorkerImageTag, we could avoid creating full_tag attribute?

rasswanth-s: to retrieve the full tag, we could do
image_tag.full_tag

packages/syft/src/syft/service/worker/worker_pool.py
```diff
@@ -43,6 +65,7 @@ class SyftWorker(SyftObject):
     image_hash: Optional[str]
     healthcheck: Optional[WorkerHealth]
     status: WorkerStatus
+    full_image_tag: Optional[str]
```
rasswanth-s: MAYBE: instead of full_image_tag we could have it as image_tag?

packages/syft/src/syft/service/worker/worker_pool.py
```diff
@@ -79,11 +109,76 @@ class WorkerPool(SyftObject):
...
+        "name",
+        "syft_worker_image_id",
+        "max_count",
+        "workers",
+        "created_at",
+    ]
 
     name: str
     syft_worker_image_id: UID
+    syft_worker_image_name_tag: Optional[str]
```
kiendang: This should just be `syft_worker_image_tag`. I also think just `worker_image_id` and `worker_image_tag` is better, or even just `image_id` and `image_tag`. @rasswanth-s what do you think?

packages/syft/src/syft/service/worker/worker_pool.py
rasswanth-s: Awesome 🙌

packages/syft/src/syft/service/worker/worker_pool.py
rasswanth-s: MAYBE: For image_tag, we could pass SyftWorkerImageTag class from SyftWorkerImage

packages/syft/src/syft/service/worker/worker_pool.py
rasswanth-s: MAYBE: we could name it `syft_worker_image_tag`  and we could initiliaze with SyftWorkerImageTag class to mantain consistency.

packages/syft/src/syft/service/worker/worker_image_service.py
```diff
@@ -108,16 +111,15 @@ def build(
...
         One image one docker file for now
-        TODO: change repr when listing
         """
         result = self.stash.get_all(credentials=context.credentials)
         if result.is_err():
             return SyftError(message=f"{result.err()}")
-
-        return result.ok()
+        images: List[SyftWorkerImage] = result.ok()
+        return DictTuple((im.full_tag, im) for im in images)
```
rasswanth-s: Q: What is the functionality of DictTuple, even though , we have specific columns, I think __repr_attrs__ overrides as the full_tag field is not displayed.


Screenshot 2023-12-22 at 10 44 56 AM

packages/syft/src/syft/service/worker/worker_pool_service.py
```diff
@@ -104,6 +108,7 @@ def create_pool(
             syft_worker_image_id=image_uid,
             max_count=number,
             workers=workers,
+            syft_worker_image_name_tag=worker_image.full_tag,
```
rasswanth-s: MAYBE: We could set the image tag to `worker_image.image_tag` ?

packages/syft/src/syft/service/worker/worker_image.py
```diff
@@ -91,6 +101,10 @@ class SyftWorkerImage(SyftObject):
     image_hash: Optional[str]
     created_at: DateTime = DateTime.now()
     created_by: SyftVerifyKey
+    source_file: Optional[str]
```
rasswanth-s: Q: Is source file attribute being used, I could find it being used?

packages/syft/src/syft/service/worker/worker_pool.py
rasswanth-s: @kiendang  image_id and image_tag is more better 👍

khoaguin:
We have a new convention for syft image tag, repo, name and registry (after reading the [Docker docs](https://docs.docker.com/engine/reference/commandline/tag/#tag-an-image-referenced-by-name-and-tag) and disscuss with Rasswanth and Kien) like below: 
For example if we have the image identifier as `docker.io/openmined/test-nginx:0.7.8` then
```
registry = docker.io
repo = openmined/test-nginx
tag = 0.7.8
repo_with_tag = openmined/test-nginx:0.7.8
full_name = docker.io/openmined/test-nginx
full_name_with_tag = docker.io/openmined/test-nginx:0.7.8
```
This will reflect the following changes in the codebase:
- refactor `SyftWorkerImageTag` into `SyftWorkerImageIdentifier` since tag only refers to the version (e.g. `0.7.8`) and cause confusions
- replace `SyftWorkerImage.full_tag` with `SyftWorkerImageIdentifier.repo_with_tag` everywhere
- add the following properties to `SyftWorkerImageIdentifier`: `repo_with_tag`, `full_name_with_tag`

khoaguin:
@rasswanth-s @shubham3121 I need some inputs on this:  
Irina wants to have an empty list returned when the image is not built, but for now the way we build is to get the `image_id` first with 
```python
dockerfile_list = domain_client.images
workerimage = dockerfile_list[0]
domain_client.api.services.worker_image.build(uid=workerimage.id, tag=docker_tag)
```
So I patched it currently with return a list of images with empty `image_hash` and `image_identifier`
image

khoaguin:
Updated `client.images`
image

packages/syft/src/syft/service/worker/worker_image.py
khoaguin: It's being used when user pass in a Dockerfile and now it holds the name of that file

shubham3121:
> @rasswanth-s @shubham3121 I need some inputs on this: Irina wants to have an empty list returned when the image is not built, but for now the way we build is to get the `image_id` first with
> 
> ```python
> dockerfile_list = domain_client.images
> workerimage = dockerfile_list[0]
> domain_client.api.services.worker_image.build(uid=workerimage.id, tag=docker_tag)
> ```
> 
> So I patched it currently with return a list of images with empty `image_hash` and `image_identifier` image

Yes that sounds good, in case the image_tag or image_hash is not available we should show them as None.

packages/syft/src/syft/service/worker/utils.py
```diff
@@ -62,25 +70,26 @@ def run_container_using_docker(
...
     orchestration: WorkerOrchestrationType,
 ) -> List[ContainerSpawnStatus]:
-    image_tag = worker_image.image_tag
+    image_tag = worker_image.image_identifier
 
     results = []
     if not orchestration == WorkerOrchestrationType.DOCKER:
         return SyftError(message="Only Orchestration via Docker is supported.")
 
-    with contextlib.closing(docker.from_env()) as client:
```
shubham3121: This seems inconsistent with `dev` branch. This line shouldn't be deleted, and the next loop should be within this scope.

khoaguin: Thanks @shubham3121. It was added back

packages/syft/src/syft/service/worker/worker_pool.py
khoaguin: Now pass the whole `SyftWorkerImage` 
image

khoaguin: Done. Now the `SyftWorkerImage` is passed
image

packages/syft/src/syft/service/worker/worker_pool.py
khoaguin: Done 
image

packages/syft/src/syft/service/worker/worker_pool_service.py
khoaguin: Done. Now we pass `worker_image: SyftWorkerImage` into `WorkerPool`

packages/syft/src/syft/service/worker/worker_image.py
khoaguin: Done. Removed `full_tag` property and change `SyftWorkerImageTag` to `SyftWorkerImageIdentifier`

khoaguin:
Updated `client.worker_pools`
image

khoaguin:
Updated single worker pool repr
image
Indexing with string using image repo and tag
image

khoaguin:
Showing a list of workers
image

packages/syft/src/syft/service/worker/worker_pool.py
```diff
@@ -78,12 +109,77 @@ class WorkerPool(SyftObject):
...
+        "image_id",
+        "max_count",
+        "workers",
+        "created_at",
+    ]
 
     name: str
-    syft_worker_image_id: UID
+    image_id: UID
+    image: SyftWorkerImage
```
kiendang: `image: SyftWorkerImage` already contains `image_id` here (`image_id` here is the same as `image.id` right?) so `image_id: UID` is redundant here.

packages/syft/src/syft/service/worker/worker_pool.py
```diff
@@ -43,6 +66,7 @@ class SyftWorker(SyftObject):
     image_hash: Optional[str]
     healthcheck: Optional[WorkerHealth]
     status: WorkerStatus
+    image: SyftWorkerImage
```
kiendang: Similarly, we no longer need `image_hash: Optional[str]` here since `image: SyftWorkerImage` already contains it.

packages/syft/src/syft/service/worker/worker_image.py
kiendang: Hmm I still don't see where this attribute is used?

packages/syft/src/syft/service/worker/worker_pool.py
khoaguin: That's true, thanks @kiendang. Updated the code

packages/syft/src/syft/service/worker/worker_pool.py
khoaguin: That's true, thanks @kiendang . Updated the code

packages/syft/src/syft/service/worker/worker_image.py
khoaguin: It's only used if we pass the a `.dockerfile`. For now in the api notebook we pass directly the string so we are not using it yet

kiendang: If we're not using it yet we should not include it in this PR. We should only add it in when we use it and implement support for it.

shubham3121: This shouldn't be tracked at SyftWorkerImage level, this is more of a WorkerConfig thing.

packages/syft/src/syft/service/worker/worker_pool.py
```diff
@@ -62,14 +85,30 @@ def get_job_repr(self):
...
 
-    def _coll_repr_(self):
-        return {
-            "name": self.name,
-            "container id": self.container_id,
-            "status": str(self.status.value.lower()),
-            "job": self.get_job_repr(),
-            "created at": str(self.created_at),
-        }
+    def get_status_healthcheck(self) -> None:
```
shubham3121: Health checks should be done via API call to the server not on the client side.

packages/syft/src/syft/service/worker/worker_pool.py
```diff
@@ -78,12 +117,85 @@ class WorkerPool(SyftObject):
...
     name: str
-    syft_worker_image_id: UID
+    image: Optional[SyftWorkerImage]
     max_count: int
     workers: List[SyftWorker]
+    created_at: DateTime = DateTime.now()
+
+    @property
+    def running_workers(self) -> List[SyftWorker]:
+        with contextlib.closing(docker.from_env()) as client:
```
shubham3121: Same here, these should be triggered over an API call instead.

packages/syft/src/syft/service/worker/worker_pool.py
```diff
@@ -78,12 +117,85 @@ class WorkerPool(SyftObject):
...
+            return [
+                worker
+                for worker in self.workers
+                if _get_worker_container_status(client, worker) == WorkerStatus.RUNNING
+            ]
+
+    @property
+    def healthy_workers(self) -> List[SyftWorker]:
+        with contextlib.closing(docker.from_env()) as client:
```
shubham3121: Same here, this should be triggered over an API call instead.

khoaguin:
Updated after having default worker pool in dev

image

khoaguin:
image

khoaguin:
image

packages/syft/src/syft/service/worker/worker_pool.py
khoaguin: Thanks @shubham3121, done!

packages/syft/src/syft/service/worker/worker_pool.py
khoaguin: Thanks @shubham3121, done!

packages/syft/src/syft/service/worker/worker_pool.py
khoaguin: Thanks @shubham3121, done!

khoaguin:
When using `os.environ["ORCHESTRA_DEPLOYMENT_TYPE"] = "container_stack"`, we get

image
image
image
image
image

shubham3121 (approved):
Awesome work @khoaguin !!! 🎉🎊

PR #8347, opened 2023-12-14
Added api for worker details

PeterChung241:
- [x] Get a worker
 - [x] Get worker status 

Ready for review.

PR #8357, opened 2023-12-18
Always close DockerClient after use

kiendang:
Always close `DockerClient` after use, with `contextlib.closing` or `finally` to ensure resource safety.

```py
with contextlib.closing(docker.from_env) as client:
    ...
```

Also pass a `client: DockerClient` as argument to functions that need it instead of creating a new one inside the function body.

For example, from this

```py
def start_worker_container(worker_num: int):
    client = docker.from_env()
    ...

start_worker_container(1)
```

to this

```py
def start_worker_container(client: DockerClient, worker_num: int):
    ...

with contextlib.closing(docker.from_env()) as client:
    start_worker_container(client, 1)
```

This also make some parts of the code more efficient as well since we avoid creating multiple `DockerClient`s. 
For example currently

```py
for worker_num in range(n_workers):
    start_worker_container(worker_num)
```

will create `n_workers` `DockerClient`s

After refactoring,

```py
with contextlib.closing(docker.from_env()) as client:
    for worker_num in range(n_workers):
        start_worker_container(client, worker_num)
```

only uses 1 client.

### DIFF
--- a/packages/syft/src/syft/service/worker/worker_pool_service.py
+++ b/packages/syft/src/syft/service/worker/worker_pool_service.py
@@ -1,4 +1,6 @@
 # stdlib
+import contextlib
+from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Tuple
@@ -146,28 +148,14 @@ class SyftWorkerPoolService(AbstractService):
         worker_pool, worker = worker_pool_worker
 
         # delete the worker using docker client sdk
-        docker_container = _get_worker_container(worker)
-        if isinstance(docker_container, SyftError):
-            return docker_container
+        with contextlib.closing(docker.from_env()) as client:
+            docker_container = _get_worker_container(client, worker)
+            if isinstance(docker_container, SyftError):
+                return docker_container
 
-        try:
-            # stop the container
-            docker_container.stop()
-            # Remove the container and its volumes
-            docker_container.remove(force=force, v=True)
-        except docker.errors.APIError as e:
-            if "removal of container" in str(e) and "is already in progress" in str(e):
-                # If the container is already being removed, ignore the error
-                pass
-            else:
-                # If it's a different error, return it
-                return SyftError(
-                    message=f"Failed to delete worker with id: {worker_id}. Error: {e}"
-                )
-        except Exception as e:
-            return SyftError(
-                message=f"Failed to delete worker with id: {worker_id}. Error: {e}"
-            )
+            stopped = _stop_worker_container(worker, docker_container, force)
+            if stopped is not None:
+                return stopped
 
         # remove the worker from the pool
         worker_pool.workers.remove(worker)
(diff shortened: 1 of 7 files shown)


packages/syft/src/syft/service/worker/worker_service.py
```diff
@@ -240,19 +241,19 @@ def stop(
...
-            except Exception as e:
-                # we dont throw an error here because apparently the container was already killed
-                print(f"Failed to kill container {e}")
+                # stop container
+                try:
+                    client.containers.list(filters={"id": w.container_id})[0].stop()
+                    # also prune here?
+                except Exception as e:
+                    # we dont throw an error here because apparently the container was already killed
+                    print(f"Failed to kill container {e}")
```
kiendang: @rasswanth-s I noticed for this part, if deleting a worker fails, the function will return immediately, without trying to delete the remaining workers. Should we attempt to delete all workers first, then return a `SyftError` listing the ones that fails if any?

Also after stopping the containers with `stop()` should we remove (`Container.remove()`, equivalent to `docker rm`) them too?

packages/syft/src/syft/service/worker/worker_pool_service.py
```diff
@@ -324,24 +325,41 @@ def _get_worker_container(
...
+
+def _remove_worker_container(container: Container, **kwargs: Any) -> None:
+    try:
+        container.remove(**kwargs)
+    except docker.errors.NotFound:
+        return
+    except docker.errors.APIError as e:
+        if "removal of container" in str(e) and "is already in progress" in str(e):
+            # If the container is already being removed, ignore the error
+            return
```
kiendang, line 365: @rasswanth-s This is a more robust way to remove the container. Error message from docker server is dependent on the specific docker server implementation. I tried with another dockerd not Docker Desktop and this does not correctly detect the container no longer existing. Checking for a `docker.errors.NotFound` (which is the same as a 404 `docker.errors.APIError`) is more reliable.

packages/syft/src/syft/service/worker/worker_pool_service.py
```diff
@@ -195,10 +184,21 @@ def get_worker(
 
         worker_pool, worker = worker_pool_worker
 
-        worker_status = _get_worker_container_status(worker)
+        with contextlib.closing(docker.from_env()) as client:
+            container = _get_worker_container(client, worker)
```
rasswanth-s, line 188: @kiendang  I think it would be good to have a separate function utility function for retrieving the worker status.
Because I saw in Khoa's PR we used utility funciton in displaying statuses.
I will defer the decision to you post merge which we could implement in Khoa's PR

rasswanth-s (approved):
Awesome Work @kiendang

rasswanth-s:
Merging the PR, as the tests are blocked by the flaky mongo tests which are passing in other runners of different OS.

packages/syft/src/syft/service/worker/worker_pool_service.py
kiendang: As I explained here https://github.com/OpenMined/PySyft/pull/8347#discussion_r1429890863 I don't think a function like `def _get_worker_container_status(worker: Worker) -> WorkerStatus` would be good. We need to get the `Container` from the `worker` first and we might need to use the `Container` for something else beside getting the status, so this would be better `def _get_worker_container_status(container: Container) -> WorkerStatus` and this is not much different from just having a mapping `_CONTAINER_STATUS_TO_WORKER_STATUS: dict[str, WorkerStatus]` anyway and call `_CONTAINER_STATUS_TO_WORKER_STATUS.get(worker_container.status)`. We could wrap that in a function but it doesn't necessarily simplify things.

kiendang: On second thoughts, it's not a bad idea to wrap everything in a function, just need to take notes whether we've already got the worker container object. Will make changes in Khoa's PR.

PR #8363, opened 2023-12-21
Add build and push to client.images

kiendang:
instead of having `client.build_image` and `client.push_image`

Added a password prompt on top of `SyftClient.api.services.worker_image.push` when the API endpoint is returned by `client.images`.

### DIFF
--- a/packages/syft/src/syft/client/client.py
+++ b/packages/syft/src/syft/client/client.py
@@ -11,9 +11,11 @@ from typing import Callable
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import ParamSpec
 from typing import TYPE_CHECKING
 from typing import Tuple
 from typing import Type
+from typing import TypeVar
 from typing import Union
 from typing import cast
 
@@ -664,9 +666,13 @@ class SyftClient:
 
     @property
     def images(self) -> Any:
-        if self.api.has_service("worker_image"):
-            return self.api.services.worker_image
-        return None
+        if not self.api.has_service("worker_image"):
+            return None
+
+        api = self.api.services.worker_image
+        api.push = _prompt_for_password(api.push)
+
+        return api
 
     def login_as_guest(self) -> Self:
         _guest_client = self.guest()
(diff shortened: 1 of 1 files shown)


packages/syft/src/syft/client/client.py
```diff
@@ -664,40 +666,13 @@ def registries(self) -> Any:
...
             return None
 
-        return self.api.services.worker_image.build(
-            image=image,
-            tag=repo_name,
-            version=version,
-            registry=registry,
-        )
+        api = self.api.services.worker_image
+        api.push = _prompt_for_password(api.push)
```
kiendang, line 673: we have to clone `self.api.services.worker_image` here somehow. so that `client.api.services.worker_image.push` is not modified, only `client.images.push`.

yashgorana: maybe we can make this a bit more generic?

```python
prompt_map = {
    "push": dict(password="Enter registry password"),
    "delete": dict(password="Enter password to delete the registry"),
    "api_func": dict(arg="Enter arg to pass through to api_func"),
}
return with_prompts(self.api.services.worker_image, prompt_map )
```

this basically scales to any api, any func, any prompt (as long as it's logical)

kiendang: We should make it more generic later on when we need to use similar things in different places. Currently this is the only place where we need this so I think this is sufficient, best to avoid premature abstraction. Prompting for password requires using `getpass` while `input` is more suitable for other arguments so to make it work for any arbitrary arg we would need to handle that too and that would be too much for now.

packages/syft/src/syft/client/client.py
```diff
@@ -1060,3 +1035,17 @@ def get_client(
...
+
+
+def _prompt_for_password(f: Callable[P, T]) -> Callable[P, T]:
+    def inner(*args: P.args, **kwargs: P.kwargs):
+        if kwargs.get("password") is None:
+            kwargs["password"] = getpass("Password: ")
+
+        return f(*args, **kwargs)
+
+    return inner
```
kiendang, line 1051: One small issue with is now if the user wants to pass password directly as an argument to `client.images.push` they have to pass it as a kwarg. Passing it as a positional arg would not work, but this should be fixable (by getting the signatures of `f` and check which argument is `password`).

packages/syft/src/syft/client/client.py
kiendang: I do think it's not unlikely that we need to use more of this pattern. As I understand `client.<service>` is more suited for interactivity and `client.api.services.<service>` is more for scripting/programmable stuff and there are definitely frictions between them and we might need to adapt the latter to better fit the former. Just think that as more concrete use cases surface we would have a better idea on how to make the right abstraction.

kiendang:
Close this. After discussing with @shubham3121 and team, we decide to modify `SyftClient.api.service.worker_image.push` to add a prompt on top, and there would be no difference between `SyftClient.images.push` and `SyftClient.api.service.worker_image.push`. After further investigation I decide the best way is to just modify `SyftWorkerImageService.push` to add the prompt directly rather than adding it inside `SyftClient.images` since if doing the latter we would modify `push` every time someone calls `SyftClient.images`.

PR #8377, opened 2024-01-08
syncing l0

koenvanderveen:
## Description
Enables syncing requests and results between nodes, full flow in separate notebook

## Affected Dependencies
None

## How has this been tested?
- Added new tests for requests with multiple nodes
- Test full flow in separate notebook

## Checklist


packages/syft/src/syft/node/node.py
```diff
@@ -1243,21 +1243,76 @@ def add_queueitem_to_queue(
             return result
         return job
 
+    def _can_enqueue_user_code_job(
```
koenvanderveen: I think this function might be problematic in some cases. In the case of nested jobs we follow the logic that: once you can execute the outer job, you will automatically get permissions to execute inner jobs. This information (that you come from an outer job), will be added to the context. I am not sure if we currently have a test that tests this properly, but this might break that

koenvanderveen: Also, I wonder whether we need this at all. Don't we already check permission once we try to execute the code?

packages/syft/src/syft/node/node.py
```diff
@@ -1243,21 +1243,76 @@ def add_queueitem_to_queue(
...
 
         is_user_code = unsigned_call.path == "code.call"
 
-        service, method = unsigned_call.path.split(".")
+        service_str, method_str = unsigned_call.path.split(".")
 
         action = None
         if is_user_code:
             action = Action.from_api_call(unsigned_call)
+            if self.node_side_type == NodeSideType.LOW_SIDE:
```
koenvanderveen, line 1270: what is the reason to not handle this within the service? This seems like its service specific logic to me

packages/syft/src/syft/service/job/job_service.py
```diff
@@ -57,6 +61,27 @@ def get_all(self, context: AuthedServiceContext) -> Union[List[Job], SyftError]:
...
+        self, context: AuthedServiceContext, user_code_id: UID
+    ) -> Union[List[Job], SyftError]:
+        all_jobs = self.stash.get_all(context.credentials)
+        if all_jobs.is_err():
+            return SyftError(message=all_jobs.err())
+
+        all_jobs = all_jobs.ok()
+
+        res = []
+        for job in all_jobs:
```
koenvanderveen: this should probably be a stash method that does the querying, in that case you also search by unique keys I think

packages/syft/src/syft/service/job/job_service.py
```diff
@@ -158,3 +202,53 @@ def get_active(self, context: AuthedServiceContext) -> Union[List[Job], SyftErro
...
+    @service_method(
+        path="job.create_job_for_user_code_id",
+        name="create_job_for_user_code_id",
+        roles=DATA_OWNER_ROLE_LEVEL,
+    )
+    def create_job_for_user_code_id(
+        self, context: AuthedServiceContext, user_code_id: UID
+    ) -> Union[Job, SyftError]:
+        """
+        When DO submits job from high side for DS,
```
koenvanderveen: maybe remove this comment

packages/syft/src/syft/service/job/job_stash.py
```diff
@@ -111,6 +133,9 @@ class Job(SyftObject):
     job_worker_id: Optional[UID] = None
     updated_at: Optional[DateTime] = None
 
+    # Jobs created by DO on low side have no action yet, but do have user_code_id
+    user_code_id_override: Optional[UID] = None
```
koenvanderveen: Can you explain me how this works?

packages/syft/src/syft/service/request/request.py
```diff
@@ -456,10 +459,57 @@ def save(self, context: AuthedServiceContext) -> Result[SyftSuccess, SyftError]:
         save_method = context.node.get_service_method(RequestService.save)
         return save_method(context=context, request=self)
 
+    def _get_or_create_job(self) -> Union[Job, SyftError]:
```
koenvanderveen: lets add "last" to the function name

packages/syft/src/syft/service/request/request.py
```diff
@@ -505,6 +555,17 @@ def accept_by_depositing_result(self, result: Any, force: bool = False):
...
                 return result
+
+            if job is not None:
+                job_info = job.info
+            else:
+                job_info = JobInfo(
+                    status=JobStatus.COMPLETED,
+                    resolved=True,
+                )
+            res = self._set_job_result(result=result, job_info=job_info)
```
koenvanderveen: this means that currently be default it syncs everything?

packages/syft/src/syft/service/request/request.py
```diff
@@ -535,12 +597,44 @@ def accept_by_depositing_result(self, result: Any, force: bool = False):
...
+                return approved
+
+            if job is not None:
+                job_info = job.info
+            else:
+                job_info = JobInfo(
+                    status=JobStatus.COMPLETED,
+                    resolved=True,
+                )
+            res = self._set_job_result(result=result, job_info=job_info)
```
koenvanderveen: why are we doing this twice?

packages/syft/tests/syft/request/request_multiple_nodes_test.py
```diff
@@ -0,0 +1,216 @@
...
+
+@pytest.fixture(scope="function")
+def client_ds_1(node_1, client_do_1):
+    client_do_1.register(
+        name="test_user", email="test@us.er", password="1234", password_verify="1234"
+    )
+    return node_1.login(email="test@us.er", password="1234")
+
+
+# @pytest.fixture(scope="function")
```
koenvanderveen: lets remove these

packages/syft/tests/syft/request/request_multiple_nodes_test.py
```diff
@@ -0,0 +1,216 @@
...
+    compute_sum.code = dedent(compute_sum.code)
+
+    client_ds_1.code.request_code_execution(compute_sum)
+
+    # Submit + execute on second node
+    request_1_do = client_do_1.requests[0]
+    client_do_2.code.sync_code_from_request(request_1_do)
+
+    # DO executes + syncs
+    client_do_2._fetch_api(client_do_2.credentials)
```
koenvanderveen, line 129: maybe we need to fix this?

packages/syft/tests/syft/request/request_multiple_nodes_test.py
```diff
@@ -0,0 +1,216 @@
...
+
+    # DS checks job info
+    job_1_ds = client_ds_1.code.compute_mean.jobs[-1]
+    assert job_1_ds.info == job_2.info
+
+    # DO finishes + syncs job result
+    result = job_2.wait().get()
+    assert result == dataset_2.data.mean()
+    assert job_2.status == JobStatus.COMPLETED
+    res = request_1_do.accept_by_depositing_result(job_2)
```
koenvanderveen: didnt we agree here to sync a job info object instead of a job?

eelcovdw: The JobInfo object is only for syncing the metadata, not the result (line 170)

eelcovdw: here we want to sync both metadata + result -> sync whole job

packages/syft/tests/syft/request/request_multiple_nodes_test.py
eelcovdw: Lets do it in separate PR?

packages/syft/tests/syft/request/request_multiple_nodes_test.py
eelcovdw: the JobInfo is only the metadata, not the result (line 170)

packages/syft/tests/syft/request/request_multiple_nodes_test.py
eelcovdw: Lets do it in separate PR?

packages/syft/src/syft/node/node.py
koenvanderveen: two options:
1. keep this implementation
2. return a job instead that contains the error

For the long term we may consider refactoring the permission logic and check where we want it to happen: node.py user_code_service.py, action_service.py

eelcovdw: Decided to keep it like this for now, context: https://openmined.slack.com/archives/C05TVM4RZKQ/p1705646834228349?thread_ts=1705593124.565069&cid=C05TVM4RZKQ

packages/syft/tests/syft/request/request_multiple_nodes_test.py
eelcovdw: Resolved with new jobinfo

packages/syft/tests/syft/request/request_multiple_nodes_test.py
eelcovdw: Doing this in new PR, took a bit more time

packages/syft/src/syft/node/node.py
eelcovdw: Leaving this here for now, see https://github.com/OpenMined/PySyft/pull/8377#discussion_r1458603133

eelcovdw (approved):

packages/syft/src/syft/service/job/job_stash.py
```diff
@@ -92,7 +96,7 @@ class JobV2(SyftObject):
 
 
 @serializable()
-class Job(SyftObject):
+class JobV3(SyftObject):
```
shubham3121, line 99: I think we don't need to create a new job class version. Since Version 3 is the latest one being edited in dev.

eelcovdw: Ah I see, fixed it here https://github.com/OpenMined/PySyft/pull/8407

PR #8380, opened 2024-01-09
Fix bugs in Default Worker Pool Spinning

shubham3121:
## Description

- [x] Fixing default worker not being spawned on hot-reload
- [x] Fix Query key for worker config in image stash `get_by_docker_config` method

### DIFF
--- a/packages/syft/src/syft/node/node.py
+++ b/packages/syft/src/syft/node/node.py
@@ -1463,9 +1463,7 @@ def create_default_worker_pool(node: Node) -> Optional[SyftError]:
         # Else add a worker to existing worker pool
         add_worker_method = node.get_service_method(SyftWorkerPoolService.add_workers)
         result = add_worker_method(
-            context=context,
-            number=worker_count,
-            pool_name=DEFAULT_WORKER_POOL_NAME
+            context=context, number=worker_count, pool_name=DEFAULT_WORKER_POOL_NAME
         )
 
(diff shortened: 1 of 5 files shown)


packages/syft/src/syft/node/node.py
```diff
@@ -1435,26 +1435,36 @@ def create_default_worker_pool(node: Node) -> Optional[SyftError]:
...
 
-    # Build the Image for given tag
-    result = image_build_method(
-        context, image_uid=default_image.id, tag=DEFAULT_WORKER_IMAGE_TAG
-    )
-
-    if isinstance(result, SyftError):
-        print("Failed to build default worker image: ", result.message)
-        return
+    if not default_image.built_at:
```
yashgorana: you can now call `if not default_image.is_built`

shubham3121: Fixed in https://github.com/OpenMined/PySyft/pull/8380/commits/04d116caa158474f482206191f792ca499999520

kiendang:
If I understand correctly, this

https://github.com/OpenMined/PySyft/blob/5563cbe6e1e38ddf8f87acba09ab5f36d497dcd2/packages/syft/src/syft/node/node.py#L1475-L1482

only prints the message for the first container that errors, even if there are multiple containers failing to build? Should it aggregate all the containers that fail to build and print out all the error messages instead? I understand that this logic exists prior to the PR.

PR #8396, opened 2024-01-16
[UX] Rename Container Workload Services

khoaguin:
Issue: https://github.com/orgs/OpenMined/projects/88/views/1?pane=issue&itemId=49774845


packages/syft/src/syft/service/worker/worker_pool_service.py
```diff
@@ -111,7 +111,7 @@ def create_pool(
             context=context,
```
rasswanth-s, line 111: NITPICK: could we also change the doc string from `number` to `num_workers`

rasswanth-s (approved):
LGTM, left one small comment, we could merge after the comment.

PR #8415, opened 2024-01-24
fix nested syncing

koenvanderveen:


packages/syft/src/syft/client/client.py
```diff
@@ -527,6 +528,29 @@ def create_project(
         project = project_create.start()
         return project
 
+    def sync_code_from_request(self, request):
+        code = deepcopy(request.code)
+        code.worker_pool_id = None
```
shubham3121: `code.worker_pool_id` has been renamed to `code.worker_pool_name`.
Also, any reason we're setting worker_pool_id to None ? Don't we want the code to be executed on custom workers ?

packages/syft/src/syft/client/client.py
```diff
@@ -527,6 +528,29 @@ def create_project(
...
+
+        def get_nested_codes(code):
+            result = []
+            for __, (linked_code_obj, _) in code.nested_codes.items():
+                nested_code = linked_code_obj.resolve
+                nested_code = deepcopy(nested_code)
+                nested_code.worker_pool_id = None
```
shubham3121: Same here `worker_pool_id` attribute renamed to `worker_pool_name` in UserCode and SubmitUserCode

packages/syft/src/syft/client/client.py
koenvanderveen: no but there is a difference between the low side name/id and high side

packages/syft/src/syft/client/client.py
```diff
@@ -527,6 +528,27 @@ def create_project(
         project = project_create.start()
         return project
 
+    def sync_code_from_request(self, request):
+        code = deepcopy(request.code)
```
shubham3121: [Nitpick] Can we also check if request.code is a UserCode obj and not None or SyftError

shubham3121 (approved):
Lgtm !! 🎉 Just left a minor comment.

PR #8417, opened 2024-01-24
Kubernetes custom worker builder

yashgorana:
## Description

Closes https://github.com/OpenMined/Heartbeat/issues/892

Kaniko can build and push in a single job. Since build & push are two separate Syft APIs, it becomes a bit non-trivial to have kaniko build twice... first with `--no-push` and then with `--destination`. To keep things aligned with the API-stack, this PR adds build and push as two separate steps without the need of an internal registry:
- [Kaniko](https://github.com/GoogleContainerTools/kaniko) builds the image and saves it as a tar file on a shared volume
- [Crane](https://github.com/google/go-containerregistry) pushes the tar image to the provided registry (without rebuilding)

There are some tradeoffs when using crane or internal registry, hence chose this method as it's the simplest and requires the least amount of maintenance. 

PS - Crane is used by Kaniko itself.

## Affected Dependencies

## How has this been tested?

## Checklist


packages/syft/src/syft/custom_worker/builder_k8s.py
```diff
@@ -0,0 +1,302 @@
...
+from pathlib import Path
+from time import time
+from typing import Dict
+from typing import Optional
+
+# third party
+import kr8s
+from kr8s.objects import APIObject
+from kr8s.objects import ConfigMap
+from kr8s.objects import Job
```
khoaguin, line 12: should we name this something like `K8sJob` to make it not confused with the Job class we have in syft?

yashgorana: Since the scope of this `Job` is limited to `builder_k8s`, it's completely fine to keep it as-is.

packages/syft/src/syft/custom_worker/builder_k8s.py
```diff
@@ -0,0 +1,318 @@
...
+
+class InvalidImageDigest(Exception):
+    pass
+
+
+class KubernetesBuilder(BuilderBase):
+    def __init__(self):
+        self.client = kr8s.api(namespace=KUBERNETES_NAMESPACE)
+
+    def build_image(
```
rasswanth-s, line 35: NITPICK: We need to add the interface separation in worker_image_service .py `def build` method for branching kubernetes builds, currently it builds using only docker

packages/syft/src/syft/custom_worker/builder_k8s.py
```diff
@@ -0,0 +1,318 @@
...
+            # TODO: check job status, raise with logs
+
+            image_digest = self._get_image_digest(job)
+            if not image_digest:
+                raise InvalidImageDigest("Did not get any image digest from the job")
+
+            # get logs
+            logs = self._get_logs(job)
+        except Exception:
+            raise
```
rasswanth-s, line 75: MAYBE: We could return a SyftError here in the except block of the `exception raised`, that is automatically handled in worker_image_service.py build method

rasswanth-s: maybe , we could also add 
```python3
# Create and start the job
        job = self._create_kaniko_build_job(
            job_id=job_id,
            tag=tag,
            job_config=config,
            build_args=buildargs,
        )
```
as part of the try catch block

packages/syft/src/syft/custom_worker/builder_k8s.py
```diff
@@ -0,0 +1,318 @@
...
+        self,
+        tag: str,
+        username: str,
+        password: str,
+        registry_url: str,
+        **kwargs,
+    ) -> ImagePushResult:
+        # Create and start the job
+        job_id = self._new_job_id(tag)
+        job = self._create_push_job(
```
rasswanth-s, line 95: MAYBE: We could add a try catch block also for push , like we have for build_image above

packages/syft/src/syft/custom_worker/builder_k8s.py
yashgorana: Sorry I didn't get you. The call stack is as follows: `def build()` (worker_image_service.py) -> `docker_build()` (utils.py) -> `CustomWorkerBuilder.build_image()` -> `self.builder.build_image()` which uses a builder determined by runtime.

I think the problem here is the function `docker_build()` in utils.py. Should be called as `worker_build()` and `worker_push()` But i won't address them here

packages/syft/src/syft/custom_worker/builder_k8s.py
yashgorana: SyftError should not be returned by core modules like `custom_worker`. Instead they should be bare-bones python with minimal dependencies. SyftError is a API+Service-specific implementation, and sits above the core module.

If you look at the code, the code `docker_build()` and `docker_push()` in utils.py manages the exception handling and wraps them as SyftError. So shouldn't be an issue.

The reason why I handled Exception here is to implement the `finally` block.

packages/syft/src/syft/custom_worker/builder_k8s.py
yashgorana: Not needed. Same as https://github.com/OpenMined/PySyft/pull/8417#discussion_r1469387426

yashgorana: To summarize, API->Service->Handlers must return SyftSuccess/SyftError, meanwhile core modules that power the Handlers should be pure python. This makes sure the code is design to be tested in isolation and independent of any unexpected cross-deps.

packages/syft/src/syft/custom_worker/builder_k8s.py
rasswanth-s: Sorry, my bad, with the function name, I assumed , it is only docker build

rasswanth-s: Yep , we could change it later.

packages/syft/src/syft/custom_worker/builder_k8s.py
rasswanth-s: I see the try catch block is part of docker push/docker build

rasswanth-s (approved):
Well Done @yashgorana 🚀

PR #8419, opened 2024-01-25
Fix bug in jobs and launching of workers

shubham3121:
## Description
Fixed a couple of bugs:
- [x] Orchestra not able to launch queue on given port with given n_consumers
- [x] Execution of nested functions failed in non-python mode environments
- [x] Code Request have different node_uid and user_verify_key during syncing from request


packages/syft/src/syft/service/queue/zmq_queue.py
```diff
@@ -246,7 +246,16 @@ def unwrap_nested_actionobjects(self, data):
                 key: self.unwrap_nested_actionobjects(obj) for key, obj in data.items()
             }
         if isinstance(data, ActionObject):
-            return data.get()
```
shubham3121, line 249: This doesn't work on the server side, because no API exists for given credentials on the server. Seems more robust if we use the action_service instead.

packages/syft/src/syft/service/queue/zmq_queue.py
```diff
@@ -378,16 +387,21 @@ def update_consumer_state_for_worker(
             )
             return
 
-        res = self.worker_stash.update_consumer_state(
-            credentials=self.worker_stash.partition.root_verify_key,
-            worker_uid=syft_worker_id,
-            consumer_state=consumer_state,
-        )
-        if res.is_err():
+        try:
```
shubham3121, line 390: Just some extra try catching if something fails during update, it doesn't kill the consumer.

packages/syft/src/syft/client/client.py
```diff
@@ -529,24 +529,36 @@ def create_project(
...
+        from ..store.linked_obj import LinkedObject
+
+        code: Union[UserCode, SyftError] = request.code
         if isinstance(code, SyftError):
             return code
-        elif code is None:
-            return SyftError(message="no code inside request")
 
         code = deepcopy(code)
+        code.node_uid = self.id
```
shubham3121, line 541: node_uid and user_verify_key should be updated to current node and credentials, otherwise it reflects the same of the low-side node.

koenvanderveen (approved):

PR #8431, opened 2024-01-29
Fix Hagrid/PySyft compatibility issue with for Hagrid:latest - PySyft:0.8.2

IonesioJunior:
## Description
This PR addresses a compatibility issue identified in the PySyft 0.8.3 release, particularly concerning the `sy.serve_node` function. The recent introduction of the **'syft functions'** feature in version 0.8.3 necessitated an update to `sy.serve_node` to support new worker parameters. However, this update has led to a backward compatibility issue with `PySyft 0.8.2`.

## Problem
In detail, when launching a node via the latest version of Hagrid, `sy.serve_node` is invoked with the new worker parameters. Since these parameters were absent in `PySyft 0.8.2`, this results in an error, hindering the seamless functionality across different versions.

## Solution
To resolve this issue, I propose implementing a syft version check at the start of the node launching process. Specifically, the code will assess the PySyft version in use. If the version is identified as greater than `0.8.2`, the new worker parameters will be passed to `sy.serve_node`. This approach ensures compatibility with PySyft versions both before and after 0.8.3, thereby maintaining the robustness and versatility of the software.

### DIFF
--- a/packages/hagrid/hagrid/orchestra.py
+++ b/packages/hagrid/hagrid/orchestra.py
@@ -1,4 +1,5 @@
 """Python Level API to launch Docker Containers using Hagrid"""
+
 # future
 from __future__ import annotations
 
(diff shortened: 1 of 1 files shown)


madhavajay:
I just spoke with census and figured id try to fix this, looks like we can do something more general for future changes:
https://github.com/OpenMined/PySyft/pull/8435

madhavajay:
Ill close this for now

PR #8434, opened 2024-01-30
Seaweed large file fix 

koenvanderveen:


packages/grid/default.env
```diff
@@ -71,7 +71,8 @@ S3_ROOT_PWD="admin" # needs randomizing
 S3_REGION="us-east-1"
  #not-using
 S3_PRESIGNED_TIMEOUT_SECS=1800
-S3_VOLUME_SIZE_MB=1024
+S3_VOLUME_SIZE_MB=40000
```
madhavajay: Careful, this will allocate that space, which the system might not have.

We might need to add a flag to hagrid to make this a little clearer for people deploying without kubernetes.

In the helm chart its `s3VolumeSizeMB` and is currently defaulted to 1000 as well.

madhavajay:
Just so we know what `large_disk` is doing, it looks like its using an extra byte to allocate a larger volume increasing from 30 GB to 8000 GB (8 Terrabytes). I guess this is just an unfortunate side effect of having files > 30GB, but I think for now we can just stick to the larger one and not worry about this for the foreseeable future.

@koenvanderveen can you add a comment that maybe links to the source: https://github.dev/seaweedfs/seaweedfs/blob/d6e0cae6e7bf97534593842ba60fedf8b7f8b229/weed/command/update.go#L138-L139

Note to future self coming back here in 2034 (yes I know, we thought 8 TB files were too big). 😭

Screenshot 2024-01-31 at 1 03 00 pm

PR #8444, opened 2024-02-01
Add pullImageSecrets for authenticated registries

yashgorana:
## Description

Closes https://github.com/OpenMined/Heartbeat/issues/961

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/packages/syft/src/syft/custom_worker/runner_k8s.py
+++ b/packages/syft/src/syft/custom_worker/runner_k8s.py
@@ -1,5 +1,7 @@
 # stdlib
+import base64
 import copy
+import json
 import os
 from time import sleep
 from typing import List
@@ -10,6 +12,7 @@ from typing import Union
 import kr8s
 from kr8s.objects import APIObject
 from kr8s.objects import Pod
+from kr8s.objects import Secret
 from kr8s.objects import StatefulSet
 
 # relative
@@ -27,13 +30,24 @@ class KubernetesRunner:
         tag: str,
         replicas: int = 1,
         env_vars: Optional[dict] = None,
+        reg_username: Optional[str] = None,
+        reg_password: Optional[str] = None,
+        reg_url: Optional[str] = None,
         **kwargs,
     ) -> StatefulSet:
+        if reg_username and reg_password and reg_url:
+            pull_secret = self._create_image_pull_secret(
+                pool_name,
+                reg_username,
+                reg_password,
+                reg_url,
+            )
         deployment = self._create_stateful_set(
             pool_name,
             tag,
             replicas,
             env_vars,
+            pull_secret=pull_secret,
             **kwargs,
         )
         self.wait(deployment, available_replicas=replicas)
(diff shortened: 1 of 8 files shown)


shubham3121 (approved):

shubham3121:
Great work @yashgorana @rasswanth-s with registry integration to requests 🎉🎉

packages/syft/src/syft/custom_worker/runner_k8s.py
```diff
@@ -198,12 +263,14 @@ def _create_stateful_set(
...
                                 {
                                     "name": pool_name,
+                                    "imagePullPolicy": "IfNotPresent",
                                     "image": tag,
                                     "env": env_clone,
                                     "volumeMounts": [creds_volume_mount],
                                 }
                             ],
                             "volumes": [creds_volume],
+                            "imagePullSecrets": pull_secret_obj,
```
rasswanth-s, line 273: During my testing, I found that the deployment fails if the imagePullSecrets is None

We could either set it to null or 
omit completey
Screenshot 2024-02-02 at 11 05 14 AM
Sample run setting the value as None

rasswanth-s: I did the testing on trying to do helm install by setting imagepullsecrets as None in backend-statefulset.yml

yashgorana: Wroker pod spec isn't created & applied until we call the API. This definitely has to do with the static YAML files.

yashgorana: Also you're setting it as `None` which is a string in YAML, it should be `null`. Python `None` maps to YAML `null` automatically.

packages/syft/src/syft/service/worker/utils.py
```diff
@@ -337,6 +344,9 @@ def run_workers_in_kubernetes(
             worker_count,
             queue_port,
             debug,
+            reg_username,
```
rasswanth-s: NITPICK: Since we are passing lot of arguments to this function, instead of positional arguments,could we pass arguments by name
```python3
create_kubernetes_pool(runner=runner,
worker_image=worker_image,
...
) 
```

packages/syft/src/syft/custom_worker/runner_k8s.py
```diff
@@ -217,3 +284,23 @@ def _create_or_get(self, obj: APIObject) -> APIObject:
...
+                reg_url: {
+                    "username": reg_username,
+                    "password": reg_password,
+                    "auth": base64.b64encode(
+                        f"{reg_username}:{reg_password}".encode()
+                    ).decode(),
+                }
+            }
+        }
+        print(json.dumps(config, indent=4))
```
rasswanth-s: MAYBE:  we could print this only in logger debug mode?

rasswanth-s (approved):
Lets go

packages/syft/src/syft/custom_worker/runner_k8s.py
yashgorana: Ah good catch

packages/syft/src/syft/service/worker/utils.py
yashgorana: +1

PR #8447, opened 2024-02-02
Fix kubernetes cloud deployment issues

yashgorana:
## Description

Closes https://github.com/OpenMined/Heartbeat/issues/964, https://github.com/OpenMined/Heartbeat/issues/966, https://github.com/OpenMined/Heartbeat/issues/971, https://github.com/OpenMined/Heartbeat/issues/968

This change fixes PVC MultiAttach error with the following fixes:
- `credentials-data` volume is now a Secret that is mounted to all the worker pods. This way credentials-data will become `ReadWriteOnce` mounted only by the backend pod.
- `worker-builds` PVC is removed and internal registry is added as a cache. Build always pushes to internal registry. Push copies it from internal to remote registry.

Additionally refactored core kubernetes helper functions 

## Affected Dependencies

## How has this been tested?

## Checklist


packages/syft/src/syft/custom_worker/runner_k8s.py
```diff
@@ -166,67 +147,49 @@ def _create_image_pull_secret(
...
-            objs=_pod.spec.volumes,
-            name="credentials-data",
-        )
-        creds_volume_mount = self._get_obj_from_list(
-            objs=_pod.spec.containers[0].volumeMounts,
-            name="credentials-data",
-        )
-
-        env = _pod.spec.containers[0].env.to_list()
-        env_clone = copy.deepcopy(env)
```
yashgorana, line 218: Before we were copying volumes and mount points over here, but now i've made this function more generic. The responsibility for worker-specific env provisioning is moved to `service/workers/utils.py`. That way we can tweak around a bunch of things at a higher level in the call stack.

packages/syft/src/syft/service/worker/utils.py
```diff
@@ -289,6 +328,15 @@ def create_kubernetes_pool(
                 "CREATE_PRODUCER": "False",
                 "INMEMORY_WORKERS": "False",
             },
+        )
```
yashgorana, line 331: This helps provide the env vars + mount points for the worker.

packages/syft/src/syft/custom_worker/k8s.py
```diff
@@ -1,24 +1,36 @@
...
-
 # Kubernetes namespace
 KUBERNETES_NAMESPACE = os.getenv("K8S_NAMESPACE", "syft")
 
 # Kubernetes runtime flag
 IN_KUBERNETES = os.getenv("CONTAINER_HOST") == "k8s"
 
+# Internal registry URL
+DEFAULT_INTERNAL_REGISTRY = f"registry.{KUBERNETES_NAMESPACE}.svc.cluster.local"
+INTERNAL_REGISTRY_HOST = os.getenv("INTERNAL_REGISTRY_HOST", DEFAULT_INTERNAL_REGISTRY)
```
yashgorana, line 32: Optionally, in dev mode, we can simply set `INTERNAL_REGISTRY_HOST=k3d-registry.localhost:5000`

packages/grid/helm/syft/templates/backend-service-account.yaml
```diff
@@ -2,7 +2,6 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: backend-service-account
-  namespace: {{ .Release.Namespace }}
```
rasswanth-s, line 5: Q: Why did we remove this?

packages/grid/helm/syft/templates/registry-statefulset.yaml
```diff
@@ -0,0 +1,47 @@
...
+      app.kubernetes.io/managed-by: Helm
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: {{ .Chart.Name }}
+        app.kubernetes.io/component: registry
+        app.kubernetes.io/managed-by: Helm
+    spec:
+      containers:
+      - image: registry:2
```
rasswanth-s, line 25: NITPICK: we could give the full identifier `docker.io/library/registry:2`
as with other container runtimes, the default registry assumption might be different.

scripts/build_images.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+docker image build -f ./packages/grid/backend/backend.dockerfile --target backend -t syftpilot.azurecr.io/openmined/grid-backend:$1 ./packages
+docker image build -f ./packages/grid/frontend/frontend.dockerfile --target grid-ui-development -t syftpilot.azurecr.io/openmined/grid-frontend:$1 ./packages/grid/frontend
```
rasswanth-s: NITPICK: could we make the registry also as an input parameter

packages/syft/src/syft/custom_worker/k8s.py
```diff
@@ -82,3 +94,167 @@ def from_status_dict(cls: "PodStatus", status: dict):
...
+        component: str,
+        registries: Iterable[Tuple[str, str, str]],
+    ) -> Secret:
+        auths = {}
+
+        for url, uname, passwd in registries:
+            auths[url] = {
+                "username": uname,
+                "password": passwd,
+                "auth": base64.b64encode(f"{uname}:{passwd}".encode()).decode(),
```
rasswanth-s: NITPICK: We could use here again the above static method
```python3
"auth": KubeUtils.b64encode_secret(f"{uname}:{passwd}")
```

packages/syft/src/syft/custom_worker/runner_k8s.py
```diff
@@ -166,67 +125,49 @@ def _create_image_pull_secret(
...
+                            "secretName": secret_name,
+                        },
+                    }
+                )
+                volume_mounts.append(
+                    {
+                        "name": secret_name,
+                        "mountPath": mount_opts.get("mountPath"),
+                        "subPath": mount_opts.get("subPath"),
+                        "readOnly": True,
```
rasswanth-s, line 168: Good catch , on making it readonly 👍

rasswanth-s (approved):
Stellar Work @yashgorana  💯

packages/syft/src/syft/service/worker/utils.py
```diff
@@ -254,9 +257,46 @@ def run_workers_in_threads(
...
+    creds_path = os.getenv("CREDENTIALS_PATH")
+    if not creds_path:
+        raise ValueError(message="Credentials path not provided")
+
+    creds_path = Path(creds_path)
+    if not creds_path.exists():
+        raise ValueError(message="Credentials file does not exist")
+
+    # create a secret for the node credentials owned by the backend, not the pool.
+    node_secret = KubeUtils.create_secret(
```
rasswanth-s, line 276: We could add a TODO or a github issue, where we are able to remove the credentials.json creation at backend using Init containers which we discussed or any other methods, for later cycles.

packages/grid/helm/syft/templates/backend-service-account.yaml
shubham3121: I think we still need this, we use it for generating helm notes.

packages/syft/src/syft/custom_worker/utils.py
```diff
@@ -0,0 +1,24 @@
+# stdlib
+from typing import Optional
+from typing import Tuple
+
+
+class ImageUtils:
+    @staticmethod
+    def parse_tag(tag: str) -> Tuple[Optional[str], str, str]:
```
shubham3121, line 8: NITPICK: I think we can make them as independent functions, instead of static methods.

packages/grid/helm/syft/templates/backend-service-account.yaml
yashgorana: redundant. the helm deploys all the resources it in the namespace we provide as args. additionally, this was the only resource which had this value set, so just cleaning it up.

packages/grid/helm/syft/templates/registry-statefulset.yaml
yashgorana: not required, defaults to docker anyway.

packages/syft/src/syft/service/worker/utils.py
```diff
@@ -254,9 +257,46 @@ def run_workers_in_threads(
     return results
 
 
+def prepare_worker_pool_env(runner: KubernetesRunner, env_vars: dict):
```
shubham3121: can we rename this specific to k8s, `k8s_prepare_worker_pool_env`

shubham3121 (approved):
looks good, just minor comments.

packages/syft/src/syft/service/worker/utils.py
yashgorana: https://github.com/OpenMined/Heartbeat/issues/975

packages/syft/src/syft/custom_worker/utils.py
yashgorana: This is just a way to group similarly related functions in a namespace in this case "ImageUtils", so i get all the functions in one import. Also keeps the code a bit neater. Is it a much-needed change?

PR #8451, opened 2024-02-05
[Refactor] Fixing mypy issues (01)

khoaguin:


packages/syft/src/syft/client/api.py
```diff
@@ -252,7 +252,7 @@ def prepare_args_and_kwargs(
 
         return args, kwargs
 
-    def __call__(self, *args, **kwargs):
+    def __call__(self, *args: Tuple[Any, ...], **kwargs: Dict[str, Any]) -> Any:
```
kiendang:
```suggestion
    def __call__(self, *args: Any, **kwargs: Any) -> Any:
```

packages/syft/src/syft/client/api.py
```diff
@@ -394,13 +397,15 @@ def generate_remote_lib_function(
...
-):
+) -> Any:
     if "blocking" in signature.parameters:
         raise Exception(
             f"Signature {signature} can't have 'blocking' kwarg because its reserved"
         )
 
-    def wrapper(*args, **kwargs):
+    def wrapper(
+        *args: Tuple[Any, ...], **kwargs: Dict[str, Any]
```
kiendang:
```suggestion
        *args: Any, **kwargs: Any
```

packages/syft/src/syft/serde/recursive.py
```diff
@@ -269,11 +269,11 @@ def rs_proto2object(proto: _DynamicStructBuilder) -> Any:
     # clean this mess, Tudor
     module_parts = proto.fullyQualifiedName.split(".")
     klass = module_parts.pop()
-    class_type: Type = type(None)
+    class_type: Type[Any] = type(None)
```
kiendang: Is this needed? Isn't `Type` is the same as `Type[Any]`?

khoaguin: I think it's mostly the same but `Type[Any]` is more detailed and explicitly states that `class_type` can hold any class

kiendang: For generic type usually `Any` is omitted. For example we usually just write `List`, `Dict` instead of `List[Any]` and `Dict[Any, Any]`. If `Type[Any]` is the same as `Type` we should just use the latter for consistency, unless there is a difference that I'm not aware of.

khoaguin: Yeah sure sounds good. Thanks @kiendang

packages/syft/src/syft/client/api.py
```diff
@@ -961,7 +974,9 @@ def __repr__(self) -> str:
         return f"NodeIdentity <name={self.node_name}, id={self.node_id.short()}, 🔑={str(self.verify_key)[0:8]}>"
 
 
-def validate_callable_args_and_kwargs(args, kwargs, signature: Signature):
+def validate_callable_args_and_kwargs(
+    args: List, kwargs: Dict, signature: Signature
+) -> Union[SyftError, Tuple[List, Dict]]:
```
kiendang:
```suggestion
) -> Union[Tuple[List, Dict], SyftError]:
```

packages/syft/src/syft/util/util.py
```diff
@@ -97,20 +97,23 @@ def get_mb_size(data: Any) -> float:
...
+        try:
+            fqn: str = matches[1]
+            if "." in fqn:
+                return fqn.split(".")[-1]
+            return fqn
+        except Exception as e:
+            print(f"Failed to get klass name {klass}")
+            raise e
+    else:
+        return None
```
kiendang: nitpick: can skip the last else statement in a function, cleaner

```py
def factorial(n):
    if n <= 1:
        return 1
    
    return n * factorial(n - 1)
```

kiendang (approved):
Nice one @khoaguin!

PR #8458, opened 2024-02-06
[Refactor] Fixing mypy issues of `service/warnings.py`

khoaguin:


packages/syft/src/syft/service/warnings.py
```diff
@@ -30,7 +30,7 @@ class APIEndpointWarning(SyftBaseModel):
     message: Optional[str] = None
     enabled: bool = True
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: object) -> bool:
```
kiendang:
```suggestion
    def __eq__(self, other: Any) -> bool:
```

kiendang (approved):

PR #8461, opened 2024-02-06
Scale worker pool in kubernetes

yashgorana:
## Description

Closes https://github.com/OpenMined/Heartbeat/issues/959

## Affected Dependencies

## How has this been tested?

## Checklist


packages/syft/src/syft/custom_worker/runner_k8s.py
```diff
@@ -16,6 +16,8 @@
 from .k8s import get_kr8s_client
 
 JSONPATH_AVAILABLE_REPLICAS = "{.status.availableReplicas}"
+CREATE_POOL_TIMEOUT_SEC = 60
+SCALE_POOL_TIMEOUT_SEC = 60
```
rasswanth-s, line 20: Since during Create or scale, the kubelet pulls image from an external registry, 
If people build from our base image , which is around 2 GB, due to slow network connectivity, it might timeout sometimes.
Shall we set a higher value?

notebooks/api/0.8/11-container-images-k8s.ipynb
```diff
@@ -92,6 +92,16 @@
...
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "de9872be",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "domain_client.api.services.worker_pool.scale(number=2, pool_name=\"default-pool\")"
```
rasswanth-s: MAYBE: Could we also do a scale_down check
and we could also add an additional check to see if the number of SyftWorkers in the worker pool match the ones we up/down scaled.

rasswanth-s (approved):
Well Done @yashgorana

packages/syft/src/syft/custom_worker/runner_k8s.py
yashgorana: These are pool create & scale timouts - not build & push timeouts. That value is in parity with the ones we have set in docker.

Regarding the pull, on cloud this should be faster as they'll be using registry in the same resource group/region.

Again, values can be changed as we see how it see how things go when we deploy them.

PR #8463, opened 2024-02-06
Output policy error handling fix

koenvanderveen:


madhavajay (approved):
Looks reasonable. From what I can tell this is just a message change with a simple property accessor being added to clean up the code.

madhavajay:
I added a review as a paper trail for if I decide to force merge this to work around CI flapping issues.

Issue #6791, opened 2022-09-28
Incorrect typing of `*args` and `**kwargs`


kiendang:
## Description

There are a few places in the codebase where `*args` and `**kwargs` are incorrectly typed.

If you type them like so 

```python
def f(*args: T, **kwargs: T):
```

then `args` will have type `tuple[T, ...]` and `kwargs` `dict[str, T]`.

This means for example if you want `kwargs` to actually have type `dict[str, Any]` you should write `**kwargs: Any` instead of `**kwargs: dict[str, Any]`. The latter will give `kwargs` type `dict[str, dict[str, Any]]` instead. It gives problems later on for example,

```python
def g(x: str) -> None:
    pass


def f(**kwargs: dict[str, Any]) -> None:
    g(kwargs["some_thing"])
```

This doesn't type check because `kwargs['some_thing']` will be typed `dict[str, Any]` instead of `Any`.

```
error: Argument 1 to "g" has incompatible type "Dict[str, Any]"; expected "str"
```

Example of this in the codebase

https://github.com/OpenMined/PySyft/blob/62cd6194984fdb12f8eba242d34935b7bd75323b/packages/hagrid/hagrid/cli.py#L319

## Are you interested in working on this improvement yourself?
- Yes, I am.

## Additional Context
Add any other context or screenshots.

PR #8464, opened 2024-02-07
[Refactor] Fixing mypy issues of `service/dataset`

khoaguin:


packages/syft/src/syft/service/dataset/dataset.py
```diff
@@ -138,7 +138,9 @@ class Asset(SyftObject):
     __repr_attrs__ = ["name", "shape"]
 
     def __init__(
-        self, description: Optional[Union[MarkdownDescription, str]] = "", **data
+        self,
+        description: Optional[Union[MarkdownDescription, str]] = "",
+        **data: Dict[str, Any],
```
kiendang:
```suggestion
        **data: Any,
```

`*args` and `**kwargs` should be typed as `*args: Any` and `**kwargs: Any`, not `*args: Tuple[Any]` and `**kwargs: Dict[str, Any]`. see #6791

packages/syft/src/syft/service/dataset/dataset.py
```diff
@@ -329,7 +331,7 @@ class CreateAsset(SyftObject):
     class Config:
         validate_assignment = True
 
-    def __init__(self, description: Optional[str] = "", **data):
+    def __init__(self, description: Optional[str] = "", **data: Dict[str, Any]) -> None:
```
kiendang:
```suggestion
    def __init__(self, description: Optional[str] = "", **data: Any) -> None:
```

packages/syft/src/syft/service/dataset/dataset.py
```diff
@@ -470,14 +472,16 @@ class Dataset(SyftObject):
     __repr_attrs__ = ["name", "url", "created_at"]
 
     def __init__(
-        self, description: Optional[Union[str, MarkdownDescription]] = "", **data
-    ):
+        self,
+        description: Optional[Union[str, MarkdownDescription]] = "",
+        **data: Dict[str, Any],
```
kiendang:
```suggestion
        **data: Any,
```

kiendang (changes requested):

packages/syft/src/syft/service/dataset/dataset.py
khoaguin: Thanks @kiendang

kiendang (approved):

PR #8466, opened 2024-02-07
[Refactor] Fixing mypy issues of `syft/service/worker`

khoaguin:


packages/syft/src/syft/service/worker/image_registry_service.py
```diff
@@ -62,7 +62,7 @@ def delete(
         self,
         context: AuthedServiceContext,
         uid: UID = None,
-        url: str = None,
+        url: str = "",
```
kiendang: Check whether this should be `Optional[str]` or just `url: str` (without default value).

packages/syft/src/syft/service/worker/utils.py
```diff
@@ -257,20 +272,22 @@ def run_workers_in_threads(
...
+) -> Tuple[List, Dict]:
     # get current backend pod name
     backend_pod_name = os.getenv("K8S_POD_NAME")
     if not backend_pod_name:
-        raise ValueError(message="Pod name not provided in environment variable")
+        raise ValueError("Pod name not provided in environment variable")
 
     # get current backend's credentials path
-    creds_path = os.getenv("CREDENTIALS_PATH")
+    creds_path: Union[str, None, Path] = os.getenv("CREDENTIALS_PATH")
```
kiendang:
```suggestion
    creds_path: Optional[Union[str, Path]] = os.getenv("CREDENTIALS_PATH")
```

a bit clearer imo

packages/syft/src/syft/service/worker/utils.py
```diff
@@ -283,15 +300,15 @@ def prepare_kubernetes_pool_env(runner: KubernetesRunner, env_vars: dict):
 
     # clone and patch backend environment variables
     backend_env = runner.get_pod_env_vars(backend_pod_name) or []
-    env_vars = KubeUtils.patch_env_vars(backend_env, env_vars)
+    env_vars_list = KubeUtils.patch_env_vars(backend_env, env_vars)
```
kiendang: A bit nitpick but when I encounter this I do this instead:

```py
env_vars_ = ... # or _env_vars
env_vars = ...
```

It's clearer that `env_vars_` and `env_vars` mean the same thing. Also I'm not a fan of having type in variable names haha.

packages/syft/src/syft/service/worker/utils.py
```diff
@@ -374,28 +391,33 @@ def run_workers_in_kubernetes(
...
     pool_name: str,
     queue_port: int,
-    start_idx=0,
+    start_idx: int = 0,
     debug: bool = False,
     reg_username: Optional[str] = None,
     reg_password: Optional[str] = None,
     reg_url: Optional[str] = None,
-    **kwargs,
+    **kwargs: Dict[str, Any],
```
kiendang:
```suggestion
    **kwargs: Any,
```

See https://github.com/OpenMined/PySyft/pull/8464#discussion_r1481464164

packages/syft/src/syft/service/worker/image_registry_service.py
shubham3121: This needs to be Optional[str]

packages/syft/src/syft/service/worker/utils.py
khoaguin: yeah that's true

packages/syft/src/syft/service/worker/utils.py
khoaguin: Thanks @kiendang. Also, is the underscore at the end of variable a convention in our codebase?

kiendang: I don't think there's a codebase wide convention more like my own convention, but I certainly did this in various places in the codebase. Anw this is not too important and highly opinionated so do as you see fit.

khoaguin: This is probably the best option `env_vars_: List = ...`

packages/syft/src/syft/service/worker/utils.py
khoaguin: Fixed. Thanks @kiendang

packages/syft/src/syft/service/worker/worker_stash.py
```diff
@@ -59,13 +59,13 @@ def get_worker_by_name(
 
     def update_consumer_state(
         self, credentials: SyftVerifyKey, worker_uid: UID, consumer_state: ConsumerState
-    ):
+    ) -> Result[Ok, Err]:
```
kiendang: `Result[T, E]` is an alias for `Union[Ok[T], Err[E]]`. This returns `Union[Ok[str], Err[str]]` which should be `Result[str, str]`.

packages/syft/src/syft/service/worker/worker_pool_service.py
```diff
@@ -443,7 +443,7 @@ def scale(
         number: int,
         pool_id: Optional[UID] = None,
         pool_name: Optional[str] = None,
-    ):
+    ) -> Union[SyftError, SyftSuccess, List[ContainerSpawnStatus]]:
```
kiendang: We should change the code so that this returns `Union[SyftSuccess, SyftError]`. There is this line `return self.add_workers(...` which should be change to

```py
result = self.add_workers(...
if isinstance(result, SyftError):
    return result
```

packages/syft/src/syft/service/worker/worker_service.py
```diff
@@ -317,6 +317,7 @@ def _stop_worker_container(
         container.stop()
         # Remove the container and its volumes
         _remove_worker_container(container, force=force, v=True)
+        return None
```
kiendang, line 320: hmm is this needed at all? non-returning should already means `return None`. Did mypy give an error/warning?

kiendang: We can keep this though. Explicitly returning `None` is also good. I'm just curious about if mypy raises a warning for this.

packages/syft/src/syft/service/worker/worker_stash.py
khoaguin: thanks @kiendang. This is very informative

packages/syft/src/syft/service/worker/worker_service.py
khoaguin: Yes @kien without `return None` mypy will give `error: Missing return statement  [return]`

kiendang (approved):
Nice work @khoaguin

packages/syft/src/syft/service/worker/worker_pool.py
```diff
@@ -117,14 +117,20 @@ def refresh_status(self) -> None:
...
         self.refresh_status()
+
         if self.image and self.image.image_identifier:
             image_name_with_tag = self.image.image_identifier.full_name_with_tag
         else:
             image_name_with_tag = "In Memory Worker"
+
+        healthcheck: str = ""
+        if self.healthcheck is not None:
+            healthcheck = self.healthcheck.value
```
kiendang:
```py
healthcheck = self.healthcheck.value if self.healthcheck is not None else ""
```

would be a bit cleaner

PR #8467, opened 2024-02-07
[Refactor] Fixing mypy issues for `syft/service/`

khoaguin:
Fix mypy issues for the whole `syft/service` folder


packages/syft/src/syft/service/request/request.py
```diff
@@ -420,7 +421,7 @@ def _repr_html_(self) -> Any:
 
             """
 
-    def _coll_repr_(self):
+    def _coll_repr_(self) -> Dict[str, Union[str, Dict[str, str]]]:
```
kiendang, line 424: should just be
```suggestion
    def _coll_repr_(self) -> Dict[str, str]:
```
right?

khoaguin: No @kiendang since the `"Status"` key of the returned dictionary is `status_badge` which is another dictionary

kiendang: I see.

kiendang (approved):
LGTM

kiendang:
@khoaguin I think can fix the rest of `syft/service` in this PR? To shorten the regex file matcher line also.

khoaguin:
Yes sure @kiendang I think opening less PRs will save resources and mental efforts. But I am afraid that if we fix all in one PR it will make the PR very big, and we also agreed previously that we will fix one folder in one PR, right?
I am thinking about fixing multiple folders in `/service/` in one PR, but once it gets too large I will open another one. What do you think?

khoaguin:
Change to this PR: https://github.com/OpenMined/PySyft/pull/8492

khoaguin:
Change to this PR: https://github.com/OpenMined/PySyft/pull/8492

kiendang:
> But I am afraid that if we fix all in one PR it will make the PR very big
In this case it's fine. The changes are very similar, and I would need to go through the same amount of code to review any way. Having it all in one PR makes it more convenient.

> we also agreed previously that we will fix one folder in one PR, right?
Yup we still do that just with a bigger folder.

PR #8478, opened 2024-02-12
Use pytest<8 to maintain compatibility with pytest-lazy-fixure

kiendang:
pytest>=8.0 broke pytest-lazy-fixture which doesn't seem to be actively maintained
temporarily pin to pytest<8
see pytest-dev/pytest#11890


rasswanth-s (approved):

rasswanth-s:
Well Done @kiendang

PR #8488, opened 2024-02-14
3 -  ADD Notifier Settings for Admin view

IonesioJunior:
## Description
- [x] Add notifier settings for admin view


jcardonnet (approved):
Return type should be something like `Union[NotifierSettingsView, SyftError]` instead of `Union[NotifierStash, SyftError]`, but I'll fix it myself after I create the `NotifierSettingsView` class.

PR #8489, opened 2024-02-14
4 - ADD Notifier turn_on service

IonesioJunior:
## Description
- [x] Enable Notifier turn_on service
- [x] Rename/Refactor Base Notifier Classes


jcardonnet (commented):
Is there anything specific to Postmark that would require it's own class or would be possible to have a generic SMTP notifier class passing the server name on object creation? 

Maybe use a helper method like

def build_postmak_notifier(user, pass, token):
    return SMTPNotifier(user,pass,token,"smtp.postmark.com")

jcardonnet (approved):
LGTM

PR #8490, opened 2024-02-14
5 - ADD Notifier turn off service

IonesioJunior:
## Description
- [x] Enable Notifier Service to be turned off.


jcardonnet (commented):
Should we ensure the stash update was successful/check for errors before returning Syft Success/Error?

IonesioJunior:
> Should we ensure the stash update was successful/check for errors before returning Syft Success/Error?

Yup! Good catch! Fixed it in the last commit! :)

jcardonnet (approved):
LGTM

PR #8497, opened 2024-02-15
ADD Hagrid/K8s email token flag

IonesioJunior:
## Description
- [x] ADD a feature that allows a node to be started with email notifications active using hagrid cli.


jcardonnet (approved):
LGTM

PR #8504, opened 2024-02-16
Integrate Notifier Service with Postmark

IonesioJunior:
## Description
This PR aims to integrate our current notifier and notification service with postmark. Allowing PyGrid Notification events to be sent via email to PyGrid registered users.

### DIFF
--- a/packages/syft/src/syft/service/notifier/smtp_client.py
+++ b/packages/syft/src/syft/service/notifier/smtp_client.py
@@ -1,8 +1,8 @@
 # stdlib
-from typing import List
 from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
 import smtplib
+from typing import List
 from typing import Optional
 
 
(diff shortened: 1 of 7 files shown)


packages/syft/src/syft/service/notifier/smtp_client.py
jcardonnet, line 1: The idea behind token parameter was to allow authentication using an OAuth2 token, which is an alternative to username/password.

Some mail platforms create credentials which they call "token" but it's not a proper token, but a username/password pair in disguise revealed by the fact it needs to be applied to those fields during login.

A token has specific semantics so it's confusing if we use the term interchangeably with user/pwd.

If we don't want to support OAuth we can remove the associated code and keep only the username/password parameters and use them with the proper semantics. 

When the username and password are the same because that's the vendor generates credentials shouldn't be treated as a special case, and values should be assigned `username=<key>, password=<key>`

jcardonnet: I'll approve+merge it as-is and we can discuss the changes later.

PR #8506, opened 2024-02-16
Enable notifications via settings api

IonesioJunior:
Using the new Notifier Service, the node admin must be able to enable/disable the node notifier.

But to achieve that, there are some edge cases to keep in mind.

- [X] By UX/UI reasons, the api used to enable / disable it will be set at client.settints.
  - To enable: `client.settings.enable_notifications(token=<TOKEN>)`
  - To disable: `client.settings.disable_notifications()`
- [X] If email token isn't provided, returns a `SyftError` informing that **email_token** parameter is mandatory. 
- [X] If a regular user tries to use this API, it must receive the regular PySyft non authorized error.
- [X] If admin provides all the informations properly, it will return a `SyftSuccess` message.


jcardonnet (approved):
LGTM

PR #8521, opened 2024-02-21
ADD Activate/Deactivate notifier aliases for the notification service

IonesioJunior:
## Description
The notifier service provides an api to enable / disable notifications to an specific account. But for UX reasons we'd like to have these features available in the notification service.


jcardonnet (approved):
LGTM

PR #8522, opened 2024-02-21
Replace email_token -> username/password credentials

IonesioJunior:
## Description
This PR aims to replace all email_token occurencies both in hagrid and PyGrid,


jcardonnet (approved):
LGTM

PR #8523, opened 2024-02-22
[Refactor] Fixing mypy issues of `syft/`

khoaguin:


packages/syft/src/syft/client/domain_client.py
```diff
@@ -88,28 +91,32 @@ def upload_dataset(self, dataset: CreateDataset) -> Union[SyftSuccess, SyftError
...
-            metadata.show_warnings
-            and metadata.node_side_type == NodeSideType.HIGH_SIDE.value
-        ):
-            message = (
-                "You're approving a request on "
-                f"{metadata.node_side_type} side {metadata.node_type} "
-                "which may host datasets with private information."
-            )
-            prompt_warning_message(message=message, confirm=True)
+        if self.api.connection:
```
shubham3121: We can skip checking for None in self.api.connection since self.api is casted to SyftAPI

khoaguin: But Shubham here `SyftAPI.connection` can still be None, so that's why we check it there

shubham3121 (approved):
Monumental work @khoaguin !!! 🔥 🥳  Much needed heavy lifting !! 🦾 
Finally we can have typing checks back 💯

packages/syft/src/syft/client/domain_client.py
shubham3121: Type casted to NodeConnection: 0596219b8584c888c056f9084e0f590bc1075e7a

PR #8556, opened 2024-03-06
Storage permissions

koenvanderveen:


packages/syft/src/syft/client/api.py
```diff
@@ -862,6 +862,9 @@ def __repr__(self) -> str:
         return _repr_str
 
 
+SyftAPI.model_rebuild()
```
koenvanderveen: why is this needed?

packages/syft/src/syft/client/syncing.py
```diff
@@ -82,114 +78,182 @@ def resolve(
...
-                    else:
-                        print("Found multiple matches for provided id, exiting")
-                        break
-                else:
-                    print("invalid input")
+                )
+
+                remaining_private_high_diffs.remove(diff)
+                private_high_diffs_to_share.append(diff)
+                # new_permissions_lowside.append(
```
koenvanderveen: lets remove these comments

packages/syft/src/syft/protocol/protocol_version.json
```diff
@@ -98,7 +98,7 @@
       "SignedSyftAPICall": {
```
koenvanderveen: a lot of changes here? probably need to reset?

packages/syft/src/syft/service/action/action_object.py
```diff
@@ -1195,11 +1205,13 @@ def wrapper(
 
         return wrapper
 
-    def send(self, client: SyftClient) -> Self:
+    def send(self, client: SyftClient, add_storage_permission: bool = True) -> Self:
```
koenvanderveen: I get that this is here, but at the same time I would have hoped that we could hide this complexity from such an essential user facing api

koenvanderveen: maybe we can add a ._send() which has this argument, but .send() does not?

packages/syft/src/syft/service/action/action_object.py
```diff
@@ -700,11 +706,15 @@ class ActionObjectV2(SyftObject):
 
 
 @serializable(without=["syft_pre_hooks__", "syft_post_hooks__"])
-class ActionObject(SyftObject):
+class ActionObject(SyncableSyftObject):
     """Action object for remote execution."""
 
     __canonical_name__ = "ActionObject"
     __version__ = SYFT_OBJECT_VERSION_3
+    __private_sync_attrs__: ClassVar[dict[str, Any]] = {
```
koenvanderveen: I think this name is slightly confusing, because its actually not the private version of this data. Maybe __private_sync_attr_mocks__?

packages/syft/src/syft/service/code/user_code_service.py
```diff
@@ -302,10 +302,13 @@ def get_results(
...
 
                 if len(output_history) > 0:
-                    return resolve_outputs(
+                    res = resolve_outputs(
                         context=context,
                         output_ids=output_history[-1].output_ids,
                     )
+                    if res.is_err():
+                        return res
+                    return Ok(delist_if_single(res.ok()))
```
koenvanderveen: nitpick: splitting this in 2 lines is maybe more readable

packages/syft/src/syft/service/sync/sync_service.py
```diff
@@ -135,8 +146,22 @@ def add_permissions_for_item(
...
 
+    def add_storage_permissions_for_item(
+        self,
+        context: AuthedServiceContext,
+        item: SyftObject,
+        permissions_other: Set[UID],
+    ) -> None:
+        _id = item.id.id
+        permissions = [
+            StoragePermission(uid=_id, node_uid=p) for p in permissions_other
```
koenvanderveen, line 153: is this correct? we should not always give the same permissions on this side as the other node right?

packages/syft/src/syft/types/syncable_object.py
```diff
@@ -0,0 +1,36 @@
...
+from .syft_object import SyftObject
+
+
+class SyncableSyftObject(SyftObject):
+    __canonical_name__ = "SyncableSyftObject"
+    __version__ = SYFT_OBJECT_VERSION_1
+    # mapping of private attributes and their mock values
+    __private_sync_attrs__: ClassVar[dict[str, any]] = {}
+
+    from_mock_sync: bool = False
```
koenvanderveen: I wonder if this should be on the object or outside

packages/syft/src/syft/service/sync/sync_service.py
```diff
@@ -145,8 +170,14 @@ def set_object(
         if exists:
             res = stash.update(creds, item)
         else:
-            #                 res = stash.delete_by_uid(node.python_node.verify_key, item.id)
-            res = stash.set(creds, item)
+            # If the item is a mock object, do not add storage permissions
+            add_storage_permission = not item.from_mock_sync
```
koenvanderveen: can we not infer this from the storage permissions we are passing?

packages/syft/src/syft/service/sync/sync_service.py
```diff
@@ -159,20 +190,29 @@ def sync_items(
...
         items: List[Union[ActionObject, SyftObject]],
         permissions: Dict[UID, Set[str]],
+        storage_permissions: Dict[UID, Set[UID]],
     ) -> Union[SyftSuccess, SyftError]:
-        permissions = defaultdict(list, permissions)
+        permissions = defaultdict(set, permissions)
+        storage_permissions = defaultdict(set, storage_permissions)
         for item in items:
             other_node_permissions = permissions[item.id.id]
+            other_node_storage_permissions = storage_permissions[item.id.id]
```
koenvanderveen: Not sure if this is correct?

packages/syft/src/syft/service/sync/sync_service.py
```diff
@@ -185,39 +225,32 @@ def sync_items(
     def get_permissions(
```
koenvanderveen, line 218: maybe split in two methods

packages/syft/src/syft/store/mongo_document_store.py
```diff
@@ -293,6 +294,10 @@ def _set(
             if add_permissions is not None:
                 self.add_permissions(add_permissions)
 
+            if add_storage_permission:
+                # TODO: add storage permissions to Mongo store
```
koenvanderveen, line 293: Would be great if we can do this before the weekend, if that doesnt work, lets merge without it and just test with single container stack nodes or dev nodes

packages/syft/tests/syft/service/sync/sync_flow_test.py
```diff
@@ -198,3 +200,168 @@ def compute_mean(data) -> float:
     )
     low_worker.close()
     high_worker.close()
+
+@pytest.mark.skipif(sys.platform == "win32", reason="does not run on windows")
+@pytest.mark.flaky(reruns=5, reruns_delay=1)
```
koenvanderveen, line 207: awesome!

packages/syft/src/syft/service/sync/sync_service.py
koenvanderveen: do this logic client side

packages/syft/src/syft/service/sync/sync_service.py
koenvanderveen: move to client side

packages/syft/src/syft/service/sync/sync_service.py
koenvanderveen: rename and refactor init logic

koenvanderveen: `new_storage_permissions`?

eelcovdw (approved):

PR #8558, opened 2024-03-07
Veilid Chunking

rasswanth-s:
Re-opening @itstauq  veilid chunking PR.

## Description
Veilid can send at most 32 kb of data in one network call. However the requests-responses from PySyft can be larger than than. This PR contains utilities to send and receive larger data by chunking the data into packets of max 32 kb.

### Items to cover in later cuts:

- Timeout the buffers to prevent memory leaks.
- Write unit and integration tests
- Reliability improvements and error recovery (resending failed chunks or handling network interruptions gracefully).
- What happens in case the servers are behind a load balancer? The VeilidStreamer request can land in any instance but the buffer will not be shared among the instances as everything is in-memory. 
    - Does this even needs to be handled?
    - Idea 1: Use sticky connections in the load balancer.
    - Idea 2: Use a pull-based approach for sending the chunks. Sender sends start stream request -> Receiver makes subsequent requests to get all the chunks from the sender.

This PR closes https://github.com/OpenMined/Heartbeat/issues/1076

## Affected Dependencies
None

## How has this been tested?
- [x] Using tests present in `notebooks/Testing/Veilid/Large-Message-Testing.ipynb`

## Checklist


packages/grid/veilid/server/veilid_core.py
```diff
@@ -151,7 +153,11 @@ async def app_call(vld_key: str, message: bytes) -> bytes:
         async with await get_routing_context(conn) as router:
             route = await get_route_from_vld_key(vld_key, conn, router)
 
-            result = await router.app_call(route, message)
+            result = (
+                await router.app_call(route, message)
```
rasswanth-s: I think , we might also need to handle the case, when the request could be less than 32 KB, but the response might be a large value.

packages/grid/veilid/server/veilid_streamer.py
```diff
@@ -0,0 +1,393 @@
...
+        await self._send_response(connection, call_id, ResponseType.OK)
+
+    async def _send_error_response(
+        self, connection: veilid.VeilidAPI, call_id: veilid.OperationId
+    ) -> None:
+        await self._send_response(connection, call_id, ResponseType.ERROR)
+
+    def _calculate_chunks_count(self, message_size: int) -> int:
+        max_chunk_size = self.chunk_size - self.stream_chunk_header_struct.size
+        total_no_of_chunks = message_size // max_chunk_size + 1
```
rasswanth-s: MAYBE: I think this formula could be 
as 
```python3
total_no_of_chunks = math.ceil( message_size / max_chunk_size)
```
as the former formula would fail if the message size is perfectly divisible.

example:

message size = 10
chunk size = 2
total chunks = 5 (the previous formula would return 6)

packages/grid/veilid/server/veilid_streamer.py
```diff
@@ -0,0 +1,393 @@
...
+        if prefix == RequestType.STREAM_START:
+            await self._handle_receive_stream_start(connection, update)
+        elif prefix == RequestType.STREAM_CHUNK:
+            await self._handle_receive_stream_chunk(connection, update)
+        elif prefix == RequestType.STREAM_END:
+            await self._handle_receive_stream_end(connection, router, update, callback)
+        else:
+            logger.error(f"[Bad Message] Message with unknown prefix: {prefix}")
+
+    @retry(veilid.VeilidAPIError, tries=4, delay=1, backoff=2)
```
rasswanth-s, line 294: ❤️

packages/grid/veilid/server/veilid_streamer.py
```diff
@@ -0,0 +1,441 @@
...
+
+logger = logging.getLogger(__name__)
+logger.setLevel(level=logging.INFO)
+
+# An asynchronous callable type hint that takes bytes as input and returns bytes
+AsyncReceiveStreamCallback = Callable[[bytes], Coroutine[Any, Any, bytes]]
+CallId = bytes
+
+
+class RequestType(bytes, ReprEnum):
```
rasswanth-s: Q: What is the use of ReprEnum?

packages/grid/veilid/server/veilid_streamer.py
```diff
@@ -0,0 +1,441 @@
...
+            value = value.ljust(size, b"\x00")
+        self._value_ = value
+
+    def __eq__(self, __other: object) -> bool:
+        return self._value_ == __other
+
+
+class ResponseType(bytes, ReprEnum):
+    OK = b"@VS@OK"
+    ERROR = b"@VS@ER"
```
rasswanth-s, line 52: MAYBE:  we could keep the bytes value: 
```python3
b"@VS@ERR"
```

packages/grid/veilid/server/veilid_streamer.py
itstauq: This will create a sort of `BytesEnum` class, similar to the inbuilt `enum.IntEnum` and `enum.StrEnum`. Helps in validation and allows directly accessing members instead of needing to use `enum_member.value` in all places (otherwise we would need to do `RequestType.STREAM_SINGLE.value` and `RequestType.STREAM_CHUNK.value` instead of `RequestType.STREAM_SINGLE` and `RequestType.STREAM_CHUNK`.

To further improve readability, I'll add a separate `BytesEnum` class and inherit the `RequestType` class from that instead.

packages/grid/veilid/server/veilid_streamer.py
itstauq: I wanted to keep the length same for both the `ResponseType`s

packages/grid/veilid/server/veilid_streamer.py
```diff
@@ -0,0 +1,441 @@
...
+    buffers: dict[CallId, Buffer]
+
+    def __new__(cls) -> "VeilidStreamer":
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+            cls._instance.buffers = {}
+        return cls._instance
+
+    def __init__(self) -> None:
+        self.chunk_size = MAX_MESSAGE_SIZE
```
rasswanth-s: MAYBE: we could rename this variable to 
```python3
self.max_message_size 
```
because in other places we  use chunk_size for formula calculation.

packages/grid/veilid/server/veilid_streamer.py
```diff
@@ -0,0 +1,441 @@
...
+        call_id: bytes,
+        chunk_number: int,
+        message: bytes,
+    ) -> bytes:
+        chunk_header = self.stream_chunk_header_struct.pack(
+            RequestType.STREAM_CHUNK,
+            call_id,
+            chunk_number,
+        )
+        max_actual_message_size = self.chunk_size - self.stream_chunk_header_struct.size
```
rasswanth-s: MAYBE: we could re-name to a common variable name (for max_actual_message_size)
we do the same formula calculation at function `_calculate_chunks_count`
but the variable name is `max_chunk size`

rasswanth-s: Also, we could have the max chunk size calculation as a separate method or present as an instance variable in the class for uniformity?

packages/grid/veilid/server/veilid_streamer.py
rasswanth-s: Ahh ok makes sense

packages/grid/veilid/server/veilid_streamer.py
```diff
@@ -0,0 +1,477 @@
...
+        elif prefix == RequestType.STREAM_CHUNK:
+            await self._handle_receive_stream_chunk(connection, update)
+        elif prefix == RequestType.STREAM_END:
+            await self._handle_receive_stream_end(connection, router, update, callback)
+        else:
+            logger.error(f"[Bad Message] Message with unknown prefix: {prefix}")
+
+    @retry(veilid.VeilidAPIError, tries=4, delay=1, backoff=2)
+    async def _send_request(
+        self, router: veilid.RoutingContext, dht_key: str, request_data: bytes
```
rasswanth-s: MAYBE: Could we re-name in all the places from dht_key to vld_key with the new notation

rasswanth-s: for the streamer part

packages/grid/veilid/server/veilid_streamer.py
itstauq: Sure, that'll make more sense! How about renaming it to `self.max_single_veilid_message_size`?

packages/grid/veilid/server/veilid_streamer.py
itstauq: Adding a `self.max_data_in_each_stream_chunk_request` instance variable

packages/grid/veilid/server/veilid_streamer.py
itstauq: Because I think the former might imply a global maximum message size limit...

packages/grid/veilid/server/veilid_streamer.py
```diff
@@ -0,0 +1,480 @@
...
+            cls._instance = super().__new__(cls)
+            cls._instance.buffers = {}
+        return cls._instance
+
+    def __init__(self) -> None:
+        # Structs for serializing and deserializing metadata as bytes of fixed length
+        # https://docs.python.org/3/library/struct.html#format-characters
+        BYTE_ORDER = "!"  # big-endian is recommended for networks as per IETF RFC 1700
+        REQUEST_TYPE_PREFIX = f"{RequestType.SIZE}s"
+        CALL_ID = "16s"
```
rasswanth-s: MAYBE: On reading the code, since veilid also has a call_id attribute for each app call and we use it in functions like _send_response. To make it easy for new users navigating the code.

We could re-name  our `call_id` to `stream_id` ?

packages/grid/veilid/server/veilid_streamer.py
```diff
@@ -0,0 +1,480 @@
...
+        connection: veilid.VeilidAPI,
+        update: veilid.VeilidUpdate,
+    ) -> None:
+        """Handles receiving STREAM_CHUNK request."""
+        message = update.detail.message
+        chunk_header_len = self.stream_chunk_header_struct.size
+        chunk_header, chunk = message[:chunk_header_len], message[chunk_header_len:]
+        _, call_id, chunk_number = self.stream_chunk_header_struct.unpack(chunk_header)
+        buffer = self.buffers[call_id]
+        buffer.receive_chunk(chunk_number, chunk)
```
rasswanth-s: MAYBE: we could re-name the method as `buffer.add_chunk` ?

packages/grid/veilid/server/veilid_streamer.py
```diff
@@ -0,0 +1,480 @@
...
+    ERROR = b"@VS@ER"
+
+
+class Buffer:
+    def __init__(self, holds_reply: bool = False) -> None:
+        self.msg_hash: bytes
+        self.chunks: list[bytes | None]
+        self.message: asyncio.Future[bytes] = asyncio.Future()
+        self.holds_reply: bool = holds_reply
+        # TODO add mechanism to delete/timeout old buffers
```
rasswanth-s, line 61: MAYBE: we could have another variable like
self.chunks_received : int

which stores the number of chunks received.

This could be incremented in the `receive_chunk` method when we add each chunk.

During handling stream end request before `re-assembling` , we could use this number
to 


```python3
if len(self.chunks) != self.chunks_received
       raise Exception("did not receive all the chunks")
```

until we have Error handling and Re-transmission integrated in Future.

packages/grid/veilid/server/veilid_streamer.py
```diff
@@ -0,0 +1,480 @@
...
+    _instance = None
+    buffers: dict[CallId, Buffer]
+
+    def __new__(cls) -> "VeilidStreamer":
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+            cls._instance.buffers = {}
+        return cls._instance
+
+    def __init__(self) -> None:
```
rasswanth-s, line 193: MAYBE: The __init__ method has a lot of initialization, we could something like this

```python3
def __init__(self) -> None:
        self._init_structs()
        self._init_message_sizes()
        self._init_semaphores()

def _init_structs(self) -> None:
    BYTE_ORDER = "!"  
    REQUEST_TYPE_PREFIX = f"{RequestType.SIZE}s"
    CALL_ID = "16s"
    MESSAGE_HASH = "32s"
    TOTAL_CHUNKS_COUNT = "Q"
    CURRENT_CHUNK_NUMBER = "Q"

    self.stream_start_struct = Struct(
        BYTE_ORDER
        + REQUEST_TYPE_PREFIX
        + CALL_ID
        + MESSAGE_HASH
        + TOTAL_CHUNKS_COUNT
    )
    self.stream_chunk_header_struct = Struct(
        BYTE_ORDER + REQUEST_TYPE_PREFIX + CALL_ID + CURRENT_CHUNK_NUMBER
    )
    self.stream_end_struct = Struct(BYTE_ORDER + REQUEST_TYPE_PREFIX + CALL_ID)
    self.stream_single_struct = Struct(BYTE_ORDER + REQUEST_TYPE_PREFIX + CALL_ID)

def _init_message_sizes(self) -> None:
    self.max_single_veilid_message_size = MAX_SINGLE_VEILID_MESSAGE_SIZE
    self.max_data_in_each_stream_chunk_request = (
        self.max_single_veilid_message_size - self.stream_chunk_header_struct.size
    )
    self.max_data_in_each_stream_single_request = (
        self.max_single_veilid_message_size - self.stream_single_struct.size
    )

def _init_semaphores(self) -> None:
    self._send_request_semaphore = asyncio.Semaphore(MAX_STREAMER_CONCURRENCY)
    self._send_response_semaphore = asyncio.Semaphore(MAX_STREAMER_CONCURRENCY)
```

packages/grid/veilid/server/veilid_streamer.py
```diff
@@ -0,0 +1,480 @@
...
+        message_size = len(message)
+
+        if is_request_stream:
+            # This is a new request stream, so we need to generate a new call_id
+            call_id = uuid.uuid4().bytes
+            # Set up a buffer for holding the reply after the end of this request stream
+            buffer_for_holding_reply = Buffer(holds_reply=True)
+            self.buffers[call_id] = buffer_for_holding_reply
+
+        if message_size <= self.max_data_in_each_stream_single_request:
```
rasswanth-s: We could do a split of the functions to keep it more readable like

```python3
if message_size <= self.max_single_chunk_size:
    await self._stream_single_chunk_request(router, vld_key, message, stream_id)
else:
    await self._stream_multi_chunk_request(router, vld_key, message, stream_id)

if is_request_stream:
    await self._wait_for_reply(stream_id)


```

packages/grid/veilid/server/veilid_streamer.py
rasswanth-s: Keep variable name more descriptive is great.but we could try shortening like 
maybe
1. max_stream_single_msg_size
2. max_stream_chunk_msg_size

packages/grid/veilid/server/veilid_streamer.py
```diff
@@ -0,0 +1,480 @@
...
+    async def _send_request(
+        self, router: veilid.RoutingContext, vld_key: str, request_data: bytes
+    ) -> None:
+        """Send an app call to the Veilid server and return the response."""
+        async with self._send_request_semaphore:
+            response = await router.app_call(vld_key, request_data)
+            if response != ResponseType.OK:
+                raise Exception("Unexpected response from server")
+
+    async def _send_response(
```
rasswanth-s, line 304: MAYBE: we could also add retry logic to responses?

packages/grid/veilid/server/veilid_streamer.py
itstauq: This was super needed 🙌

packages/grid/veilid/server/veilid_streamer.py
itstauq: It does makes sense to add error-handling for empty chunks 💯 
However, since receiving chunks is a highly parallel operation, adding an incrementing integer will introduce the need to add locks, adding synchronization overhead and potentially slowing down the whole process. Instead, we can implement a check for empty chunks while handling the stream end request before re-assembling. How does that sound?

packages/grid/veilid/server/veilid_streamer.py
itstauq: Retrying a failed `app_call_reply` makes Veilid throw some operation id mismatch errors. We need to retry the original request in that case.

rasswanth-s: Ahh ok 👍 .We could file a bug with veilid team on the same later?

packages/grid/veilid/server/veilid_streamer.py
rasswanth-s: Good catch, we could do it at the stream end 💯

packages/grid/veilid/server/veilid_streamer.py
itstauq: Yep 💯

itstauq (approved):

rasswanth-s:
Stellar Work @itstauq  ⭐ 
It's great to Chunking working with great reliability
test

PR #8560, opened 2024-03-09
Suppress capnp path warning upon initially importing syft

iamtrask:
## Description
When first importing syft in a fresh Jupyter Notebook, sometimes the capnp package's underlying C++ code appears to be raising a warning about path inconsistencies. This creates a somewhat flaky and unsettling feeling when first using PySyft. The warning looks as follows:

Screenshot 2024-03-09 at 00 21 39

With this PR, this warning is suppressed, such that importing syft looks like:

Screenshot 2024-03-09 at 00 22 15

## Affected Dependencies
Adds a dependency on a python package called hide_warnings.

## How has this been tested?
- Tested on my laptop's jupyter notebook installation
- install this version of syft, open a jupyter notebook, and import syft.

## Checklist


iamtrask:
I recognize 0.8.5 is about to go out the door. Not necessarily a hard requirement but given it's a small PR, cosmetically it feels like a win for UX. I defer to the judgement of the team.

kiendang:
Thanks @iamtrask this warning must have driven everyone crazy. After a bit of investigation I do agree this is the only solution that works. Some context for anyone else looking into this, since the warning is raised by a C/C++ library, changing `sys.stdout`, `sys.stderr` or using `contextlib.redirect_stdout`, `contextlib.redirect_stderr` doesn't work. I also tried using the [`importlib.resources.files`](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#accessing-data-files-at-runtime) api to load the `capnp` files instead to bypass the issue but this also does not work due to `capnp.load` quirkiness.

However there is still an issue with this solution. The `hide-warnings` library depends on [`py`](https://py.readthedocs.io/en/latest/io.html), which has been deprecated for a while and raises a security alert as you can see in the CI. `py` is also a heavy dependency to add to `syft` just to suppress a warning (it used to be `pytest` backend).

Nevertheless suppressing this warning is definitely something we need to do. I think the right approach is just to port/copy `hide-warnings` and the `py` code it depends on here. This is not a lot of code (only 2 short files [1](https://github.com/Rainbow-Dreamer/hide_warnings/blob/main/hide_warnings.py) [2](https://github.com/pytest-dev/py/blob/master/py/_io/capture.py)).

If you think this makes sense I can try to resolve this during some down time in the 0.8.6 cycle.

iamtrask:
@kiendang thanks for taking all the time to investigate! I defer to the wisdom of the group on timing and prioritization. Thanks for finding the root security issue in there. I did notice the failing test and wasn't sure what that was all about. Sounds like you've got a workable plan forward! 🙏

khoaguin:
@iamtrask @kiendang I worked on Kien's proposed solution in [this PR](https://github.com/OpenMined/PySyft/pull/8563), and seems like it's working 
image

kiendang:
superseded by #8563

PR #8561, opened 2024-03-09
Import orchestra directly in syft/__init__.py so that tab completion works

iamtrask:
## Description
New users (and even some old users) learn about frameworks by navigating tab completion. However, one of our key noob-friendly features (orchestra) hasn't been added in a way that tab completion works. This PR tries to fix that.

## Affected Dependencies
n/a

## How has this been tested?
n/a

## Checklist
- [x] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [x] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [x] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [x] My changes are covered by tests

PR #8562, opened 2024-03-10
Improve the launch UX for domains.

iamtrask:
## Description
In our test notebooks, our orchestra-based launch APIs are a bit verbose and could use some cleanup. In this PR, I've cleaned up some of the dev_mode=False logging, and added a sytactic sugar method in syft/__init__.py for launch() that's designed for first-time-users. I didn't change the normal orchestra API so our existing notebooks and stuff should still work fine. Subtasks include:

- Suppress the capnp warning for RoleService
- add verbose=True flags to login and login_as_guest, default to True
- add a DomainClient.me.set_name() method
- change the launch/land terminology to always use launch/land (it used to be a mix of start/stop/launch/kill/etc.)
- for dev_mode=False, I changed the launch() logging to only use one line.
- create a syft.launch() method which asks for name, email, and password before launching with dev_mode=False.
- setup a couple of default SMPT servers for the syft.launch() feature. Meant mostly for first-time users.

Overall this changes login UX from:

Screenshot 2024-03-09 at 22 33 03

to

Screenshot 2024-03-09 at 22 33 25

## Affected Dependencies
This PR relies upon the following other PRs being merged first:
- https://github.com/OpenMined/PySyft/pull/8560
- https://github.com/OpenMined/PySyft/pull/8561

## How has this been tested?
- Tested on my laptop in my jupyter notebook.

## Checklist


madhavajay:
@iamtrask nice work finally hiding the C++ bug.
```python
# from the library source
      capture = py.io.StdCaptureFD(out=out, in_=in_)
```

That explains why no other way we tried to catch those warnings worked.

Regarding the test email, thats a pretty handy way to do it, would this best be in `dev` mode only? Also we might want to check to see what happens if auth fails if that account gets blocked to make sure it doesnt stop people running notebook examples down the track.

iamtrask:
Great question. On the email thing. I tried to setup an OpenMined-paid twilio or something but ran into mysterious issues around authenticating my account (sent an email to support — we'll see). If it wasn't crazy expensive, I'm not totally against taking the risk and putting a rate-limited private key in there (maybe 1000 emails a day?... we could increase its limits as people are using it until it gets unreasonable) and recommending it for use by first-time users. 

Longer term, it seems like people can setup a new email account for popular email providers (gmail, outlook, etc.), put in the credentails to PySyft, and have that email forward to their normal email account. This feels like the real solution long-term because it also means that if a user replies to the email they end up talking to the domain owner (And it means they can skip setting up twilio). (I think Ionesio might have had this kind of idea in mind in his video as well). Anyway, from a product perspective, it seems desirable for people to have a free option that just involves entering email credentails for a forwarding account. Feels accessible and secure. (note: strictly speaking I think this is a better security practice to setup a forwarder email than to use one's email directly — although that seems difficult for us to enforce. we could have the dummy/test email it sends upon Domain Setup specifically tell them to setup a forwarder i suppose.)

Anyway - main thing I'm hoping for here is a near-turn-key option for both dev and non-dev mode. I had such a big trouble setting up a twilio (or similar) platform, seems ideal if we can help users skip that step with good defaults, and an easy way to use their own email (by setting up a forwarder)

madhavajay:
@iamtrask we have some paid testing accounts if you reach out to @IonesioJunior he can show you how to use them.

PR #8563, opened 2024-03-10
[UX] Suppress `capnp` path warning upon initially importing syft

khoaguin:
A possible solution for https://github.com/OpenMined/PySyft/pull/8560

### DIFF
--- a/packages/syft/src/syft/serde/capnp.py
+++ b/packages/syft/src/syft/serde/capnp.py
@@ -6,23 +6,13 @@ from pathlib import Path
 import capnp
 
 # relative
-from ..util.hide_warnings import hide_warnings
-
-
-@hide_warnings
-def load_capnp(capnp_path: str) -> type:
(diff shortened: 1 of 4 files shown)


iamtrask (approved):

iamtrask (commented):
thanks for taking the time to put this together!

kiendang (approved):
Thanks @khoaguin I made a few changes

Let's not copy code from `hide-warnings` here. The package is LGPL-licensed. Besides we don't need it for this. We can just use `StdCaptureFD` directly.

Use a context manager instead of a function decorator. It's simpler this way.

Put the code in `syft.util._std_stream_capture` to indicate that this is internal code not to be used by users.

madhavajay:
This has been pretty high on my personal "unsolveable annoying problems" list! @kiendang and @khoaguin great work getting it fixed. 🙏

PR #8572, opened 2024-03-13
speed up syft unit tests

abyesilyurt:
## Description

I did some profiling on the unit tests and cleared up some bottlenecks.

There is roughly 35% speed up.

Local:
- before: 2 min.
- after: 1 min.

CI:
- before: 6 min.
- after: 4 min.

## Affected Dependencies
~~Added orjson library for faster json processing.~~ Hashing relies on the json file formatting and orjson is missing white spacing which is not configurable. So this change was reverted although there still some potential gains to be made.

## How has this been tested?
- unit tess

## Checklist


packages/syft/src/syft/service/user/user.py
```diff
@@ -106,6 +108,7 @@ def salt_and_hash_password(password: str, rounds: int) -> tuple[str, str]:
     return salt_bytes, hashed_bytes
 
 
+@cache
```
eelcovdw: Is the caching only to speed up CI? Not sure we want to do the caching with real passwords

abyesilyurt: moved caching to conftest

packages/syft/tests/conftest.py
```diff
@@ -204,3 +208,9 @@ def mongo_client(testrun_uid):
     "tests.syft.action_graph.fixtures",
     "tests.syft.serde.fixtures",
 ]
+# patching compute heavy hashing to speed up tests
+SyftClientSessionCache._get_key = (
+    lambda email, password, connection: f"{email}{password}{connection}"
+)
+user.salt_and_hash_password = cache(user.salt_and_hash_password)
+user.check_pwd = cache(user.check_pwd)
```
yashgorana: This should be provisioned through `monkeypatch.setattr` in the fixture instead of making it a global patch?

abyesilyurt: Good point! I applied the changes.

eelcovdw (approved):
Looks good!

PR #8585, opened 2024-03-14
SQLite WAL + store cleanups + fix some unit test fixes

yashgorana:
## Description

- SQLite now operates in WAL mode
- SQLite partitions no longer need a lock (experimental)
- Each worker will now work in it's own temp dir workspace
- Pytest finalizers + cleanup for fixtures like worker and stores

## Affected Dependencies

## How has this been tested?

## Checklist


packages/syft/src/syft/node/node.py
```diff
@@ -611,60 +621,10 @@ def named(
         migrate: bool = False,
         in_memory_workers: bool = True,
     ) -> Self:
+        uid = UID.with_seed(name)
         name_hash = hashlib.sha256(name.encode("utf8")).digest()
-        name_hash_uuid = name_hash[0:16]
```
koenvanderveen, line 615: this is much better

packages/syft/src/syft/node/node.py
```diff
@@ -879,22 +840,24 @@ def init_stores(
         self,
         document_store_config: StoreConfig | None = None,
         action_store_config: StoreConfig | None = None,
+        sqlite_path: Path | str | None = None,
     ) -> None:
+        # if there's no sqlite path, we'll use the tmp dir
+        if not sqlite_path:
```
koenvanderveen: I wonder, do we even need sqlite_path in the first place. It feels weird to me that we have it both in the StoreConfig but also as a separate argument

packages/syft/src/syft/node/node.py
```diff
@@ -1037,6 +996,24 @@ def _get_service_method_from_path(self, path: str) -> Callable:
 
         return getattr(service_obj, method_name)
 
+    def get_temp_dir(self, dir_name: str = "") -> Path:
```
koenvanderveen, line 985: great

packages/syft/src/syft/store/sqlite_document_store.py
```diff
@@ -138,7 +134,7 @@ def _connect(self) -> None:
 
     def create_table(self) -> None:
         try:
-            with SyftLock(self.lock_config):
+            with self.lock:
```
koenvanderveen, line 139: if this becomes basically a passthrough, shouldnt we add a busy_timeout then? What would prevent concurrent writes from failing in this setup?

packages/syft/tests/conftest.py
```diff
@@ -91,16 +118,16 @@ def stage_protocol(protocol_file: Path):
...
+@pytest.fixture(scope="session")
 def faker():
     return Faker()
 
 
 @pytest.fixture()
-def worker(faker) -> Worker:
-    worker = sy.Worker.named(name=faker.name())
+def worker() -> Worker:
+    worker = sy.Worker.named(name=token_hex(8))
```
koenvanderveen, line 128: do we want to reset here just in case?

packages/syft/tests/syft/worker_test.py
```diff
@@ -222,15 +220,25 @@ def post_add(context: Any, name: str, new_result: Any) -> Any:
...
+def test_worker_serde(worker) -> None:
     ser = sy.serialize(worker, to_bytes=True)
     de = sy.deserialize(ser, from_bytes=True)
 
     assert de.signing_key == worker.signing_key
     assert de.id == worker.id
 
 
+@pytest.fixture(params=[0])
+def worker_with_proc(request):
```
koenvanderveen, line 232: how is this different from the other worker fixture?

koenvanderveen (approved):

koenvanderveen:
great PR :heart:

packages/syft/src/syft/store/sqlite_document_store.py
yashgorana: I think busy_timeout has a default of 30000ms, so shouldn't be an issue. I'll double check.

packages/syft/tests/conftest.py
yashgorana: Sure

packages/syft/tests/syft/worker_test.py
yashgorana: This is a parameterized worker fixture that is made to configure the process count, only required for worker_test. I just refactored whatever was there before

packages/syft/src/syft/node/node.py
yashgorana: Sorry didn't get you. Do you prefer the UID.with_seed()? or do you want me to revert to this name_hash[0:16] method?

packages/syft/src/syft/node/node.py
yashgorana: Yes it's actually used to place the database in a separate dir `$HOME/data/db` in `single_container_mode` so that we can mount volume on that dir.

packages/syft/src/syft/node/node.py
koenvanderveen: no this whole node.named method is much better no you refactored it is what I meant

packages/syft/src/syft/node/node.py
koenvanderveen: yes, but do we still need it as this level, can we not sort out that logic earlier and then just put it in the storconfig earlier? Because what would happen now if you would have both for instance? If we would have that logic higher in the chain of calls it might be easier to separate deployment logic from the logic in this node class

yashgorana: agreed - i'll clean this up

packages/syft/src/syft/store/sqlite_document_store.py
yashgorana: Added busy_timeout = 5s

packages/syft/src/syft/node/node.py
yashgorana: So i made the following changes:
- Completely removed this `sqlite_path`
- If you haven't provided either of action stores, it defaults to DictStoreConfig(). If `local_db=True` then uses a default SqliteStoreConfig
- Cleaned up init_store.
- Added default filename to SqliteStoreClientConfig to fix nonetype exceptions in single container mode

packages/syft/src/syft/node/node.py
```diff
@@ -877,51 +851,22 @@ def reload_user_code() -> None:
 
     def init_stores(
         self,
-        document_store_config: StoreConfig | None = None,
-        action_store_config: StoreConfig | None = None,
+        document_store_config: StoreConfig,
+        action_store_config: StoreConfig,
     ) -> None:
```
koenvanderveen, line 856: great that this is so much cleaner

PR #8598, opened 2024-03-18
Improve helm secrets template 

yashgorana:
## Description

Part of https://github.com/OpenMined/Heartbeat/issues/1113

- Helm chart no longer includes dev credentials anymore for deployment safety.
- `global.useDefaultSecrets` is changed to `global.randomizedSecrets`.
- `common.secrets.set` is should be a bit more clear when dealing with using generated passwords vs. default static passwords
- All dev k8s deployments will use `packages/grid/helm/values.dev.yaml` which includes the dev secrets

## Affected Dependencies

## How has this been tested?

## Checklist


packages/grid/helm/syft/values.yaml
```diff
@@ -24,9 +24,9 @@ mongo:
   # Mongo secret name. Override this if you want to use a self-managed secret.
   secretKeyName: mongo-secret
 
-  # Dev mode default passwords
-  defaultSecret:
-    rootPassword: example
+  # custom secret values
+  secret:
+    rootPassword: null
```
yashgorana, line 29: Charts will no longer ship with a dev default value. It must be explicitly provided by the cluster owner if `randomizedSecrets=false` else helm will error out because of [this line](https://github.com/OpenMined/PySyft/pull/8598/files#diff-7b312d1bb67bfa5583d9e85276c7b51439bf9f0e19e4f7e4b69f38d53759e6f1R57)

tox.ini
```diff
@@ -811,59 +809,38 @@ setenv =
...
+    bash -c './scripts/wait_for.sh service proxy --context k3d-$CLUSTER_NAME --namespace syft'
+    bash -c '(kubectl logs service/backend --context k3d-$CLUSTER_NAME --namespace syft -f &) | grep -q "Application startup complete" || true'
 
     # Run Notebook tests
     tox -e e2e.test.notebook
 
-    # Cleanup
-    bash -c "k3d cluster delete syft || true"
-    bash -c "docker volume rm k3d-syft-images --force || true"
+    bash -c "k3d cluster delete ${CLUSTER_NAME} || true"
```
rasswanth-s, line 843: MAYBE: we could use tox -e dev.k8s.destroy itself here?

packages/grid/helm/syft/templates/_secrets.tpl
```diff
@@ -24,17 +24,24 @@ Params:
...
+
+  Use a static default value (with random disabled)
+  {{- include "common.secrets.set " (dict "secret" "some-secret-name" "default" "default-value" "randomDefault" false "context" $ ) }}
 
 Params:
   secret - String (Required) - Name of the 'Secret' resource where the key is stored.
   key - String - (Required) - Name of the key in the secret.
-  default - String - (Optional) - Default value to use if the secret does not exist.
-  length - Int - (Optional) - The length of the generated secret. Default is 32.
+  randomDefault - Bool - (Optional) - If true, a random value will be generated if secret does note exit.
```
rasswanth-s, line 42: MAYBE: we could keep useDefault to make it very intuitive or any other name, as randomDefault is not very intuitive

packages/grid/helm/syft/templates/backend/backend-secret.yaml
```diff
@@ -11,6 +11,7 @@ data:
   defaultRootPassword: {{ include "common.secrets.set" (dict
     "secret" $secretName
     "key" "defaultRootPassword"
-    "default" .Values.node.defaultSecret.defaultRootPassword
+    "randomDefault" .Values.global.randomizedSecrets
```
rasswanth-s, line 14: MAYBE: it would be good to setup application level secret randomization
instead of global.randomizedSecrets.

it could be 
.Values.global.node.randomizedSecrets


Where for example if users , need to set a custom secret only  for syft container, they  could do it easily.

rasswanth-s: This could be done similarly for  mongo, seaweedfs

tox.ini
yashgorana: Destroy calls cleanup. Cleanup = remove images, deployment and namespace. Destroy = destroy the whole k3d cluster.

packages/grid/helm/syft/templates/backend/backend-secret.yaml
yashgorana: Hmm ... looks like it didn't get pushed

yashgorana: > Where for example if users , need to set a custom secret only for syft container, they could do it easily.

for that they can use what you had implemented, provide a custom Secret and use it through `values.backend.secretKeyName`.

rasswanth-s (approved):
Well Done @yashgorana

PR #8605, opened 2024-03-20
Use `uv` in backend container

yashgorana:
## Description

- Use uv as package manager
- Cleaned up dockerfile & dockerignores

## Affected Dependencies

## How has this been tested?

## Checklist


yashgorana:
Haven't tested this rigorously so please don't merge

packages/grid/backend/backend.dockerfile
```diff
@@ -1,94 +1,43 @@
...
-# copy skeleton to do package install
-COPY --chown=$USER_GRP \
-    syft/setup.py \
-    syft/setup.cfg \
-    syft/pyproject.toml \
-    syft/MANIFEST.in \
-    syft/
+RUN --mount=type=cache,target=/root/.cache,sharing=locked \
+    uv venv && \
+    uv pip install torch==2.2.1+cpu --index-url https://download.pytorch.org/whl/cpu
```
yashgorana: `uv` needs this identifier explicitly - https://github.com/astral-sh/uv/blob/main/PIP_COMPATIBILITY.md#local-version-identifiers. They are working to improve this eventually https://github.com/astral-sh/uv/issues/2541

packages/grid/backend/backend.dockerfile
```diff
@@ -1,94 +1,43 @@
...
-    syft/src/syft/
+RUN --mount=type=cache,target=/root/.cache,sharing=locked \
+    uv pip install jupyterlab==4.1.5
 
-# Install all dependencies together here to avoid any version conflicts across pkgs
-RUN --mount=type=cache,id=pip-$UID,target=$HOME/.cache/pip,uid=$UID,gid=$UID,sharing=locked \
-    pip install --user --default-timeout=300 torch==2.2.1 -f https://download.pytorch.org/whl/cpu/torch_stable.html && \
-    pip install --user pip-autoremove jupyterlab -e ./syft[data_science] && \
-    pip-autoremove ansible ansible-core -y
+COPY --chown=nonroot:nonroot \
```
rasswanth-s: Q: If possible could you detail, why we switched from USER_GRP to nonroot:nonroot?

yashgorana: Draft PR - has a bunch of leftover stuff. Please ignore for now. I'm dropping rootless support temporarily else it creates a new layer of permission issues.

packages/grid/backend/backend.dockerfile
```diff
@@ -1,94 +1,43 @@
...
-# uncomment for creating rootless user
-# && adduser -D -u $UID $USER
+FROM cgr.dev/chainguard/wolfi-base as backend
 
-# ==================== [BUILD STEP] Install Syft Dependency ==================== #
+ARG PYTHON_VERSION="3.12"
 
-FROM python_dev as syft_deps
+RUN apk update && apk upgrade && \
+    apk add git bash python-$PYTHON_VERSION-default uv=0.1.22-r0
```
rasswanth-s: MAYBE: we could make the uv version as an  arg variable like python version.?
recently uv release 0.1.23 , we could upgrade to that, or does it work only in this specific version?

packages/grid/backend/grid/start.sh
```diff
@@ -1,43 +1,35 @@
...
-# For debugging permissions
-ls -lisa $HOME/data
-ls -lisa $APPDIR/syft/
-ls -lisa $APPDIR/grid/
-
 if [[ ${DEV_MODE} == "True" ]];
 then
-    echo "DEV_MODE Enabled"
     RELOAD="--reload"
-    pip install --user -e "$APPDIR/syft[telemetry,data_science]"
```
rasswanth-s: Perfect

packages/grid/backend/grid/start.sh
```diff
@@ -1,43 +1,35 @@
...
 DEBUG_CMD=""
 
-# For debugging permissions
-ls -lisa $HOME/data
-ls -lisa $APPDIR/syft/
-ls -lisa $APPDIR/grid/
-
 if [[ ${DEV_MODE} == "True" ]];
 then
-    echo "DEV_MODE Enabled"
```
rasswanth-s, line 24: We could keep this echo statement

packages/grid/backend/backend.dockerfile
rasswanth-s: In my mac, this command failed
Replacing and 
Executing
```sh
uv pip install torch==2.2.1 --index-url https://download.pytorch.org/whl/cpu
```
made it work.

I think also, we could remove the +cpu , as we are directly indexing from a cpu URL?

I also after installing by the command executed

```python3
import torch
torch.cuda.is_available() -> return False
```
confirming that it is a cpu version.

packages/grid/backend/backend.dockerfile
```diff
@@ -1,94 +1,43 @@
...
+RUN --mount=type=cache,target=/root/.cache,sharing=locked \
+    uv venv && \
+    uv pip install torch==2.2.1+cpu --index-url https://download.pytorch.org/whl/cpu
 
-COPY --chown=$USER_GRP \
-    syft/src/syft/VERSION \
-    syft/src/syft/capnp \
-    syft/src/syft/
+RUN --mount=type=cache,target=/root/.cache,sharing=locked \
+    uv pip install jupyterlab==4.1.5
```
rasswanth-s: jupyterlab or plain jupyter depends on psutil.

and Currently psutil does not support aarch64 wheels.

This would make it fail on mac silicon chips, if python3-dev is not installed.

The error message, which I received during building.
Screenshot 2024-03-21 at 1 29 39 PM

packages/grid/backend/backend.dockerfile
yashgorana: Apparently that doesn't work on linux :|

packages/grid/backend/backend.dockerfile
yashgorana: Ah yeah i figured this was going to break - nevermind i have the python-dev way implemented as well, i'll just update the file to use it

packages/grid/backend/backend.dockerfile
yashgorana: fixed

packages/grid/backend/backend.dockerfile
yashgorana: fixed

packages/grid/backend/grid/start.sh
yashgorana: fixed

packages/grid/backend/backend.dockerfile
yashgorana: fixed

packages/grid/backend/backend.dockerfile
yashgorana: fixed

tox.ini
```diff
@@ -886,7 +886,8 @@ commands =
     bash -c 'k3d --version'
 
     ; create registry
-    bash -c 'k3d registry create registry.localhost --port 5800 -v $HOME/.k3d-registry:/var/lib/registry || true'
```
yashgorana, line 971: local path `$HOME/.k3d-registry` gets mounted on the docker macos VM through virtiofs. Might be a reason why docker push gets stuck at some point. For now just use docker-native volume to make sure there's no virtiofs involved happening during a push.

packages/grid/backend/backend.dockerfile
```diff
@@ -1,94 +1,70 @@
...
-    pip-autoremove ansible ansible-core -y
+RUN apk update && apk upgrade && \
+    apk add build-base gcc python-$PYTHON_VERSION-dev-default uv=$UV_VERSION
+
+WORKDIR /root/app
+
+ENV UV_HTTP_TIMEOUT=600
+
+# keep static deps separate to have each layer cached independently
+# if amd64 then we need to append +cpu to the torch version
```
madhavajay, line 22: I wish PyTorch would fix this somehow.

yashgorana: Yeah that --index-url for cpu installs is annoying because setup.cfg doesn't support it - else things would've been easier.

Plus uv also has this issue with +cpu monikers at the moment. I've mentioned it here - https://github.com/astral-sh/uv/issues/1497#issuecomment-2098896853 & they're looking into it https://github.com/astral-sh/uv/issues/3437.

packages/grid/devspace.yaml
```diff
@@ -36,7 +36,7 @@ images:
   backend:
     image: "${CONTAINER_REGISTRY}/${DOCKER_IMAGE_BACKEND}"
     buildKit:
-      args: ["--platform", "linux/${PLATFORM}"]
+      args: ["--target", "backend", "--platform", "linux/${PLATFORM}"]
     dockerfile: ./backend/backend.dockerfile
     target: "backend"
```
rasswanth-s: Q: We already have the target as backend here, why do we also include it in the args above?

packages/grid/backend/backend.dockerfile.dockerignore
```diff
@@ -0,0 +1,67 @@
+# Syft
+tests/
+*.md
```
rasswanth-s: I think this would be omitted as docker always looks for the .dockerignore in the build context directory

In the case of backend container, the build context is: `packages`
we could place it under packages folder

packages/grid/frontend/frontend.dockerfile.dockerignore
```diff
@@ -0,0 +1,10 @@
+# Frontend
```
rasswanth-s, line 3: MAYBE: we could keep the file names short like
frontend.dockerignore?

rasswanth-s: similar for backend and frontend

yashgorana: That's how docker wants it - https://docs.docker.com/build/building/context/#filename-and-location

Or we move it under packages and combine all ignores into one - which might create conflicts in future. So better to keep it separate and explicit.

packages/grid/backend/backend.dockerfile.dockerignore
yashgorana: See https://github.com/OpenMined/PySyft/pull/8605#discussion_r1597470335

packages/grid/devspace.yaml
yashgorana: Keeping it explicit and in-parity with cd-syft-*.yaml

```yaml

      - name: Build and push `grid-backend` image to DockerHub
        id: grid-backend-build
        uses: docker/build-push-action@v5
        with:
          context: ./packages
          file: ./packages/grid/backend/backend.dockerfile
          platforms: ${{ steps.release_metadata.outputs.release_platform }}
          target: backend # <---

```

rasswanth-s: Ahh ok, but why do we repeat in args and as well as in the target in devspace.yaml?

packages/grid/frontend/frontend.dockerfile.dockerignore
rasswanth-s: Sure the naming makes sense 👍

packages/grid/backend/backend.dockerfile.dockerignore
rasswanth-s: Correct, referring to the same document, the .dockerignore file is looked for in the root directory of context, in the case of backend build, since the context in packages directory, I think it should be present inside packages directory instead of packges/grid/backend

rasswanth-s: Screenshot 2024-05-14 at 12 39 35 PM

packages/grid/backend/grid/start.sh
```diff
@@ -28,15 +21,17 @@ fi
 # only set by kubernetes to avoid conflict with docker tests
 if [[ ${DEBUGGER_ENABLED} == "True" ]];
 then
-    pip install --user debugpy
+    uv pip install debugpy
     DEBUG_CMD="python -m debugpy --listen 0.0.0.0:5678 -m"
 fi
 
-set +e
+source $APPDIR/.venv/bin/activate
```
rasswanth-s: Q: Should the venv be activate before installation of debupy, as it would not be available for later commands?

packages/grid/backend/backend.dockerfile
```diff
@@ -1,94 +1,70 @@
...
+    uv pip install torch==$TORCH_VERSION --index-url https://download.pytorch.org/whl/cpu
+
+COPY syft/setup.py syft/setup.cfg syft/pyproject.toml ./syft/
+
+COPY syft/src/syft/VERSION ./syft/src/syft/
+
+RUN --mount=type=cache,target=/root/.cache,sharing=locked \
+    # remove torch because we already have the cpu version pre-installed
+    sed --in-place /torch==/d ./syft/setup.cfg && \
+    uv pip install -e ./syft[data_science,telemetry] && \
```
rasswanth-s: I think we could omit telemetry for the default docker build.

Currently, we mainly use telemetry for dev purposes (The telemetry in the codebase is also old, not sure if is functional)

maybe we could be keep it as part of start.sh, where we have a new env variable called TELEMETRY_ENABLED in backen-statefulset instead of having it in devmode.


```sh
if [[ ${TELEMETRY_ENABLED} == "True" ]];
then
    uv pip install ./syft[telemetry]
    
fi
```
Thoughts ?

rasswanth-s:
Great work , I tested the builds 

on DEV branch without Cache:
Took 211 s

on this branch without Cache:
took 115 s

Screenshot 2024-05-14 at 12 56 38 PM
That's a 2x a speed up great work 


mcqueen

packages/grid/backend/backend.dockerfile
yashgorana: Let's avoid runtime pip installs because it creates race conditions that we miss out on - in this case a delayed DEV_MODE pip install resulted in seaweedfs race that got caught because of fast container startup time.

And yes, telemetry is old and we should definitely relook into it later. Happy to remove telemetry for now.

packages/grid/backend/grid/start.sh
yashgorana: uv pip install works regardless if you've activated the environment or not

yashgorana: Actually this is not required. I think there's some lines that should not be in the code, probably merge conflicts that got pushed in. I'll sit and review again

packages/grid/devspace.yaml
yashgorana: ah yes - merge conflict got pushed in

yashgorana: Ah my bad - I didn't see that. will fix.

packages/grid/backend/backend.dockerfile.dockerignore
yashgorana: Because the context keeps on changing for different build processes. In some it's packages/ in some it's application-specific like frontend/ & seaweedfs/. So this makes stuff way more clear no? So better to keep it next to the dockerfile so context can be anywhere as long as the paths are correct.

madhavajay: Its only in `/package/` for the `backend` because it copies syft directly into the container. We could "build" a syft wheel first and then copy that in and install it instead and then the context path could be `backend`.

rasswanth-s: Correct, since for backend the context is in `packages` but we have the .dockeringore inside `packagres/grid/backend` as docker looks for .dockerignore in the root directory of the context, so It would not take into account the `backend.dockerfile.dockerignore`  , so we could place this under `packages` ?

yashgorana: > We could "build" a syft wheel first and then copy that in and install it instead and then the context path could be backend.

that would increase the build time so all gains would be lost :( And then we have the issue with torch where the cuda libs might be bundled with the wheel.

> ... so It would not take into account the backend.dockerfile.dockerignore , so we could place this under packages 

_"A Dockerfile-specific ignore-file takes precedence over the .dockerignore file at the root of the build context if both exist."_ https://docs.docker.com/build/building/context/#filename-and-location

The goal here is to keep things predictable and follow a pattern and have ignores next to their dockerfiles than have them spread over at different places based on their context.

rasswanth-s: That statement mentions only about precedence over multiple dockerignore files
But the dockerignore itself would not be considered , if it is not in the root directory of the context.

rasswanth-s (approved):
Excellent work @yashgorana

packages/grid/backend/backend.dockerfile.dockerignore
yashgorana: Tested this on chat and ignore is working as expected - marking this as resolved

PR #8606, opened 2024-03-20
smartopen any cloud storage

yashgorana:
## Description

Part of https://github.com/OpenMined/Heartbeat/issues/1149

## Affected Dependencies

## How has this been tested?

## Checklist


yashgorana:
Demo code

kiendang:
Had a discussion with @tcp the other day around blob storage abstractions (client, connection, config). I do see now that the `BlobStorageConnection` interface does not really work here, at least not yet. The original intention behind `BlobStorageConnection` is that you can do 

```py
with blob_client.connect() as conn:
    conn.write(...)

with blob_client.connect() as conn:
    conn.read(...)
```

where `blob_client.connect()` returns a `BlobStorageConnection` via `BlobStorageConnection.__init__()`

I see that here we can't really do that yet since you need to pass different parameters to the connection to Smart Open depending on whether you write or read.

We could change the abstraction a bit to allow `client.connect()` and `BlobStorageConnection.__init__()` to accept extra parameters but that would leak abstraction. We probably needs a rework/rethinking around this.

kiendang:
@yashgorana while working on integration with different blob storage services, if you think of any tweaks to these abstractions that better fit our use cases please push them.

yashgorana:
@kiendang this is a spike PR to see if we can use smartopen, but as you rightly pointed out it just doesn't fit in well. Additionally, smartopen doesn't support deleting files which renders it partially useful. Meanwhile, @tcp has been working on using fsspec which feels much more complete and in-line with our use-case. There still remains some gaps, specifically how to upload blob from client to bucket (through backend) without seaweed's `/blob` API. As we solve it we may actually end up doing some much needed rework/refactoring of our blob store APIs. 

Regardless, thank you for your insights!

PR #8612, opened 2024-03-21
Moving protocol_version.json to gitignore

snwagh:
Moving `packages/syft/src/syft/protocol/protocol_version.json` to gitignore


kiendang:
I don't think this should be gitignored since it has to be tracked and synced. @shubham3121

PR #8621, opened 2024-03-24
Improve launch UX (mostly for first-time users)

iamtrask:
(this PR is based on https://github.com/OpenMined/PySyft/pull/8562 — creating a new PR without the latest dev change so I can see CI between now and when dev is fixed)


iamtrask:
UX is coming together 
Screenshot 2024-03-24 at 02 42 08

iamtrask:
Added the ability to upload demo datasets (MNIST so far)

Screenshot 2024-03-24 at 03 33 04

iamtrask:
Polished logging a bit more.

Screenshot 2024-03-24 at 03 47 36

iamtrask:
Added dummy users to demo.

Screenshot 2024-03-24 at 04 11 35

Screenshot 2024-03-24 at 04 11 45
Screenshot 2024-03-24 at 04 14 47

packages/syft/src/syft/client/domain_client.py
```diff
@@ -89,6 +91,119 @@ class DomainClient(SyftClient):
     def __repr__(self) -> str:
         return f"<DomainClient: {self.name}>"
 
+    def demo(self) -> tuple[DomainClient, DomainClient, DomainClient]:
```
koenvanderveen, line 94: maybe client.demo("mnist") ?

packages/syft/src/syft/__init__.py
```diff
@@ -159,3 +161,100 @@ def _orchestra() -> Orchestra:
 
 def search(name: str) -> SearchResults:
     return Search(_domains()).search(name=name)
+
+
+def launch(
```
koenvanderveen, line 163: why do we need this here? isnt this duplicating code?

packages/syft/src/syft/__init__.py
```diff
@@ -159,3 +161,100 @@ def _orchestra() -> Orchestra:
...
+            password = getpass("\tAdmin Password:")
+
+            client.me.set_name(name)
+            client.me.set_email(email)
+            client.me.set_password(password)
+
+        # set default email notifier (note: each email account can only send 300 emails a day,
+        # and it might go to people's junk folders.)
+        emails_and_passwords = list()  # noqa: C408
+        emails_and_passwords.append(
```
koenvanderveen, line 238: this is part of syft.launch now, does not sound like a good idea to hardcode things here. If we really want this can we add environment variables for this?

yashgorana: Hmm agreed... Since hagrid will be deprecated soon, we anyway need to refactor `sy.orchestra.launch` and give it a proper home within syft. Additionally, this gives us the opportunity to implement a bootstrapping mechanism that @madhavajay was asking.

I'm thinking on the lines of a `bootstrap` module (a more generic variant of `demo.py`) that allows us to setup arbitrary/static data in the node. This entails a couple of questions:

- Currently `.demo()` is being called by the client, but who should really bootstrap the data? Syft backend on node start as per args?  Client-DS? Client-DO? And how do we control what data gets fed into the backend?
- How does this operate in different modes of sy.launch, namely - inmemory, process, single_container, k8s. Because in k8s you can only "connect" to a node and not "launch" it, so we should expect some baseline information to be already present.

madhavajay: Yes, I think we should split this up into two PRs:
- small textual changes / flags which we want to change
- adding some kind of demo mode, which to me might be better handled by the planned work to add syft demo data fixutres into the code base that let you easily run "scenarios" without having to load up a notebook and copy and paste to get to a known good state

packages/syft/src/syft/__init__.py
iamtrask: We have two users for launch commands:
- very advanced PySyft developers who want all the bells and whistles, logging, config, and argument
- noobs who want to launch a server with reasonable defaults

Advanced users can keep using syft.orchestra.launch(many, arguments, little, default, configuration, and, returns, a, node, object, instead, of client).

Beginners can focus on admin_client = syft.launch("my-already-configured-node")

iamtrask: The alternative is to break the syft.orchestra.launch() command to ship with all the batteries-included features beginners will want.

packages/syft/src/syft/__init__.py
iamtrask: Down to split this into two PRs

> "can we add environment variables for this"

While I think it's a good idea to add an environment variable at some point, from a UX perspective, I find env variables to be the low on the list of configuration option UXes for any system. IMO, from the perspective of a data scientist/statistician (especially one not trained in computer science or software engineering), there's a UX order of preference which works something like:

1) BEST CASE: a default i never have to know about or change
2) a default which is managed by a public registry other people worry about (like syft.networks)
3) a default which pings me when i need to set it myself (like resetting the admin password)
4) a default I can override via a python API
5) A default I have to go onto a website, find documentation, open up a config file of env variables, edit, save, reboot my environment/cluster, and then somehow check to see if the environment variable was picked up by the system (or ignored because of some scoping/virtualenv/container magic.
6) WORST CASE: a default i wish i could change but can't because it's hardcoded.

Everything can have an environment variable override, but I don't think we should start there. We should start higher up on the UX list and work our way down as necessary.

In this case, there's an element of this that is (6), except there's a python override already... but if there was a problem with these services it might be difficult to debug. But if I added a flag to the launch() sequence to skip this registration then this bumps everything up to a (4). In a perfect world this would pull from a registry like syft.networks which would bump it to a (2)... with a backup of a (4) if you wanted to override it yourself.

packages/syft/src/syft/client/domain_client.py
iamtrask: I like it. Will add (and with a default that likely includes MNIST so the user doesn't have to read documentation to know how to use it). Also might make a syft.networks like default registry of demo datasets so that it's updatable after the fact.

PR #8627, opened 2024-03-25
Fix data protocol bug

khoaguin:
Fixed the bug below in dev and syft 0.8.5

image

Thank you so much @teo-milea for the solution


khoaguin:
@shubham3121 is this because we cleaned up the older versions of the classes and made the codebase stop being compatible with older versions after updated to pydantic v2

PR #8659, opened 2024-04-03
Attestation service in Syft

snwagh:
## Description
Adding an attestation endpoint in Syft


packages/grid/veilid/start.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
snwagh: Delete this folder

packages/grid/enclave/attestation/enclave-development.md
```diff
@@ -102,3 +102,107 @@ print ("[RemoteGPUTest] node name :", client.get_name())
 client.add_verifier(attestation.Devices.GPU, attestation.Environment.REMOTE, NRAS_URL, "")
 client.attest()
 ```
+
+### Instructions for using helm charts
+
+- The attestation container runs inside the backend pod (so backend pod has two containers now). However, in order to run the attestation container, you need to uncomment the attestation flags in `packages/grid/helm/values.dev.yaml`
+- Next, we run the deployment. Since k3d creates an intermediate layer of nesting, we need to mount some volumes from host to k3d registry. Thus, when launching, use the following tox command `tox -e dev.k8s.start -- --volume /sys/kernel/security:/sys/kernel/security --volume /dev/tmprm0:/dev/tmprm0`
+- Finally, note that the GPU privilages/drivers etc. have not been completed so while the GPU attestation endpoints should work, they will not produce the expected tokens. To test the GPU code, follow the steps provided in [For GPU Attestation
```
itstauq: Minor typo in `privilages` 🫣

packages/grid/enclave/attestation/enclave-development.md
```diff
@@ -102,3 +102,107 @@ print ("[RemoteGPUTest] node name :", client.get_name())
...
+  ](#for-gpu-attestation) to look at the tokens.
+
+### Local Client-side Verification
+
+Use the following function to perform local, client-side verification of tokens. They expire quick.
+
+```python3
+def verify_token(token: str, type: str):
```
itstauq: Naming the second parameter `type` might cause a bit of confusion since it shadows the built-in `type()` function. How about `token_type` or `kind` instead?

packages/grid/enclave/attestation/start.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 export PATH="/root/.local/bin:${PATH}"
 
-APP_MODULE=server.main:app
+APP_MODULE=server.attestation_main:app
```
itstauq, line 5: ❤️

packages/grid/enclave/attestation/server/gpu_attestation.py
```diff
@@ -1,11 +1,27 @@
...
+# stdlib
+import io
+import re
+import sys
+
 # third party
 from loguru import logger
 from nv_attestation_sdk import attestation
 
 NRAS_URL = "https://nras.attestation.nvidia.com/v1/attest/gpu"
```
itstauq: Should we move this to a new `constants.py` file?

packages/syft/src/syft/__init__.py
```diff
@@ -154,5 +154,11 @@ def _orchestra() -> Orchestra:
     return Orchestra
 
 
+@module_property
+def hello_baby() -> None:
+    print("Hello baby!")
+    print("Welcome to the world. \u2764\ufe0f")
+
+
```
itstauq, line 162: Is this required?

packages/syft/src/syft/service/attestation/attestation_endpoints.py
```diff
@@ -0,0 +1,5 @@
+ATTESTATION_SERVICE_URL = (
+    "http://localhost:4455"  # Replace with "http://attestation:4455"
+)
+ATTEST_CPU_ENDPOINT = "/attest/cpu"
+ATTEST_GPU_ENDPOINT = "/attest/gpu"
```
itstauq: Should we rename this file as `constants.py` to make it more generic and future-proof?

packages/syft/src/syft/service/attestation/__init__.py
itstauq, line 1: Should we keep these files under ...service/`enclave/`attestation/ directory instead? This will help keep the directory structure consistent with `packages/grid/`, and it also kind of makes sense to nest it like that. Thoughts?

itstauq (changes requested):

packages/syft/src/syft/__init__.py
snwagh: Kinda, wanted to sneak this in for Andrew.

packages/syft/src/syft/service/attestation/attestation_endpoints.py
snwagh: Good idea, making the changes.

packages/syft/src/syft/service/attestation/__init__.py
snwagh: @rasswanth-s What are your thoughts on this? I think it's a fair comment from Tauquir's but also feel that each backend pod is fairly independently structured in terms of the services (looking at frontend, backend, seaweedfs) so maybe it is fine to keep it as is.

packages/syft/src/syft/service/attestation/attestation_service.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def perform_request(
+        self, method: Callable, endpoint: str, raw: bool = False
+    ) -> SyftSuccess | SyftError | str:
+        try:
+            response = method(f"{ATTESTATION_SERVICE_URL}{endpoint}")
+            response.raise_for_status()
+            message = response.json().get("result")
+            raw_token = response.json().get("token")
+            return raw_token if raw else SyftSuccess(message=message)
```
rasswanth-s: Leaving a Note, even during the situations, when the output if False, it is wrapped in SyftSuccess message, in future it would be good to wrap it in Syft Error.

rasswanth-s:
Tested the Workflow, works great
Screenshot 2024-05-03 at 4 01 10 PM

packages/grid/enclave/attestation/server/gpu_attestation.py
```diff
@@ -1,11 +1,29 @@
+# standard imports
```
rasswanth-s: NIT: we could remove standard imports, as we have stdlib below

packages/grid/enclave/attestation/server/gpu_attestation.py
```diff
@@ -1,11 +1,29 @@
...
+import re
+import sys
+
 # third party
 from loguru import logger
 from nv_attestation_sdk import attestation
 
-NRAS_URL = "https://nras.attestation.nvidia.com/v1/attest/gpu"
+# relative
+# relative imports
```
rasswanth-s: NIT:  we could remove relative imports, as we have relative above added by the linter.

packages/grid/enclave/attestation/server/cpu_attestation.py
```diff
@@ -14,7 +14,16 @@ def attest_cpu() -> str:
...
 
     logger.info("Attestation Return Code: {}", cpu_report.returncode)
+    res = "False"
     if cpu_report.returncode == 0 and cpu_report.stdout == "true":
-        return "True"
+        res = "True"
 
-    return "False"
+    # Fetch token from Micrsoft Attestation library
+    cpu_token = subprocess.run(
```
rasswanth-s, line 22: Leaving a note , for later, we do two calls, to the attestation client library , we could combine them in to a single call in later PR's

packages/syft/src/syft/service/attestation/__init__.py
rasswanth-s: Good point Tauquir, I think for now, we could keep it same, as we are yet to decide  on how generic the attestation would be and how it would fit in syft.

rasswanth-s (approved):
Great work @snwagh 🎸

packages/syft/src/syft/service/attestation/attestation_service.py
snwagh: Added this change.

packages/grid/enclave/attestation/server/cpu_attestation.py
snwagh: Will shelf this as a TODO item, good point though.

packages/syft/src/syft/service/attestation/attestation_service.py
```diff
@@ -0,0 +1,65 @@
...
+        self, method: Callable, endpoint: str, raw: bool = False
+    ) -> SyftSuccess | SyftError | str:
+        try:
+            response = method(f"{ATTESTATION_SERVICE_URL}{endpoint}")
+            response.raise_for_status()
+            message = response.json().get("result")
+            raw_token = response.json().get("token")
+            if raw:
+                return raw_token
+            elif message:
```
rasswanth-s: Since we return a string as true or false, we could use the utility function called str_to_bool to make it such that  we convert it to a boolean value.

rasswanth-s: Resolved by [249e309](https://github.com/OpenMined/PySyft/pull/8659/commits/249e309cf92d0d9c32cba0dd5b4fcdbdb71f3a1b)

rasswanth-s: Also Tested it on Cloud on the same.

itstauq (approved):
LGTM 🚀

PR #8643, opened 2024-03-31
Attestation Side Car Container

rasswanth-s:
## Description
The PR aims to add a attestation side car container to retrieve the Attestation reports  from Enclaves
1. CPU
2. GPU

Pending Task List
- [x] Add FastAPI Server endpoint to retrieve Attestation results
- [ ] [NICE TO HAVE] Add the ability to set appraisal policy by Relying party.
- [ ] Minimize the Docker Image Size

## How has this been tested?
Locally

## Checklist
- [x] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [x] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [x] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [ ] My changes are covered by tests

PR #8670, opened 2024-04-05
Add helm charts for enclave attestation service

itstauq:
## Description
1. Add attestation container to `devspace.yaml`
2. Create a new folder for attestation in `packages/grid/helm`
3. Create attestation deployment and service
5. Add an entry in `values.yaml`, by default disabling attestation.

Closes OpenMined/Heartbeat#1168

## Affected Dependencies
TODO

## How has this been tested?
TODO

## Checklist


packages/grid/helm/syft/values.yaml
```diff
@@ -146,6 +146,20 @@ ingress:
   class: null
   className: null
 
+# Enclave attestation Service
+attestation:
+  enabled: false
+  serverFlags: null
```
snwagh: Is this used?

packages/grid/helm/syft/templates/attestation/attestation-deployment.yaml
```diff
@@ -0,0 +1,61 @@
...
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        {{- include "common.labels" . | nindent 8 }}
+        app.kubernetes.io/component: attestation
+    spec:
+      containers:
+        - name: attestation-container
+          image: {{ .Values.global.registry }}/openmined/grid-attestation:{{ .Values.global.version }}
```
yashgorana: this image is not yet in the repo? is it a separate PR?

yashgorana: ah i see there are two PRs:
- https://github.com/OpenMined/PySyft/pull/8643/files#diff-d58bb1b2d58992c665bb779ba6eddfc2e6ad56c9fe064b92d0f63f5f752dd882
- https://github.com/OpenMined/PySyft/pull/8659/files#diff-09319b2675ff5b9deb32aadf2fca875581cdbebadb62ed71f3bed02e95916b2e

packages/grid/helm/syft/values.yaml
itstauq: Nope, removing this 👍🏻

packages/grid/devspace.yaml
```diff
@@ -93,19 +101,24 @@ dev:
...
       app.kubernetes.io/name: syft
       app.kubernetes.io/component: backend
-    env:
-      - name: RELEASE
-        value: development
-      - name: DEV_MODE
-        value: "True"
-      - name: DEBUGGER_ENABLED
-        value: "True"
     ports:
```
rasswanth-s, line 108: NIT: We would need to expose these ports only in backend-container, maybe we could  add it only to the backend-container-section?

packages/grid/devspace.yaml
```diff
@@ -93,19 +101,24 @@ dev:
...
+            value: "True"
+          - name: DEBUGGER_ENABLED
+            value: "True"
+        sync:
+          - path: ./backend/grid:/root/app/grid
+          - path: ../syft:/root/app/syft
+      enclave-attestation:
+        sync:
+          - path: ./enclave/attestation/server:/app/server
+      ssh: {}
```
rasswanth-s, line 126: MAYBE: could we add a commented out code in 
`values.dev.yaml`
like

Screenshot 2024-04-16 at 4 55 47 PM

So it is easy to re-enable it for testing by devspace

rasswanth-s (approved):
Well Done @itstauq , left some minor comments

packages/grid/devspace.yaml
itstauq: @rasswanth-s I tried doing this originally but this is not possible. As per [devspace's docs](https://www.devspace.sh/docs/configuration/dev/selectors/labels#multiple-dev-containers-per-pod)

> since pods share the same network stack, the ports option needs to be on the top-level instead of for each container individually.

packages/grid/devspace.yaml
itstauq: Sure, that'll be helpful. Adding!

packages/grid/devspace.yaml
rasswanth-s: Right makes sense 👍

tox.ini
```diff
@@ -878,6 +878,7 @@ passenv = HOME, USER
 setenv =
     CLUSTER_NAME = {env:CLUSTER_NAME:syft-dev}
     CLUSTER_HTTP_PORT = {env:CLUSTER_HTTP_PORT:8080}
+    CLUSTER_EXTRA_ARGS = {env:CLUSTER_EXTRA_ARGS:}
```
rasswanth-s: cc: @yashgorana 

For context:
This is required, as for attestation, we need additional volume mounts from the host during cluster creation.

Example:

```sh
k3d cluster create test --volume sys/kernel/security:/sys/kernel/security ...
```

itstauq: I’ve replaced this with `{posargs}`, so now we need to run `tox -e dev.k8s.start -- --volume sys/kernel/security:/sys/kernel/security --volume /dev/tmprm0:/dev/tmprm0` instead

snwagh: Command has a missing `/`, the right command is:
`tox -e dev.k8s.start -- --volume /sys/kernel/security:/sys/kernel/security --volume /dev/tmprm0:/dev/tmprm0`

snwagh (approved):
Tested CPU (and GPU), looks good to me.

packages/grid/helm/syft/templates/backend/backend-statefulset.yaml
```diff
@@ -152,6 +157,53 @@ spec:
...
+        {{ end }}
+      {{- if .Values.attestation.enabled }}
+      volumes:
+        - name: kernel-security
+          hostPath:
+            path: /sys/kernel/security
+        - name: tpm-device
+          hostPath:
+            path: /dev/tpmrm0
+      {{- end }}
```
yashgorana, line 206: Any idea what would happen/what error will be thrown if we run this on a node that does not have these files?

tox.ini
```diff
@@ -889,7 +889,7 @@ commands =
     tox -e dev.k8s.registry
 
     ; for NodePort to work add the following --> -p "NodePort:NodePort@loadbalancer"
-    bash -c 'k3d cluster create ${CLUSTER_NAME} -p "${CLUSTER_HTTP_PORT}:80@loadbalancer" --registry-use k3d-registry.localhost:5800 && \
+    bash -c 'k3d cluster create ${CLUSTER_NAME} -p "${CLUSTER_HTTP_PORT}:80@loadbalancer" --registry-use k3d-registry.localhost:5800 {posargs} && \
```
yashgorana, line 892: amazing

yashgorana (approved):

packages/grid/helm/syft/templates/backend/backend-statefulset.yaml
itstauq: It will ignore that and create empty directories within the containers instead. As a nice side-effect, this will also ensure that the attestation endpoint return false as expected without having to handle anything in the code.

PR #8710, opened 2024-04-15
Bugfix: Fix adding data subject not working if there is no member

khoaguin:
Fix some bugs reported by user: https://openmined.slack.com/archives/C6EEFN3A8/p1712836116859599
Issue: https://github.com/OpenMined/Heartbeat/issues/1178

Progress:
- Fix adding data subject not working if there is no member
- Add a debug log when default worker pool can't be launched

### DIFF
--- a/packages/syft/src/syft/node/node.py
+++ b/packages/syft/src/syft/node/node.py
@@ -17,6 +17,7 @@ import traceback
 from typing import Any
 
 # third party
+from loguru import logger
 from nacl.signing import SigningKey
 from result import Err
 from result import Result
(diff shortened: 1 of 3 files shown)


packages/syft/src/syft/node/node.py
```diff
@@ -1608,6 +1609,10 @@ def create_default_worker_pool(node: Node) -> SyftError | None:
             image_uid=default_image.id,
             num_workers=worker_count,
         )
+    elif isinstance(default_worker_pool, SyftError):
```
shubham3121: Lets move this check before the `create_default_image` method itself (Move to Line 1567)

shubham3121 (changes requested):

shubham3121 (approved):
💯

PR #8712, opened 2024-04-17
Add podLabels podAnnotations to Helm Charts

yashgorana:
## Description

This is to support GKE and any cloud-specific configurations

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/packages/grid/helm/syft/templates/backend/backend-statefulset.yaml
+++ b/packages/grid/helm/syft/templates/backend/backend-statefulset.yaml
@@ -20,6 +20,12 @@ spec:
       labels:
         {{- include "common.labels" . | nindent 8 }}
         app.kubernetes.io/component: backend
+        {{- if .Values.node.podLabels }}
+        {{- toYaml .Values.node.podLabels | nindent 8 }}
+        {{- end }}
+      {{- if .Values.node.podAnnotations }}
+      annotations: {{- toYaml .Values.node.podAnnotations | nindent 8 }}
+      {{- end }}
     spec:
       {{- if .Values.node.nodeSelector }}
       nodeSelector: {{- .Values.node.nodeSelector | toYaml | nindent 8 }}
@@ -45,7 +51,7 @@ spec:
             - name: NODE_TYPE
               value: {{ .Values.node.type | default "domain" | quote }}
             - name: NODE_NAME
-              value: {{ .Values.node.name | default .Release.Name | quote }}
+              value: {{ .Values.node.name | default (printf "%s-node" .Release.Name) | quote }}
             - name: NODE_SIDE_TYPE
               value: {{ .Values.node.side | quote }}
             - name: DEFAULT_ROOT_EMAIL
--- a/packages/grid/helm/syft/templates/backend/backend-statefulset.yaml
+++ b/packages/grid/helm/syft/templates/backend/backend-statefulset.yaml
@@ -21,6 +21,9 @@ spec:
         {{- include "common.labels" . | nindent 8 }}
         app.kubernetes.io/component: backend
     spec:
+      {{- if .Values.node.nodeSelector }}
+      nodeSelector: {{- .Values.node.nodeSelector | toYaml | nindent 8 }}
+      {{- end }}
       containers:
         - name: backend-container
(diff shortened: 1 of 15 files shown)


yashgorana:
@eelcovdw i have relaxed the resourcePresets in this PR.
backend=xlarge (4 cpu 8gb max)
seaweed=xlarge (4 cpu 8gb max)
mongo=large (2cpu 4gb max)
proxy=small (0.5cpu 1GB max)
frontend=medium (1cpu 2GB max)

in a separate pr i'll update worker pools to use 4xlarge presets (8cpu and 16gb min)

packages/grid/helm/syft/templates/seaweedfs/seaweedfs-statefulset.yaml
```diff
@@ -37,8 +46,6 @@ spec:
                 secretKeyRef:
                     name: {{ .Values.seaweedfs.secretKeyName | required "seaweedfs.secretKeyName is required" }}
                     key: s3RootPassword
-            - name: MOUNT_API_PORT
```
rasswanth-s, line 40: Q: Why are we removing this ?

packages/grid/helm/syft/templates/backend/backend-statefulset.yaml
```diff
@@ -20,7 +20,16 @@ spec:
       labels:
         {{- include "common.labels" . | nindent 8 }}
         app.kubernetes.io/component: backend
+        {{- if .Values.node.podLabels }}
+        {{- toYaml .Values.node.podLabels | nindent 8 }}
```
rasswanth-s, line 24: When we input mutiple podsLabels , it takes the last one.

Example Values files

```
node:
    podLabels:
        test/label: 5
        test/label: 2
```

Output

Screenshot 2024-04-17 at 3 00 27 PM

rasswanth-s: But if I reverse the order
Screenshot 2024-04-17 at 3 01 58 PM

It takes the last one

Screenshot 2024-04-17 at 3 02 19 PM

rasswanth-s: It should take into account all the pod labels?

packages/grid/helm/syft/templates/seaweedfs/seaweedfs-statefulset.yaml
yashgorana: Not really required. Because (1) not exposed to the net and (2) the idea is to avoid dynamic port assignment for simplicity and follow standard port numbers for internal apis.

packages/grid/helm/syft/templates/backend/backend-statefulset.yaml
yashgorana: This is valid & expected. `metadata.labels` & `metadata.annotations` are a map and not list. Both kubernetes and golang templating will hold only one value because underlying struct is a go map. You also see a squigly error line under your YAML because that's validation failure 😉

rasswanth-s: Ahh ok, is there a way such that a user could provide multiple pod annotations to pod spec by helm values?

rasswanth-s (approved):

packages/grid/helm/syft/templates/backend/backend-statefulset.yaml
yashgorana: Yes it is supported and this change implements it ([example](https://github.com/OpenMined/PySyft/pull/8712/files#diff-80ac10ca85a1ad75067177b222c9053f62d08a940e816bf23ebd47cf66b530fcR14-R17)), but k8s spec has [explicit directives  for the type of values](https://github.com/OpenMined/PySyft/pull/8712/files#diff-80ac10ca85a1ad75067177b222c9053f62d08a940e816bf23ebd47cf66b530fcR14-R17) the labels and annotations must have. Since keys-values must be strings, often times k8s providers will expect a JSON string as values ([GKE example](https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-configuration#create_service))

tcp (approved):

PR #8721, opened 2024-04-19
fix: sync widget timestamp entries

abyesilyurt:
There are still a few missing.


abyesilyurt:
widgets need last updated timestamp instead of last synced.

PR #8738, opened 2024-04-23
Fix container tests

yashgorana:
## Description

## Affected Dependencies

## How has this been tested?

## Checklist


notebooks/api/0.8/10-container-images.ipynb
yashgorana, line 1: Some changes in the notebook

- Drop pydicom image build. just use opendp and recordlinkage and to improve test runtime
- Remove workers after each worker pool is tested
- `worker_pool.workers` returns a list of workers, but `worker_pool.worker_list` returns a list of LinkedObject. Unsure if worker remove API actually updates the pool's LinkedObject list. @shubham3121 and maybe that's why this code was failing.

packages/syft/src/syft/service/worker/utils.py
```diff
@@ -82,21 +81,36 @@ def extract_config_from_backend(
...
         mode = parts[2]
-        if "/storage" in bind:
+
+        if "/root/data/creds" in vol:
             # we need this because otherwise we are using the same node private key
             # which will make account creation fail
-            worker_postfix = worker_name.split("-", 1)[1]
-            key = f"{key}-{worker_postfix}"
-        extracted_config["volume_binds"][key] = {"bind": bind, "mode": mode}
+            key = f"{key}-{worker_name}"
```
yashgorana, line 104: There is an issue in WSL2 here.
- The DEV_MODE=True mount is `./path/to/local/creds:/root/data/creds` (when mounted becomes absolute - /home/user/dir/path/to/local/creds`)
- This is available on Linux, but docker desktop windows fails to resolve this path as it's not a windows path and it mounts against the docker-desktop-data WSL2 distro and not the distro it was invoked from. This makes syft function call fails - probably because credentials sharing not working as expected?

@shubham3121

packages/syft/src/syft/service/worker/utils.py
```diff
@@ -535,47 +549,20 @@ def create_default_image(
     tag: str,
     in_kubernetes: bool = False,
 ) -> SyftError | SyftWorkerImage:
-    # TODO: Hardcode worker dockerfile since not able to COPY
-    # worker_cpu.dockerfile to backend in backend.dockerfile.
-
-    # default_cpu_dockerfile = get_syft_cpu_dockerfile()
-    # DockerWorkerConfig.from_path(default_cpu_dockerfile)
-
     if not in_kubernetes:
```
yashgorana, line 552: Kubernetes and Docker are now in parity - no need to build a new worker image - just run the backend in worker mode

packages/syft/src/syft/util/util.py
```diff
@@ -888,19 +888,6 @@ def set_klass_module_to_syft(klass: type, module_name: str) -> None:
...
-
-
-def get_grid_src_path() -> Path:
-    syft_path = get_syft_src_path()
-    return syft_path.parent.parent / "grid"
-
-
-def get_syft_cpu_dockerfile() -> Path:
-    return get_grid_src_path() / "backend" / "worker_cpu.dockerfile"
-
```
yashgorana, line 902: unused and should instead use `CustomWorkerBuilder.find_worker_image(type)`

notebooks/api/0.8/10-container-images.ipynb
shubham3121: Hmm.. the `api.services.worker.delete` API does remove the worker from the worker_pool list. Although, I see we're not doing the same thing in the `worker_pool.scale` service. I will create a ticket to handle this.

packages/syft/src/syft/service/worker/utils.py
shubham3121: Nice catch @yashgorana. We can think about mounting according to os type, but again since we will be kubernetes for now, not sure if this is an urgent fix.

shubham3121 (approved):
@yashgorana this LGTM !! Nice work fixing the container test.

packages/syft/src/syft/service/worker/utils.py
yashgorana: yeah not an urgent fix

notebooks/api/0.8/10-container-images.ipynb
```diff
@@ -33,7 +33,47 @@
     "\n",
     "# syft absolute\n",
     "from syft.service.worker.image_registry import SyftImageRegistry\n",
-    "from syft.service.worker.worker_image import SyftWorkerImage"
+    "from syft.service.worker.worker_image import SyftWorkerImage\n",
+    "\n",
+    "# Local registry to test external registry\n",
+    "\n",
+    "\n",
+    "class LocalRegistryContainer:\n",
```
rasswanth-s, line 41: It would be good to do the optimization we did in this notebook to also the K8s notebooks in a   later PR.

tox.ini
```diff
@@ -1066,18 +1070,25 @@ changedir = {toxinidir}/packages
 allowlist_externals =
     docker
     bash
+    env
 setenv =
     PIP_PACKAGES = {env:PIP_PACKAGES:llama-index opendp}
     SYSTEM_PACKAGES = {env:SYSTEM_PACKAGES:curl wget}
+    BUILD_PLATFORM = {env:BUILD_PLATFORM:linux/amd64}
```
rasswanth-s, line 1077: Q: Why have used explicitly linux/amd64, as it would slow the build on arm 64 based systems?

yashgorana: Oh this is just to make sure if I want to test this locally on a different platform, i can pass in a buildx platform through env var.

yashgorana: Also I moved to buildx build - so it needs a platform arg which defaults to linux/amd64

rasswanth-s (approved):
Great work @yashgorana

notebooks/api/0.8/10-container-images.ipynb
```diff
@@ -1277,37 +1278,47 @@
...
-    "    try:\n",
-    "        result = builder.build_image(\n",
-    "            config=config, tag=tag, pull=pull, rm=True, forcerm=True, **kwargs\n",
-    "        )\n",
-    "        return SyftSuccess(message=result.logs)\n",
-    "    except Exception as e:\n",
-    "        return SyftError(message=f\"Failed to build image !! Error: {str(e)}.\")\n",
+    "RUN pip install recordlinkage\n",
+    "\"\"\".strip()\n",
     "\n",
```
rasswanth-s, line 1305: A Quick question, many a times, this tests flakes , due to the fact , that, it was not able to delete a worker with a given UID.

were we able to fix that?

yashgorana: i have not been able to repro this after this fix - it was likely an out of resource error. hence i decided to remove the workers once the test is done.

PR #8739, opened 2024-04-24
Extract container worker tags from env variable

khoaguin:
## Description
Extract container worker tags from `HAGRID_FLAGS` for `stack.test.integration`


khoaguin:
Closed since `hagrid` will be deprecated soon

PR #8754, opened 2024-04-25
add a method to delete/kill the consumer in the thread

shubham3121:
Fixes:

- Fix: For in memory workers, consumer thread is not killed when worker is deleted
https://files.slack.com/files-pri/T6963A864-F0704PM06J3/screenshot_2024-04-25_at_2.49.54_pm.png

- Fix: Skip updating consumer state when corresponding worker object is deleted from the stash
image

### DIFF
--- a/packages/syft/src/syft/node/node.py
+++ b/packages/syft/src/syft/node/node.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 # stdlib
 from collections import OrderedDict
+from collections import defaultdict
 from collections.abc import Callable
 from datetime import datetime
 from functools import partial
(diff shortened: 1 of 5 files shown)


packages/syft/src/syft/node/node.py
```diff
@@ -567,6 +571,18 @@ def add_consumer_for_service(
         )
         consumer.run()
 
+    def remove_consumer_with_id(self, syft_worker_id: UID) -> None:
+        for _, consumers in self.queue_manager.consumers.items():
```
koenvanderveen, line 575: I am not sure how to parse the following datastructure, we have a: Client that has multiple consumers. How does that work? Doesnt every consumer need its own client?

koenvanderveen (approved):

PR #8777, opened 2024-05-01
Single container launching added to tox and workflows

snwagh:
## Description
This adds a new tox environment to run a backend container. The workflows will automatically run this before running all the tests. 

Note: Created the PR even though it's not fully bug free since it is blocking Tauquir.


snwagh:
## Notes about the failing tests
- `10-container-images.ipynb`. So when I run the following commands:
```
cd packages/
docker build -f grid/backend/backend.dockerfile . -t backend:0.1
docker run -d -e SINGLE_CONTAINER_MODE=true -p 8080:80 --name mybackend backend:0.1
tox -e stack.test.notebook
```
the test fails because it tries to pull an image from `openmined/grid-backend:tag` where `tag` gets set to `local-dev` because when we run the tox command, we modified this setenv `ORCHESTRA_DEPLOYMENT_TYPE = {env:ORCHESTRA_DEPLOYMENT_TYPE:k8s}`. But when I visit OpenMined's dockerhub page after logging into my account, I do not see any image tagged `local-dev` (I don't think it's a perms issue with my docker account). That is what is making the test fail since it's unable to load that base image in this command:
```
custom_dockerfile_str = f"""
FROM openmined/grid-backend:{syft_base_worker_tag}

RUN pip install pydicom

""".strip()
```
Not sure what the right fix to this is. 

- `12-custom-api-endpoint.ipynb` This test I have not been able to figure out a way to debug the error stack. Will continue working on this.

rasswanth-s (approved):
Well Done @snwagh

tox.ini
```diff
@@ -449,6 +451,25 @@ commands =
     ; pytest --nbmake tutorials/pandas-cookbook -p no:randomly -vvvv
 
 
+[testenv:single_container.launch]
+description = Launch a single backend container using the dockerfile
+changedir = {toxinidir}/packages
+allowlist_externals =
+    bash
```
shubham3121, line 463: suggestion: can we accept NODE_NAME,  NODE_TYPE and PORT as env, (how we do it k8s.dev.deploy) so that users can set them instead of using the default. We can also set a label
```
--label orgs.openmined.syft="this is a syft worker container"
```
This label together with NODE_NAME can later be used to remove containers, volumes, etc.

tox.ini
```diff
@@ -449,6 +451,25 @@ commands =
...
+commands =
+    bash -c 'tox -e single_container.destroy'
+    bash -c 'docker build -f grid/backend/backend.dockerfile . -t openmined/grid-backend:local-dev'
+    bash -c 'docker run -d  -e N_CONSUMERS=1 -e SINGLE_CONTAINER_MODE=true -e CREATE_PRODUCER=true -e INMEMORY_WORKERS=true -p 8080:80 --add-host=host.docker.internal:host-gateway --name test_backend openmined/grid-backend:local-dev'
```
shubham3121: suggestion: port number can be derived from env

tox.ini
```diff
@@ -449,6 +451,25 @@ commands =
...
+
+[testenv:single_container.destroy]
+description = Destroy the single backend container run using single_container.launch
+changedir = {toxinidir}/packages
+allowlist_externals =
+    bash
+commands =
+    # Image is not cleaned up
```
shubham3121, line 485: suggestion: This can also be update to accept NODE_NAME and use label to stop container.

shubham3121 (changes requested):
Left some minor comments, otherwise looks good. 🎉

tox.ini
rasswanth-s: Resolved in [c03f06f](https://github.com/OpenMined/PySyft/pull/8777/commits/c03f06fb6c336b57ffbada66345c63ce3644c808)

tox.ini
rasswanth-s: Resolved in [c03f06f](https://github.com/OpenMined/PySyft/pull/8777/commits/c03f06fb6c336b57ffbada66345c63ce3644c808)

tox.ini
rasswanth-s: Resolved in [c03f06f](https://github.com/OpenMined/PySyft/pull/8777/commits/c03f06fb6c336b57ffbada66345c63ce3644c808)

rasswanth-s: The label might be useful when we have multiple contianers, and to land all the containers.
I think for single containers, the node name would be unique identifier for landing.
We could omit labels for now?
I have added the remaining suggesstions.

shubham3121 (approved):
Great work guys !!

PR #8787, opened 2024-05-07
Optimize attaching attributes to `SyftBaseObject`s

kiendang:
## Description
Optimize the `attach_attribute_to_syft_object` function used to attach attributes (`syft_node_location` and `syft_client_verify_key`) to `SyftBaseObject`s by avoiding creating new/reconstructing objects.

The function also now works out of the box for data structures *eg* `tuple` and `DictTuple` which we currently have to special case.

Does seem to shave seconds off running the whole unit test suit, but nothing dramatic.


shubham3121 (approved):

shubham3121:
Nice work @kiendang, this simplifies the code !! 💯

PR #8790, opened 2024-05-07
Reenable ruff

kiendang:
## Description
ruff pre-commit was misconfigured and as a result disabled for the whole codebase

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -78,6 +78,7 @@ repos:
 
   - repo: https://github.com/nbQA-dev/nbQA
     rev: 1.8.3
+    files: "^notebooks/(api|tutorials|admin)"
     hooks:
       - id: nbqa-isort
       - id: nbqa-black
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -84,7 +84,7 @@ repos:
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: "v0.3.0"
+    rev: "v0.4.3"
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix, --show-fixes]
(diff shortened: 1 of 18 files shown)


kiendang:
@abyesilyurt Could you help fix the few remaining ones? Thanks!

```
packages/syft/tests/syft/service/sync/sync_resolve_single_test.py:90:16: B011 Do not `assert False` (`python -O` removes these calls), raise `AssertionError()`
packages/syft/tests/syft/stores/kv_document_store_test.py:47:10: B017 `pytest.raises(Exception)` should be considered evil
packages/syft/tests/syft/stores/kv_document_store_test.py:82:10: B017 `pytest.raises(Exception)` should be considered evil
tests/integration/local/twin_api_sync_test.py:191:16: B011 Do not `assert False` (`python -O` removes these calls), raise `AssertionError()`
```

.pre-commit-config.yaml
```diff
@@ -88,7 +88,7 @@ repos:
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix, --show-fixes]
-        files: "^notebooks/(api|tutorials|admin)"
+        exclude: packages/syft/tests/mongomock
```
kiendang, line 92: @yashgorana help me check pls. Thanks!

packages/syft/src/syft/service/settings/settings.py
```diff
@@ -1,5 +1,5 @@
 # stdlib
-from typing import Callable
+from collections.abc import Callable
```
tcp, line 2: Is this intended @kiendang?

kiendang: yes `typing.Callable` is deprecated, replaced by `collections.abc.Callable` https://docs.python.org/3/library/typing.html#typing.Callable

.pre-commit-config.yaml
kiendang: I think `files: "^notebooks/(api|tutorials|admin)"` was added wrongly to the `ruff` step instead of the `nbqa` one right above.

yashgorana: mmm weird - had just made changes to clean up the regex in 4d1c349 but this line was introduced in 2b31864. unsure why but good that this was caught!

yashgorana (approved):

PR #8792, opened 2024-05-07
Fix client autocomplete

koenvanderveen:


packages/syft/src/syft/__init__.py
```diff
@@ -109,6 +109,40 @@
     pass  # nosec
 
 
+try:
+    # third party
+    from IPython.core.guarded_eval import EVALUATION_POLICIES
+
+    ipython = get_ipython()  # type: ignore
+    ipython.Completer.evaluation = "limited"
+    ipython.Completer.use_jedi = False
```
koenvanderveen: I disabled jedi for now, once you use a property getter, it will only be able to autcomplete class level attributes (like methods), not object level attributes (properties). It was way harder to monkeypatch that properly.

packages/syft/src/syft/__init__.py
```diff
@@ -109,6 +109,40 @@
...
+    policy = EVALUATION_POLICIES["limited"]
+
+    # this allow for dynamic attribute getters for autocomplete
+    policy.allowed_getattr_external.update(
```
koenvanderveen: maybe we want to have this in more places eventually, but these seemed to be the most dynamic

eelcovdw (approved):
🎉

packages/syft/src/syft/__init__.py
```diff
@@ -109,6 +109,40 @@
...
+        if hasattr(value, attr_name) and attr in getattr(value, attr_name, []):
+            return True
+        else:
+            return original_can_get_attr(value, attr)
+
+    # this allows property getters to be used in nested autocomplete
+    policy.can_get_attr = patched_can_get_attr
+
+except Exception as e:
+    print(e)
```
kiendang: In non-jupyter/ipython mode importing `syft` will always print some error

kiendang: fixed

PR #8797, opened 2024-05-09
[wip] remove connection pool from sqlite

abyesilyurt:
Sqlite connection is not thread-safe. We use complicated and unstable pooling to make connections thread-safe. We don't need to do any locking if we just open the connection per each thread.

The tests are also much more stable now.


packages/syft/tests/syft/service/sync/sync_resolve_single_test.py
```diff
@@ -76,7 +76,7 @@ def compute() -> int:
 
     client_low_ds.refresh()
     res = client_low_ds.code.compute(blocking=True)
-    assert res == compute(blocking=True).get()
+    assert res == compute(syft_no_node=True)
```
abyesilyurt: We start threads when `syft_no_node=False`, which leads to unstable tests.

packages/syft/src/syft/store/document_store.py
```diff
@@ -348,18 +348,18 @@ def store_query_keys(self, objs: Any) -> QueryKeys:
 
     # Thread-safe methods
     def _thread_safe_cbk(self, cbk: Callable, *args: Any, **kwargs: Any) -> Any | Err:
-        locked = self.lock.acquire(blocking=True)
-        if not locked:
-            print("FAILED TO LOCK")
-            return Err(
-                f"Failed to acquire lock for the operation {self.lock.lock_name} ({self.lock._lock})"
-            )
+        # locked = self.lock.acquire(blocking=True)
```
abyesilyurt: We don't need locks for sqlite anymore. Thread safety is done by sqlite on file level.
Mongodb should handle thread safety as well.

packages/syft/src/syft/store/sqlite_document_store.py
```diff
@@ -100,102 +66,53 @@ def __init__(
...
+
+        cur = con.cursor()
+        try:
+            yield cur.execute(sql, *args)
+            con.commit()
+        except Exception as e:
+            con.close()
+            raise e
+        finally:
+            con.close()
```
kiendang: Suggestion: would be a bit nicer and safer this way since the `con.execute("PRAGMA journal_mode = WAL")... cur = con.cursor()` part would be covered by the `con` context manager as well.

```suggestion
        with contextlib.closing(
            sqlite3.connect(
                self.file_path,
                timeout=timeout,
                check_same_thread=False,  # do we need this if we use the lock?
                # check_same_thread=self.store_config.client_config.check_same_thread,
            )
        ) as con:
            # Set journal mode to WAL.
            con.execute("PRAGMA journal_mode = WAL")
            con.execute("PRAGMA busy_timeout = 5000")
            con.execute("PRAGMA temp_store = 2")
            con.execute("PRAGMA synchronous = 1")

            cur = con.cursor()

            yield cur.execute(sql, *args)
            try:
                con.commit()
            finally:
                cur.close()
```

kiendang: Also at this point should we just `yield con.cursor()` and write
```py
with self._execute() as cur:
    cur.execute(insert_sql, [str(key), _repr_debug_(value), data, str(key)])
```
(and maybe change `def _execute` to `def _cursor`)
instead of
```py
with self._execute(...) as _:
    pass
```
?

packages/syft/tests/syft/stores/sqlite_document_store_test.py
```diff
@@ -22,6 +23,31 @@ def test_sqlite_store_partition_sanity(
     assert hasattr(sqlite_store_partition, "searchable_keys")
 
 
+def test_sqlite_store_partition_benchmark(
```
abyesilyurt, line 26: test to benchmark performance. to be deleted after reviews.

kiendang (commented):
Could `def _cursor` just `yield con.cursor()` instead of 

```py
cur = con.cursor()
yield cur.execute(sql, *args)
```
?

We can rewrite this

```py
with self._cursor(
    insert_sql, [str(key), _repr_debug_(value), data, str(key)]
) as _:
    pass
```

into 

```py
with self._cursor as cursor:
    cursor.execute(insert_sql, [str(key), _repr_debug_(value), data, str(key)])
```

looks a bit more idiomatic to me.

abyesilyurt:
> Could `def _cursor` just `yield con.cursor()` instead of
> 
> ```python
> cur = con.cursor()
> yield cur.execute(sql, *args)
> ```
> 
> ?
> 
> We can rewrite this
> 
> ```python
> with self._cursor(
>     insert_sql, [str(key), _repr_debug_(value), data, str(key)]
> ) as _:
>     pass
> ```
> 
> into
> 
> ```python
> with self._cursor as cursor:
>     cursor.execute(insert_sql, [str(key), _repr_debug_(value), data, str(key)])
> ```
> 
> looks a bit more idiomatic to me.

I agree that it looks better, but I think this change is out of scope for this PR as it leads to much bigger refactoring. I will create an issue for this one though.

packages/syft/src/syft/store/document_store.py
madhavajay: What about overwriting the same data with a stale copy from a different thread can we add a test for this?

@shubham3121 will this effect the thing we are discussing on meets right now?

abyesilyurt:
I figured out some problems with this implementation and database locked errors are still there, when several jobs are launched.

PR #8801, opened 2024-05-12
Update README.md with CPU/RAM requirements

iamtrask:
## Description
I installed PySyft on a fresh (ish) machine and my defaults weren't right. Worth adding to README.


SamirPS:
For this it's because of Docker default setting, i think

PR #8809, opened 2024-05-14
test: update tests to not use ephemeral node

abyesilyurt:
Ephemeral node leads to flaky tests on windows due to threading.


teo-milea (approved):
LGTM!

PR #8811, opened 2024-05-14
Fix request repr

teo-milea:
Just a small correction for requests repr, especially for tables.


abyesilyurt (approved):

packages/syft/src/syft/service/request/request.py
```diff
@@ -470,6 +470,7 @@ def _coll_repr_(self) -> dict[str, str | dict[str, str]]:
             "Request time": str(self.request_time),
             "Description": self.html_description,
             "Requested By": "\n".join(user_data),
+            "Creation Time": str(self.request_time),
```
eelcovdw, line 472: Duplicate of line 470

PR #8813, opened 2024-05-14
fix nightlies 🟢 

abyesilyurt:
- Replace any "undefined" security severity values with 0.
- Use macOS 13, due to issues with macos 14 with docker.

I manually run the failing actions under nightlies and they pass now.


madhavajay (approved):
Great work @abyesilyurt !!! 🙌

PR #8816, opened 2024-05-15
rename windows runner

abyesilyurt:
There is no runner named "windows" anymore, so the actions do not start.
list of available runner tags https://github.com/actions/runner-images


.github/workflows/pr-tests-stack-public.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       max-parallel: 99
       matrix:
         # issues with macos 14 arm https://github.com/crazy-max/ghaction-setup-docker/pull/53
-        os: [ubuntu-latest, macos-13, windows]
```
rasswanth-s, line 22: We used to use windows for self hosted runners, currently we do not have the same.Good catch

rasswanth-s (approved):

PR #8825, opened 2024-05-16
Deprecate Action Graph Service

khoaguin:
[Remove Action Graph Service, tests and libraries](https://github.com/OpenMined/Heartbeat/issues/1329)


shubham3121 (approved):
@khoaguin, thanks this removes dead code. 
Let's bring it back soon with much more robustness.

PR #8827, opened 2024-05-16
[WIP] Consistent crud

teo-milea:
## Description

## Affected Dependencies

## How has this been tested?

## Checklist


packages/syft/src/syft/service/action/action_store.py
```diff
@@ -210,7 +210,8 @@ def take_ownership(
         return Err(f"UID: {uid} already owned.")
 
     def delete(self, uid: UID, credentials: SyftVerifyKey) -> Result[SyftSuccess, str]:
```
madhavajay: should the signature be uid: UID | LineageID?

packages/syft/src/syft/types/newdatetime.py
```diff
@@ -0,0 +1,45 @@
...
+        return cls(utc_timestamp=datetime.utcnow().timestamp())
+
+    def __str__(self) -> str:
+        utc_datetime = datetime.utcfromtimestamp(self.utc_timestamp)
+        return utc_datetime.strftime("%Y-%m-%d %H:%M:%S")
+
+    def __hash__(self) -> int:
+        return hash(self.utc_timestamp)
+
+    def __sub__(self, other: "NewDateTime") -> "NewDateTime":
```
madhavajay: use from typing_extensions import Self?

packages/syft/src/syft/types/newdatetime.py
```diff
@@ -0,0 +1,45 @@
...
+from typing_extensions import Self
+import pydantic
+
+# relative
+from ..serde.serializable import serializable
+from datetime import datetime
+
+@serializable()
+@total_ordering
+class NewDateTime(pydantic.BaseModel):
```
madhavajay, line 16: If the old Datetime was a syft object with a version then you can rename the class since the lookup is from the canonical name, which means you can call this one `DateTime`.

I do still think removing the lookup / versioning aspect of this type just means that we corner ourselves for future changes.

This could easily be fixed by adding another object to the .mro()

SyftObjectBase(pydantic.BaseModel, SyftHashableObject) <- no versioning
SyftObjectVersioned(SyftObjectBase) <- new class with __canonical_name__ and __version__
SyftObject(SyftObjectVersioned) <- all the crazy new fields that are being added mostly for our larger Document store style objects, which I agree feels overkill for a DateTime object.

```
# the old class
class DateTimeV2(SyftObject):
    __canonical_name__ = "DateTime"
    __version__ = SYFT_OBJECT_VERSION_2

# your new class here
class DateTime(SyftObjectVersioned):
    __canonical_name__ = "SyftObject"
    __version__ = SYFT_OBJECT_VERSION_3
```

@koenvanderveen thoughts?

packages/syft/src/syft/types/newdatetime.py
madhavajay: Also this type departing from every other type means it can be serialized but it can never be stored, like you could never have a `NewDateTime` object as the result or input to anything else...

I do think we need the bare minimum of:
```
class SyftType:
    uid: UID
    value: Thing
    __canonicalname__: str
    __version__: int
```
Which means we can reference it, version it and then get the actual value we care about since this is essentially a versioned networked python object.

madhavajay: Except that we could just wrap it in an ActionObject which... is also probably fine, but does beg the question how and why we differentiate between document objects and action objects other than applying well known math operations `__method__` for which most document objects have no well defined implementation.

packages/syft/src/syft/types/syft_object.py
```diff
@@ -395,6 +397,13 @@ class SyftObject(SyftBaseObject, SyftObjectRegistry, SyftMigrationRegistry):
 
     # all objects have a UID
     id: UID
+    _created_date: NewDateTime | None = None
+    _updated_date: NewDateTime | None = None
+    _deleted_date: NewDateTime | None = None
+
+    # @property
+    def is_deleted(self) -> bool:
```
madhavajay, line 405: @koenvanderveen can you make sure that we address the actual permanent deletion of objects seperately. For example if there was some dangerous code or a very large blob we want to remove that should be possible. I figure the version upgrade surgery tools will be the best place to add this kind of behavior that allows for precise manual changes to data.

packages/syft/src/syft/service/service.py
```diff
@@ -50,10 +53,159 @@
...
+            if kwargs.get("skip_method", False):
+                continue
+            wrap_service(service_name=service_name, kwargs=kwargs)
+    
+    @property
+    def object_type(self) -> str:
+        return self.stash.object_type.__canonical_name__.lower()
+    
+    def set(self, context: AuthedServiceContext, obj: Any) -> SyftError | SyftSuccess:
+        # if hasattr(obj, id):
```
madhavajay, line 163: How are our permissions being checked here. Are they admin only or do they allow any user? Do we check for `owner`, or `write`. Also are we allowing for the `uid` to be set by the user as we do in action objects (where we make sure its unclaimed first).

Many of the `CreateModelType` objects are uid free and generate one during the transform. I dont 100% think this matters too much either way as long as perform the permission checks, but if we do allow users to inject their own UIDs we actually need to make sure they are unique in the entire system (the action store as well) otherwise they could create an object with the same uid as an object in the action store and... do something weird probably.

It could be that the generic `set` is admin only and is only used by developers to write their own `create` object / methods in services although I imagine at least building a consistent pattern on top of `set` for everyone is a good idea, or creating a super().create() might also be helpful.

teo-milea: The permissions are checked once when the API is created and once in the store, we could also add some safeguards there as well, though I think for some methods we already have that, but I am not sure how consistent they are. From my understanding, this PR is only for the services and we will refactor `store` as well sometime in the (hopefully near) future. 

Regarding the UID issue, indeed you are right that it might be a problem, we can leave `set` for admin only and add create as well, with the assumption that we have `CreateModelType` object for every object. I could work on adding that if you agree.

packages/syft/src/syft/types/newdatetime.py
teo-milea: This new interface would indeed fix this problem, but at the same time we are currently looking into how to properly handle third-party objects in serialization as in the migration PR we have started to use canonical names instead of fully qualified names. This could be a solution for versioning as well, by adding an exception for `NewDateTime` without it having to inherit from `SyftObject`

rasswanth-s:
Hello 👋 

Thanks so much for this PR — I really appreciate the work you put into it.

A quick heads-up: the repo just went through a significant refactor, so the codebase has changed quite a bit from what this PR was built on. As a result, these changes no longer apply cleanly to the new structure, and I'm going to close this PR for now.

if you're interested in contributing to the current version, kindly have a look at the [README](https://github.com/OpenMined/PySyft/blob/dev/README.md)

Thanks again for contributing — and sorry for the churn!

PR #8830, opened 2024-05-17
[DX] print out SQLite database path in dev mode

khoaguin:
- Link to issue: https://github.com/OpenMined/Heartbeat/issues/1303.
- Description: Print the path to the SQLite database in `DEV` mode in python so that we can easily get it and check it during development.


khoaguin:
For now we are printing 2 paths (one for `document_store`, and one for `action_store`)
image

shubham3121:
@khoaguin possible to indicate in the prints which db is mapped to which Store

khoaguin:
Thanks @shubham3121 for the feed back. Updated accordingly

image

shubham3121 (approved):
👍🏾

PR #8831, opened 2024-05-17
[UX] change tqdm progress bar color to green

khoaguin:
## Description
Making the progress bar green so it looks less like errors and more like progresses

Before: 
image

After:
image
image


shubham3121 (approved):
💯

PR #8837, opened 2024-05-20
Remove Jax and Haiku

khoaguin:
## Description
Since Haiku is deprecated, we should remove Jax and update the notebooks where Jax is used to use Torch

[Link to Issue](https://github.com/orgs/OpenMined/projects/88/views/1?pane=issue&itemId=61944948)

## Affected Dependencies

- Removed:
```
jaxlib==0.4.20
jax==0.4.20
dm-haiku==0.0.10
```

## Notebook Changes
The 5th notebook in `notebooks/api` and the 4 notebooks in `notebooks/tutorials/models_training/` are modified to use `torch` instead of `jax` and `haiku`


packages/syft/src/syft/serde/third_party.py
```diff
@@ -180,13 +179,42 @@ def serialize_bytes_io(io: BytesIO) -> bytes:
...
     pass
 
-# jax
-recursive_serde_register(
-    ArrayImpl,
-    serialize=lambda x: serialize(np.array(x), to_bytes=True),
-    deserialize=lambda x: jnp.array(deserialize(x, from_bytes=True)),
-)
 
+try:
```
shubham3121, line 183: 👍🏾

shubham3121 (approved):
Nice work @khoaguin as always !! 💯 🎉

packages/syft/src/syft/serde/third_party.py
khoaguin: Thank you very much for the kind words @shubham3121

PR #8843, opened 2024-05-22
Fix settings update for side type

teo-milea:
Added a way to update the node side type.

```
from syft.abstract_node import NodeSideType
from syft.service.settings.settings import NodeSettingsUpdate
new_settings = NodeSettingsUpdate(node_side_type=NodeSideType.LOW_SIDE)
client.settings.update(settings=new_settings)
client.refresh()
```


koenvanderveen (approved):

packages/syft/src/syft/service/settings/settings.py
```diff
@@ -66,6 +71,18 @@ class NodeSettingsUpdate(PartialSyftObject):
     admin_email: str
     association_request_auto_approval: bool
     welcome_markdown: HTMLObject | MarkdownDescription
+    node_side_type: str
+
+    @field_validator("node_side_type", check_fields=False)
+    @classmethod
+    def validate_node_side_type(cls, v: str) -> None:
+        msg = f"You cannot update 'node_side_type' through NodeSettingsUpdate. \
```
koenvanderveen, line 80: you could add here that you should watch out to not have private data on a high side if you change it to low side

PR #8679, opened 2024-04-08
Peer health checks

shubham3121:
## Description
Please include a summary of the change, the motivation, and any additional context that will help others understand your PR. If it closes one or more open issues, [please tag them as described here](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

## Affected Dependencies
List any dependencies that are required for this change.

- track ping status, ping message and ping timestamp at NodePeer level
- add a utility track peer route healthcheck
- add a network service to find peer
- add an integration test for peers health check

## How has this been tested?
- Describe the tests that you ran to verify your changes.
- Provide instructions so we can reproduce.
- List any relevant details for your test configuration.

## Checklist
- [ ] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [ ] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [ ] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [ ] My changes are covered by tests

PR #8851, opened 2024-05-23
Fix concurrency issue with NodePeer update

shubham3121:
## Description
- Now, we are running peers health check periodically in background task (https://github.com/OpenMined/PySyft/pull/8679). This PR tries to get the list of online domains using the results of this peers health check feature instead of trying to login the domains as guest clients to check if they are online every time we call `sy.domains`.
- Fix concurrency issues when updating the same objects in the document store
- Using `ParitialSyftObject` for `update`


packages/syft/src/syft/service/network/network_service.py
```diff
@@ -87,13 +88,18 @@ def get_by_name(
     def update(
         self,
         credentials: SyftVerifyKey,
-        peer: NodePeer,
+        peer: NodePeer | NodePeerUpdate,
```
shubham3121: Let just use NodePeerUpdate, instead of NodePeer

packages/syft/src/syft/service/network/network_service.py
```diff
@@ -113,9 +119,7 @@ def create_or_update_peer(
         valid = self.check_type(peer, NodePeer)
         if valid.is_err():
             return SyftError(message=valid.err())
-        existing: Result | NodePeer = self.get_by_uid(
-            credentials=credentials, uid=peer.id
-        )
+        existing = self.get_by_uid(credentials=credentials, uid=peer.id)
         if existing.is_ok() and existing.ok():
             existing = existing.ok()
             existing.update_routes(peer.node_routes)
```
shubham3121: We can make use of NodePeerUpdate here

khoaguin:
image

shubham3121:
Nice work @khoaguin !!! 🎉 🔥 💯

PR #8866, opened 2024-05-29
[bugfix] Add error handlings for `Job.wait()`

khoaguin:
## Description
Add error handlings for `Job.wait()` so it does not hang forever when something wrong happens.  
Related issues:
- https://github.com/OpenMined/Heartbeat/issues/1376
- https://github.com/OpenMined/Heartbeat/issues/1301


khoaguin:
Hi @IrinaMBejan and @shubham3121, I made some changes to the `job.wait()` and `ActionObject.get()`:
- For now if a DS does not have mock data execution permission, but the DO approves the code request, then the error message will be like below 
image

- If the user has the permission to run the code on mock, but the code contains error, then the result of `job.wait()` and `job2.result` will be like below
Screenshot from 2024-06-03 16-26-40

Note that now, `job2` will be an `Err` instead of an `ActionObject` that wraps `Err` like in `dev`

- If everything is fine, then we get back the result normally like below 
image

koenvanderveen (approved):

PR #8883, opened 2024-06-02
Refactor: syft error handling

tcp:
## Description
Improves error handling in PySyft by implementing our own result and custom exception.

Proposed changelist for this PR:

- [x] Our own result implementation
- [x] Our own custom error type (SyftException)
- [x] Exception filtering
- [x] Delete PySyftException
#8919
- [x] Client and server side handling of service results
#9012
- [x] Update BaseStash and BaseUIDStoreStash
#9013
- [x] Refactor services and stashes
#9018 
- [x] Fully integrate with SyftResponseMessage
  - [x] SyftSuccess
  - [x] SyftError
  - [x] Remove `SyftException` in service/response.py
  - [x] Other `SyftResponseMessage` subclasses
- [x] Tests are green
- [x] Notebooks are green
- [x] Integration tests are green
## Related PRs
- #8919 
- #9012
- #9013
- #9018 

## Notable changes (TODO)
- {client,server}-side handling of service results
- error handling of client-side methods

## How has this been tested?
- [x] Added new tests for Result
- [x] Added new tests for SyftException
- [x] Added tests for ExceptionFilter
- [x] Updated tests
- [x] Current tests are green

## Checklist


packages/syft/src/syft/types/result.py
```diff
@@ -0,0 +1,98 @@
...
+            raise self.value
+        raise TypeError("Error is not a BaseException")
+
+
+Result: TypeAlias = Ok[T] | Err[E]
+
+
+def catch(
+    *exceptions: type[BE],
+) -> Callable[[Callable[..., T]], Callable[..., Result[T, BE]]]:
```
eelcovdw: Can use Paramspec here instead of ... to keep the params as typehints, else they get replaced by a general ... in IDE

https://peps.python.org/pep-0612/

```
P = ParamSpec("P")

def catch(
    *exceptions: type[BE],
) -> Callable[[Callable[P, T]], Callable[P, Result[T, BE]]]:
```

packages/syft/src/syft/types/errors.py
```diff
@@ -0,0 +1,51 @@
...
+]
+
+
+class SyftError(Exception):
+    __match_args__ = "code"
+
+    def __init__(
+        self,
+        message: str,
+        code: SyftErrorCodes,
```
eelcovdw: I think this `code` field is a bit of an antipattern compared to the standard python way of doing it. Instead of `SyftError(code="usercode-bad-input-policy")`, we should subclass:

```
class SyftException(Exception): ...

class InputPolicyException(SyftException): ...
```

packages/syft/src/syft/types/errors.py
```diff
@@ -0,0 +1,51 @@
...
+    "serde-deserialization-error",
+    "stash-error",
+    "worker-invalid-pool",
+    "usercode-not-approved",
+    "usercode-bad-input-policy",
+    "usercode-bad-output-policy",
+]
+
+
+class SyftError(Exception):
```
eelcovdw: Perhaps `SyftException` to avoid shadowing SyftSuccess/SyftError?

packages/syft/src/syft/types/errors.py
```diff
@@ -0,0 +1,51 @@
...
+    "serde-serialization-error",
+    "serde-deserialization-error",
+    "stash-error",
+    "worker-invalid-pool",
+    "usercode-not-approved",
+    "usercode-bad-input-policy",
+    "usercode-bad-output-policy",
+]
+
+
```
eelcovdw, line 20: I think having our own exception types is a really good change over the current Result type, it moves our error handling a lot to the standard python way of doing it.

I think this is something we should zoom in on as main part of the error handling re-vamp: this is how you're supposed to specify and propagate errors in Python, the result is just some nice on top to reduce boilerplate big try-catch blocks

packages/syft/src/syft/types/result.py
```diff
@@ -0,0 +1,98 @@
...
+        if isinstance(self.value, BaseException):
+            raise self.value
+        raise TypeError("Error is not a BaseException")
+
+
+Result: TypeAlias = Ok[T] | Err[E]
+
+
+def catch(
+    *exceptions: type[BE],
```
eelcovdw: A bit worried about always having to pass all exception types, this can get verbose very fast. 

Example: a function like `UserCodeService._call` calls 10+ other internal syft functions, if all return their own error types, we need a `@catch` with 10+ different exception types above the `_call` function

packages/syft/src/syft/types/result.py
tcp: Yup, I have pushed it in the other branch, I'll move it here ;)

packages/syft/src/syft/types/errors.py
tcp: Hey @eelcovdw :)

This is not the final name. `SyftException` also exists... and so does `PySyftException`! By cleaning up, my aim is to take over `SyftException`.

Let's see :)

packages/syft/src/syft/types/result.py
tcp: It is also one of my concerns.

With `catch`, we don't necessarily have to carry all exceptions from one level to the other. However, if we're not carrying the exceptions over, why are we using result anyway? :)

One option is to carry "major" subclasses that groups many types of exceptions. For example, `@catch(StashException, UserCodeException)` instead of the other 10+ exceptions that will be thrown down the stack.

packages/syft/src/syft/types/errors.py
tcp: Hey @eelcovdw! Thanks for the comment.

That's one of the approaches we've discussed so far, and is probably how we'll do it.

packages/syft/src/syft/types/result.py
jcardonnet: An alternative to exception subclasses is using Enums to define custom exception groups (as Rust does)
This is more flexible than a class tree because it let's you predefine lists of specific exceptions which can be referenced by a single name while still allowing static type checking.
Something like

```
class StashErrors(Enum):
    VALIDATION_ERROR = ValidationError
    DATABASE_ERROR = DatabaseError
    NETWORK_ERROR = NetworkError
    UNKNOWN_ERROR = Exception
```

packages/syft/src/syft/types/errors.py
```diff
@@ -0,0 +1,55 @@
+# stdlib
+from typing import Any
+
+# relative
+from ..service.context import AuthedServiceContext
+from ..service.user.user_roles import ServiceRole
+
+
```
madhavajay, line 28: How do we send this between worker and backend or other nodes?
Should we consider having a single message exception and then keeping the public / private as kind of twin somewhere?

koenvanderveen (approved):

PR #8894, opened 2024-06-10
[ux] fix tqdm prints multiple lines when `client.upload_dataset`

khoaguin:
Issue: https://github.com/OpenMined/Heartbeat/issues/1274

If the dataset contains multiple assets, tqdm prints multiple lines:
Screenshot from 2024-06-10 09-31-19

The fix makes it only one line:
a

### DIFF
--- a/packages/syft/src/syft/client/domain_client.py
+++ b/packages/syft/src/syft/client/domain_client.py
@@ -123,25 +123,35 @@ class DomainClient(SyftClient):
             )
             prompt_warning_message(message=message, confirm=True)
 
-        for asset in tqdm(dataset.asset_list, colour="green"):
-            print(f"Uploading: {asset.name}")
-            try:
-                twin = TwinObject(
-                    private_obj=asset.data,
-                    mock_obj=asset.mock,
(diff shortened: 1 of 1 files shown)


packages/syft/src/syft/client/domain_client.py
```diff
@@ -123,25 +123,35 @@ def upload_dataset(self, dataset: CreateDataset) -> SyftSuccess | SyftError:
...
+        ) as pbar:
+            for asset in dataset.asset_list:
+                try:
+                    twin = TwinObject(
+                        private_obj=asset.data,
+                        mock_obj=asset.mock,
+                        syft_node_location=self.id,
+                        syft_client_verify_key=self.verify_key,
+                    )
+                    twin._save_to_blob_storage()
```
khoaguin, line 137: We want to only call `_save_to_blob_storage()` when the object's size is more than 16 Mb. 
Link to the issue: https://github.com/OpenMined/Heartbeat/issues/1274

shubham3121 (approved):
LGTM !! ✅

PR #8896, opened 2024-06-10
fix mongo issues

koenvanderveen:


packages/syft/src/syft/types/twin_object.py
```diff
@@ -87,9 +87,11 @@ def _save_to_blob_storage(self) -> SyftError | None:
...
         )
-        # self.mock_obj._set_obj_location_(
-        #     self.syft_node_location,
-        #     self.syft_client_verify_key,
-        # )
+        self.mock_obj._set_obj_location_(
+            self.syft_node_location,
+            self.syft_client_verify_key,
+        )
+        mock_store_res = self.mock_obj._save_to_blob_storage()
```
shubham3121: We need to validate if the blob_storage_entry_id has read permissions for mock objects. 
https://github.com/OpenMined/PySyft/pull/8896/files#diff-4ef06b0973fb9fa22d186235a17d570c846c26cb8dc87ed184dd470f04f81fc6L123

here if during action.set, if its a TwinObject, and the `mock_obj` has blob_storage_entry_id then we should add READ permission to it.

shubham3121 (dismissed):

packages/syft/src/syft/service/action/action_service.py
```diff
@@ -102,9 +102,12 @@ def _set(
 
         if isinstance(action_object, ActionObject):
             action_object.syft_created_at = DateTime.now()
+            action_object._clear_cache(clear_reprs=True)
```
shubham3121: Do we expect to clear repr in all cases ? I think we want to store the repr but truncate it if the repr is too large.

koenvanderveen: hmm yes maybe we can just truncate the repr indeed

madhavajay: Yes, lets set a max repr to like 100 characters or something.

teo-milea (approved):

PR #8897, opened 2024-06-11
Send small variables to storage without blob storage

khoaguin:
Issue: https://github.com/OpenMined/Heartbeat/issues/1274

- [x] Only save `ActionObject` to blob storage if it is bigger than a `min_size` variable (e.g. 16 MB)
- [x] The min size can be set with an environment variable (e.g. `MIN_SIZE_BLOB_STORAGE_MB`) so it can be configured when we change database implementations
- [x] Add tests for 2 cases: `ActionObject` bigger than `MIN_SIZE_BLOB_STORAGE_MB` and smaller than `MIN_SIZE_BLOB_STORAGE_MB`


packages/syft/src/syft/client/domain_client.py
```diff
@@ -134,6 +136,11 @@ def upload_dataset(self, dataset: CreateDataset) -> SyftSuccess | SyftError:
                         syft_node_location=self.id,
                         syft_client_verify_key=self.verify_key,
                     )
+                    serialized: bytes = serialize(twin, to_bytes=True)
+                    size_mb: float = sys.getsizeof(serialized) / 1024 / 1024
+                    if size_mb < 16:
```
madhavajay: Can we make this configurable so that the max size can be set in an environment variable for when we change database implementations and allow this to be configurable?

khoaguin: Yes sure, thanks Madhava. It should be that way

madhavajay:
onhold for now.

Merge this later after this pr:
https://github.com/OpenMined/PySyft/pull/8896

packages/syft/src/syft/service/action/action_object.py
```diff
@@ -508,44 +511,33 @@ def convert_to_pointers(
...
-                    syft_action_data=arg,
-                    syft_client_verify_key=api.signing_key.verify_key,
-                    syft_node_location=api.node_uid,
-                )
-                arg.syft_node_uid = node_uid
-                r = arg._save_to_blob_storage()
-                if isinstance(r, SyftError):
-                    logger.error(r.message)
-                arg = api.services.action.set(arg)
+    def process_arg(arg: ActionObject | Asset | UID | Any) -> Any:
```
shubham3121, line 514: 👍🏾

packages/syft/src/syft/service/blob_storage/util.py
```diff
@@ -0,0 +1,22 @@
...
+from typing import Any
+
+# relative
+from ...util.util import get_mb_serialized_size
+from ..metadata.node_metadata import NodeMetadata
+from ..metadata.node_metadata import NodeMetadataJSON
+
+
+def min_size_for_blob_storage_upload(metadata: NodeMetadata | NodeMetadataJSON) -> int:
+    if not isinstance(metadata, (NodeMetadata | NodeMetadataJSON)):
```
shubham3121: Why do we need this check ?

packages/syft/src/syft/service/action/action_object.py
```diff
@@ -851,9 +851,40 @@ def _save_to_blob_storage_(self, data: Any) -> SyftError | None:
...
+                self._clear_cache()
+            return SyftSuccess(
+                message=f"Saved action object {self.id} to the blob store"
+            )
+        except Exception as e:
+            print(
+                f"Failed to save action object {self.id} to the blob store. Error: {e}"
+            )
+
         self.syft_action_data_cache = data
```
shubham3121: This step should happen under `_save_to_blob_storage_`. Similarly, these three vars should also be set even if saving to memory.

```
self.syft_action_data_type = type(data)
self._set_reprs(data)
self.syft_has_bool_attr = hasattr(data, "__bool__")
```

All of these can still be set under the _save_to_blob_storage_ method beforehand (in the else statement)

packages/syft/src/syft/service/action/action_object.py
```diff
@@ -801,17 +793,25 @@ def reload_cache(self) -> SyftError | None:
...
         # relative
         from ...types.blob_storage import BlobFile
         from ...types.blob_storage import CreateBlobStorageEntry
 
+        api = APIRegistry.api_for(self.syft_node_location, self.syft_client_verify_key)
+        if api is None:
+            raise ValueError(
+                f"api is None. You must login to {self.syft_node_location}"
+            )
+        if not can_upload_to_blob_storage(data, api.metadata):
```
shubham3121: This check should be moved to else statement line:816. Since for BlobFile we have a different flow and for ActionDataEmpty we don't want to set any values.

shubham3121 (changes requested):
@khoaguin left some comments.

packages/syft/src/syft/service/blob_storage/util.py
khoaguin: yeah actually we don't need this. Thanks Shubham

khoaguin:
Thank you very much for the review, @shubham3121. I changed accordingly

shubham3121 (approved):

PR #8920, opened 2024-06-17
Allow User to set custom headers

IonesioJunior:
# Description
For certain use cases, it is crucial to enable users to pass custom headers in their requests. This pull request (PR) aims to implement a native solution that allows users to embed custom headers directly within Syft API calls. By doing so, we enhance the flexibility and security of the API interactions, catering to specific requirements such as authentication, content negotiation, and user-defined metadata.


jcardonnet (approved):
LGTM

PR #8929, opened 2024-06-18
implement `UserCode.__call__` and fix code_history

abyesilyurt:
- implement `UserCode.__call__`, so that Data Owner can do `admin_client.code_histories["user@email.org"].user_func[0]()`.
- `submit` creates code history.
- `request_code_execution` calls `submit` under the hood.

closes https://github.com/OpenMined/Heartbeat/issues/1499


packages/syft/src/syft/service/code/user_code_service.py
```diff
@@ -75,6 +75,19 @@ def _submit(
     ) -> Result[UserCode, str]:
         if not isinstance(code, UserCode):
             code = code.to(UserCode, context=context)  # type: ignore[unreachable]
+            result = self._post_user_code_transform_ops(context, code)
+            if isinstance(result, SyftError):
+                # if the validation fails, we should remove the user code status
```
koenvanderveen, line 80: does this also remove the version though?

koenvanderveen (approved):

packages/syft/src/syft/service/code/user_code_service.py
abyesilyurt: Code version is not created if `_post_user_code_transform_ops` fails, since creating a new version is the last step. So if that fails, no history created.

PR #8936, opened 2024-06-19
Function factories demo

koenvanderveen:


eelcovdw (approved):

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -1141,19 +1152,27 @@ def syft_function_single_use(
     )
 
 
+def replace_func_name(src: str, new_func_name: str) -> str:
+    pattern = r"\bdef\s+(\w+)\s*\("
```
madhavajay, line 1156: We should mark this as tech debt to update and use the ast library so its a bit more strict.

PR #8939, opened 2024-06-20
Distributed project result refactor (check all parties get result)

snwagh:
This adds a check to `DistributedProject.request_execution()` to return only when all domains get the result.

### DIFF
--- a/packages/syft/src/syft/service/enclave/enclave_service.py
+++ b/packages/syft/src/syft/service/enclave/enclave_service.py
@@ -310,6 +310,9 @@ class EnclaveService(AbstractService):
     ) -> Any:
         if not context.node or not context.node.signing_key:
             return SyftError(message=f"{type(context)} has no node")
+        
+        # if context.node.name == "italy-domain":
+        #     return SyftError(message="For testing purposes, italy-domain execution is blocked")
 
         code_service = context.node.get_service("usercodeservice")
         code: UserCode = code_service.get_by_service_name(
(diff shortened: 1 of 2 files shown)


packages/syft/src/syft/service/enclave/enclave_service.py
```diff
@@ -311,6 +311,9 @@ def request_execution(
         if not context.node or not context.node.signing_key:
             return SyftError(message=f"{type(context)} has no node")
 
+        # if context.node.name == "italy-domain":
+        #     return SyftError(message="For testing purposes, italy-domain execution is blocked")
+
```
itstauq, line 316: Should we remove this?

packages/syft/src/syft/service/project/distributed_project.py
```diff
@@ -166,17 +166,24 @@ def request_execution(self, blocking: bool = True) -> Any:
...
         result_parts = []
+        errors = []
         for client in self.clients.values():
             result = client.api.services.enclave.request_execution(
                 service_func_name=self.code.service_func_name
             )
             if isinstance(result, SyftError):
-                raise SyftException(result.message)
-            result_parts.append(result)
+                errors.append(result.message)
```
itstauq: We can also return early as soon as there is an error. This pattern will be more consistent with the rest of the codebase.

itstauq (approved):

itstauq:
@snwagh approved with some minor comments 🚀

rasswanth-s (approved):

PR #8943, opened 2024-06-20
Create EnclaveInstance using validators

snwagh:
Comments: 
* We are not setting the `SyftObject` `id` (base class) right now (commented out) and thus `node_uid` and `id` are different, let me know if I should change the behaviour.

Tested
* Ran `00-do-setup-domain.ipynb`

Closes [1510](https://github.com/OpenMined/Heartbeat/issues/1510)


packages/syft/src/syft/service/enclave/enclave.py
```diff
@@ -45,20 +45,25 @@ class EnclaveInstance(SyftObject):
...
+    def initialize_values(cls, values: dict[str, Any]) -> dict[str, Any]:
+        if "route" in values:
+            route = values["route"]
+            metadata = login_as_guest(url=route.host_or_ip, port=route.port).metadata
+            if not metadata:
+                raise SyftException("Failed to fetch metadata from the node")
+
+            values.update(
+                {
+                    # 'id': UID(metadata.id),
```
itstauq: `id` is the identifier for this object's instance in the owner domain, while `node_uid` identifies the Enclave node that this object points to.

itstauq: Since this comment implies the wrong thing, let's remove it to avoid any confusion.

itstauq (approved):
I've added a small comment. Overall, this PR looks great to me. Approved! 💪🏻

rasswanth-s:
Well done @snwagh

PR #8947, opened 2024-06-21
Fix new admin unable to list user code

kiendang:
~Fix permission in `UserCodeService.get_all` to allow `ServiceRole.ADMIN`, similar to [`UserService.get_all`](https://github.com/OpenMined/PySyft/blob/d5476da5b49df11ed903a6b3b2b74fe8a6d138d5/packages/syft/src/syft/service/user/user_service.py#L113-L116).~

Fix at the Store/Partition level so admins have permissions for all stashes/services.


kiendang:
eb07fc9a615a28fd13ad9d12f68eb2750eba56f7 fixed an issue that caused a [CI failure](https://github.com/OpenMined/PySyft/actions/runs/9833392421/job/27143546068?pr=8947) that's not specific to this PR. `__getattr__` should raise `AttributeError` instead of bare `Exception` so that `hasattr` returns `False` instead of raising an exception here https://github.com/OpenMined/PySyft/blob/8e1c7c0848bf8322634c0d9e6370f6cd0a5cd809/packages/syft/src/syft/service/queue/queue.py#L212.

kiendang:
Oops closed by mistake.

tcp (approved):

PR #8962, opened 2024-06-24
Reloadable uvicorn app

snwagh:
Run the different uvicorn servers as follows:
```
bash scripts/reloadable_run.sh --port 9081 --name "canada-domain"
bash scripts/reloadable_run.sh --port 9082 --name "italy-domain" 
bash scripts/reloadable_run.sh --port 9083 --name "canada-enclave" --node_type "enclave"
```

The `NodeHandler`'s are buggy so workaround is to directly login into the Nodes using `sy.login` as follows: 
```
do_canada_client = sy.login(email="info@openmined.org", password="changethis", 
                            url="http://0.0.0.0", port=CANADA_DOMAIN_PORT)
```
or 
```
do_italy_client = sy.login(email="info@openmined.org", password="changethis", 
                            url="localhost", port=ITALY_DOMAIN_PORT)
```

If you really need to use the NodeHandle, then we can do it this way:
```
from syft.orchestra import NodeHandle

canada_node = NodeHandle(
        node_type="domain",
        deployment_type="python",
        name="canada-domain",
        port=9081,
        url="localhost",
        node_side_type="high",
    )
```

### DIFF
--- a/packages/syft/src/syft/client/client.py
+++ b/packages/syft/src/syft/client/client.py
@@ -14,6 +14,8 @@ from typing import cast
 
 # third party
 from argon2 import PasswordHasher
+from cachetools import TTLCache
+from cachetools import cached
 from pydantic import field_validator
 import requests
 from requests import Response
@@ -200,6 +202,8 @@ class HTTPConnection(NodeConnection):
         return self.session_cache
 
     def _make_get(self, path: str, params: dict | None = None) -> bytes:
+        if params is None:
+            return self._make_get_no_params(path)
         url = self.url.with_path(path)
         response = self.session.get(
             str(url),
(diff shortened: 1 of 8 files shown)


scripts/reloadable_run.sh
```diff
@@ -0,0 +1,58 @@
...
+# And get the NodeHandlers as follows (note that the handlers point to the correct objects but are buggy in their attributes)
+# canada_domain = sy.orchestra.launch(port=9081, deploy_to="remote")
+# italy_domain = sy.orchestra.launch(port=9082, deploy_to="remote")
+# canada_enclave = sy.orchestra.launch(port=9083, deploy_to="remote")
+
+# Default values
+PORT=9694
+HOST="0.0.0.0"
+FACTORY="--factory"
+RELOAD="--reload"
```
madhavajay: I tried this in jupyter and it seemed to cause a lot of weird behavior, where different code versions were being loaded at the same time, possibly because of the way jupyter manages threads of memory, where as when we do this in the container the entire process restarts so there is no mismatch between code versions.

I would love it if you have solved this! Has it been working for you?

snwagh: @madhavajay It worked for me but I'm making a few changes

- Will point it to `dev`, since it is more generally and widely useful than Enclave stuff
- Will remove the script so that it works in python (so that all existing notebooks should work as is)

Will ping you when I have this ready.

snwagh: For some reason, unknown to mankind, this has moved to [this PR](https://github.com/OpenMined/PySyft/pull/8971)

PR #8974, opened 2024-06-26
Add linter and error for code submissions that use global scope variables / methods

khoaguin:
[Issue](https://github.com/OpenMined/Heartbeat/issues/1282)


khoaguin:
image

khoaguin:
Hi @IrinaMBejan, currently the fix is like below: if the user put a `global` inside the code, it will display a warning. And if it's submitted to the server, then there will be an error showing the exception 
image

khoaguin:
Hi @IrinaMBejan, currently the fix is like below
image

Also the way we do it now is to put the check in the `syft_function` decorator to run it locally first before submitting to the server

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -1285,26 +1294,48 @@ def generate_unique_func_name(context: TransformContext) -> TransformContext:
     return context
 
 
-def process_code(
-    context: TransformContext,
+def _check_global(code_tree: ast.Module) -> GlobalsVisitor | SyftWarning:
```
shubham3121: we can remove `_` from function name, possibly rename to `check_for_global_vars` ?

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -1285,26 +1294,48 @@ def generate_unique_func_name(context: TransformContext) -> TransformContext:
...
+    try:
+        v.visit(code_tree)
+    except Exception:
+        raise SyftException(
+            "Your code contains (a) global variable(s), which is not allowed"
+        )
+    return v
+
+
+def _parse_code(raw_code: str) -> ast.Module | SyftWarning:
```
shubham3121: Remove `_` from function name. Let's move `parse_code` and `check_for_global_vars` to `service/code/utils.py`

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -1329,6 +1360,25 @@ def process_code(
     return unparse(wrapper_function)
 
 
+def process_code(
```
shubham3121, line 1372: 👍🏾

shubham3121 (changes requested):
@khoaguin left a few minor comments, otherwise LGTM !! 💯

packages/syft/src/syft/service/code/user_code.py
khoaguin: Thanks for the review @shubham312. Done!

shubham3121 (approved):

PR #8979, opened 2024-06-26
Split EnclaveService into DomainEnclaveService and EnclaveService

itstauq:
## Description

Closes OpenMined/Heartbeat#1556

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/packages/syft/src/syft/service/enclave/domain_enclave_service.py
+++ b/packages/syft/src/syft/service/enclave/domain_enclave_service.py
@@ -1,0 +1,220 @@
+# stdlib
+import itertools
+from typing import Any
+
+# relative
+from ...serde.serializable import serializable
+from ...service.action.action_permissions import ActionObjectPermission
+from ...service.action.action_permissions import ActionPermission
+from ...service.network.routes import NodeRouteType
+from ...service.response import SyftError
+from ...service.response import SyftSuccess
+from ...service.user.user_roles import ADMIN_ROLE_LEVEL
+from ...service.user.user_roles import DATA_SCIENTIST_ROLE_LEVEL
+from ...store.document_store import DocumentStore
+from ...types.uid import UID
+from ..code.user_code import UserCode
+from ..context import AuthedServiceContext
+from ..service import AbstractService
+from ..service import service_method
+from .enclave import EnclaveInstance
+from .enclave_stash import EnclaveInstanceStash
+
+
+@serializable()
+class DomainEnclaveService(AbstractService):
+    """Contains service methods for Domain -> Enclave communication."""
+
+    store: DocumentStore
+    stash: EnclaveInstanceStash
+
+    def __init__(self, store: DocumentStore) -> None:
+        self.store = store
+        self.stash = EnclaveInstanceStash(store=store)
+
+    @service_method(
+        path="enclave.add",
+        name="add",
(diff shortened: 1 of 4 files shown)


packages/syft/src/syft/service/enclave/domain_enclave_service.py
```diff
@@ -0,0 +1,220 @@
...
+
+    store: DocumentStore
+    stash: EnclaveInstanceStash
+
+    def __init__(self, store: DocumentStore) -> None:
+        self.store = store
+        self.stash = EnclaveInstanceStash(store=store)
+
+    @service_method(
+        path="enclave.add",
```
BrendanSchell, line 36: I wonder about the pathing. I think having the user-facing ones like enclave.add maintain their pathing as `enclave.*` makes sense. Maybe the other service methods that are called only from distributed project should be in a different namespace so it's more obvious which ones are expected to be called from a notebook directly vs via distributed project.

BrendanSchell (approved):
LGTM - one comment, but I can go either way with the naming on it and leave that up to you

packages/syft/src/syft/service/enclave/domain_enclave_service.py
rasswanth-s: That is a valid point, I am also  not fully sure, for now, we could leave it and we could think about it, when we shift to Old Project class ?

rasswanth-s (approved):
Great work @itstauq

packages/syft/src/syft/service/enclave/domain_enclave_service.py
itstauq: +1 to this. I am not able to come up with a decent name though. Any suggestions will be appreciated. Otherwise we can park this one for later.

snwagh: Changing to `enclave_internal.*` for internal paths and `enclave.*` for open paths?

packages/syft/src/syft/service/enclave/enclave_service.py
```diff
@@ -148,74 +48,6 @@ def setup_enclave_for_code_execution(
         result = context.node.get_service("usercodeservice").submit(root_context, code)
```
snwagh, line 48: I know we just merged the other PR but should we consider changing the `root_context` to using `node.root_client` for uniformity? Same comment to the other two internal enclave service_method's: `enclave.setup_enclave_for_code_execution`, `enclave.upload_input_data_for_code`, and `enclave.execute_code`

snwagh (commented):
Minor comments but looks good.

packages/syft/src/syft/service/enclave/domain_enclave_service.py
itstauq: I like `enclave_internal.*`. It might not work if we decide to give direct Enclave access to the DS though.

packages/syft/src/syft/service/enclave/enclave_service.py
itstauq: We ideally don't want to use the client inside the service methods. This was a temporary workaround for v1. In the next iteration we'll find a way to create jobs without using clients.

PR #8983, opened 2024-06-27
Show requesting domain (for DistributedProject as well as base Requests)

snwagh:
This changes the output to show the domain name at which the request was made. Tagging for PM review since I change the base object `Request`'s repr. The output looks as follows (Requested on added):

Screenshot 2024-06-27 at 10 51 39 AM

The implementation may not be optimal since I am making the API call (I can take a look at the relation between `_coll_repr_` and `_repr_html_` to see if I can do this efficiently if that is promising).

### DIFF
--- a/packages/syft/src/syft/service/project/distributed_project.py
+++ b/packages/syft/src/syft/service/project/distributed_project.py
@@ -14,8 +14,9 @@ from ...client.api import NodeIdentity
 from ...client.client import SyftClient
 from ...client.client import SyftClientSessionCache
 from ...types.uid import UID
-from ...util import options
-from ...util.colors import SURFACE
+from ...util.notebook_ui.components.tabulator_template import (
+    build_tabulator_table_with_data,
+)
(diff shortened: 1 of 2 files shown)


itstauq (approved):
LGTM

itstauq (changes requested):
Edit: On a closer look, I was wondering if we could directly patch the Request objects in DistributedProject.requests property? Otherwise merging this will change the Request object for existing single projects as well.

snwagh:
@itstauq Since the DistributedProjects uses the same objects inside it, I was wondering if this is a better fix as long as we run it by PM. It changes the Request object for the single projects too but it's just a change to the __repr__ so not a conflicting change IMO.

snwagh:
@itstauq To address your comment, added another commit that does not create a breaking change.

madhavajay:
Closing for now, we will move the domain information to Project code in this PR: https://github.com/OpenMined/PySyft/pull/8999

PR #8984, opened 2024-06-27
fix syft function result repr

abyesilyurt:
`has_result_read_permission` was incorrectly set in the code.call endpoint, leading it to return `result.as_empty()` instead of the actual result. Consequently, the output appeared as None.

Closes https://github.com/OpenMined/Heartbeat/issues/1427

before
Screenshot 2024-07-08 at 07 32 38

after
Screenshot 2024-07-08 at 07 32 55


packages/syft/src/syft/service/action/action_object.py
```diff
@@ -1969,30 +1970,30 @@ def _repr_markdown_(self, wrap_as_python: bool = True, indent: int = 0) -> str:
         elif not self.is_twin:
             res = "Pointer"
 
-        if isinstance(self.syft_action_data_cache, ActionDataEmpty):
+        if isinstance(self.syft_action_data, ActionDataEmpty):
```
eelcovdw: Not sure about these changes, not using the cache here changes the behaviour

'if the cache is empty use the cached repr, else get repr from action data'

abyesilyurt: `syft_action_data` handles the caching, so it shouldn't be a problem.

abyesilyurt: changed the implementation

teo-milea (approved):

PR #8985, opened 2024-06-27
Add migrations for latest UserCode and CodeHistory

BrendanSchell:
## Description
Adds missing migrations from https://github.com/OpenMined/PySyft/pull/8969 for UserCode and CodeHistory.

## Affected Dependencies

## How has this been tested?
Tested locally using `obj.migrate_to(version_no)`.

## Checklist


packages/syft/src/syft/service/code/user_code.py
```diff
@@ -1933,3 +1933,15 @@ def migrate_usercode_v5_to_v4() -> list[Callable]:
         drop("origin_node_side_type"),
         drop("l0_deny_reason"),
     ]
+
+
+@migrate(UserCodeV5, UserCode)
```
rasswanth-s: NIT: could we delete UserCodeV4 from the codebase?

packages/syft/src/syft/service/code_history/code_history.py
```diff
@@ -171,3 +175,15 @@ def _repr_html_(self) -> str | None:
             "icon": None,
         }
         return build_tabulator_table_with_data(rows, metadata)
+
+
+@migrate(CodeHistoryV2, CodeHistory)
```
rasswanth-s, line 180: NIT: could we delete CodeHistory V1 from the codebase?

rasswanth-s: I see it is already deleted.

packages/syft/src/syft/service/code/user_code.py
rasswanth-s: and also its corresponding migrations.

rasswanth-s: as since we need to maintain migrations for only last two versions.

packages/syft/src/syft/service/code/user_code.py
```diff
@@ -1933,3 +1933,15 @@ def migrate_usercode_v5_to_v4() -> list[Callable]:
         drop("origin_node_side_type"),
```
rasswanth-s: Can we also add migrations for SubmitUserCode ?

BrendanSchell: ya I can. I didn't see any migrate decorators in the codebase for any client-side code so that's why I wasn't sure, but it looks like we call migrations on all args and kwargs so I'll add.

packages/syft/src/syft/service/code/user_code.py
BrendanSchell: ya I can, but it kind of seems like bad practice in a way to lose the ability to migrate from old versions. Especially since all of these updates to UserCode are within such a short timeframe (V4 was just two weeks ago), migrations would be broken from UserCode from less than a month ago with this approach. Going to raise in slack since I think it would be good to understand why we do it this way.

snwagh (approved):
LGTM

shubham3121 (approved):
LGTM !! Fixed changes to UserCode version

rasswanth-s (approved):
Great work @BrendanSchell  and @shubham3121

PR #9000, opened 2024-07-01
show asset details on code repr

abyesilyurt:
We now show assets in UserCode and show extra data in the assets.
image

closes https://github.com/OpenMined/Heartbeat/issues/1284


packages/syft/src/syft/service/code/user_code.py
```diff
@@ -744,31 +744,49 @@ def byte_code(self) -> PyCodeObject | None:
...
     @property
-    def assets(self) -> list[Asset]:
-        # relative
-        from ...client.api import APIRegistry
-
-        api = APIRegistry.api_for(self.node_uid, self.syft_client_verify_key)
-        if api is None:
-            return SyftError(message=f"You must login to {self.node_uid}")
+    def assets(self) -> list[Asset] | SyftError:
+        if not self.input_policy:
```
eelcovdw, line 749: This will be an issue with custom input policies, they are None until the code is approved.

Alternative is to look at input_policy_input_kwargs, but I don't know if we enforce some structure on those when using custom input policies

eelcovdw (approved):

PR #9030, opened 2024-07-10
add server and client traceback to syftexception

abyesilyurt:
SyftException repr is updated to show both client side and server side traces, depending on the role.

1. DS: Widget is first shown as collapsed and a public message is displayed. When expanded client side traceback is displayed.
2. DO or Dev mode: Widget is first shown as expanded, private message is displayed. When expanded both client and server side traceback is displayed.

Screenshot 2024-07-10 at 10 04 41

Screenshot 2024-07-10 at 10 04 36

Co-authored-by: @eelcovdw


packages/syft/src/syft/client/api.py
```diff
@@ -583,17 +582,6 @@ def jobs(self) -> list[Job] | SyftError:
         result = self.make_call(api_call=api_call)
         return self.post_process_result(api_call.path, result)
 
-    def post_process_result(self, path, result):
```
abyesilyurt: duplicate

PR #9049, opened 2024-07-15
Added notification_enabled attribute to ServerSettings

snwagh:
Closes [1612](https://github.com/OpenMined/Heartbeat/issues/1626)

Tested as: 

https://github.com/user-attachments/assets/38154ef3-cdc7-41a4-864a-b8675785e60f


packages/syft/src/syft/service/settings/settings.py
```diff
@@ -98,8 +165,29 @@ def _repr_html_(self) -> Any:
...
+                <p><strong>Notifications enabled: </strong>{notification_print_str}</p>
                 <p><strong>Admin email: </strong>{self.admin_email}</p>
             </div>
 
             """
+
+
+@migrate(ServerSettingsV1, ServerSettings)
+def migrate_server_settings_v1_to_v2() -> list[Callable]:
+    return []
```
shubham3121: I think we should set a default value to `notifications_enabled` of `False` when upgrading the old ServerSettings version to new one, since `notifications_enabled` is a required field in the new version (and old version will not have it)

snwagh: Right, makes sense. Also, you meant the `ServerSettingsUpdate` class (and it's old version) as opposed to the `ServerSettings` class?

shubham3121: No I meant `ServerSettings` class itself. Cause when we migrate old ServerSettings data to the new version, in that case, when we create the new ServerSettings class, we need to specify some value for the `notifications_enabled` attr, but since the old ServerSettings version doesn't have this attr defined, we will fail to migrate to the new ServerSetting class. So to fix this we need to define a migration that adds the `notification_enabled` attr with some default value to it.

packages/syft/src/syft/server/server.py
```diff
@@ -1585,6 +1585,7 @@ def create_initial_settings(self, admin_email: str) -> ServerSettings | None:
                     show_warnings=self.enable_warnings,
                     association_request_auto_approval=self.association_request_auto_approval,
                     default_worker_pool=get_default_worker_pool_name(),
+                    notifications_enabled=True,
```
shubham3121: Quick Question: We set `notifications_enabled` True here but we're not setting the `set_notifier_active_to_true` ? We're setting this the Settings update service https://github.com/OpenMined/PySyft/pull/9049/files#diff-38e8350faa70ff82eb9473d049ab0f1142ca23e9fad17094f6361ea6d7a852daR121 . Which means, that user will have to call the settings.update service again to enable notifier.

IrinaMBejan:
If I am following the conversation - I believe the notifications should be False by default. Setting it to true requires plugging in the credentials, so this could be quite confusing to users if they are active, but do not work

packages/syft/src/syft/server/server.py
snwagh: Good catch. The way the code is structured is that every time the settings are viewed, the `notification_enabled` is set to the appropriate value from the notifier stash so the defaults might be less critical.

IonesioJunior (approved):

PR #9052, opened 2024-07-16
ADD reset password api

IonesioJunior:
This PR adds two features

1. Password reset feature if you forget your password (`[your_client].account.set_password([new_password])` is to change password)
3. Ability to limit total number of daily notifications


packages/syft/src/syft/service/user/user.py
```diff
@@ -68,10 +72,65 @@ class User(SyftObject):
...
+    __attr_unique__ = ["email", "signing_key", "verify_key"]
+    __repr_attrs__ = ["name", "email"]
+
+
+@migrate(UserV1, User)
+def migrate_server_user_update_v1_current() -> list[Callable]:
+    return [
+        make_set_default("reset_token", None),
+        make_set_default("reset_token_date", None),
+    ]
```
BrendanSchell, line 120: should `__attr_searchable__` be updated here too to add reset_token on upgrade?

packages/syft/src/syft/service/user/user.py
```diff
@@ -68,10 +72,65 @@ class User(SyftObject):
...
+
+
+@migrate(User, UserV1)
+def migrate_server_user_downgrade_current_v1() -> list[Callable]:
+    return [drop("reset_token"), drop("reset_token_date")]
```
BrendanSchell: similar comment to above with dropping reset_token from `__attr_searchable__`

packages/syft/src/syft/service/user/user_service.py
```diff
@@ -84,6 +91,212 @@ def create(
...
+    @service_method(
+        path="user.forgot_password", name="forgot_password", roles=GUEST_ROLE_LEVEL
+    )
+    def forgot_password(
+        self, context: AuthedServiceContext, email: str
+    ) -> SyftSuccess | SyftError:
+        result = self.stash.get_by_email(credentials=context.credentials, email=email)
+        # Isn't a valid email
+        if result.is_err():
+            return SyftSuccess(
```
BrendanSchell: 1. Is this a success because we don't want someone to be able to infer if someone is a member of the server or not by spamming? 
2. It looks like we reuse the same message for the SyftSuccess each time so maybe we can just have that syft success as a var and return that so the messages don't drift (the last message is slightly different from this one).

packages/syft/src/syft/service/user/user_service.py
```diff
@@ -84,6 +91,212 @@ def create(
...
+        self, token_config: PwdTokenResetConfig
+    ) -> str:
+        valid_characters = ""
+        if token_config.ascii:
+            valid_characters += string.ascii_letters
+
+        if token_config.numbers:
+            valid_characters += string.digits
+
+        # valid_characters = string.ascii_letters + string.digits
```
BrendanSchell: I assume this is leftover and can be removed?

packages/syft/src/syft/service/settings/settings.py
```diff
@@ -176,22 +225,138 @@ def _repr_html_(self) -> Any:
...
+    association_request_auto_approval: bool
+    eager_execution_enabled: bool = False
+    default_worker_pool: str = DEFAULT_WORKER_POOL_NAME
+    welcome_markdown: HTMLObject | MarkdownDescription = HTMLObject(
+        text=DEFAULT_WELCOME_MSG
+    )
+    notifications_enabled: bool
+    pwd_token_config: PwdTokenResetConfig = PwdTokenResetConfig()
+
+    def _repr_html_(self) -> Any:
```
BrendanSchell, line 213: I think we can remove the `_repr_html_` method from ServerSettingsV2 right?

packages/syft/src/syft/service/notification/email_templates.py
```diff
@@ -22,6 +23,88 @@ def email_body(notification: "Notification", context: AuthedServiceContext) -> s
...
+                    margin-top: 20px;
+                }
+            </style>
+        </head>"""
+        body = f"""<body>
+            <div class="container">
+                <h1>Password Reset</h1>
+                <p>Hello,</p>
+                <p>We received a request to reset your password. Your new temporary token is:</p>
+                <h1>{user.reset_token}</h1>
```
BrendanSchell, line 99: There would be a follow-up action after this where the user needs to call `user.reset_password` with this token and the new password right? Should we make that instruction explicit in this template?

packages/syft/src/syft/service/notifier/notifier.py
```diff
@@ -161,6 +170,9 @@ class NotifierSettings(SyftObject):
     email_username: str | None = ""
     email_password: str | None = ""
 
+    email_activity: dict[str, dict[SyftVerifyKey, UserNotificationActivity]] = {}
```
BrendanSchell, line 205: should we add migrations for NotifierSettings?

BrendanSchell (changes requested):
Great work on this Ionesio - looks really good! Just a few small things to address. Also a test for the password reset flow would be nice too.

packages/syft/src/syft/service/notifier/notifier.py
IonesioJunior: Yup!! Thanks for notificing that. I've added them in [74f1b9d](https://github.com/OpenMined/PySyft/pull/9052/commits/74f1b9da8124b0468a2c59fca716d17e103f8916)

packages/syft/src/syft/service/settings/settings.py
IonesioJunior: Done! [74f1b9d](https://github.com/OpenMined/PySyft/pull/9052/commits/74f1b9da8124b0468a2c59fca716d17e103f8916)

packages/syft/src/syft/service/user/user_service.py
IonesioJunior: Yup! Removed at: [74f1b9d](https://github.com/OpenMined/PySyft/pull/9052/commits/74f1b9da8124b0468a2c59fca716d17e103f8916)

packages/syft/src/syft/service/user/user_service.py
IonesioJunior: 1 - Yup
2 - Good catch! I've added a new variable on top of this service and used it everywhere.

packages/syft/src/syft/service/user/user.py
IonesioJunior: Yup, solved it in  [74f1b9d](https://github.com/OpenMined/PySyft/pull/9052/commits/74f1b9da8124b0468a2c59fca716d17e103f8916)

packages/syft/src/syft/service/user/user.py
IonesioJunior: Solved it in [74f1b9d](https://github.com/OpenMined/PySyft/pull/9052/commits/74f1b9da8124b0468a2c59fca716d17e103f8916)

packages/syft/src/syft/service/notification/email_templates.py
```diff
@@ -22,6 +23,91 @@ def email_body(notification: "Notification", context: AuthedServiceContext) -> s
         return ""
 
 
+@serializable(canonical_name="PasswordResetTemplate", version=1)
```
snwagh, line 26: Curious: Is this an alternative class versioning we do?

packages/syft/src/syft/service/user/user_service.py
```diff
@@ -84,6 +91,207 @@ def create(
...
+        self, context: AuthedServiceContext, email: str
+    ) -> SyftSuccess | SyftError:
+        success_msg = (
+            "If the email is valid, we sent a password "
+            + "reset token to your email or a password request to the admin."
+        )
+        result = self.stash.get_by_email(credentials=context.credentials, email=email)
+        # Isn't a valid email
+        if result.is_err():
+            return SyftSuccess(message=success_msg)
```
snwagh, line 107: Should this be a `SyftError`?

packages/syft/src/syft/service/user/user_service.py
```diff
@@ -84,6 +91,207 @@ def create(
...
+
+            result = method(context=root_context, notification=message)
+            if isinstance(result, SyftError):
+                return result
+
+            return SyftSuccess(message=success_msg)
+
+        # Email notification is Enabled
+        # Therefore, we can directly send a message to the user with its new password.
+        message = CreateNotification(
```
snwagh: Should this be inside an else clause of `if not_receive_emails:`

packages/syft/src/syft/service/user/user_service.py
```diff
@@ -84,6 +91,207 @@ def create(
...
+        # If notifier service is not enabled.
+        if not_receive_emails:
+            message = CreateNotification(
+                subject="You requested password reset.",
+                from_user_verify_key=root_key,
+                to_user_verify_key=user.verify_key,
+                linked_obj=link,
+            )
+
+            method = context.server.get_service_method(NotificationService.send)
```
snwagh, line 149: If `User` notifications are off, it seems that we try to send notifications to the `User` and the `admin`. Should we send only to the `admin`?

packages/syft/src/syft/service/user/user_service.py
```diff
@@ -84,6 +91,207 @@ def create(
...
+        time_difference = now - user.reset_token_date
+
+        # If token expired
+        expiration_time = context.server.settings.pwd_token_config.token_exp_min
+        if time_difference > timedelta(minutes=expiration_time):
+            return invalid_token_error
+
+        if not validate_password(new_password):
+            return SyftError(
+                message="Your new password must have at least 8 \
```
snwagh, line 257: Note: I think the `validate_password` does not check for these constraints. 

Also: Given that we generate the random password using Ascii and digits, there may be a possibility that random passwords don't follow these guidelines.

packages/syft/src/syft/service/user/user_service.py
```diff
@@ -84,6 +91,207 @@ def create(
...
+            return SyftError(
+                message=(
+                    f"Failed to update user with UID: {uid}. Error: {str(result.err())}"
+                )
+            )
+
+        return user.reset_token
+
+    @service_method(
+        path="user.reset_password", name="reset_password", roles=GUEST_ROLE_LEVEL
```
snwagh, line 225: Does the User have some method to generate a `token`? Otherwise they might not be able to use this endpoint right?

snwagh (commented):
Generally looks good to me, I added some comments on parts I was a bit confused on.

packages/syft/src/syft/service/notification/email_templates.py
IonesioJunior: Yes, I believe this is the next step in migrating classes. However, it doesn't seem to be compatible with all classes yet. I tried it with other classes, but it didn't work as expected.

packages/syft/src/syft/service/user/user_service.py
IonesioJunior: The goal is to avoid disclosing too much information, which could allow attackers to infer whether someone is a member of the server through spam. So, in this case, we always return a success message, saying that if the email is valid, a message was sent to it.

packages/syft/src/syft/service/user/user_service.py
IonesioJunior: Oh, good catch! During the development process, @jcardonnet and I thought it would be reasonable to also notify the account owner about any password change attempts. This way, if there is a malicious attempt, the account owner can spot it and take proper action.

packages/syft/src/syft/service/user/user_service.py
IonesioJunior: The token used here is the one sent to the account owner's email or provided by the node admin in the previous step (during the forgot_password API call).

packages/syft/src/syft/service/user/user_service.py
IonesioJunior: We only generate temporary password reset tokens. The password itself is only set by the account owner.

packages/syft/src/syft/service/notification/email_templates.py
snwagh: Right, I noticed the PR had both. I'll try these out as well, thanks!

packages/syft/src/syft/service/user/user_service.py
snwagh: Ah, I see, the flow makes more sense now. Thanks!

packages/syft/src/syft/service/user/user_service.py
IonesioJunior: Yes, this approach also allows us to use a single return statement for success, reducing the risk of inconsistent behavior in both scenarios. Thanks for pointing that out. ❤️

IonesioJunior: Fixed by: [4322021](https://github.com/OpenMined/PySyft/pull/9052/commits/43220215f24f0fc8b1b174ef28d5a26c3973dfda)

packages/syft/src/syft/service/user/user.py
```diff
@@ -112,12 +112,23 @@ def migrate_server_user_update_v1_current() -> list[Callable]:
     return [
         make_set_default("reset_token", None),
         make_set_default("reset_token_date", None),
+        make_set_default(
```
BrendanSchell, line 116: do you have to do the drop first like in the reverse migration?

BrendanSchell (approved):
Great work on this Ionesio! 🚀

packages/syft/src/syft/service/user/user.py
IonesioJunior: Fixed by: [470cae8](https://github.com/OpenMined/PySyft/pull/9052/commits/470cae83db0c59070c5ebbeef48f4d8d8a76a24c)

PR #9053, opened 2024-07-16
Enclave model upload optimization

itstauq:
## Description

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/packages/syft/src/syft/client/domain_client.py
+++ b/packages/syft/src/syft/client/domain_client.py
@@ -157,6 +157,10 @@ class DomainClient(SyftClient):
                 model_size += get_mb_size(asset.data)
                 model_ref_action_ids.append(twin.id)
 
+                # Clear the Data and Mock , as they are uploaded as twin object
+                asset.data = None
+                asset.mock = None
+
                 # Update the progress bar and set the dynamic description
                 pbar.set_description(f"Uploading: {asset.name}")
                 pbar.update(1)
@@ -182,7 +186,6 @@ class DomainClient(SyftClient):
             return valid
 
         # Step 4. Upload Model to Model Stash
-        model.asset_list = []
         return self.api.services.model.add(model=model)
 
     def upload_dataset(self, dataset: CreateDataset) -> SyftSuccess | SyftError:
--- a/packages/syft/src/syft/client/domain_client.py
+++ b/packages/syft/src/syft/client/domain_client.py
@@ -182,6 +182,7 @@ class DomainClient(SyftClient):
             return valid
 
         # Step 4. Upload Model to Model Stash
+        model.asset_list = []
         return self.api.services.model.add(model=model)
 
     def upload_dataset(self, dataset: CreateDataset) -> SyftSuccess | SyftError:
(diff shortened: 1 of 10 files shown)


packages/syft/src/syft/client/domain_client.py
```diff
@@ -182,6 +182,7 @@ def upload_model(self, model: CreateModel) -> SyftSuccess | SyftError:
             return valid
 
         # Step 4. Upload Model to Model Stash
+        model.asset_list = []
```
itstauq: @rasswanth-s the changes in `on_disk.py` alone reduces the `upload_model` time from ~1m20s to ~30s. Adding this line further reduces the total time to ~5s.

From what I gather (and I might be wrong here), we are already uploading these assets to the action store in step 2. Can we have some sort of lazy loading for this asset list instead, where they contain only the action ids, and the actual objects can be fetched on demand instead?

packages/syft/src/syft/store/blob_storage/on_disk.py
itstauq, line 1: Earlier, most of the time was being spent on serde and signing of the very large assets. Directly writing the file onto the disk and bypassing the API call layer reduces the time for uploading GPT2 model (~500 MB) from 1m20s to ~30s. Great for prototyping. Potential long term fix would be using pre-signed URLs instead of API calls.

madhavajay: We might want this is an env variable or something for local mode only, because otherwise this would trigger when you run a python web server that you want to run as a normal server right?

rasswanth-s: I did a performance and found that , signing is the root bottleneck.

when using other libraries like `cryptography` it took under a second.

pynacl's, internal hashing method maybe slow.

packages/syft/src/syft/client/domain_client.py
rasswanth-s: This is because of the data and mock field in the asset, we could delete them after uploading to blob storage.
I will push a commit fixing the same.

rasswanth-s: Done [2358ede](https://github.com/OpenMined/PySyft/pull/9053/commits/2358ede3116ac276d17b1ff62bcacd72b20160bc)

packages/syft/src/syft/node/routes.py
```diff
@@ -120,7 +120,7 @@ async def stream_upload(peer_uid: str, url_path: str, request: Request) -> Respo
         status_code=200,
         response_class=JSONResponse,
     )
-    def root() -> dict[str, str]:
+    async def root() -> dict[str, str]:
```
rasswanth-s: Pyinstrument requires async routes , as it currently does not work with sync routes.

packages/syft/src/syft/store/blob_storage/on_disk.py
rasswanth-s: This fix is okay for now, 

This would fails for single container scenario, as the path is not local to a client.
but we could fix it later.
I think  real fix would be to optimize the signing part, for which I will create an issue shortly.
and also there are some performance overhead in pydantic v2.

rasswanth-s (approved):
Great work @itstauq

PR #9057, opened 2024-07-17
bugfix: DS can access unshared `Job.result` after syncing

khoaguin:
Issue: https://github.com/OpenMined/Heartbeat/issues/1602

### DIFF
--- a/packages/syft/src/syft/client/syncing.py
+++ b/packages/syft/src/syft/client/syncing.py
@@ -3,10 +3,12 @@
 # stdlib
 from collections.abc import Collection
 import logging
+from typing import cast
 
 # relative
 from ..abstract_server import ServerSideType
 from ..server.credentials import SyftVerifyKey
+from ..service.job.job_stash import Job
 from ..service.response import SyftError
 from ..service.response import SyftSuccess
 from ..service.sync.diff_state import ObjectDiffBatch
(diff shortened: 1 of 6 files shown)


packages/syft/src/syft/client/syncing.py
```diff
@@ -193,6 +195,14 @@ def handle_sync_batch(
...
         share_private_data_for_diff = share_private_data[diff.object_id]
         mockify_for_diff = mockify[diff.object_id]
+        if diff.object_type == "Job":
+            job = cast(Job, diff.get_obj())
+            if (
+                UID(job.result.id) not in share_private_data
+                or share_private_data[UID(job.result.id)] is False
+            ):
+                job.result = None
+                diff.set_obj(obj=job)
```
khoaguin, line 205: Hi @abyesilyurt and @IrinaMBejan, I found that the problem lies here where in the `share_private_data` dict we have the job's value is True (we can share the job), but its result's value (which is an `ExecutionOutput` instance) is `False` (can't share the result). However, since the low-side clients (both DS and Admin) have READ access to the synced job, they can access the `job.result` and hence can see the private result.  
For now I added a possibly temporary fix that set the `job.result` to `None` if its `ExecutionOutput` is not shared, making both the low-side DS and Admin seeing the `job.result=None`. In the future, we can turn `job.result` to return an `ActionObject` / `ExecutionOutput` (instead of `Any` for now) and manage its permissions better.
Screenshot from 2024-07-17 15-46-41

khoaguin:
Needs more planning

PR #9061, opened 2024-07-18
revert asset.data change

abyesilyurt:
It breaks `code.run`.


packages/syft/src/syft/service/dataset/dataset.py
```diff
@@ -289,8 +289,6 @@ def data(self) -> Any:
         if api is None or api.services is None:
             return None
         res = api.services.action.get(self.action_id)
-        if isinstance(res, str):
```
abyesilyurt, line 292: We show a warning below instead. ```warning = SyftWarning(
                message="You do not have permission to access private data."
            )```

shubham3121 (approved):

PR #9069, opened 2024-07-19
Suppressing guest login message after each login

snwagh:
Fixing [1631](https://github.com/OpenMined/Heartbeat/issues/1631)

The issue is that the `server.login` first launches a guest connection and then uses that guest connection to connect as the user. Not sure why this is the case but I changed the default verbose flag for login_as_guest() to be False.

Screenshot 2024-07-19 at 1 17 43 PM


snwagh:
Not sure this is the right fix. `login_as_guest` is used other places where the print is useful. Need to understand the if else conditions of `server.login` a bit more.

IonesioJunior (approved):

BrendanSchell (approved):

PR #9077, opened 2024-07-23
Fix multiple warning when accessing self.data in Assets

shubham3121:
## Description

**Problem**
accessing self.data property in Asset class shows warning at each call, since self.data is invoked at multiple places in Asset.

**Solution**

Clear display every time self.data is called. This ensures there is only one instance of the warning at each call.

image

## Checklist

### DIFF
--- a/packages/syft/src/syft/service/dataset/dataset.py
+++ b/packages/syft/src/syft/service/dataset/dataset.py
@@ -2,7 +2,6 @@
 from collections.abc import Callable
 from datetime import datetime
 from enum import Enum
-from functools import cached_property
 import logging
 import textwrap
 from typing import Any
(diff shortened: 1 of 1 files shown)


packages/syft/src/syft/service/dataset/dataset.py
```diff
@@ -249,7 +250,7 @@ def pointer(self) -> Any:
         if api is not None and api.services is not None:
             return api.services.action.get_pointer(self.action_id)
 
-    @property
+    @cached_property
```
abyesilyurt: If `mock` returns `SyftError` at some point, the result will be cached and it won't be possible to recover.

How about ignoring the consequent `display(warning)` calls with the same warning?

shubham3121: Hmm.. yeah that's true.

shubham3121: Another way is to clear the display on each of these calls

shubham3121: On seconds thoughts, I think if self.data is not a SyftError, then we should think about caching the result bcz it might help in improving the performance as well.

abyesilyurt (approved):
looks good!

PR #9072, opened 2024-07-22
Dataset Upload with Force Replace

khoaguin:
This PR closes the following issues:
- https://github.com/OpenMined/Heartbeat/issues/1277
- https://github.com/OpenMined/Heartbeat/issues/1640
- https://github.com/OpenMined/Heartbeat/issues/1684
- https://github.com/OpenMined/Heartbeat/issues/1683

PR #9082, opened 2024-07-23
Exclude soft-deleted datasets from get_all results

jcardonnet:
image

### DIFF
--- a/packages/syft/src/syft/service/dataset/dataset_stash.py
+++ b/packages/syft/src/syft/service/dataset/dataset_stash.py
@@ -1,6 +1,8 @@
 # stdlib
 
 # third party
+from result import Err
+from result import Ok
 from result import Result
 
 # relative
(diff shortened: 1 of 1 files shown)


khoaguin:
Hi @jcardonnet I am working on this feature in [this PR](https://github.com/OpenMined/PySyft/pull/9072) where the `get_all` end point of dataset service has the `include_deleted` flag which default to `False` but can be set to `True` to list the deleted datasets, and also improved repr html for deleted datasets so the DS can see some metadata of the deleted datasets. I think this way we make use of the soft delete feature better in my opinion. Please check out the PR for screenshots

IonesioJunior (approved):
LGTM

IonesioJunior:
Hi @khoaguin ,

Seems like your PR  is more robust and covers other issues too, which I think is near to the ideal solution for the whole dataset deletion problem. For now, I'll merge this specific fix to resolve a small but critical issue that needs to be addressed before our next release.

Please feel free to continue developing in this direction. If you need any assistance with reviews or development, don't hesitate to reach out. =)

khoaguin:
Thanks @IonesioJunior. This sounds good

PR #9095, opened 2024-07-25
Remove ALL_READ permission for requests

BrendanSchell:
## Description
If a DS user calls requests.get_all, it currently retrieves all requests regardless of whether the calling user has read permission on the linked code. This causes issues in the repr when the user is not able to resolve the linked code. 

This PR removes ALL_READ from being added to the permissions for requests so that when get_all is called, it'll function similarly to code.get_all where it only retrieves the requests where that user has read permissions (unless they are an admin in which case it will be all requests). 

Closes https://github.com/OpenMined/Heartbeat/issues/1561 
(and most likely https://github.com/OpenMined/Heartbeat/issues/1637 too). 

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/packages/syft/tests/conftest.py
+++ b/packages/syft/tests/conftest.py
@@ -190,6 +190,27 @@ def guest_datasite_client(root_datasite_client) -> DatasiteClient:
 
 
 @pytest.fixture
+def ds_client(
+    faker: Faker, root_datasite_client: DatasiteClient, guest_client: DatasiteClient
+):
+    guest_email = faker.email()
+    password = "mysecretpassword"
+    root_datasite_client.register(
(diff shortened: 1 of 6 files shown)


packages/syft/tests/conftest.py
```diff
@@ -189,6 +189,27 @@ def guest_datasite_client(root_datasite_client) -> DatasiteClient:
     yield root_datasite_client.guest()
 
 
+@pytest.fixture
+def ds_client(
```
snwagh, line 193: Glad you added the fixtures. Removing tech debt one step at a time. <3

snwagh (approved):
Awesome, LGTM!

PR #9125, opened 2024-08-03
Simplify syft.datasites registry to operate against a static file

iamtrask:
## Description
I've changed syft.datasites to point to a static file of datasites which people can apply to be added to. We'll operate with this simpler model until the more sophisticated approaches are ready.

## Affected Dependencies
none


madhavajay:
@iamtrask giddyup!

PR #9141, opened 2024-08-06
Fixing cli launch bug

snwagh:
Fixes [1725](https://github.com/OpenMined/Heartbeat/issues/1725)

Screenshot 2024-08-06 at 10 02 32 AM

### DIFF
--- a/packages/syft/src/syft/server/run.py
+++ b/packages/syft/src/syft/server/run.py
@@ -24,7 +24,7 @@ def run() -> ServerHandle | None:
         "--server-type",
         help="server type",
         type=str,
-        default="python",
+        default="datasite",
         dest="server_type",
     )
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


packages/syft/src/syft/server/run.py
```diff
@@ -82,7 +82,7 @@ def run() -> ServerHandle | None:
     args.reset = str_to_bool(args.reset)
     args.local_db = str_to_bool(args.local_db)
     args.tail = str_to_bool(args.tail)
-    args.cmd = str_to_bool(args.cmd)
+    # args.cmd = str_to_bool(args.cmd)
```
BrendanSchell: can we remove instead of commenting out? Doesn't look like it's being used anyway

BrendanSchell (approved):
Looks good to me! 🔥

packages/syft/src/syft/server/run.py
snwagh: Just never have the confidence to do these things. :D Fixed in [4d08b7f](https://github.com/OpenMined/PySyft/pull/9141/commits/4d08b7f8d5f6b14b106ba530cff8281eac51e04c)

PR #9163, opened 2024-08-13
Fix `dataset.getall()` shows `DictTuple()` when there are no datasets

khoaguin:
Issue: https://github.com/OpenMined/Heartbeat/issues/1685

Before: 
image


khoaguin:
Hi @kiendang, I tried to make an empty table with no rows when there are no datasets as we discussed, however, this is challenging since we need an object in the `DictTuple` iterable so we can grab that object and find its `_coll_repr_` (in the `_create_table_rows()` function. However, when the `DictTuple` is empty, we don't have any objects. 
For now, I make the `DictTuple`'s repr to be an empty list when there are no items inside, like in the picture
image

This will make the repr consistent with other APIs when the results are empty: 
image

kiendang:
I see. We do need a refactoring of `_coll_repr_` to be able to display an empty table. I still think it is the desirable behavior, but let me check with product. For now I think let's display `0 records` instead of `[]`. A `DictTuple` is not a list and showing `[]` is misleading.

packages/syft/src/syft/types/dicttuple.py
```diff
@@ -236,7 +236,16 @@ def __getitem__(self, __key: _KT | slice | SupportsIndex, /) -> _VT | Self:
         return super().__getitem__(self.__mapping[__key])
 
     def __repr__(self) -> str:
-        return f"{self.__class__.__qualname__}{super().__repr__()}"
+        # return f"{self.__class__.__qualname__}{super().__repr__()}"
+        return "[]"
```
kiendang: We don't need to modify this.

packages/syft/src/syft/types/dicttuple.py
```diff
@@ -236,7 +236,16 @@ def __getitem__(self, __key: _KT | slice | SupportsIndex, /) -> _VT | Self:
...
+
+    def _repr_html_(self) -> str:
+        super_repr_html = getattr(super(), "_repr_html_", None)
+        if super_repr_html is None:
+            return super().__repr__()
+        if super_repr_html() == self.__repr__():  # empty DictTuple
+            return "[]"
```
kiendang: Checking for empty DictTuple with `len(self) == 0` is better. So something like
```py
if len(self) == 0 or super_repr_html is None:
    return "0 records"
```

khoaguin: Thank you @kiendang very much for the review. Fixed accordingly
image

packages/syft/src/syft/types/dicttuple.py
```diff
@@ -238,6 +238,14 @@ def __getitem__(self, __key: _KT | slice | SupportsIndex, /) -> _VT | Self:
     def __repr__(self) -> str:
         return f"{self.__class__.__qualname__}{super().__repr__()}"
 
+    def _repr_html_(self) -> str:
+        super_repr_html = getattr(super(), "_repr_html_", None)
+        if super_repr_html is None:
+            return super().__repr__()
```
kiendang: nitpick: should be `self.__repr__()` here, or more pythonic

```suggestion
            return repr(self)
```

PR #9169, opened 2024-08-13
Fix Inconsistent use of UTC time in DateTime class

BrendanSchell:
## Description
Main change is to update the DateTime class to treat timezone consistently as UTC to avoid some inconsistencies between the use of UTC and local time.

Other minor updates:
- replace use of utctimestamp, utcnow functions in DateTime, BaseDateTime classes
- update sub op of DateTime / BaseDateTime to return timedelta instead of DateTime (not sure if it made sense before or was even being used)
- fixed Asset repr when either private or mock data was empty action object (noticed incidentally trying to replicate bug)

Closes https://github.com/OpenMined/Heartbeat/issues/1475.

## Affected Dependencies

## How has this been tested?
- Tested in notebook to make sure times / delta were correct in syncing widget.

## Checklist
- [] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)

### DIFF
--- a/packages/syft/src/syft/service/dataset/dataset.py
+++ b/packages/syft/src/syft/service/dataset/dataset.py
@@ -134,8 +134,12 @@ class Asset(SyftObject):
         else:
             private_data_obj = private_data_res.ok_value
             if isinstance(private_data_obj, ActionObject):
-                df = pd.DataFrame(self.data.syft_action_data)
-                data_table_line = itable_template_from_df(df=private_data_obj.head(5))
+                if isinstance(private_data_obj.syft_action_data, ActionDataEmpty):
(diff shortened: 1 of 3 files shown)


snwagh (approved):
Awesome work, LGTM!

PR #9171, opened 2024-08-14
Repr truncation for asset's primitive types 

khoaguin:
Issue: 
https://github.com/OpenMined/Heartbeat/issues/1704

### DIFF
--- a/packages/syft/src/syft/util/util.py
+++ b/packages/syft/src/syft/util/util.py
@@ -27,6 +27,7 @@ from pathlib import Path
 import platform
 import random
 import re
+import reprlib
 import secrets
 from secrets import randbelow
 import socket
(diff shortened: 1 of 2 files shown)


khoaguin:
Before

khoaguin:
After:
Screenshot from 2024-08-14 14-51-50

Screenshot from 2024-08-14 15-45-30

packages/syft/src/syft/util/util.py
```diff
@@ -1082,3 +1082,49 @@ def get_nb_secrets(defaults: dict | None = None) -> dict:
         print(f"Unable to load {filename}")
 
     return defaults
+
+
+def repr_truncation(obj: Any, max_elements: int = 10) -> str:
```
BrendanSchell, line 1095: Looks great Khoa! So that this is a little standard, maybe we can use pprint.pformat with compact=True to get the truncated collection reprs?

packages/syft/src/syft/util/util.py
```diff
@@ -1082,3 +1082,49 @@ def get_nb_secrets(defaults: dict | None = None) -> dict:
...
+    Return a truncated string representation of the object if it is too long.
+
+    Args:
+    - obj: The object to be represented (can be str, list, dict, set).
+    - max_elements: Maximum number of elements to display before truncating.
+
+    Returns:
+    - A string representation of the object, truncated if necessary.
+    """
+    if isinstance(obj, str):
```
BrendanSchell: I don't think there's a default truncated repr for long strings, but imo maybe it's more intuitive for the string case to keep just the first n letters instead of first n//2 and last n//2. Probably more than 10 chars by default too so that there can be a few words at least.

packages/syft/src/syft/util/util.py
khoaguin: Thanks for the review @BrendanSchell, I tried with `pprint.pformat` but I think just using the function is not enough 
image

packages/syft/src/syft/util/util.py
khoaguin: Thanks @BrendanSchell, yes we do not have a truncated repr for long strings.
And I totally agree and changed accordingly. Now for strings we will show the first 100 characters
image

packages/syft/src/syft/util/util.py
BrendanSchell: ya nvm, I think reprlib.repr is probably the one I was thinking of https://docs.python.org/3/library/reprlib.html. Should work for strings too.

khoaguin: Thanks @BrendanSchell
Indeed `reprlib.repr` provides a more general solution to this problem. I fixed the code accordingly

khoaguin:
- For np array: image

- For list:  image

- For strings: image

- For sets:  image

- For dict:  image

- For torch tensors: image

BrendanSchell (approved):
Looks great Khoa! 🚀

khoaguin:
Thank you very much @BrendanSchell

PR #9172, opened 2024-08-14
Settings validation

jcardonnet:
Fixes [#1664](https://github.com/OpenMined/Heartbeat/issues/1664), [#1659](https://github.com/OpenMined/Heartbeat/issues/1659)


packages/syft/src/syft/service/settings/settings.py
```diff
@@ -265,6 +270,30 @@ class ServerSettings(SyftObject):
     pwd_token_config: PwdTokenResetConfig = PwdTokenResetConfig()
     allow_guest_sessions: bool = True
 
+    @field_validator("admin_email")
+    def admin_email_format(cls, v: str) -> str:
+        try:
+            validate_email(v)
+        except EmailNotValidError as e:
+            raise ValueError(f"Invalid email format: {e}")
+        return v
```
kiendang: Shouldn't this be achieved with just

```py
admin_email: EmailStr
```
?

jcardonnet: Actually that was my first approach but it caused other issues like unclear error messages when updating the settings due to autosplat, or the validation not being triggered because of how `ServerSettingsUpdate` data is transfered to the `ServerSettings` object, so I went back to `str`

The `field_validator` was the first alternative I tried (which didn't fix the side effects either). 

Eventually, the fixes I implemented for the other issues allowed the validator to do it's job but I didn't revisit it to check if I could replace it with a simpler `EmailStr`.

However, prompted by your comment I realized that now I can, so I will. Thanks!

packages/syft/src/syft/service/settings/settings.py
```diff
@@ -265,6 +269,22 @@ class ServerSettings(SyftObject):
...
+
+    @field_validator("name")
+    def name_length(cls, v: str) -> str:
+        if len(v) < MIN_SERVER_NAME_LENGTH:
+            raise ValueError(
+                f'"name" must be at least {MIN_SERVER_NAME_LENGTH} characters long'
+            )
+        return v
+
```
kiendang, line 272: pydantic has [`constr`](https://docs.pydantic.dev/2.0/usage/types/string_types/#arguments-to-constr) which supports validation for min and max str length, e.g `name: constr(min_length = MIN_SERVER_NAME_LENGTH)` which I think would be cleaner here, but needs to check for autosplat and serialization.

packages/syft/src/syft/service/settings/settings_service.py
```diff
@@ -111,42 +95,50 @@ def _update(
...
+        notifier_service = context.server.get_service("notifierservice")
+        if not notifier_service.settings(context):
+            return SyftError(
+                message="Create notification settings using enable_notifications from user_service"
+            )
+        # If notifications_enabled is present in the update, we need to update the notifier settings
+        if settings.notifications_enabled is True:
+            result = notifier_service.set_notifier_active_to_true(context)
+        elif settings.notifications_enabled is False:
+            result = notifier_service.set_notifier_active_to_false(context)
```
kiendang: This is a little bit unrelated to this PR, but we can replace `set_notifier_active_to_false` and `set_notifier_active_to_true` with just 1 function `set_notifier_active(yes: bool)` and then this could be just `result = notifier_service.set_notifier_active(settings.notifications_enabled)`.

khoaguin:
Maybe this is also related to this ticket https://github.com/OpenMined/Heartbeat/issues/1649, since after updating the settings with 
```python
client.api.services.settings.update(
    name="a",
    organization="b",
    description="c",
    admin_email="info2@openmined.org",
    association_request_auto_approval=True,
    eager_execution_enabled=False,
    notifications_enabled=True,
)
client.refresh()
```
doing `guest_client.register` will produce the below error
image

IonesioJunior (approved):

PR #9173, opened 2024-08-14
Adding pip install quotes for zsh support

snwagh:
Adding quotes to README install directions

Also pinning tox versions in our workflows to 4.16.0 for the failing tests


kiendang (commented):
👍

btw I think we should get rid of the `$`. Reason is that users won't have to manually delete it after copying the command using the GitHub markdown click to copy button.

jcardonnet (approved):

kiendang (approved):

IonesioJunior (approved):

PR #9176, opened 2024-08-15
Refactor DB as document store, add Postgres, remove mongo

eelcovdw:
## Description

## Affected Dependencies

## How has this been tested?

## Checklist


packages/syft/src/syft/service/notification/notification_stash.py
```diff
@@ -102,14 +69,10 @@ def get_notification_for_linked_obj(
...
     ) -> Notification:
-        qks = QueryKeys(
-            qks=[
-                LinkedObjectPartitionKey.with_obj(linked_obj),
-            ]
-        )
-        return self.query_one(credentials=credentials, qks=qks).unwrap(
-            public_message=f"Notifications for Linked Object {linked_obj} not found"
-        )
+        # TODO does this work?
```
koenvanderveen: can we not just match on the id?

packages/grid/backend/grid/core/config.py
```diff
@@ -138,7 +138,7 @@ def get_emails_enabled(self) -> Self:
         True if os.getenv("CREATE_PRODUCER", "false").lower() == "true" else False
     )
     N_CONSUMERS: int = int(os.getenv("N_CONSUMERS", 1))
-    SQLITE_PATH: str = os.path.expandvars("$HOME/data/db/")
+    SQLITE_PATH: str = os.path.expandvars("/tmp/syft/")
```
khoaguin: Hi @shubham3121 I changed this to `/tmp/syft` to solve sqlalchemy can't open database file in `/root/data/db`, but not sure it's optimal. Also in `postgres-statefulset.yaml`  the `mountPath` is set to `/data/db`, should be change it to be the same to value of `SQLITE_PATH`?

shubham3121: What error do we get with `/root/data/db` ? Does it complain about directory doesn't exists or some sort of permission error ? Earlier we used to dir if doesn't exist, I think sqlalchemy could do that same maybe via some config.

khoaguin: The error is like below. I think it's a permission issue. This made the `stack.test.notebook` failed since the backend container was not running

Screenshot from 2024-09-13 10-55-57

shubham3121: Its seems like a permission issue to create a directory at the root level. Maybe we can create the director beforehand (while changing ownership of directory to user)

khoaguin: but is there any problem with using `/tmp/syft` since we use it in the local python version?

packages/syft/src/syft/protocol/protocol_version.json
```diff
@@ -1,5 +1,1172 @@
 {
-  "1": {
-    "release_name": "0.9.1.json"
```
shubham3121, line 3: @eelcovdw @abyesilyurt Its seems like we have reset the protocol file. Do we want to ensure compatibility with previous version ?

packages/syft/src/syft/orchestra.py
```diff
@@ -183,7 +183,9 @@ def deploy_to_python(
     log_level: str | int | None = None,
     debug: bool = False,
     migrate: bool = False,
+    store_client_config: dict | None = None,
```
koenvanderveen: is this convenient to pass, do we have an example of how we would pass it? Also, how do I know the default value?

packages/syft/src/syft/serde/json_serde.py
```diff
@@ -0,0 +1,457 @@
...
+
+
+class JSONSerdeError(SyftException):
+    pass
+
+
+@dataclass
+class JSONSerde(Generic[T]):
+    klass: type[T]
+    serialize_fn: Callable[[T], Json] | None = None
```
koenvanderveen, line 49: why are these optional if the only usage is that it always has them

koenvanderveen: I think this is a bit clunky to read, can we not just have a type check in the register function that checks if its a default json type and then passes a no-op into this?

packages/syft/src/syft/serde/json_serde.py
```diff
@@ -0,0 +1,457 @@
...
+
+    # JSON recursive types
+    # only strictly annotated iterables and mappings are supported
+    # example: list[int] is supported, but not list[int | str]
+    elif _is_serializable_iterable(annotation):
+        result = _serialize_iterable_to_json(value, annotation)
+    elif _is_serializable_mapping(annotation):
+        result = _serialize_mapping_to_json(value, annotation)
+    else:
+        result = _serialize_to_json_bytes(value)
```
koenvanderveen, line 399: why do we need this? We also still store the bytes separately I think? would we ever query on the bytes?

packages/syft/src/syft/serde/json_serde.py
```diff
@@ -0,0 +1,457 @@
...
+    cls = type(obj)
+    attr = getattr(cls, attr_name, None)
+
+    if isinstance(attr, property):
+        return attr.fget.__annotations__.get("return", None)
+
+    return None
+
+
+def _serialize_searchable_attrs(
```
koenvanderveen: we should probably rename this to something that indicates that those are added

packages/syft/src/syft/serde/json_serde.py
```diff
@@ -0,0 +1,457 @@
...
+    searchable_attrs: list[str] = getattr(obj, "__attr_searchable__", [])
+    unique_attrs: list[str] = getattr(obj, "__attr_unique__", [])
+
+    attrs_to_add = set(searchable_attrs) | set(unique_attrs)
+    for attr in attrs_to_add:
+        if attr not in obj_dict:
+            try:
+                value = getattr(obj, attr)
+            except Exception as e:
+                if raise_errors:
```
koenvanderveen, line 224: when would you not want to do this?

packages/syft/src/syft/server/server.py
```diff
@@ -1717,17 +1722,32 @@ def create_initial_settings(self, admin_email: str) -> ServerSettings:
             ).unwrap()
 
 
-def create_admin_new(
+def create_root_admin(
```
koenvanderveen: we should rename this to reflect the logic

koenvanderveen: it also feels like there is a third option: the root admin exists, with this email

packages/syft/src/syft/orchestra.py
khoaguin: it's old code and should be removed

khoaguin: Removed in [3b3023d](https://github.com/OpenMined/PySyft/pull/9176/commits/3b3023d3aa3f3a9914b276a0663f28c4feebe0a0)

packages/syft/src/syft/serde/json_serde.py
eelcovdw: Many types in syft are not serializable to json, either because they are custom types or because they are untypedd (and cannot be serialize back to the python obj). they are stored inside the json as bytestring.

for example: UserCodeStatusCollection is not serializable because the keys are some custom object, so the db would look like:

```
{
    "__canonical_name__": "UserCodeStatusCollection",
    "__version__": 1,
    "status_dict": <bytestring>,
}
```

koenvanderveen: yes, but isnt that redundant?

packages/syft/src/syft/serde/json_serde.py
eelcovdw: searchable attrs can be `@property` methods, and those can throw errors. Example: `Job.code_id` was throwing errors when the Job was for a twinAPI. This flag is to control this behaviour, 'do i want to skip writing if some method throws an error'

we're using raise_errors=False right now to just skip searchable properties that hit an exception

koenvanderveen (approved):

packages/syft/src/syft/server/server.py
eelcovdw: Then it would fail the first check, root verify key user already exists

packages/syft/src/syft/serde/json_serde.py
eelcovdw: There is no other place we save the full blob, the json in Table.fields is the only data we're saving in the db

packages/grid/backend/grid/core/config.py
khoaguin: after discussion changed it to `/tmp/data/db` for both single container and k8s mode in this commit [cc8e976](https://github.com/OpenMined/PySyft/pull/9176/commits/cc8e9762d75e300ce522daf007b40db65c4eacee)

packages/syft/setup.cfg
```diff
@@ -111,7 +114,6 @@ telemetry =
     opentelemetry-instrumentation==0.48b0
     opentelemetry-instrumentation-requests==0.48b0
     opentelemetry-instrumentation-fastapi==0.48b0
-    opentelemetry-instrumentation-pymongo==0.48b0
```
shubham3121, line 114: We need to add `opentelemetry-instrumentation-sqlalchemy==0.48b0`

shubham3121: Resolved by: https://github.com/OpenMined/PySyft/pull/9176/commits/7a6458883145d1efd971513921cb83a4e4ca8114

PR #9231, opened 2024-08-30
Notebook tests - Create and test API endpoints

khoaguin:
https://github.com/OpenMined/Heartbeat/issues/1794

- separate configuring and testing api endpoints into 2 notebooks
- test endpoint security (DS has no private access to private endpoints)
- test endpoint rate limiter on both the DO and DS account (configure it to be really short so its easy to test, then change it back)
- test code blocks queries that are too big (inside `create_long_running_query_job`)
- test timeout of long query doesn't break system (not sure about this since in if setting `should_succeed=False` for `create_long_running_query_job`, the request will be denied)


yashgorana:
@khoaguin what is the state of this PR?

khoaguin:
Hi @yashgorana I think it's mostly out of date. There are some things like the rate limiter that can be integrated to the current scenario test, but can be done later. 
I'll close this PR now.

PR #9235, opened 2024-08-30
Adding Scikit-learn as additional dependency in Data Science Setup Extra

leriomaggio:
Scikit-learn is necessary to complete the (full) getting started tutorial that we feature in the documentation.

Moreover, recordlinkage has been tempoarily commented out as it is not required for documentation so far.


madhavajay:
@leriomaggio we need to add the serde code more than the lib itself, can we discuss this in planning?

leriomaggio:
Yeah, sounds good to me. Closing this one to re-open in case, then :)

PR #9244, opened 2024-09-02
ignore rate limit exceptions in request.[submit, approve, deny]

abyesilyurt:


packages/syft/src/syft/service/request/request_service.py
```diff
@@ -179,55 +181,64 @@ def apply(
...
-                    notifier_types=[NOTIFIERS.EMAIL],
                     email_template=RequestUpdateEmailTemplate,
                 )
+        return result
 
-                send_notification = context.server.get_service_method(
-                    NotificationService.send
-                )
-                send_notification(context=context, notification=notification)
+    def _send_email_notification(
```
koenvanderveen, line 203: wouldnt it be better to resultify this and do the error handling from where this is called? By implementing it like this we remove the ability to build logic on it in the case of an error

koenvanderveen (approved):

PR #9274, opened 2024-09-05
Added test to notebook for password token expiry

snwagh:
Added test to check password token expiry. 
- Reduce token expiry time to 3 seconds and test token expiry
- Update email body to include the token expiration time

Closes https://github.com/OpenMined/Heartbeat/issues/1808

Repurposed variable `token_exp_min` to encode seconds.


packages/syft/src/syft/service/notification/email_templates.py
```diff
@@ -50,6 +50,28 @@ def email_body(notification: "Notification", context: AuthedServiceContext) -> s
...
+        # Option 1
+        # relative
+        # from ..settings.settings_service import SettingsService
+
+        # method = context.server.get_service_method(SettingsService.get)
+        # settings = method(context=context)
+        # expiry_time = settings.pwd_token_config.token_exp_min
+
+        # Option 2
+        settings_service = context.server.get_service("settingsservice")
```
BrendanSchell: should we update to the context.server.services approach and remove the rest of this comment?

BrendanSchell (approved):
Looks good! Just the one nitpicky comment to change how the service is called

packages/syft/src/syft/service/notification/email_templates.py
snwagh: Good point, will merge it with Teo's PR. Will wait for that to be merged so as to not create conflicts for his PR.

PR #9276, opened 2024-09-09
add synchronous message handler

abyesilyurt:
Notebook 1 - Start the server as usual, but add handler_type=Handler.Synchronous so the consumer does not use any threads.
```
server = sy.orchestra.launch(
    name="my_node",
    n_consumers=0,
    create_producer=True,
    handler_type=Handler.Synchronous,
)
# Prints: Adding producer for queue: api_call on: tcp://localhost:53022
```

Notebook 2 - Consume requests in the same thread.
```
consumer = ZMQConsumer.default()
# Queue port is auto deduced based on creation time.
consumer._run()
```

### DIFF
--- a/packages/syft/src/syft/orchestra.py
+++ b/packages/syft/src/syft/orchestra.py
@@ -28,6 +28,7 @@ from .server.datasite import Datasite
 from .server.enclave import Enclave
 from .server.gateway import Gateway
 from .server.uvicorn import serve_server
+from .service.queue.queue import Handler
 from .service.response import SyftInfo
 from .types.errors import SyftException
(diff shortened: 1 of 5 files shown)


packages/syft/src/syft/orchestra.py
```diff
@@ -182,6 +183,7 @@ def deploy_to_python(
     log_level: str | int | None = None,
     debug: bool = False,
     migrate: bool = False,
+    handler_type: Handler | None = None,
```
eelcovdw: Should we rename this to ConsumerHandler?

eelcovdw (approved):

PR #9285, opened 2024-09-12
Update otel instrumentation

yashgorana:
## Description
Instead of deploying it in the current cluster in a separate namespace, we can deploy to a separate cluster instead.

TODOs:
- [x] Plumbing syft otel endpoint to use this one.
- [x] Multi-cluster: Partition OTEL services by adding cluster name prefix
- [ ] Log correlation with trace ids

## Affected Dependencies

## How has this been tested?

## Checklist


tox.ini
```diff
@@ -1132,7 +1188,7 @@ allowlist_externals =
     bash
 commands =
     bash -c 'devspace purge --force-purge --kube-context k3d-${CLUSTER_NAME} --no-warn --namespace syft; sleep 3'
-    bash -c 'devspace cleanup images --kube-context k3d-${CLUSTER_NAME} --no-warn --namespace syft --var CONTAINER_REGISTRY=k3d-registry.localhost:5800 || true'
+    ; bash -c 'devspace cleanup images --kube-context k3d-${CLUSTER_NAME} --no-warn --namespace syft --var CONTAINER_REGISTRY=k3d-registry.localhost:5800 || true'
```
yashgorana: todo: move this to `.destroyall`?

yashgorana:
@rasswanth-s can you please review this.

tox.ini
rasswanth-s: 👍

packages/syft/src/syft/store/blob_storage/seaweedfs.py
```diff
@@ -42,26 +42,16 @@
...
-
-        BotocoreInstrumentor().instrument()
-        message = "> Added OTEL BotocoreInstrumentor"
-        print(message)
-        logger.info(message)
-    except Exception:  # nosec
-        pass
+logger = logging.getLogger(__name__)
+
+instrument_botocore()
```
rasswanth-s, line 54: Q: Instead of calling library specific instrumentation, at the specific locations, could we not call all of them during FastAPI startup?

packages/syft/src/syft/service/worker/utils.py
```diff
@@ -353,9 +353,6 @@ def create_kubernetes_pool(
                 "CREATE_PRODUCER": "False",
                 "INMEMORY_WORKERS": "False",
                 "OTEL_SERVICE_NAME": f"{pool_name}",
-                "OTEL_PYTHON_LOG_CORRELATION": os.environ.get(
```
rasswanth-s, line 356: Q: We are removing this, as currently the log co-relation is not working?

packages/syft/src/syft/util/trace_decorator.py
```diff
@@ -39,9 +44,6 @@ def set_default_attributes(cls, attributes: dict[str, str] | None = None) -> Non
                 cls.default_attributes[att] = attributes[att]
 
 
-T = TypeVar("T", bound=Callable | type)
-
-
 def instrument(
```
rasswanth-s, line 47: While Testing I found that, we do not capture blob storage , as we upload them directly.
I think it would be good to enable otel at the ingress controller.

For now, we could enable at the internal traefik proxy  level.
Traefik has native support: https://traefik.io/blog/monitor-your-production-at-a-glance-with-traefik-3-0-and-opentelemetry/

when we remove traefik as internal proxy, we could enable at ingress controller when we switch to nginx.

rasswanth-s: Enabling at ingress controller, would additionally give us full view of the traces .

rasswanth-s (approved):
LGTM, left some comments

packages/syft/src/syft/service/worker/utils.py
yashgorana: because it configures `logging.basicConfig` but not uvicorn logger through the YAML file.

packages/syft/src/syft/store/blob_storage/seaweedfs.py
yashgorana: no let it load in the order of module imports.

packages/syft/src/syft/util/trace_decorator.py
yashgorana: Locally, that's taken care of by signoz otel collector installed in the cluster `default` namespace.
On cloud, this is already taken care of by default.

yashgorana: > While Testing I found that, we do not capture blob storage , as we upload them directly.

isn't botocore instrumentation doing that for us?

PR #9288, opened 2024-09-12
Moving helpers into a better location

BrendanSchell:
This changes the structure of notebook helpers. 

- `from syft import get_helpers` will provide you access to all helpers in `notebooks/notebook_helpers` (so no more path manipulation)
- This import has to be done before third party import but isort will flip the order so add `isort:on, off` in your notebook cell with imports


snwagh (approved):
Looks incredible!

PR #9298, opened 2024-09-17
Stateful Notebooks

shubham3121:
## Description

## Affected Dependencies

## How has this been tested?

## Checklist


packages/grid/devspace.yaml
yashgorana, line 1: using dev-latest creates issues with latest image not getting pulled during k8s development.. can we avoid adding this?

yashgorana (commented):
are we planning on adding the checkpoint code to all the notebooks in this PR?

packages/grid/devspace.yaml
shubham3121: What are issues we face ? Right now the problem is that we want a way to spin up a cluster with the same tag as checkpoint is created, but since the DEVSPACE_TIMESTAMP is dynamic we cannot create images with a static tag. Also, Currently we're tagging an image with both DEVSPACE_TIMSTAMP tag and tag-latest.

notebooks/api/0.8/11-container-images-k8s.ipynb
```diff
@@ -493,8 +493,7 @@
     "assert workerimage is not None, str([image.__dict__ for image in image_list])\n",
     "assert workerimage.is_built is not None, str(workerimage)\n",
     "assert workerimage.built_at is not None, str(workerimage)\n",
-    "assert workerimage.image_hash is not None, str(workerimage)\n",
-    "assert image_list[workerimage.built_image_tag] == workerimage"
```
rasswanth-s, line 497: Q: Is this assert statement not needed anymore?

notebooks/api/0.8/11-container-images-k8s.ipynb
```diff
@@ -1037,10 +1036,6 @@
     "assert workerimage_opendp.built_at is not None, str(workerimage_opendp.__dict__)\n",
     "assert workerimage_opendp.image_hash is not None, str(workerimage_opendp.__dict__)\n",
     "\n",
-    "assert _images[workerimage_opendp.built_image_tag] == workerimage_opendp, str(\n",
```
rasswanth-s, line 1040: Q:  Is this assert statement not needed anymore?

notebooks/scenarios/bigquery/001-scale-delete-worker-pools.ipynb
```diff
@@ -67,13 +81,30 @@
     "    port=\"8080\",\n",
     "    n_consumers=num_workers,  # How many workers to be spawned\n",
     "    create_producer=True,  # Can produce more workers\n",
+    "    reset=True,\n",
```
rasswanth-s: MAYBE: we could remove this reset=True, as currently we reset db and load checkpoints?

notebooks/api/0.8/11-container-images-k8s.ipynb
shubham3121: Nope not required. Earlier we could index the worker image list with the tag name. But that implementation is buggy,  when you have different images with the same tag. So now we return all images but let go of indexing them via tag.

notebooks/api/0.8/11-container-images-k8s.ipynb
shubham3121: Refer to 
https://github.com/OpenMined/PySyft/pull/9298#discussion_r1780393223

shubham3121:
> are we planning on adding the checkpoint code to all the notebooks in this PR? 

Yes planning to do that as a new PR. We will create checkpoint at end of each notebook and then upload them as artifacts in CI.

packages/syft/src/syft/service/worker/worker_image_service.py
```diff
@@ -192,18 +194,7 @@ def get_all(self, context: AuthedServiceContext) -> DictTuple[str, SyftWorkerIma
         One image one docker file for now
         """
         images = self.stash.get_all(credentials=context.credentials).unwrap()
-
-        res = {}
-        # if image is built, index it by full_name_with_tag
-        for im in images:
-            if im.is_built and im.image_identifier is not None:
-                res[im.image_identifier.full_name_with_tag] = im
```
rasswanth-s, line 200: Q:  Are we deleting this method of indexing images by tag, as it is not being used much, do we need to confirm with product ?

packages/syft/src/syft/service/migration/migration_service.py
```diff
@@ -259,14 +266,35 @@ def _create_migrated_objects(
         ignore_existing: bool = True,
         skip_check_type: bool = False,
     ) -> dict[type[SyftObject], list[SyftObject]]:
+        # These tables are considered as startup tables, if the object is in this table and already exists,
+        # we need to overwrite the existing object with the migrated object.
+        STARTUP_TABLES = [
```
rasswanth-s: MAYBE: we could omit this logic of startup tables from this PR and keep it for another PR.

As I generally think, reset_and_restore, which we created, should be the defacto approach for migrations if we are doing it in a separate cluster . As I think the current migrations are mainly kind of backup and restore. if there are other data apart from startup tables the migration would fail, as we are enforcing the constraint that the migration happens just after the cluster is created .

packages/syft/src/syft/service/worker/worker_image_service.py
shubham3121: I thing is that indexing by tag is a buggy with different images with same tag. This effects images.get_all() functionality. Let's remove this and inform this to the product, we can ask product if we need re-implementing it as a separate ticket.

packages/syft/src/syft/service/migration/migration_service.py
shubham3121: Yes, I think so with the reset functionality we can drop this STARTUP_TABLE logic, will make the change accordingly.

shubham3121: Resolved by: [eb6cc1e](https://github.com/OpenMined/PySyft/pull/9298/commits/eb6cc1ec99c83eb18753389ba2877b34b533d344)

packages/grid/devspace.yaml
yashgorana: The issue is that `image:dev-latest` does not always get updated if you re-deploy after any code changes, breaking development experience and creating confusion. I previously had to revert this exact change from Rasswanth as well - because it just never worked.

`dev-latest` might work in CI because there's only one image, but let's not add `dev-latest` anywhere as it ends up creating confusion, and just causes divergence in expectation while debugging locally, in notebooks and CI.

yashgorana:
> > are we planning on adding the checkpoint code to all the notebooks in this PR?
> 
> Yes planning to do that as a new PR. We will create checkpoint at end of each notebook and then upload them as artifacts in CI.

Let's not use artifacts; reason being 1) issues with conflicting caches across multiple version/branches/nightlies/pr like we have with current pip cache 2) goals was so that local testing can leverage these checkpoint files directly.

I'd recommend keeping it simple and checking them into the repo itself. That way everyone can run it locally almost instantly and we don't have to worry about wrong checkpoint versions being loaded for some other syft version on CIs.

packages/grid/devspace.yaml
shubham3121: But if the image tag is constantly changing, then the info about the image tag in the db is stale right ? Means the pools cannot use the redeployed versions anyways

shubham3121:
> > > are we planning on adding the checkpoint code to all the notebooks in this PR?
> > 
> > 
> > Yes planning to do that as a new PR. We will create checkpoint at end of each notebook and then upload them as artifacts in CI.
> 
> Let's not use artifacts; reason being 1) issues with conflicting caches across multiple version/branches/nightlies/pr like we have with current pip cache 2) goals was so that local testing can leverage these checkpoint files directly.
> 
> I'd recommend keeping it simple and checking them into the repo itself. That way everyone can run it locally almost instantly and we don't have to worry about wrong checkpoint versions being loaded for some other syft version on CIs.

But don't we do that for K8s logs ? For each k8s run in CI the logs are available for us to download at end of tests. We were thinking of a similar behaviour for checkpoints as well, atleast when tests fail. 

I also agree local testing is the priority but I still think we can leverage it in CI atleast for notebooks and maybe just scenario notebooks.

Also caching uses a different plugin and we don't want to persist checkpoints across PRs, instead have a copy available if things break in CI for local debugging.

packages/syft/src/syft/store/db/stash.py
```diff
@@ -180,6 +180,54 @@ def _print_query(self, stmt: sa.sql.select) -> None:
     def unique_fields(self) -> list[str]:
         return getattr(self.object_type, "__attr_unique__", [])
 
+    @as_result(SyftException, StashException, NotFoundException)
+    @with_session
+    def get_by_unique_fields(
```
abyesilyurt: Is this method used anywhere? I can't tell what it does.

shubham3121: Not any longer, good catch @abyesilyurt. Thanks.

shubham3121: Resolved in: [bc23abc](https://github.com/OpenMined/PySyft/pull/9298/commits/bc23abcb9eb47835459db9729e130a61347b27c9)

packages/grid/devspace.yaml
shubham3121: As per discussion:
1. set the tags such that timestamp is always on top.
   - dev-${DEVSPACE_TIMESTAMP}
   - dev-latest
2. don't change global.version value - keep it dev-${DEVSPACE_TIMESTAMP},
3. use dev-latest for worker pool 

Resolved in: [9e98cf4](https://github.com/OpenMined/PySyft/pull/9298/commits/9e98cf4e43b46f9a9ae72e0d6d508b76db51c93b)

yashgorana:
> > > > are we planning on adding the checkpoint code to all the notebooks in this PR?
> > > 
> > > 
> > > Yes planning to do that as a new PR. We will create checkpoint at end of each notebook and then upload them as artifacts in CI.
> > 
> > 
> > Let's not use artifacts; reason being 1) issues with conflicting caches across multiple version/branches/nightlies/pr like we have with current pip cache 2) goals was so that local testing can leverage these checkpoint files directly.
> > I'd recommend keeping it simple and checking them into the repo itself. That way everyone can run it locally almost instantly and we don't have to worry about wrong checkpoint versions being loaded for some other syft version on CIs.
> 
> But don't we do that for K8s logs ? For each k8s run in CI the logs are available for us to download at end of tests. We were thinking of a similar behaviour for checkpoints as well, atleast when tests fail.
> 
> I also agree local testing is the priority but I still think we can leverage it in CI atleast for notebooks and maybe just scenario notebooks.
> 
> Also caching uses a different plugin and we don't want to persist checkpoints across PRs, instead have a copy available if things break in CI for local debugging.

ah i must've confused cache with artifacts. okay this looks cool.

yashgorana (approved):
awesome work @shubham3121 @rasswanth-s =)

PR #9300, opened 2024-09-18
Initial email changes

snwagh:
Adding all the email servers and tests to the sync notebooks
- low side email server is launched at port 9025, high side on port 9026
- After the admin on high side approves code and syncs to low side, it is not clear if our system should send a notification to the low side admin and the associated DS. Right now it does not and the tests are incomplete in notebook 040. 

Working to close https://github.com/OpenMined/Heartbeat/issues/1833

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,5 +86,6 @@ packages/grid/helm/examples/dev/migration.yaml
 # dynaconf settings file
 **/settings.yaml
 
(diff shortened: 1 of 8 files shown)


snwagh:
Work moved into [this PR](https://github.com/OpenMined/PySyft/pull/9306).

PR #9333, opened 2024-10-01
Shift to Single Checkpoint Files

rasswanth-s:
## Description
The PR shifts checkpoints to single files, for easy portability.
It also adds the ability to load checkpoints from custom paths

## Affected Dependencies

## How has this been tested?

## Checklist

### DIFF
--- a/packages/syft/src/syft/util/test_helpers/checkpoint.py
+++ b/packages/syft/src/syft/util/test_helpers/checkpoint.py
@@ -1,7 +1,8 @@
 # stdlib
-import datetime
 import os
 from pathlib import Path
+import tempfile
+import zipfile
 
 # syft absolute
 from syft import SyftError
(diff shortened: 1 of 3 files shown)


packages/syft/src/syft/util/test_helpers/checkpoint.py
```diff
@@ -71,37 +74,34 @@ def create_checkpoint(
     if isinstance(migration_data, SyftError):
         raise SyftException(message=migration_data.message)
 
-    checkpoint_dir = create_checkpoint_dir(
-        server_uid=client.id.to_string(), chkpt_name=name
-    )
+    checkpoint_path = chkpt_dir / f"{name}.zip"
+
+    # get a temporary directory to save the checkpoint
+    temp_dir = Path(tempfile.mkdtemp())
```
yashgorana, line 80: why change to tempfile?

shubham3121 (approved):

packages/syft/src/syft/util/test_helpers/checkpoint.py
rasswanth-s: The migrations, save api, requires a path to save checkpoint.yaml and checkpoint.blob, we save it in a temp directory and create a zipfile from the temp directory.

The alternative would be to save the checkpoint files (checkpoint.yaml, checkpoint.blob) in the same directory of checkpoints directory, and if multiple notebooks are running at same time, it might cause race conditions.

shubham3121: @yashgorana , here instead of persisting a checkpoint directory, we're trying to create a tmp folder to save the data over there and then zip it and make it available as a .zip file under `~/.syft/data/checkpoints/`,  so  e.g. `~/.syft/data/checkpoints/1-checkpoint.zip`.

PR #9339, opened 2024-10-04
Fix status for requests high side after syncing

teo-milea:
Currently running the notebooks under scenarios/bigquery will incorrectly show that requests on the high side are not approved while being approved through syncing and displaying the correct status on the low side. The issue had to do with some missing permissions on the high side. I added a test for this in the notebooks as well.


packages/syft/src/syft/service/output/output_service.py
```diff
@@ -315,6 +315,23 @@ def get_by_output_policy_id(
     def get(self, context: AuthedServiceContext, id: UID) -> ExecutionOutput:
         return self.stash.get_by_uid(context.credentials, id).unwrap()
 
+    # @service_method(
+    #     path="output.set_permission",
```
koenvanderveen: can we remove before merge?

teo-milea: done!

koenvanderveen (approved):

PR #9392, opened 2026-01-13
Fix output policy validation in UserCodeService.is_execution_allowed

### LINKED ISSUE #9391: UserCodeService.is_execution_allowed() allows invalid output policy execution
## Description

The `UserCodeService.is_execution_allowed()` method contains a logic error that causes it to ignore the boolean return value from `output_policy.is_valid()`. This results in execution being allowed even when the output policy should deny it.

## Current Behavior

The method uses a try/except block but only checks for exceptions raised by `output_policy.is_valid()`:
- Exception is raised → returns `IsExecutionAllowedEnum.INVALID_OUTPUT_POLICY`
- No exception is raised → **always** returns `IsExecutionAllowedEnum.ALLOWED`

The boolean return value from `is_valid()` is ignored.

**Code Reference:**
https://github.com/OpenMined/PySyft/blob/73901fe937c221ae3f12b2fed3a280e3ee97fa8f/packages/syft/src/syft/service/code/user_code_service.py#L359-L364

## Expected Behavior

Based on the method signatures of `OutputPolicy` classes, the implementation should:
1. Call `output_policy.is_valid()` and capture its boolean return value
2. Return `IsExecutionAllowedEnum.ALLOWED` only if `is_valid()` returns `True`
3. Return `IsExecutionAllowedEnum.INVALID_OUTPUT_POLICY` if `is_valid()` returns `False` OR raises an exception (e.g., `NotImplementedError`)

## Impact

This bug allows policy violations, including multiple executions of `SingleExecutionExactOutput` which run without the server returning cached `ExecutionOutput` values to the client in violation of the intended policy control that should limit execution to a single run

## Suggested Fix

Modify the method to check the boolean return value:

```python
try:
    is_valid = output_policy.is_valid()
    if not is_valid:
        return IsExecutionAllowedEnum.INVALID_OUTPUT_POLICY
    return IsExecutionAllowedEnum.ALLOWED
except Exception:
    return IsExecutionAllowedEnum.INVALID_OUTPUT_POLICY
```

Nitya-Choudhary:
Fixes #9391

### What changed
- Updated `UserCodeService.is_execution_allowed()` to respect the boolean return
  value of `output_policy.is_valid()`.
- Execution is now allowed only when `is_valid()` returns `True`.
- Execution is correctly rejected when `is_valid()` returns `False` or raises
  an exception.

### Why this change is needed
Previously, `output_policy.is_valid()` was called only for its side effects.
If it returned `False`, execution was still allowed, which bypassed output
policy enforcement. This fix ensures output policies are properly enforced
as intended.

### Tests added
- Added unit tests to verify execution is allowed for valid output policies.
- Added unit tests to verify execution is rejected for invalid output policies.

### Impact
- Improves correctness and security of output policy validation.
- No breaking changes to public APIs.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,3 +1,12 @@
+
+Welcome to the PySyft documentation! **PySyft** is an open-source library for privacy-preserving machine learning, enabling secure data science workflows. This guide helps you set up the docs locally for development or contribution. 
+
+### Prerequisites
+- Python 3.8 or higher
+- These instructions assume a Unix-like environment (Linux/Mac or WSL on Windows). On    Windows with PowerShell, you may need to install tools like `make` via Chocolatey or adapt commands accordingly.
+
+
+
 # PySyft Documentation
 
 Welcome to the PySyft docs. You can setup the PySyft docs locally via 2 methods currently,
@@ -46,12 +55,13 @@ Welcome to the PySyft docs. You can setup the PySyft docs locally via 2 methods 
 
 3. Voila! Now visit the PySyft/docs/build/html/index.html to view the docs locally.
 
+## Viewing the Docs
+
+After running either setup method, open `docs/build/html/index.html` in your web browser to view the generated documentation locally.
+
 ## Debugging
 
 If you want to start a fresh build, run:
 
 ```sh
 make clean
-```
-
-as this will remove all the pre-exisitng files or directories in the build/ directory.


rasswanth-s:
Hello 👋 

Thanks so much for this PR — I really appreciate the work you put into it.

A quick heads-up: the repo just went through a significant refactor, so the codebase has changed quite a bit from what this PR was built on. As a result, these changes no longer apply cleanly to the new structure, and I'm going to close this PR for now.

if you're interested in contributing to the current version, kindly have a look at the [README](https://github.com/OpenMined/PySyft/blob/dev/README.md)

Thanks again for contributing — and sorry for the churn!

PR #9418, opened 2026-06-16
feat: add pre-write immutability filter for private datasets on enclave

rasswanth-s:
## Summary

- Adds an optional `pre_write_filter` hook to `DataSiteWatcherCache` that gates every file write/delete in `apply_event_message()`
- Implements `make_private_dataset_immutability_filter()` in the enclave package — blocks overwrites/deletes of private dataset files that already exist on disk
- Wires the filter in `SyftEnclaveClient.from_config()` so all enclave instances enforce immutability automatically

**Threat addressed:** Once a DO shares private dataset files with the enclave, subsequent re-sends or sync cycles cannot overwrite them — ensuring the data the enclave computes on remains the same as what was originally delivered.

## Test plan

- [x] Unit tests for `is_private_dataset_path()` (positive, public path, wrong subdir, too short)
- [x] Unit tests for the filter (allows first write, blocks overwrite, blocks delete, allows non-dataset paths, allows remaining files in batch)
- [x] Integration test: full quad setup, DO shares dataset, DO re-shares tampered data, enclave retains original
- [x] All 67 existing enclave tests pass
- [x] All 113 syft_bg tests pass


packages/syft-enclave/src/syft_enclaves/client.py
```diff
@@ -378,6 +378,16 @@ def from_config(
         """
         manager = SyftboxManager.from_config(config)
         manager.job_client = EnclaveJobClient(manager.job_client)
+
+        if manager.datasite_watcher_syncer:
+            from syft_enclaves.immutability import (
```
rasswanth-s: Can we move the import to top?

rasswanth-s: resolved [99d1a94](https://github.com/OpenMined/PySyft/pull/9418/commits/99d1a942e048246a1045b934417cfaa25da206d0)

PR #9434, opened 2026-07-03
feat: versioned JobManager for reading/writing jobs across protocol versions

koenvanderveen:
- **Versioned `JobManager`** (nested in `JobClient`) owns all `JobState`/`JobSubmissionMetadata` filesystem IO: reads upgrade objects in memory to the latest registered version, writes downgrade them to the version/layout the reading peer understands. `JobClient`, `JobInfo`, `SyftJobRunner`, the syft-bg job monitor, and the enclave client all route through it.
- **On-disk protocol version**: new jobs live under `inbox/<ds_email>/v1/<job_name>/` (same for `review/`). Protocol 0 = last release (0.1.38): no version segment, no `canonical_name`/`version` fields in yaml; protocol-0 jobs are read, listed, and written back byte-compatibly with 0.1.38.
- **Per-peer protocol negotiation**: `JobClient` accepts optional `peer_schemas` (peer email → job `ProtocolSchema`); the version spoken to a peer is the min of ours and theirs, so an older peer receives the older layout/format. Unknown peers default to the current protocol.
- **syft-migration**: `MigrationRegistry` gains a `protocol_version`; released schemas are frozen as `ReleasedProtocol` / `ReleasedPackageProtocolInfo` artifacts (save/load, registered on import) with drift detection that fails if a released object schema changes without a new version. A hardcoded 0.1.38 artifact ships in `syft_job/migrations/history/` as if that release had emitted it.
- **`JobInfo` identity from the path, not `config.yaml`**: `datasite_owner_email`, `submitted_by`, and `name` now come from the path-derived `JobRef` (governed by syft permissions), never from the DS-writable `config.yaml` — closing a spoofing vector where a DS could claim a fake identity. `ref` is now required; `current_user_email` stays sourced from the client config.
- **Release tooling** (`packages/syft-job/scripts/`): `export_release_artifact.py` writes the JSON schema artifacts, and the new `generate_release_fixture.py` snapshots a full SyftBox tree of exactly how the release serializes jobs to disk. Both are run on every release and documented in `CLAUDE.md` / `README.md`.
- **Cross-release on-disk fixtures**: fixtures named `syft_job-<version>-protocol<p>_syftbox` (0.1.38/protocol 0 hand-authored, 0.1.39/protocol 1 generated). `test_older_protocol_compatibility.py` loops over **all** of them — keyed on release, not just protocol, since a release can change how it writes bytes without bumping the protocol — asserting load/upgrade plus byte-exact scan/approve write-back per protocol.

**Test plan**: syft-migration + syft-job migration/p2p suites (list/upgrade, scan, approve write-back byte-match, mixed listing, reserved `v<n>` job names, per-protocol peer submit, perms on the `v1` subfolder, repr smoke test, and an identity-from-path spoofing test) plus root `tests/unit`; all suites green, pre-commit clean, history JSON verified in the wheel.


packages/syft-job/src/syft_job/job_storage.py
```diff
@@ -0,0 +1,229 @@
...
+        config: SyftJobConfig,
+        registry: MigrationRegistry = job_registry,
+        peer_schemas: Optional[dict[str, ProtocolSchema]] = None,
+    ) -> None:
+        self.config = config
+        self.registry = registry
+        self.service = MigrationService(registry=registry)
+        # peer email -> job ProtocolSchema; filled in by syft-client later.
+        # Peers without an entry are assumed to run the current protocol.
+        self.peer_schemas: dict[str, ProtocolSchema] = peer_schemas or {}
```
pjwerneck, line 42: While `peer_schemas` is being assigned here and the versioning machinery treats it correctly, it's apparently not being sent by the syft-client, since `JobClient` is initialized using `.from_config` and there's no `peer_schemas` being passed anywhere.

The test cases are correct and pass `peer_schemas`, but the actual code doesn't.

packages/syft-job/src/syft_job/manager.py
```diff
@@ -0,0 +1,229 @@
...
+        if len(matches) > 1:
+            raise ValueError(
+                f"Multiple jobs named '{job_name}' found; pass user to disambiguate"
+            )
+        return matches[0]
+
+    # -- model IO ---------------------------------------------------------------
+    def read_submission(self, ref: JobRef) -> JobSubmissionMetadata:
+        """Load a submission's config.yaml, upgraded to the latest version."""
+        extra = {"submitted_by": ref.ds_email, "datasite_email": ref.datasite_email}
```
pjwerneck: Shouldn't `ref.job_name` be added to `extra` here too, since it comes from the path like `submitted_by` and `datasite_email`?

packages/syft-job/src/syft_job/job_storage.py
```diff
@@ -0,0 +1,229 @@
...
+        """The unique inbox/<ds_email>/[v<n>/]<job_name>/ ref; ``ds_email`` disambiguates."""
+        matches = [
+            ref
+            for ref in self.iter_submission_refs(datasite_email)
+            if ref.job_name == job_name
+            and (ds_email is None or ref.ds_email == ds_email)
+        ]
+        if not matches:
+            raise FileNotFoundError(f"Job '{job_name}' not found")
+        if len(matches) > 1:
```
pjwerneck, line 147: This will raise an error with jobs with the same name across protocol versions. Is that intended?

packages/syft-job/src/syft_job/client.py
```diff
@@ -577,48 +576,23 @@ def _get_all_jobs(self) -> List[JobInfo]:
...
+                            job_metadata=config,
+                            state=state,
+                            current_user_email=self.current_user_email,
+                            client=self,
+                            ref=ref,
                         )
-                    except Exception:
-                        continue
+                    )
+                except Exception:
```
pjwerneck: Blanket exception catches on a long block like this are an anti-pattern. It should catch specific exceptions, or at least log the error with the job name and path.

packages/syft-job/src/syft_job/client.py
```diff
@@ -491,24 +503,27 @@ def validate_submission(self, inbox_job_path: Path) -> tuple[bool, str]:
             return False, "'config.yaml' must be a file"
         return True, ""
 
-    def receive_job(self, ds_email: str, job_name: str) -> JobState:
+    def receive_job(
+        self,
+        ds_email: str,
+        job_name: str,
+        protocol_version: str = JOB_PROTOCOL_VERSION,
```
pjwerneck: Since `receive_job` is only called by `scan_inbox`, which already determines the correct `protocol_version` on load, it should be required parameter, not default to the constant.

packages/syft-job/src/syft_job/job_runner.py
```diff
@@ -166,17 +158,9 @@ def reset_all_jobs(self) -> None:
...
-        review_dir = self.config.get_review_dir(root_email)
-
-        for scan_dir in [inbox_dir, review_dir]:
-            if scan_dir.exists():
-                for ds_dir in scan_dir.iterdir():
-                    if ds_dir.is_dir():
-                        for item in ds_dir.iterdir():
-                            if item.is_dir():
-                                total_jobs += 1
+        total_jobs = len(list(self.manager.iter_submission_refs(root_email))) + len(
```
pjwerneck, line 161: Won't a job have both `config.yaml` and `state.yaml` files after `scan_inbox` and `receive_job` are called, therefore `total_jobs` here is counting those jobs twice?

It doesn't affect deletion, since it uses `shutil.rmtree` but the warning message below will be misleading.

packages/syft-job/src/syft_job/job_runner.py
```diff
@@ -429,52 +360,50 @@ def _execute_job_captured(
...
-        self._prepare_outputs_dir(job_name, user)
+        self._prepare_outputs_dir(ref)
 
         print(f" Executing job: {job_name}")
         print(f" Inbox: {submission_dir}")
 
         # Update state to RUNNING
-        state_file = review_dir / "state.yaml"
-        state = JobState.load(state_file)
+        state = self.manager.read_state(ref)
```
pjwerneck, line 398: `JobManager.read_state` can return `None` if the `state.yaml` no longer exists. There should be a check here to prevent an exception on the `state.status` on the next line.

packages/syft-job/src/syft_job/job_storage.py
```diff
@@ -0,0 +1,229 @@
...
+        return path
+
+    def write_state(self, ref: JobRef, state: JobState) -> Path:
+        """Write state.yaml in the version/format the submitter can read."""
+        path = self.review_dir(ref) / "state.yaml"
+        self._write_in_target_version(path, state, ref, reader_email=ref.ds_email)
+        return path
+
+    # -- internals -----------------------------------------------------------------
+    def _load_upgraded(
```
pjwerneck: Both `_load_upgraded` and `_write_in_target_version` work for only two versions, but are somewhat fragile when it comes to handling more future versions, and specially once we need to upgrade or downgrade jobs throughout multiple versions.

Instead of conditionals I would introduce a `ProtocolCodec` class that handles the serialization and loading for each version, and `JobManager` delegates to the appropriate class, keeping all migration specific code in the codec class. Once v2 is introduced, it's just a new class, instead of more conditionals.

packages/syft-job/src/syft_job/manager.py
```diff
@@ -0,0 +1,229 @@
...
+class JobRef:
+    """One job on disk: who owns it, who submitted it, and its protocol layout."""
+
+    datasite_email: str  # DO whose datasite holds the job
+    ds_email: str  # submitter
+    job_name: str
+    protocol_version: str  # "0" (no path segment) or "1"+ (v<n> segment)
+
+
+class JobManager:
```
pjwerneck: `JobManager` is a somewhat vague name that blurs the line with `JobClient`. Although the two have clearly distinct responsibilities, the suffix is not as clear. Maybe `JobRepository` or `JobStorage`.

packages/syft-job/src/syft_job/job_storage.py
```diff
@@ -0,0 +1,229 @@
...
+        return self._load_upgraded(path, "JobSubmissionMetadata", extra)
+
+    def read_state(self, ref: JobRef) -> Optional[JobState]:
+        """Load a job's state.yaml, upgraded to the latest version; None if absent."""
+        path = self.review_dir(ref) / "state.yaml"
+        if not path.exists():
+            return None
+        return self._load_upgraded(path, "JobState", {})
+
+    def write_submission(self, ref: JobRef, metadata: JobSubmissionMetadata) -> Path:
```
pjwerneck, line 179: Both `write_submission` and `write_state` could be renamed to make the asymmetry more explicit, like in the docstring. Like `write_submission_for_do` and `write_state_for_ds`.

packages/syft-bg/src/syft_bg/notify/monitors/job.py
```diff
@@ -85,61 +78,40 @@ def _maybe_process_job(self, job_path: Path):
...
-                        JobStatus.DONE,
-                        JobStatus.FAILED,
-                    ):
-                        self.state.mark_notified(metadata.name, "approved")
-                    if review_state.status == JobStatus.DONE:
-                        self.state.mark_notified(metadata.name, "executed")
-                    if review_state.status == JobStatus.FAILED:
-                        self.state.mark_notified(metadata.name, "failed")
-                count += 1
+            self.state.mark_notified(metadata.name, "new")
```
pjwerneck: `metadata.name` was replaced with `ref.job_name` in most places, but still remains in others like here.

pjwerneck (commented):
The comments mark some minor issues found in the code.

The main high-level design issue would be having a separate `ProtocolCodec` class, to avoid bundling all protocol rules in the `_load_upgraded` and `_write_in_target_version` methods, as suggested in the comments.

packages/syft-job/src/syft_job/job_storage.py
koenvanderveen: Hmm yes I have been thinking about this but didnt find a better solution than JobManager. The idea of JobManager was that its only responsibiliy was reading and writing jobs in the right version. Perhaps that didnt really succeed 100% since we have

`negotiated_protocol_version_for_peer`
`_get_write_target_schema`
`new_submission_ref`

but we could move these to `JobClient` if we want to. So I do like the solution but then I guess JobManager would just be a class that finds the right codec and passes it through. Which is fine I think. One thing that crossed my mind is that perhaps it wouldnt be amazing to have to create a new ProtocolCodec every time we make a change to a Job object, but I guess we can have a SIngle codex handle multiple job versions

packages/syft-job/src/syft_job/job_storage.py
koenvanderveen: yes thats out of scope for this PR. The Idea is that syft-client passes those in but that is follow up work

packages/syft-job/src/syft_job/manager.py
koenvanderveen: yes, good find!

packages/syft-job/src/syft_job/job_storage.py
koenvanderveen: yes it is. This was an optimistic UX request

packages/syft-job/src/syft_job/client.py
koenvanderveen: I will add a log for now. To do this properly is perhaps out of scope for this PR

packages/syft-job/src/syft_job/manager.py
koenvanderveen: Yes perhaps true, I will rename to JobStorage

packages/syft-job/src/syft_job/job_storage.py
koenvanderveen: yes actually the docstrings are not completely correct here, will update

packages/syft-job/src/syft_job/job_runner.py
koenvanderveen: hmm yes I think you are right

PR #9436, opened 2026-07-06
fix: harden syft-restrict against verifier and obfuscator bypasses

pjwerneck:
## Summary

Closes a series of bypasses in `syft-restrict`'s static verifier and obfuscator, found through
iterative adversarial review. Update documentation.

## Changes

- Ban additional dynamic-escape/reflection builtins: `type`, `__build_class__`, `print`, and the
  dunder-proxy builtins `repr`/`str`/`ascii`/`format`/`bytes`.
- Fix multiple aliasing bypasses: bare names imported via `from X import name`, homoglyph identifiers, and
  common identifier-aliasing patterns that evaded the call-target checks.
- Fix f-string handling: conversion flags (`!r`/`!s`/`!a`), the `{x=}` debug form, and plain
  interpolation (`f"{x}"`) all invoke `__format__`/`__repr__`/`__str__` with no `Call` node — all
  are now rejected instead of only the conversion-flag forms.
- Harden `self`/`cls` parameter vetting: reassigning `self`/`cls`, reusing them as an unrelated
  parameter name, and non-first-parameter/nested-function/lambda cases no longer grant the
  `self.<name>` trust exemption.
- Close `self.<attr>` trust bypasses: tuple-unpack and for-loop assignment targets are now tracked
  by the safety table, and local-variable aliasing (including multi-hop copies) is tracked so
  `tmp = self.fn; tmp(x)` is checked the same way `self.fn(x)` is.
- Ban `def`/`class` statements whose name shadows a trusted import alias or visible wrapper name
  (previously only rebinding via assignment was checked).
- Fix the obfuscator to correctly blank f-string literal text under Python 3.12+'s PEP 701
  tokenizer (previously left un-obfuscated on 3.12+).
- Rewrite `docs/disallowed-ast-examples.md` as per-construct subheadings instead of a wide table;
  add `docs/code-layout.md`; simplify `README.md`.
- Extract shared AST helpers into `astutil.py` and refactor `verifier.py`/`obfuscator.py` for
  readability (central violation-code registry, `_SelfAttrTrust` helper, split obfuscator passes),
  with no behavior change.
- Raise on a malformed/inverted private-line range instead of silently verifying nothing in it.

## Testing

- Added `tests/verify/test_bypasses.py` with a regression test per bypass class
- Added  `tests/verify/test_ranges.py` for the range-validation fix.

## Asana task

https://app.asana.com/1/1185126988600652/project/1216249688888494/task/1216266561077080?focus=true


packages/syft-restrict/src/syft_restrict/verifier.py
```diff
@@ -132,301 +161,661 @@ def verify(source: str, private, policy: Policy) -> VerifyResult:
...
+    def _check_self_cls_params(self, args: ast.arguments) -> None:
+        # Guards against: forging the self/cls exemption. self.<name> is trusted everywhere in this
+        # checker purely by matching the literal identifier "self"/"cls" -- it never verifies the name
+        # is bound to the real instance. That is only sound for the genuine first parameter of a method
+        # defined directly in a class body (where Python guarantees the binding). Anywhere else -- a
+        # non-first param, a nested function/lambda param, a *args/**kwargs of the same name -- an
+        # attacker's own object would receive the identical blanket trust. See docs/verify.md#edge-cases.
+        is_direct_method = bool(self._scope_stack) and isinstance(
+            self._scope_stack[-1], ast.ClassDef
+        )
```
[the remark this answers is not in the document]
pjwerneck: This is intentional. A lambda assigned directly as a class attribute goes through the same descriptor protocol as a def-based method, so self is genuinely bound to the real instance when called as instance.f().

The bypass this check actually guards against is a lambda assigned to a local variable and called as a plain functio, where self is just whatever gets passed positionally.

packages/syft-restrict/docs/blacklist.md
```diff
@@ -0,0 +1,133 @@
...
+
+**Dunder-proxy builtins** — same escape as calling a dunder on a value (`x.__repr__()`), spelled as
+a bare call. Code `banned-call`: `repr`, `str`, `ascii`, `format`, `bytes`. (`bytes(x)` losslessly
+serializes an array's raw memory; `print` is a stdout exfil channel.)
+
+The same escape via **any f-string interpolation** (`f"{x}"`, `f"{x!r}"`, `f"{x!s}"`, `f"{x!a}"`, `f"{x=}"`) has no `Call` node and is rejected as `method-on-value`.
+
+---
+
+## The JAX / serialization denylist
```
koenvanderveen: I dont get it, do we have a specific jax denylist?

packages/syft-restrict/docs/blacklist.md
```diff
@@ -0,0 +1,133 @@
...
+| Magic/hook method other than `setup`/`__call__`/`__post_init__` | `def __getattr__`, `def __reduce__` | `dunder-def`    |
+
+`@property` is rejected because it runs code on bare attribute access (`block.w`) — same hook
+class as a dunder def.
+
+---
+
+## Reserved-name rebinding
+
+Rebinding a name the resolver trusts would poison verification. Rejected wherever the name is
```
koenvanderveen: what does it mean to trust

packages/syft-restrict/docs/blacklist.md
```diff
@@ -0,0 +1,133 @@
...
+bound — assignment, `for`/comprehension target, or parameter. Code: `reserved-name`.
+
+- A trusted module alias (`jnp`, `nn`, `lax`, …) — rebinding makes the import table a lie, so every
+  "allow-listed path" through that name becomes attacker-controlled.
+- A visible wrapper name — rebinding `transpose = evil` defeats the wrapper's type guard.
```
koenvanderveen: what doe "defeats the wrappers type guard" mean? Can we. simplify this or just say that its not possible even though we dont know how it can be used exactly

packages/syft-restrict/docs/blacklist.md
```diff
@@ -0,0 +1,133 @@
...
+- `self` / `cls` — the exemption is trusted by identifier alone; rebinding it (or reusing it as an
```
koenvanderveen: again, I am not completely following this sentence

packages/syft-restrict/docs/blacklist.md
```diff
@@ -0,0 +1,133 @@
...
+  unrelated parameter) would grant an attacker's object the same trust. See
+  [verify.md#edge-cases](verify.md#edge-cases).
+
+---
+
+## Container / aliasing tricks
+
```
koenvanderveen, line 237: nice, is this just builtin references or would there be other attack vectors here I am wondering

packages/syft-restrict/README.md
```diff
@@ -1,263 +1,67 @@
 # syft-restrict
 
-## 1. Overview
+Static analysis for JAX/Flax inference code. You mark the parts of a file that contain the model
```
koenvanderveen: not specific for Jax/Flax

packages/syft-restrict/README.md
```diff
@@ -1,263 +1,67 @@
...
+math as _private_; `syft-restrict` checks that those lines only do trusted computation — no sneaky
```
koenvanderveen: tbh, I think this opening section/summary is worse than the original. I think we should revert or end up somewhere in the middle

packages/syft-restrict/README.md
```diff
@@ -1,263 +1,67 @@
...
 ```python
 import syft_restrict as restrict
 
 result = restrict.run(
     "gemma_inference.py",
-    obfuscate=[[22, 93], [99, 280]],                              # 1-based ranges: identifiers renamed, constants blanked
-    hide=[],                                                      # 1-based ranges: whole line replaced with ■■■■■■■■
-    allow_functions=["jax.*", "flax.linen.*"],                    # things callable BY NAME (path-resolved)
-    allow_methods=["arithmetic", "indexing", "comparison"],       # operators allowed ON A VALUE
+    obfuscate=[[22, 93], [99, 280]],                        # 1-based ranges: identifiers renamed, constants blanked
```
koenvanderveen: can you make sure this aligns with the example, I think currently this is out of sync (it already was)

packages/syft-restrict/README.md
```diff
@@ -1,263 +1,67 @@
...
 
-"Attestation" (step 5) is a cryptographically signed statement the secure enclave produces: a machine-verifiable receipt saying _"I ran exactly this check over exactly these bytes and it passed."_ The DO checks the signature instead of trusting anyone's word.
+## How it works
 
-The enclave runs steps 1–5 and attests: _"I, the enclave, parsed bytes with sha256 = H, ran whitelist policy version P over the PRIVATE lines, it PASSED, and the obfuscated artifact you're reading was derived from exactly those bytes."_
+`restrict.run()` verifies first, then obfuscates:
 
-## 5. What this whitelist does NOT stop
+1. Parse the whole file and build an import binding table (`import jax.numpy as jnp` → `jnp` maps to `jax.numpy`).
+2. Walk the private lines with default-deny. Each AST node must match an explicit rule: allowed node
```
koenvanderveen: this describes it in terms that are not introduced yet

packages/syft-restrict/README.md
```diff
@@ -1,263 +1,67 @@
...
+generated (`python examples/generate.py` to regenerate).
 
-"Attestation" (step 5) is a cryptographically signed statement the secure enclave produces: a machine-verifiable receipt saying _"I ran exactly this check over exactly these bytes and it passed."_ The DO checks the signature instead of trusting anyone's word.
+## How it works
 
-The enclave runs steps 1–5 and attests: _"I, the enclave, parsed bytes with sha256 = H, ran whitelist policy version P over the PRIVATE lines, it PASSED, and the obfuscated artifact you're reading was derived from exactly those bytes."_
+`restrict.run()` verifies first, then obfuscates:
 
-## 5. What this whitelist does NOT stop
+1. Parse the whole file and build an import binding table (`import jax.numpy as jnp` → `jnp` maps to `jax.numpy`).
```
koenvanderveen: concept of binding table not introduced yet, describe without it, more generic

packages/syft-restrict/README.md
```diff
@@ -1,263 +1,67 @@
...
-2. **Timing & cache side channels.** Execution time, memory-access and CPU-cache patterns leak information regardless of which ops ran; the whitelist can't enforce constant-time execution. → enclave-level side-channel mitigations.
-3. **A future JAX host-callback API under an already-allowed prefix.** Default-deny closes unknown _new_ paths automatically, but a newly added dangerous symbol inside an allowed prefix (e.g. a hypothetical `jax.numpy.<new_io>`) is a gap until the denylist is updated. → keep the policy versioned and reviewed against each pinned JAX release.
-4. **Bugs or supply-chain compromise in the trusted libraries.** The model only constrains the _caller's_ code; a malicious jax/flax/orbax build defeats it. → attest the exact library versions/hashes; the DO must trust those releases.
+- [docs/verify.md](docs/verify.md) — how the checker decides what's allowed, what order things run in, edge cases, and known limits.
```
koenvanderveen: checker, not introduced yet

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,313 @@
+# How verification works
+
+`syft-restrict` statically analyzes the **private** lines of a source file. The goal is to show
+those lines only perform trusted math: no file or network access, no dynamic Python, no reaching
+into interpreter internals. The code is never executed. You get a violation list (empty on success)
```
koenvanderveen: I dont thing using "You" is the tone of voice we want

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,313 @@
...
+
+Like [RestrictedPython](https://github.com/zopefoundation/RestrictedPython), the model is
+default-deny: if a node type, call, attribute, or name isn't explicitly permitted below, it's
+rejected. New Python syntax stays blocked until someone reviews it.
+
+---
+
+## The whitelist
+
+Private inference code needs to wire up JAX operations: Flax modules with `setup`/`__call__`, JAX
```
koenvanderveen: it assume inference code too much now, that needs more of a nuance

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,313 @@
...
+function calls, arithmetic, shape/dtype plumbing (lists, dicts, tuples, slices, f-strings for
+einsum strings), and control flow that doesn't depend on secret data. It should not be able to touch
+the host, filesystem, network, interpreter internals, or build code from strings.
```
koenvanderveen: I would seperate this out as one of the use cases we want to support but not make it part of the package internals explanation

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,313 @@
...
+   reviewed output schema.
+2. **Timing and cache side channels.** Execution time and memory access patterns leak regardless of
+   which ops ran. Handle at the enclave level.
+3. **New dangerous symbols under an already-allowed prefix.** Default-deny blocks unknown new paths,
+   but a bad addition inside `jax.numpy.*` is a gap until the denylist is updated. Keep the policy
+   versioned and reviewed against each pinned JAX release.
+4. **Bugs or supply-chain compromise in trusted libraries.** The checker only constrains caller
+   code; a malicious jax/flax build defeats it. Attest exact library versions and hashes.
+
+## Examples
```
koenvanderveen: I feel like this section does not make clear these are not allowed, also, if they are not why are they not on the blacklist?

koenvanderveen:
## Summary of the latest changes

This push reworks the policy denylist and simplifies/tightens the verifier. Full suite green (147 passed).

### Policy: hardcoded denylist → user-configurable disallow list
- Removed the built-in `JAX_DENYLIST` (host-callback / IO / serialization globs).
- Added a per-policy **`disallow_functions`** glob list that **beats the allow**, threaded through `run(...)` and `Policy.parse(...)`. It's a hard floor an author can set over a broad allow like `jax.*`; empty by default.
- **Behavior change (intentional):** with no disallow list, a bare `jax.*` allow now permits formerly-denied leaves such as `jax.numpy.save`. Safety now comes from a *specific* allow-list or an explicit disallow — documented and covered by `test_no_disallow_permits_leaf_under_broad_allow`.
- Renamed `Policy` fields for clarity: `functions → allowed_functions`, `methods → allowed_methods`, `reserved → reserved_names`, plus the new `disallowed_functions`. `policy_id` now folds in the disallow list.

### Verifier tightening & simplification
- **Base classes:** only an allow-listed import may be a base class. `object` and hidden-region (locally-defined) classes are no longer accepted — a base's metaclass / `__init_subclass__` runs at class-creation time, so it must resolve to something the policy vetted.
- **`_check_call`:** dropped the `BANNED_NAMES` branch (banned builtins in call position are flagged by `_check_name`, the single backstop) and removed the `hidden_defs`/`visible_defs` shadowing exclusions on the import-binding check.
- **Readability:** `dotted → dotted_name`; `_check_self_cls_params → _check_arguments_dont_abuse_self_or_cls`; per-method leading comment blocks converted to docstrings.

### Tests & docs
- Updated `test_whitelist` / `test_bypasses` for the base-class tightening (`class M(object)` → `class M:`; `test_class_bases_object_and_hidden_def` → `test_class_base_must_be_allow_listed_import`).
- Added `test_user_disallow_beats_allow` and `test_no_disallow_permits_leaf_under_broad_allow`; `make_policy` gained a `disallow` param.
- Updated README, `docs/verify.md`, `docs/blacklist.md`, `docs/code-layout.md`, and regenerated `gemma_inference.obfuscated.py` + certificate.

packages/syft-restrict/docs/blacklist.md
pjwerneck: `blacklist.md` was fully rewritten in clearer structure and language.

packages/syft-restrict/docs/blacklist.md
pjwerneck: The container/aliasing tricks are less important to enforce now, since we are default-denying all callable builtins that can't be proven safe. If someone uses one of these to work around the banned-name, it triggers the `call-unresolved` error code.

packages/syft-restrict/README.md
pjwerneck: Rewrote it based on the original, with a clearer introduction and learning curve.

packages/syft-restrict/README.md
```diff
@@ -1,263 +1,100 @@
...
-   │
-   └─ 5. EMIT (artifact = public glue + obfuscated math,
-              attestation{source_hash, policy_hash, PASS}).
-```
+This is how **[examples/gemma_inference.py](examples/gemma_inference.py)** generates
+**[examples/gemma_inference.obfuscated.py](examples/gemma_inference.obfuscated.py)** —
+see [examples/generate.py](examples/generate.py).
 
-"Attestation" (step 5) is a cryptographically signed statement the secure enclave produces: a machine-verifiable receipt saying _"I ran exactly this check over exactly these bytes and it passed."_ The DO checks the signature instead of trusting anyone's word.
+The same model marked up with comment blocks instead of numeric ranges lives in
```
koenvanderveen, line 86: I love the readability of this readme! I think mark up should be the only UX documented in the readme, having two UX's described here complicates the story somewhat, but perhaps we can move it to verify.md

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,352 @@
...
+
+---
+
+## Public vs private
+
+Every file has two regions:
+
+| Region      | Typical contents                                        | Checked?          |
+| ----------- | ------------------------------------------------------- | ----------------- |
+| **Public**  | Imports, data loading, wrappers the data owner can read | No (human review) |
```
koenvanderveen: the imports are checked I suppose?

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,352 @@
...
+| [blacklist.md](blacklist.md) | `tests/verify/test_disallowed.py`                           | Default-deny catalog   |
+| Edge / attack shapes         | `tests/verify/test_bypasses.py`                             | Multi-step regressions |
+
+---
+
+## Public vs private
+
+Every file has two regions:
+
+| Region      | Typical contents                                        | Checked?          |
```
koenvanderveen: perhaps more clear to say "checked by syft-restrict" (compared to checked by user)

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Always allowed syntax
+
+These shapes are fine in private code **when** nested pieces also obey the rules:
+
+| Category          | Examples                                                   |
+| ----------------- | ---------------------------------------------------------- |
+| Definitions       | `def`, `class`, `lambda`, `return`                         |
+| Names & constants | variables, numbers, strings (not f-strings)                |
+| Assignment        | `=`, `+=`, annotated assigns                               |
```
koenvanderveen, line 137: is this always allowed? also reassigning self for instance?

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,352 @@
...
+| Containers        | `list`, `tuple`, `dict`, `set`, comprehensions             |
```
koenvanderveen, line 138: I guess the note here is that they can only nest allowed syntax?

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,352 @@
...
+
+> [!NOTE]
+>
+> **Never** allowed: imports, `with`, `try`/`raise`, `async`, generators,
+> `assert`, `del`, f-strings, walrus `:=`, `match`/`case`. Those are listed
+> under [blacklist.md](blacklist.md).
+
+---
+
+## Per-file knobs
```
koenvanderveen: maybe its me, but I am not familiar with the term "knob"

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,352 @@
...
+Because call sites trust some names by spelling alone:
+
+| Name kind            | Example                                      |
+| -------------------- | -------------------------------------------- |
+| Import aliases       | `jnp`, `nn`                                  |
+| Public wrappers      | `transpose`                                  |
+| Private defs/classes | `Attention`, `helper`                        |
+| Safe builtins        | `list`, `range`                              |
+| `self` / `cls`       | only as the real first parameter of a method |
+
```
koenvanderveen, line 259: how about the result of x =transpose(y), are we allowed to reassign x?

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,352 @@
...
+---
+
+## `self` and Flax-style modules
+
+Private models usually look like:
+
+```python
+class Net(nn.Module):              # base must be allow-listed (e.g. flax.linen.Module)
+    def setup(self):                 # often public (data owner can read wiring)
```
koenvanderveen, line 274: Now that I read this, perhaps this is the weakest part of our setup. We assume that setup acts like a init function, something a variable is assigned here and never reassigned. But that is very jax specific, and how do we even know that setup is always called (if it wouldnt be jax)

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,352 @@
...
+> The verifier still uses public `setup` assignments when deciding
+> whether `self.<name>(...)` is safe.
+
+### Classes and hooks
+
+| Allowed                                                                      | Not allowed                                           |
+| ---------------------------------------------------------------------------- | ----------------------------------------------------- |
+| Bases that resolve to an allow-listed path (e.g. `nn.Module`)                | `object`, random private bases, non-allow-listed libs |
+| Decorators: `nn.compact`, `jax.jit`, `jax.named_scope`, `flax.linen.compact` | `@property`, `@staticmethod`, arbitrary functions     |
+| Defining `setup`, `__call__`, `__post_init__`                                | `__getattr__`, `__reduce__`, other magic methods      |
```
koenvanderveen: do we need __post_init__?

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,352 @@
...
+>
+> The verifier still uses public `setup` assignments when deciding
+> whether `self.<name>(...)` is safe.
+
+### Classes and hooks
+
+| Allowed                                                                      | Not allowed                                           |
+| ---------------------------------------------------------------------------- | ----------------------------------------------------- |
+| Bases that resolve to an allow-listed path (e.g. `nn.Module`)                | `object`, random private bases, non-allow-listed libs |
+| Decorators: `nn.compact`, `jax.jit`, `jax.named_scope`, `flax.linen.compact` | `@property`, `@staticmethod`, arbitrary functions     |
```
koenvanderveen: I think in the end we are not using any of these anymore. perhaps we should just disallow them?

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,352 @@
...
+def transpose(x):
+    return x  # data owner reviews this
+
+def private_math(w):
+    return transpose(w)
+```
+
+```python
+# library path (imports public, use private)
+import jax.numpy as jnp
```
koenvanderveen, line 334: only if allowlisted

packages/syft-restrict/docs/verify.md
```diff
@@ -0,0 +1,352 @@
+# How verification works
```
koenvanderveen, line 1: this is high quality now! It helps me reason over the logic and find edge cases we may not have covered. Great job!

packages/syft-restrict/tests/fixtures/marked_model.py
```diff
@@ -0,0 +1,42 @@
...
+    def setup(self):
+        self.weight = self.param("w", nn.initializers.ones, (CONFIG["dim"],))
+
+    def __call__(self, x):
+        sq = jnp.mean(jnp.square(x), axis=-1, keepdims=True)
+        return x * jax.lax.rsqrt(sq + CONFIG["eps"]) * (1.0 + self.weight)
+
+
+class Block(nn.Module):
+    cfg: dict
```
koenvanderveen, line 28: I didnt think about this yet, is this allowed? is it secure?

packages/syft-restrict/src/syft_restrict/markers.py
```diff
@@ -0,0 +1,162 @@
...
+from .errors import MarkerError
+
+__all__ = ["parse_markers"]
+
+# Matches the whole comment, not a substring, so a marker string that happens to appear inside an
+# ordinary sentence-like comment is never mistaken for a directive.
+_MARKER_RE = re.compile(r"^#\s*syft-restrict:\s*(obfuscate|hide)(?:-(start|end))?\s*$")
+
+
+def parse_markers(source: str) -> tuple[list[tuple[int, int]], list[tuple[int, int]]]:
```
koenvanderveen: some comments here would help me or otherwise split it up in functions

packages/syft-restrict/README.md
pjwerneck: Maybe the explicit numeric ranges shouldn't be a documented UX at all. Having the two creates unnecessary conflicts. Do the parameters override markers? Does running the verifier on a file with no private markers and explicit empty ranges issues a verified result?

I think `run()` should resolve the markers and extract the ranges, and call an internal `_run()` with the current behavior. If for some reason an user wants to call `_run()` explicitly with ranges instead of adding markers to the file, it's their choice to fiddle with internals, not the official UX.

packages/syft-restrict/src/syft_restrict/verifier.py
```diff
@@ -132,301 +164,836 @@ def verify(source: str, private, policy: Policy) -> VerifyResult:
...
+            if name_rebound:
+                self.report(
+                    node,
+                    "reserved-name",
+                    f"{node.id!r} is a private-region class/def and may not be rebound",
+                )
+            return
+
+        # methods are not bare-call targets and therefore not shadowed; shared names like `setup`` are fine
+        if isinstance(node, ast.FunctionDef) and id(node) in self.scan.method_ids:
```
koenvanderveen, line 261: perhaps nitpick and not dangerous, but it looks like you might be able to define method x on class y twice because we are not checking that

packages/syft-restrict/docs/verify.md
pjwerneck: Yes. All private imports are banned outright. Imported names in the private region must resolve to a dotted path explicitly allowed through `allow_functions`.

Public imports are trusted by default, like all public code.

packages/syft-restrict/src/syft_restrict/policy.py
```diff
@@ -80,18 +48,57 @@
...
-# access (`block.w`), the same attribute-access-hook class banned in §3.1 #6, so default-deny
+# access (`block.w`), the same attribute-access-hook class banned for dunder defs, so default-deny
 # rejects it. Pure inference needs only setup/__call__.
 ALLOWED_DECORATORS: frozenset[str] = frozenset(
     {"nn.compact", "jax.jit", "jax.named_scope", "flax.linen.compact"}
 )
 
-# The only dunder/hook methods a model class may *define* (approach-B §3.1 #6).
+# The only dunder/hook methods a model class may *define* (docs/verify.md#allow_functions--paths-callable-by-name).
 ALLOWED_DUNDER_DEFS: frozenset[str] = frozenset({"__call__", "setup", "__post_init__"})
```
koenvanderveen: I think we can remove post init

packages/syft-restrict/src/syft_restrict/policy.py
```diff
@@ -80,18 +48,57 @@
...
+        "super",
+    }
+)
+
+# Decorators allowed above a def/class in the private region (docs/verify.md#allow_functions--paths-callable-by-name).
 # NOTE: `property` (and any descriptor) is deliberately absent — it runs code on a bare attribute
-# access (`block.w`), the same attribute-access-hook class banned in §3.1 #6, so default-deny
+# access (`block.w`), the same attribute-access-hook class banned for dunder defs, so default-deny
 # rejects it. Pure inference needs only setup/__call__.
 ALLOWED_DECORATORS: frozenset[str] = frozenset(
```
koenvanderveen: we are not using these, remove

packages/syft-restrict/docs/verify.md
pjwerneck: Yes. An allowed outer construct does not  grant permission to its inner pieces. Each node is verified independently. I will make that clearer in the document.

packages/syft-restrict/docs/verify.md
pjwerneck: Good point. I'll change the section title to "Per-file policy configuration", matching the terms used in the rest of the project.

packages/syft-restrict/docs/verify.md
pjwerneck: Yes. Ordinary locals can be reassigned at will. Only reserved names are protected.

However, the verifier still tracks the source to attempt to resolve and reject if the name is used as a callable.

I'll add a line to clarify that too.

packages/syft-restrict/docs/verify.md
pjwerneck: We already block `__init__`. If we block `__post_init__` too, there's no hook for initializing dataclasses at construction time. I think there's a case for leaving it in, as convenience.

packages/syft-restrict/tests/fixtures/marked_model.py
pjwerneck: It's secure, keeping the documented limitations in mind. Malicious public wrappers still depend on human review.

However, it's still a gap because the verifier doesn't catch potential abuse of a declared field like that. It's not a security issue, since you still need malicious public code to abuse it, but it's inconsistent when compared to how we handle similar implicit assignments.

I will close this gap too. `self.cfg()` in that case shouldn't be trusted.

packages/syft-restrict/src/syft_restrict/markers.py
pjwerneck: I modified the structure to use a stack to handle the start/end and nesting and simplified the implementation. I think it's easier to understand now.

packages/syft-restrict/src/syft_restrict/verifier.py
pjwerneck: Yes. It's not exploitable in any way that matters for the security model, but they might be confusing in obfuscated code. I'll fix it.

packages/syft-restrict/docs/verify.md
pjwerneck: Yes. I'll just ban the decorator node entirely, since those were the only decorators allowed.

Anyone who  wants decorator behavior should make an explicit call instead. I'll document that.

packages/syft-restrict/docs/verify.md
pjwerneck: Assignment is always allowed as a construct, it's not banned in itself, but name and target are checked separately from the assignment. The name must not be reserved, the target must be allowed, and the assignment resets the later call-trust of the target.

I will add a line clarifying that too.

packages/syft-restrict/docs/verify.md
pjwerneck: That's a matter of correctness to be resolved at runtime, not security. In the static analysis, the `self.attr()` calls in `__call__` only pass if they were assigned a safe target somewhere in the class, or if they were not assigned anywhere and therefore presumed to come from the base class. 

Order doesn't matter. The verifier collects all assignments first and requires **all** of them to be vetted. If another method reassigns them between `setup` and `__call__`, that reassignment needs to be vetted too. If setup doesn't run before `__call__`, it's a runtime `AttributeError`, not a bypass. It's not the verifier's job to check for that. 

We can flip the rule to default-deny any `self.attr` that wasn't explicitly assigned in the class, so the assumption that undefined attributes come from the base class no longer holds. It's a stricter rule, but breaks access to base class attributes.

packages/syft-restrict/src/syft_restrict/verifier.py
```diff
@@ -132,301 +164,836 @@ def verify(source: str, private, policy: Policy) -> VerifyResult:
...
-        if dotted is not None:
-            root = dotted.split(".")[0]
+
+        path = dotted_name(node)
+
+        if path is not None:
+            root = path.split(".")[0]
             if root in ("self", "cls"):
+                # a single self.<name>/cls.<name> read is fine; a deeper chain is not
+                if self_attr_name(node) is not None:
```
koenvanderveen, line 674: I think we may want to consider making this more strict. I think its only fine to call that if  we defined it ourselves, so we know the getattr doesnt create side effects

packages/syft-restrict/src/syft_restrict/verifier.py
```diff
@@ -132,301 +164,836 @@ def verify(source: str, private, policy: Policy) -> VerifyResult:
...
+                safe.pop(t.id, None)
+
+    def _is_safe_local_source(self, value: ast.AST) -> bool:
+        """Is a local provably safe to call?"""
+        attr = self_attr_name(value)
+        if attr is None and isinstance(value, ast.Subscript):
+            attr = self_attr_name(value.value)
+        if attr is not None:
+            return self._self_attr.is_safe(attr, self._enclosing_class())
+        return _all_leaves_safe(value, self._is_safe_local_leaf)
```
koenvanderveen, line 810: I dont have an answer to this question yet, but do we need any asignment that not an attribute?

packages/syft-restrict/src/syft_restrict/verifier.py
```diff
@@ -132,301 +164,836 @@ def verify(source: str, private, policy: Policy) -> VerifyResult:
...
         for t in targets:
-            self._check_reserved_target(t)
+            if not isinstance(t, ast.Name):
+                continue
+            if value is not None and self._is_safe_local_source(value):
+                safe[t.id] = True
+            else:
+                safe.pop(t.id, None)
+
+    def _is_safe_local_source(self, value: ast.AST) -> bool:
```
koenvanderveen, line 803: we are calling this function from _track_safe_local after we already **skipped anything that is not an ast.Name, like ast.Attribute**, so in that context value here can only be an ast.Name. However, we are calling functions that sound like they should be applied to an attribute. (self_attr_name), this is confusing me a bit

packages/syft-restrict/src/syft_restrict/verifier.py
pjwerneck: That will prevent using attrs inherited from the base class. It's the same question I asked above.

https://github.com/OpenMined/PySyft/pull/9436#discussion_r3589335157

packages/syft-restrict/src/syft_restrict/verifier.py
pjwerneck: No, we don't strictly need it. It's a trade-off between convenience and attack surface. If we remove that, we lose any legitimate alias-then-call pattern, like this:

```
block = self.layers[0]
return block(x)
```

Because `block(x)` is trusted only because block was tracked as safe. 

Without local assignments, you'd have to do:

```
return self.layers[0](x)
```

I don't think the trade-off is worth it. The reduced attack surface here is minimal. The local aliased call must still resolve to a safe callable.

packages/syft-restrict/README.md
koenvanderveen: sounds good

packages/syft-restrict/docs/verify.md
koenvanderveen: perhaps that should be documented here, it reads as if the public code is not checked at all

packages/syft-restrict/docs/verify.md
koenvanderveen: I think reassignment is tricky though. Because of control flow we never know whether something actually was reassignment. So instead of tracking the "current" value of something (which we cant), we should track all assignments and check if they are all safe.

packages/syft-restrict/docs/verify.md
koenvanderveen: yes I was thinking about something similar, or we could let the user pass in certain fields that can be used but not reassigned or something like that. Perhaps out of scope for this PR

packages/syft-restrict/docs/verify.md
koenvanderveen: hmm oke

packages/syft-restrict/src/syft_restrict/verifier.py
pjwerneck: No, there's target and value. The `Name` filter is for the target name, the variable being assigned to. The value is the one being assigned and need to be tracked as safe.

I think it could be confusing because we have to iterate possible chained assignments, like `x = y = value`. I will add some comments clarifying that.

koenvanderveen (approved):

packages/syft-restrict/docs/verify.md
pjwerneck: Yes, we can't track dynamic branching, but we still block everything at the source.  It gets rejected in some way.

For example:

```
if c: g = safe
else: g = unsafe
g()
```
`unsafe` is visited last, so in theory this case gets rejected with call-unresolved.


But if we flip it:

```
if c: g = unsafe
else: g = safe
g()
````

In this `safe` is visited last, so in theory it's not rejected, even though it should have.

However, the false negative is not exploitable anyway because defining `unsafe` with something like a banned name or other unsafe construct would have triggered another error at the origin, so while the assignment tracking theoretically fails in this case, it's inconsequential in practice.

PR #9437, opened 2026-07-08
refactor: remove AutoApproveConfig.load/.save and route all I/O through SyftBgConfig

pjwerneck:
## Summary

Remove AutoApproveConfig.load()/.save(); all config I/O now goes through SyftBgConfig (CLI commands, api.py, orchestrator.py, JobApprovalHandler, save_gcp_project_id).

## Changes

- Add SyftBgConfig.edit(), a locked (fcntl) read-modify-write context manager; skips save() if the block raises.
- Fix config correctness bugs: unknown top-level YAML keys were being dropped on save (extra="allow"); merged common fields (do_email, token paths) were getting permanently baked into sub-configs on every save, defeating inheritance (merge=False for edit()); per-service token-path overrides were getting clobbered by the parent's default.
- Fix auto_approve(): copy/hash files into a private staging directory first, resolve naming and final placement atomically under one lock — eliminates a filesystem race where concurrent callers with the same auto-generated name could corrupt or orphan each other's entries, and handles a stale/orphaned target directory as a clean failure instead of an unhandled OSError.
- Extract a shared file_lock() helper (common/locking.py); dedupes the identical lock logic that used to live separately in SyftBgConfig.edit() and JsonStateManager._file_lock.
- init() now derives both gmail_token_path and drive_token_path from a single OAuth token path and returns the canonical (post-copy) path.
- Widen auto_approve()/resolve_content_files() to Sequence[str | Path] (was list[str | Path], which rejected valid list[str]/list[Path] callers due to list invariance).
- Update notebooks/ai_audit/internal/test_email_e2e.ipynb off the removed AutoApproveConfig.load() API.
- Fix a test-isolation bug in test_email_auto_approve_flow.py that was silently leaking auto-approval directories into the real ~/.syft-bg instead of a tmp dir.

## Testing

- Added 19 new test cases.
- Added `tests/unit/syft_bg/test_init.py
- Added regression tests `tests/unit/syft_bg/test_ensure_running.py`
- Existing tests modified to the new SyftBgConfig-based code.

## Asana task

https://app.asana.com/1/1185126988600652/project/1210542925864934/task/1214797316252530


bitsofsteve (approved):
LGMT!

PR #9438, opened 2026-07-08
feat: Init RDS package

rasswanth-s:
## Summary

Extracts the Remote Data Science (RDS) domain — datasets and jobs — out of `syft-client` and into a new top-level `syft-rds` package, leaving `syft-client` as a **domain-free, generic sync engine** that knows nothing about datasets or jobs.

The sync core and the RDS product now talk through two narrow, generic seams:
- **Lifecycle callbacks** — the sync engine emits events (`peer_approved`, `peers_loaded`, `collection_restored`); RDS subscribes and supplies the dataset/job-specific behavior. Dependencies are inverted so the core never imports the domain.
- **Generic collection sync** — a domain-free `CollectionSyncSpec` (prefix + local subpath + `immutable`/`owner_only` flags) describes *how* to sync a content-hashed folder without the engine knowing it's a dataset. Concrete dataset specs are supplied by the RDS layer.

`syft-bg` and `syft-enclave` are migrated to consume the new `syft-rds` client instead of reaching into `syft-client` internals.

## Changes

- **New `packages/syft-rds/` package**
  - `SyftRDSClient` — composes the sync engine (`SyftboxManager`) with the RDS-owned `JobClient`, `SyftJobRunner`, and `SyftDatasetManager`, and wires them to sync-engine lifecycle callbacks.
  - `SyftRDSClientConfig` — composes `sync` / `job` / `dataset` sub-configs from one set of primitives (email, folder, role) so paths always line up; owns the dataset collection specs and prefixes.
  - `login_do` / `login_ds` entry points mirroring the pre-split login flow.
  - Job auto-approval (`auto_approve_and_run_jobs`, `job_matches_criteria`) moved here.
- **`syft-client` made generic**
  - `SyftboxManager` slimmed substantially — all job/dataset references removed.
  - New `BaseModelCallbackMixin` (`on` / `_emit`) providing isolated, exception-safe lifecycle callbacks.
  - New `CollectionSyncSpec` primitive plus parametrized collection folder names / wire-level prefixes; `datasite_owner_syncer` / watcher cache reworked to sync generic collections.
  - `syft-client` dependencies leaned; removed as a direct dependency of `syft-bg`.
- **Consumers migrated** — `syft-bg` and `syft-enclave` (client, login, runner, orchestrators, handlers/monitors) now use the `syft-rds` client.
- **Tests relocated** — dataset/job tests moved from `syft-client` into `packages/syft-rds/tests/`; new `SyftRDSClient` unit tests added; enclave and syft-bg tests updated for the new client naming.

## Testing

- New unit tests in `packages/syft-rds/tests/` (`test_client.py`, plus relocated dataset/job/auto-approval suites).
- Existing `syft-enclave` and `syft-bg` unit suites updated to the new client and kept passing.

## Asana task
(https://app.asana.com/1/1185126988600652/project/1210542925864934/task/1216142561490291)

- ***

### PR naming convention

Your PR title **must** follow this format or merging will be blocked:

```
type: short description in lowercase
```

**Example titles:**

- `feat: add retry logic to job approval`
- `fix: handle timeout in notification sender`
- `docs: update syft-bg README`
- `chore: bump dependencies`
- `refactor: split init flow into helpers`
- `test: add criteria validation tests`
- `ci: add release train workflow`

**Allowed types:**

| Type       | When to use                             | Example                             |
| ---------- | --------------------------------------- | ----------------------------------- |
| `feat`     | New feature or capability               | `feat: add DS rejection emails`     |
| `fix`      | Bug fix                                 | `fix: handle empty peer list`       |
| `docs`     | Documentation only                      | `docs: update syft-bg README`       |
| `chore`    | Maintenance, deps, config               | `chore: bump dependencies`          |
| `refactor` | Code restructuring (no behavior change) | `refactor: split init flow`         |
| `test`     | Adding or updating tests                | `test: add approval criteria tests` |
| `ci`       | CI/CD workflow changes                  | `ci: add release train workflow`    |
| `perf`     | Performance improvement                 | `perf: cache Drive API responses`   |
| `build`    | Build system or dependency changes      | `build: pin syft-bg>=0.2.0`         |

Just edit the PR title to fix any errors — the check re-runs automatically.

### Auto-labeling

Labels are applied automatically — you don't need to add them manually:

- **Type labels** from your PR title (e.g., `feat:` adds `feature`, `fix:` adds `bugfix`)
- **Package labels** from which files you changed (e.g., editing `packages/syft-bg/` adds `pkg:syft-bg`)

These labels are used to auto-generate categorized release notes.


packages/syft-rds/src/syft_rds/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+
+from syft_rds.client import SyftRDSClient
+from syft_rds.config import SyftRDSClientConfig
+from syft_rds.job_auto_approval import auto_approve_and_run_jobs, job_matches_criteria
+from syft_rds.login import login_do, login_ds
+
+# Generic sync-stack helpers re-exported so consumers that compose the RDS product
+# (e.g. syft-bg) have a single integration surface and need not import from
+# syft_client.sync internals directly.
+from syft_client.sync.syftbox_manager import get_jupyter_default_syftbox_folder
```
rasswanth-s, line 11: This pattern of importing syft-client imports at rds is temporary, will refactor this part shortly

packages/syft-rds/src/syft_rds/config.py
```diff
@@ -0,0 +1,187 @@
...
+)
+from syft_client.sync.sync.collection_spec import CollectionSyncSpec
+from syft_job import SyftJobConfig
+from syft_datasets.config import SyftBoxConfig
+from syft_datasets.dataset_manager import (
+    DATASET_COLLECTION_PREFIX,
+    PRIVATE_DATASET_COLLECTION_PREFIX,
+)
+
+# The RDS layer OWNS the dataset collection values: both the on-wire prefixes and
```
pjwerneck: The subpaths are defined below, but the prefixes are being imported from syft_datasets.dataset_manager. This comment seems contradictory.

packages/syft-rds/src/syft_rds/config.py
```diff
@@ -0,0 +1,187 @@
...
+        email: str,
+        folder: Path,
+        has_do_role: bool,
+        sync: SyftboxManagerConfig,
+    ) -> "SyftRDSClientConfig":
+        """Build the composed config from ONE set of primitives.
+
+        ``sync`` is built by the caller (it differs per environment); ``job``
+        and ``dataset`` are derived here from the SAME email/folder/role so all
+        paths line up with the sync engine.
```
pjwerneck: The alignment is documented, but there's no validator enforcing it. Like a `@model_validator(mode="after")` asserting that `sync.email` and `dataset.email` match.

packages/syft-rds/src/syft_rds/config.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        job = SyftJobConfig(
+            syftbox_folder=folder,
+            current_user_email=email,
+            has_do_role=has_do_role,
+        )
+        dataset = SyftBoxConfig(syftbox_folder=folder, email=email)
```
pjwerneck: Why are we taking `folder` and `email` as parameters here, instead of using attributes from sync? It looks  like a potential source of drift or inconsistencies.

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+                )
+
+    def _share_any_datasets_with_peer(self, peer_email: str) -> None:
+        """Share all datasets tagged 'any' with a specific peer.
+
+        Google Drive "anyone with link" files are not discoverable via search,
+        so explicit user sharing is added. Reads the cache populated during
+        ``pull_initial_state()`` in the nested DatasiteOwnerSyncer.
+        """
+        for tag, content_hash in self._sync.datasite_owner_syncer._any_shared_collections:
```
pjwerneck: This is one example of the encapsulation violations.

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+            try:
+                self._sync._connection_router.owner_share_collection(
+                    DATASET_COLLECTION_PREFIX, tag, content_hash, [peer_email]
+                )
+            except Exception:
```
pjwerneck, line 240: This doesn't look like an acceptable "ignore all errors" case as the comment suggests, since `owner_share_collection` is not a trivial method. If a network request fails, it's going to behave as if it's already shared. This could also potentially silently real bugs in `owner_share_collection`.

Also, using a bare `except:` or a blanket `except Exception:` to silently suppress all errors is a python anti-pattern. In legitimate cases, `contextlib.suppress` is the more idiomatic way of doing that.

packages/syft-rds/src/syft_rds/client.py
pjwerneck, line 1: There's a lot of lines that violate encapsulation of the sync_engine's private attributes, most of them reaching into `self._sync._connection_router`.  If the goal of the refactoring is a "domain-free core", it should not reach into private internals. The sync_engine should have public methods exposing the functionality.

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+
+    # ------------------------------------------------------------------ #
+    # job product surface (RDS-owned)
+    # ------------------------------------------------------------------ #
+    def submit_bash_job(
+        self,
+        user: str,
+        script: str,
+        job_name: str = "",
+        sync=True,
```
pjwerneck: This method isn't honoring the `sync` parameter, like others do.

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+        self._sync.push_job_files(job_dir)
+
+    def submit_python_job(
+        self,
+        user: str,
+        code_path: str,
+        job_name: str | None = "",
+        dependencies: list[str] | None = None,
+        entrypoint: str | None = None,
+        sync=True,
```
pjwerneck: Also not honoring `sync`.

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+        self, prefix: str, tag: str, local_dir: Any
+    ) -> None:
+        """A collection was restored from the backend by the owner-syncer.
+
+        For the PRIVATE dataset collection, rewrite ``private_metadata.yaml``'s
+        machine-specific ``data_dir`` to the current local path, so job execution
+        can locate the real data after a restore onto a new machine/path. This is
+        dataset domain knowledge, so it lives here (rds) rather than the sync core.
+        """
+        from syft_datasets.dataset_manager import PRIVATE_DATASET_COLLECTION_PREFIX
```
pjwerneck: Redundant import. It's already imported at the top of the file.

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+
+        if prefix != PRIVATE_DATASET_COLLECTION_PREFIX:
+            return
+
+        import yaml
```
pjwerneck: Should be at the top too.

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+
+        metadata_path = Path(local_dir) / "private_metadata.yaml"
+        if not metadata_path.exists():
```
pjwerneck, line 204: This checks if the file exists, but if it's empty, `yaml.safe_load` below returns `None`, causing an `AttributeError` on the `data.get` call next, that will  not interrupt flow, only be logged as a callback failure by the _emit handler . Since `read_text()` is  going to be called anyway, this guard should check the content length too.

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+    def _build_rds_pair_from_managers(cls, ds_mgr, do_mgr):
+        """Wrap a paired ``(ds, do)`` ``SyftboxManager`` tuple into RDS clients.
+
+        Peers were already approved/loaded during pairing (before our callbacks
+        were registered), so we replay the DO-side setup so its owned
+        ``JobClient`` has the DS job folders + install sources.
+        """
+
+        def _build(mgr) -> "SyftRDSClient":
+            job_cfg = SyftJobConfig(
```
pjwerneck: This is duplicating the construction logic that `SyftRDSClientConfig._compose` already does.

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+        code_path: str,
+        job_name: str | None = "",
+        dependencies: list[str] | None = None,
+        entrypoint: str | None = None,
+        sync=True,
+        force_submission: bool = False,
+        ignore_peer_version: bool = False,
+    ):
+        peer_emails = {p.email for p in self._sync.peer_manager.syncable_peers}
+        if user not in peer_emails:
```
pjwerneck, line 283: The `submit_bash_job` method above does not check peer-membership the same way this does. Was that intentional?

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+        data = yaml.safe_load(metadata_path.read_text())
+        expected_dir = str(local_dir)
+        if data.get("data_dir") != expected_dir:
+            data["data_dir"] = expected_dir
+            metadata_path.write_text(yaml.safe_dump(data, indent=2, sort_keys=False))
+
+    def _on_peers_loaded(self, *args: Any, **kwargs: Any) -> None:
+        """Peers were (re)loaded: copy each peer's advertised install source
+        into our owned job client so submitted run.sh references the DO's path."""
+        for peer in self._sync.peer_manager.peer_store.syncable_peers:
```
pjwerneck: This should use `peer_manager.syncable_peers` instead of reaching inside `peer_store`, since the `syncable_peers` property exists for that.

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+
+        if not self.has_do_role:
+            raise ValueError("Only dataset owners can share datasets")
+
+        # Verify dataset exists
+        dataset = self._dataset_manager.get(name=tag, datasite=self.email)
+        if dataset is None:
+            raise ValueError(f"Dataset {tag} not found")
+
+        # Compute current content hash from local files
```
pjwerneck: This section is copy-paste from the `_upload_dataset_to_collection`. It should be extracted to a method to avoid code duplication.

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+                files[mock_file.name] = mock_file.read_bytes()
+        metadata_path = dataset.mock_dir / "dataset.yaml"
+        if metadata_path.exists():
+            files["dataset.yaml"] = metadata_path.read_bytes()
+        if dataset.readme_path and dataset.readme_path.exists():
+            files[dataset.readme_path.name] = dataset.readme_path.read_bytes()
+
+        content_hash = CollectionFolder.compute_hash(files)
+
+        # Share collection
```
pjwerneck: This is also copy-paste from `upload_dataset_to_collection`.

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+
+        print("\n✅ Job submitted successfully!")
+        print("   Status : inbox (waiting for DO to review)")
+        print(f"\n⏳ Next step: wait for {user} to approve and run it.")
+        print("   Check progress with: client.jobs")
+
+    @property
+    def jobs(self) -> Any:
+        """List of jobs. Auto-syncs first unless PRE_SYNC=false."""
+        if os.environ.get("PRE_SYNC", "true").lower() == "true":
```
pjwerneck: This `PRE_SYNC` guard checking the env var is repeated three times. It should be extracted to a property, or maybe checked once and saved.

packages/syft-rds/src/syft_rds/client.py
```diff
@@ -0,0 +1,702 @@
...
+
+            # Upload mock data to collection folder
+            mock_folder_id = self._upload_dataset_to_collection(dataset, users)
+
+            # Upload private data to a separate owner-only collection
+            if upload_private:
+                private_folder_id = self._upload_private_dataset_to_collection(dataset)
+
+            if sync:
+                self.sync()
```
pjwerneck, line 431: The word "sync" is carrying four different meanings in the file. That's a real readability issue. In most cases it refers to the `SyftboxManager` instance, but it's also used as a boolean flag meaning to run a full sync after the method, the `sync()` method itself, that delegates to `self._sync.sync()` , and `config.sync` referring to the sub-config.

I'd rename the boolean to something more explicit, like `sync_after`, and the private attribute to `_sync_engine` so it matches the public property `sync_engine`.

syft_client/sync/syftbox_manager.py
```diff
@@ -110,18 +101,29 @@ def for_colab(
...
 
+        # No collection specs by default: the generic sync engine knows nothing
+        # about custom collections (eg: datasets or jobs).
+        if collection_specs is None:
+            collection_specs = []
+
         syftbox_folder = get_colab_default_syftbox_folder(email)
         use_in_memory_cache = False
-        collections_folder = syftbox_folder / email / COLLECTION_SUBPATH
+        collections_folder = (
```
pjwerneck: This interpolation is repeated four times in the file. It should be extracted to a common function/method.

Also, it's not clear why it assumes `collection_specs[0] ` is special.  Is it because it's the public spec? What happens if they are reordered? It should have a clarifying comment here.

If it must always be the public, it should refer to it by an unambiguous value.

syft_client/sync/syftbox_manager.py
```diff
@@ -995,7 +903,7 @@ def load_peers(self, force_download: bool = False):
                 instead of using the cached copy.
         """
         cast(PeerManager, self.peer_manager).load_peers(force_download=force_download)
-        self._sync_peer_install_sources_to_job_client()
+        self._emit_peers_loaded()
```
pjwerneck, line 955: The `sync` method calls `load_peers` on every invocation, so every `sync` emits `peers_loaded`. Is that intentional? Is there any idempotency handling?

syft_client/sync/syftbox_manager.py
```diff
@@ -822,84 +802,12 @@ def add_peer(
     ):
         """Add a peer. Delegates to PeerManager."""
         self.peer_manager.add_peer(peer_email, force=force, verbose=verbose)
-        self._sync_peer_install_sources_to_job_client()
+        self._emit_peers_loaded()
         if self.has_do_role:
```
pjwerneck, line 816: Does this mean a peer added by the DO is approved automatically? If that's the case, there should be a comment clarifying it.

syft_client/sync/syftbox_manager.py
```diff
@@ -822,84 +802,12 @@ def add_peer(
...
             self._post_approve_peer_do(peer_email)
         if sync:
```
pjwerneck, line 818: This is the same issue with the word "sync" mentioned in client.py. For clarity, the boolean could be renamed to `sync_after`.

pjwerneck (commented):
The intent to create a domain-free core and define a public API is clear, but there are still violations of that boundary. It's mostly clear for the happy path, but there's still coupling and private-reaching when handling others cases.
 
There are still broken tests and conflicts, so this review is not final.

syft_client/sync/syftbox_manager.py
koenvanderveen: I think this is a valid concern but out of scope for this PR

syft_client/sync/syftbox_manager.py
koenvanderveen: can you elaborate why you think thats the case

packages/syft-rds/src/syft_rds/config.py
koenvanderveen: agreed, perhaps we can move that variable here. We do need to look into gdrive transport, which needs it for some functionality

packages/syft-rds/src/syft_rds/config.py
koenvanderveen: Added a `@model_validator(mode="after")` (`_assert_sub_configs_aligned`) that asserts `sync`/`job`/`dataset` agree on email, `syftbox_folder`, and `has_do_role`, raising if they drift. (15ccb48eef)

packages/syft-rds/src/syft_rds/config.py
koenvanderveen: Good catch — `_compose` now takes only `sync` and derives email/folder/role from it (single source of truth), instead of accepting separately-passed `email`/`folder` params that could drift. All callers updated. (15ccb48eef)

packages/syft-rds/src/syft_rds/client.py
koenvanderveen: not sure, syft-client is responsible for sharing collections (files that may be shared with many users), syft-rds is responsible for sharing datasets, which is a specific implementation of those. Given that design, do you still think this is not structured correctly? If so, can you explain why

pjwerneck: The client is reaching for private attributes of its children, `_any_shared_collections` in this case.  There are many cases where `_connection_router` is reached into. If that's intended, there's no point in prefixing these attributes with `_`. That's a convention for private attributes.

syft_client/sync/syftbox_manager.py
pjwerneck: I'm asking if that's the case. It's calling `self._post_approve_peer_do` if `self.has_do_role` is `True`. We have `self.peer_manager.approve_peer_request` making that same call, but in this case we're getting it on a `self.peer_manager.add_peer`, so it's going to emit the `peer_approved` callback, even though `approve_peer_request` was never called for this case. The workflow here isn't clear.

packages/syft-rds/src/syft_rds/client.py
koenvanderveen: Fixed in a6a51c1ad0 — removed the redundant import (`PRIVATE_DATASET_COLLECTION_PREFIX` is already imported at the top of the file).

packages/syft-rds/src/syft_rds/client.py
koenvanderveen: Fixed in a6a51c1ad0 — moved `import yaml` to the top of the file.

packages/syft-rds/src/syft_rds/client.py
koenvanderveen: Fixed in a6a51c1ad0 — after `yaml.safe_load` we now do `if not data: return`, which handles a missing/empty file (safe_load returning `None`) so the `data.get` below can't raise AttributeError.

packages/syft-rds/src/syft_rds/client.py
koenvanderveen: Fixed in a6a51c1ad0 — now uses `peer_manager.syncable_peers` instead of reaching into `peer_store`.

packages/syft-rds/src/syft_rds/client.py
koenvanderveen: Fixed in a6a51c1ad0 — extracted `_collect_mock_files` (file collection) and `_share_dataset_collection` (share logic); both `_upload_dataset_to_collection` and `share_dataset` now reuse them.

packages/syft-rds/src/syft_rds/client.py
koenvanderveen: Fixed in a6a51c1ad0 — this share block is now the shared `_share_dataset_collection` helper, reused by `_upload_dataset_to_collection` and `share_dataset`.

packages/syft-rds/src/syft_rds/client.py
koenvanderveen: Fixed in a6a51c1ad0 — extracted into a `_pre_sync_enabled` property, used at all three call sites.

syft_client/sync/syftbox_manager.py
koenvanderveen: Fixed the repeated interpolation in a6a51c1ad0 — extracted it into a module-level `default_collections_folder(syftbox_folder, email, collection_specs)` helper, now used at all four call sites.

I also documented the `collection_specs[0]` assumption in that helper's docstring (it's the public spec by convention). The broader question of referring to the public spec by an unambiguous value rather than index `[0]` I've left as a follow-up.

packages/syft-rds/src/syft_rds/client.py
koenvanderveen: Addressed in e2358cf904. Kept the storage as a `PrivateAttr` (it's a runtime cache rebuilt during sync, so it shouldn't be serialized or part of the model schema), and exposed a proper public API on `DatasiteOwnerSyncer` instead:

- `any_shared_collections` — a read-only property returning a copy of the pairs
- `register_any_shared_collection(tag, content_hash)` — a deduplicated mutator

The client now goes through those instead of touching `_any_shared_collections`, and `_update_any_shared_collections_cache` reuses the mutator (removing the duplicated dedup). The broader `_connection_router` encapsulation cleanup is tracked in the sibling thread.

packages/syft-rds/src/syft_rds/client.py
koenvanderveen: Resolved in 0d2ad9d3e4 by removing the `sync` parameter. It was never read in the method body (a no-op) and no caller passes it. On the DS submit path `push_job_files` *is* the submission — a later `sync()` only pulls down and would push nothing — so there's no post-sync step to gate here, unlike `create_dataset`/`delete_dataset`. Removing it rather than implementing it.

packages/syft-rds/src/syft_rds/client.py
koenvanderveen: Resolved in 0d2ad9d3e4 by removing the `sync` parameter. It was never read in the method body (a no-op) and no caller passes it. On the DS submit path `push_job_files` *is* the submission — a later `sync()` only pulls down and would push nothing — so there's no post-sync step to gate here, unlike `create_dataset`/`delete_dataset`. Removing it rather than implementing it.

packages/syft-rds/src/syft_rds/config.py
pjwerneck: Can't move it here. syft-rds depends on syft-client, so it'd be a cyclce. I left both copies and added a test to ensure they are in sync.

syft_client/sync/syftbox_manager.py
koenvanderveen: Then idea is that we can "approve" a peer in two ways

1) if we already go a peer request, we approve that peer request
2) if we didnt get a peer request, we pre-approve this peer such that whenever the other party adds us as a peer, everything is set up

koenvanderveen (approved):

PR #9443, opened 2026-07-15
feat: versioned DatasetStorage for reading/writing datasets across protocol versions

koenvanderveen:
## Summary
Mirrors the syft-job migration/storage architecture (PR #9434) in **syft-datasets**, adding a versioned read/write layer so datasets can evolve their on-disk format while staying compatible with older peers.

- **`DatasetStorage`** — the single choke point for all dataset filesystem IO. Reads select a `ProtocolCodec` by the on-disk protocol version and upgrade the object to the latest version in memory; writes downgrade to the target protocol and let the codec persist it. Nests a `MigrationRegistry`, `peer_schemas`, and codecs.
- **`v<n>` path segment** — new datasets can live under `public/syft_datasets/v<n>/<name>/` (and `private/…`). **Protocol 0** = the last release (syft-client 0.1.117 / syft-dataset 0.1.20): flat layout, no `v<n>`, no `canonical_name`/`version` fields — read and written back **byte-compatibly**.
- **Versioned models** — `Dataset` / `PrivateDatasetConfig` are now `MigratableObject`s under `models/<object>/v1.py`. Object schemas are unchanged since 0.1.117, so only `v1` exists; the protocol bump is purely a path-layout + identity-field change. `dataset.py` is kept as a re-export shim.
- **`DatasetRef`** — path-derived identity `(owner, name, protocol_version)`; the dataset name comes from the path, not the spoofable file.
- **Broadcast write** — a public dataset is a *single copy read by the whole audience* (unlike jobs, which target one peer), so the write protocol(s) come from the audience's versions (`target_protocol_versions_for_peers`), defaulting to the **widest-compatible (oldest)** protocol when no peers are known. `peer_schemas` is injectable but not yet populated by syft-client, so **today datasets are written in protocol 0 and the v1 layout ships dormant** (writing v1 now would hide datasets from 0.1.117 peers).
- **`SyftDatasetManager`** routes all reads/writes through `DatasetStorage` and supports **multi-version write** (one on-disk copy per protocol version in the audience); `get_all` dedupes copies across layouts, preferring the newest protocol.
- **Release tooling** — `scripts/export_release_artifact.py` + `generate_release_fixture.py`; the protocol-0 / 0.1.20 artifacts and fixtures are derived as if that release had emitted them.
- **Tests** mirror `syft-job/tests/migrations` (unit + p2p), including a byte-exact cross-release round-trip over hand-generated fixtures and a multi-version-write test. CI (`unit-tests.yml` `dataset-tests` job) and `justfile` (`test-unit-datasets`) wired.

## Scope
Package core only, per discussion. syft-client wiring (advertising each peer's *dataset* protocol version and writing in it at create/share) is a follow-up — syft-client currently exchanges only a single client-version string, not per-package schemas.

## Testing
- `uv run pytest packages/syft-datasets/tests` → **40 passed**
- `just test-unit-fast` → 406 passed
- `tests/unit` dataset + sync_manager + delete suites → green
- `pre-commit run --all-files` → clean


packages/syft-datasets/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "syft-dataset"
-version = "0.1.20"
+version = "0.1.21"
```
koenvanderveen: is this necessary?

packages/syft-datasets/scripts/generate_release_fixture.py
```diff
@@ -0,0 +1,126 @@
...
+
+DO_EMAIL = "do@test.org"
+DS_EMAIL = "ds@test.org"
+
+FIXTURES_DIR = (
+    Path(__file__).resolve().parents[1] / "tests" / "migrations" / "p2p" / "fixtures"
+)
+
+
+def _seed_syftbox(syftbox: Path, src_dir: Path, protocol_version: str) -> None:
```
koenvanderveen: double check this script

packages/syft-datasets/src/syft_datasets/config.py
```diff
@@ -20,45 +38,119 @@ def private_dir(self) -> Path:
     def public_dir(self) -> Path:
         return self.syftbox_folder / self.email / "public"
 
+    def public_datasets_root_for_datasite(self, datasite: str) -> Path:
```
koenvanderveen: it feels like this should be part of the codec

packages/syft-datasets/src/syft_datasets/dataset.py
```diff
@@ -1,243 +1,15 @@
-from datetime import datetime, timezone
-from functools import cached_property
-from pathlib import Path
-from typing import ClassVar
-from syft_permissions.spec.ruleset import PERMISSION_FILE_NAME
-from typing_extensions import Self
-from uuid import UUID, uuid4
+"""Backwards-compatible re-exports.
```
koenvanderveen: Do we still need this file at all? Where do we use these "historical import paths"

packages/syft-datasets/src/syft_datasets/dataset_manager.py
```diff
@@ -166,86 +136,106 @@ def create(
...
+        # (The stricter _validate_dataset_name regex is intentionally not enforced
+        # here to preserve historical behavior, e.g. names with spaces.)
+        self.storage.validate_dataset_name(name)
+        mock_path = to_path(mock_path)
+        private_path = to_path(private_path)
+        readme_path = to_path(readme_path) if readme_path else None
         tags = tags or []
 
-        mock_url = self.syftbox_config.get_mock_url_for_my_dataset(dataset_name=name)
+        target_versions = self._target_protocol_versions(users)
```
koenvanderveen: we should probably by default always write in the current version

packages/syft-datasets/src/syft_datasets/dataset_manager.py
```diff
@@ -80,14 +84,7 @@ def _prepare_mock_data(self, dataset_name: str, src_path: Path) -> list[Path]:
...
 
-    def _prepare_private_data(
-        self,
-        dataset_name: str,
-        src_path: Path,
-    ) -> list[Path]:
-        private_dir = self.syftbox_config.private_dir_for_my_dataset(
-            dataset_name=dataset_name
-        )
+    def _prepare_private_data(self, private_dir: Path, src_path: Path) -> list[Path]:
```
koenvanderveen: private_dir should be called private_target_dir

packages/syft-datasets/src/syft_datasets/dataset_manager.py
```diff
@@ -111,36 +108,9 @@ def _prepare_private_data(
...
-
-        private_metadata_path.parent.mkdir(parents=True, exist_ok=True)
-        private_config.save(filepath=private_metadata_path)
-
-    def _prepare_readme(self, dataset_name: str, src_file: Path | None) -> list[Path]:
-        target_mock_dir = self.syftbox_config.get_my_mock_dataset_dir(
-            dataset_name=dataset_name
-        )
+    def _prepare_readme(
+        self, target_mock_dir: Path, src_file: Path | None
```
koenvanderveen: target_mock_dir should be called target_dir here

packages/syft-datasets/src/syft_datasets/dataset_manager.py
```diff
@@ -39,11 +44,10 @@ def _validate_dataset_name(self, dataset_name: str) -> None:
...
                 f"Invalid dataset name '{dataset_name}'. Only alphanumeric characters, underscores, and hyphens are allowed."
             )
+        # Reject names that collide with a protocol-version directory (v1, v2, ...).
+        self.storage.validate_dataset_name(dataset_name)
 
-    def _prepare_mock_data(self, dataset_name: str, src_path: Path) -> list[Path]:
-        target_mock_dir = self.syftbox_config.get_my_mock_dataset_dir(
-            dataset_name=dataset_name
-        )
+    def _prepare_mock_data(self, target_mock_dir: Path, src_path: Path) -> list[Path]:
```
koenvanderveen: all these `_prepare_<x>` methods should have a ref as input, compute the target path and do the copying

packages/syft-datasets/src/syft_datasets/dataset_manager.py
```diff
@@ -345,6 +322,15 @@ def _repr_html_(self) -> str:
 
         return dataset_manager_repr_html(self.get_all())
 
+    def _datasites_to_check(self, datasite: str | None) -> list[str]:
```
koenvanderveen: this method has a weird name. Perhaps it should have no args and we should do something like datasites = datasite if datasite is not None else self._all_syftbox_datasites() when we call it

packages/syft-datasets/tests/migrations/p2p/fixtures/syft_dataset-0.1.21-protocol1_syftbox/do@test.org/private/syft_datasets/v1/demo/private.csv
koenvanderveen, line 1: this whole folder should probably not exist because this version has not been released yet

packages/syft-datasets/src/syft_datasets/dataset_manager.py
```diff
@@ -166,86 +136,106 @@ def create(
         users: list[str] | str | None = None,
         # copy_private_data: bool = True, # TODO
```
koenvanderveen, line 47: perhaps we should also have the posibility to pass in protocol verisons here in case we want to override something

packages/syft-datasets/src/syft_datasets/dataset_storage.py
```diff
@@ -0,0 +1,216 @@
...
+        path = codec.private_metadata_path(ref)
+        if not path.exists():
+            raise PrivateConfigNotFoundError(
+                f"Private config not found for dataset {ref.name}"
+            )
+        return self._upgrade(
+            codec.read(path, "PrivateDatasetConfig"), "PrivateDatasetConfig"
+        )
+
+    def write_dataset(self, ref: DatasetRef, dataset: Dataset) -> Path:
```
koenvanderveen: probably for all the writes of datasets, we first want to see which versions are already on disk. If we have multiple versions on disk already we need to make sure that they are in sync. So we first need a step that checks the existing protocols on disk, then finds the codecs for those protocols (and add the ref protocol) and then write for all codecs

packages/syft-datasets/src/syft_datasets/dataset_manager.py
```diff
@@ -166,86 +136,106 @@ def create(
...
+        # Return the newest protocol version written (richest layout).
+        newest = max(created, key=int)
+        return created[newest]
+
+    def _target_protocol_versions(self, users: list[str] | str | None) -> set[str]:
+        """The protocol versions to write so the audience can read the dataset."""
+        peer_emails = None if users in (None, SHARE_WITH_ANY) else list(users)
+        return self.storage.target_protocol_versions_for_peers(peer_emails)
+
+    def _create_one_version(
```
koenvanderveen: I think the responsibility of this should probably lie with the storage layer

packages/syft-datasets/tests/migrations/unit/test_history_artifacts.py
```diff
@@ -0,0 +1,100 @@
...
+    ReleasedPackageProtocolInfo,
+    ReleasedProtocol,
+)
+
+from syft_datasets.migrations import dataset_registry
+from syft_datasets.migrations.history import PACKAGE_ARTIFACTS_DIR, PROTOCOLS_DIR
+from syft_datasets.models import DatasetV1
+
+
+def test_0_1_20_artifact_file_loads():
```
koenvanderveen: this should be more generic, it should just load every package artifact that was released

packages/syft-datasets/tests/migrations/unit/test_history_artifacts.py
```diff
@@ -0,0 +1,100 @@
...
+        # The filename encodes the protocol version: protocol-<n>.json.
+        assert path.name == f"protocol-{schema.version}.json"
+        assert schema.protocol_name == "syft-dataset"
+        assert schema.supported_versions
+        assert set(schema.current_object_schemas) == set(schema.supported_versions)
+        for canonical_name in schema.supported_versions:
+            assert schema.current_schema(canonical_name)
+
+
+def test_protocol_0_released_protocol_loads():
```
koenvanderveen: we can remove this, we already have the one that loads all of them

packages/syft-datasets/src/syft_datasets/dataset_manager.py
```diff
@@ -353,30 +339,24 @@ def get_all(
...
-        for datasite in datasites_to_check:
-            public_datasets_dir = self.syftbox_config.public_datasets_dir_for_datasite(
-                datasite
-            )
-            if not public_datasets_dir.exists():
+        all_datasets = []
+        for ref in best_ref.values():
+            try:
+                all_datasets.append(self.storage.read_dataset(ref))
+            except Exception:
```
pjwerneck: Blanket `Exception` catch with no logging. This will silently ignore errors due to unreadable datasets in `read_dataset`. Not minor issue, since migration failures or schema inconsistencies can make the dataset vanish with no logs.

packages/syft-datasets/src/syft_datasets/dataset_manager.py
```diff
@@ -166,86 +136,106 @@ def create(
...
+        private_path = to_path(private_path)
+        readme_path = to_path(readme_path) if readme_path else None
         tags = tags or []
 
-        mock_url = self.syftbox_config.get_mock_url_for_my_dataset(dataset_name=name)
+        target_versions = self._target_protocol_versions(users)
+        dataset_uid = uuid4()
+
+        created: dict[str, Dataset] = {}
+        for protocol_version in sorted(target_versions, key=int):
```
pjwerneck: This can create a dataset in several protocol versions at once, but `delete()`  uses `get()` which uses `store.find_dataset_ref()`, which returns only a single dataset, with the newest protocol. It's going to leave older copies. Is that intended? If not, `delete()` should iterate all versions too.

packages/syft-datasets/src/syft_datasets/models/dataset/v1.py
```diff
@@ -0,0 +1,214 @@
...
+    _syftbox_config: SyftBoxConfig | None = None
+    _protocol_version: str = DATASET_PROTOCOL_VERSION
+
+    def disk_dict(self) -> dict:
+        """The on-disk form of the dataset metadata."""
+        return self.model_dump(mode="json")
+
+    @property
+    def owner(self) -> str:
+        return self.mock_url.host
```
pjwerneck: Shouldn't the canonical `owner` come from the path?

packages/syft-datasets/src/syft_datasets/dataset_manager.py
```diff
@@ -166,86 +136,106 @@ def create(
...
 
         Returns:
-            Dataset: The created Dataset object.
+            Dataset: The created Dataset object (the newest protocol version written).
         """
-        mock_path: Path = to_path(mock_path)
-        private_path: Path = to_path(private_path)
-        readme_path: Path | None = to_path(readme_path) if readme_path else None
+        # Reject names that collide with a protocol-version directory (v1, v2, ...).
+        # (The stricter _validate_dataset_name regex is intentionally not enforced
```
pjwerneck: In that case, `SyftDatasetManager._validate_dataset_name` is dead code and should be removed.

packages/syft-datasets/src/syft_datasets/protocolcodecs/v0.py
```diff
@@ -0,0 +1,52 @@
...
+    version = "0"
+    protocol_versions = ["0"]
+
+    def public_dataset_dir(self, ref: DatasetRef) -> Path:
+        return self.config.get_mock_dataset_dir(
+            ref.name, ref.owner, ref.protocol_version
+        )
+
+    def private_dataset_dir(self, ref: DatasetRef) -> Path:
+        return self.config.get_private_dataset_dir(
```
pjwerneck: Both codecs (v0 and v1) are deferring the public and private paths to config. Wasn't the point of the refactoring to have the codec own disk layout?

packages/syft-datasets/src/syft_datasets/dataset_manager.py
```diff
@@ -447,33 +427,43 @@ def delete(
...
+
+        Written in the dataset's on-disk protocol format so peers can read it.
+        """
+        config = self.storage.read_private_config(ref)
         config.data_dir = Path("")
-        return yaml.safe_dump(
-            config.model_dump(mode="json"), indent=2, sort_keys=False
-        ).encode()
+        data = config.disk_dict()
+        if ref.protocol_version == "0":
```
pjwerneck: This should be handled by the codec, not by the manager. The manager should never make internal decisions based on protocol version.

packages/syft-datasets/src/syft_datasets/dataset_storage.py
```diff
@@ -0,0 +1,216 @@
...
+        return max(matches, key=lambda r: int(r.protocol_version))
+
+    # -- model IO ------------------------------------------------------------
+    def read_dataset(self, ref: DatasetRef) -> Dataset:
+        """Load a dataset's dataset.yaml, upgraded to the latest version."""
+        codec = self._codec_for(ref.protocol_version)
+        data = codec.read(codec.metadata_path(ref), "Dataset")
+        dataset = self._upgrade(data, "Dataset")
+        # Name comes from the path (the dataset dir), never a spoofable file.
+        dataset.name = ref.name
```
pjwerneck, line 458: This is the counterpart to the comment on `owner` property in models/dataset/v1.py:69. If we assign the name from `ref.name` here, why not the owner too?

packages/syft-datasets/src/syft_datasets/config.py
```diff
@@ -1,9 +1,27 @@
+import re
+from typing import Optional
+
 from pydantic import BaseModel, Field
 from pathlib import Path
 from .url import SyftBoxURL
+from .migrations.registry import DATASET_PROTOCOL_VERSION
```
pjwerneck: The config.py now has a `protocol_version` parameter everywhere, which defaults to this imported `DATASET_PROTOCOL_VERSION`,the *current* protocol, but the storage defaults to the widest. Is that intended?

pjwerneck (commented):
A few issues found and commented on.

Overall, the high-level design is in line with #9434, but the fact `peer_schemas` is still not properly wired is a bit concerning. In production, everything still resolves to v0. I'd be more comfortable if v1 and the whole protocol mechanics was fully integrated and tested end-to-end before wiring it to the rest of syft-client.

packages/syft-datasets/src/syft_datasets/dataset_manager.py
koenvanderveen: yes that should be handled by the storage layer

packages/syft-datasets/src/syft_datasets/protocolcodecs/v0.py
koenvanderveen: yes I updated this in the latest refactor

packages/syft-datasets/src/syft_datasets/dataset_storage.py
koenvanderveen: done via making owner computed from ref and passing ref

PR #9461, opened 2026-07-21
feat: Further syft-restrict improvements and red-team hardening

pjwerneck:
## Summary

Finishes the loose ends from #9436 and closes the verifier gaps found across three independent red-team passes (kimi-k3, Grok 4.5, Opus 4.8).

### Security

- Import resolution bug: import jax.numpy (no as) bound the root jax to the full path jax.numpy, so jax.numpy.save mis-resolved to jax.numpy.numpy.save and slipped the disallow_functions floor while runtime hit the real function. Now binds the package root correctly. (astutil.py)
- Star imports (from x import *) banned anywhere — they can silently shadow a trusted name and are unreviewable. (verifier.py)
- Safe-local verdict now cleared on tuple-unpack and for-loop rebinds (b = self.safe; b, _ = evil, b; b(x) no longer trusts the stale verdict). (verifier.py)
- Boundary-straddling statements (a call opened on a public line, args in the private region) are now verified as private (default-deny) instead of skipped. (astutil.py, verifier.py)
- Block markers must be on their own line — code sharing a -start/-end marker line was escaping verification; now a MarkerError. (markers.py)

### API / UX:

- run() is markers-only (no obfuscate=/hide=); explicit line ranges moved to internal _run(). (runner.py)
- Two optional strictness flags: allow_local_assignments, allow_base_class_attributes (both default to current behavior). (policy.py, verifier.py)
- __post_init__ restored to the allowed hooks.

### Docs / examples:

- README example switched to default-deny (explicit allow_functions); widened the recommended disallow_functions floor; clarified public-region/import handling in verify.md/blacklist.md.
- Examples renamed: gemma_inference.py is now the canonical markers version; the explicit-ranges variant is gemma_inference_ranges.py (+ matching generate*.py).

## Testing

- New test_policy_flags.py plus regressions in test_bypasses, test_markers, test_disallowed, test_whitelist, test_run.

## Asana task

https://app.asana.com/1/1185126988600652/project/1216249688888494/task/1216630823225716?focus=true


koenvanderveen (approved):
Reviewed: security-focused verifier hardening. Fixes are sound and well-tested (import-binding bypass, boundary straddle, star imports, safe-local clearing, dunder-in-call, own-line markers, policy_id includes new flags). Approving.

koenvanderveen:
awesome!

PR #9485, opened 2026-08-03
Feat/enclave job sandbox

bitsofsteve:
spike


packages/syft-job/src/syft_job/job_runner.py
```diff
@@ -31,6 +32,300 @@ def get_job_timeout_seconds() -> int:
...
+            if not os.path.islink(path):
+                os.chown(path, uid, gid)
+
+    # The datasite is read-only to the job. Ownership stays with root; we only
+    # open traversal and read. Every parent must be traversable or the job
+    # cannot reach the datasets at all -- the folder lives under /root (0700)
+    # by default, since the container runs as root and that is its home.
+    for parent in list(syftbox_folder.parents)[:-1]:
+        try:
+            os.chmod(parent, os.stat(parent).st_mode | 0o011)
```
pjwerneck, line 188: I couldn't find any code restoring this permission change to the original `0700` after the job. Is that intended to be permanent? is the tree ephemeral?

packages/syft-job/src/syft_job/job_runner.py
```diff
@@ -31,6 +32,300 @@ def get_job_timeout_seconds() -> int:
...
+        except OSError:
+            pass
+    for root, dirs, files in os.walk(syftbox_folder):
```
pjwerneck, line 191: This walk is granting read for the whole syftbox folder, not just the approved datasets. Are the enclaves single-tenant?

packages/syft-job/src/syft_job/job_runner.py
```diff
@@ -31,6 +32,300 @@ def get_job_timeout_seconds() -> int:
...
+    """
+    if os.geteuid() != 0:
+        return
+
+    # The job owns its own working tree: venv, code, outputs.
+    os.chown(submission_dir, uid, gid)
+    for root, dirs, files in os.walk(submission_dir):
+        for name in dirs + files:
+            path = os.path.join(root, name)
+            if not os.path.islink(path):
```
pjwerneck, line 179: This walk over `submission_dir` skips symlinks, but  the one below, over `syftbox_folder` does not. Is that correct? Because if enclaves are not ephemeral, someone could use multiple job executions to intentionally or accidentally create symlinks that go where they shouldn't.

pjwerneck (commented):
I'm not familiarized with the constraints of the enclaves and how we use them, but I left three comments on issues that could be relevant.

koenvanderveen:
I think this was eventually not meant to go in right? If it is please let me know and I will reopen

PR #9487, opened 2026-08-07
feat: address migration gaps

pjwerneck:
## Summary

Closes most of the migration gaps that block safe upgrades: private Drive folders are adopted by rename, P2P folders are found regardless of the version in the name, and durable surfaces (checkpoints, rolling state, crypto keys, peers JSON, local caches) carry or refuse versions instead of failing silently.

Adds a per-protocol minimum-supported floor (shipped inert at "0"), stops refusing peers on client semver mismatch, and hardens release/CI (export → publish → tag → bump, client migration tests in CI, numeric version ordering). Job protocol skew delivery is covered end-to-end; dataset transport multi-copy (A1) and a few process/policy items remain open.

> [!NOTE]
>  The base branch is `stephen/syft-client-migrations`, not `dev`. This branch assumes that branch will be merged first.

## Changes

**Done**

- **A1 dataset transport.** `create_dataset` writes one copy for each protocol
  version in the audience. Each copy gets its own Drive collection, and every
  collection reaches the whole audience. A peer that upgrades later therefore
  moves to the newer layout with no action by the owner. The collection name
  holds the version as a `v<n>` infix before the separator. A client that
  predates this change therefore never lists a layout it cannot read. A
  protocol-0 name does not change. The client uploads the private data of a
  copy with that copy. A watcher keeps the newest readable layout for each
  dataset, and warns and skips the rest.
- **A2a private folders.** `_find_or_adopt_versioned_folder` renames the folder
  of the highest earlier version and keeps the data. The personal datasite
  folder, the checkpoints folder, and the rolling-state folder all use it.
- **A2b P2P folders.** `_find_p2p_folder_id` finds a folder whatever version
  its name holds. `_filter_patch_compatible` is gone.
- **A3 floor.** Each package publishes `MIN_SUPPORTED_*_PROTOCOL_VERSION`. The
  value travels as `min_supported_version`, and `negotiate_protocol_version`
  enforces it for jobs and for datasets.
- **A3 peer gate.** An incompatible client version no longer skips a peer
  (`get_peer_compatibility_status`). `force_ignore_protocol_version` is gone.
- **A3 login.** The mismatch prompt keeps local and remote data. A full
  deletion is an explicit second choice. A run with no terminal takes the
  keep-data answer, so a notebook or a scheduled run does not stop. Login
  writes the version file on both sides. A second login therefore shows no
  prompt, and a peer reads a current version.
- **A4.** The dead `ClientVersionMismatchError` and
  `ProtocolVersionMismatchError` are gone. The inert semver `PROTOCOL_VERSION`
  field stays on the wire.
- **A5.** The defaults for an unknown peer are documented, and a forced path
  logs a warning. See `test_unknown_peer_forced_path.py`.
- **B1.** The client refuses a checkpoint or a rolling state from a later
  client, and downloads all events instead. A local load that fails reports the
  failure.
- **B2.** The crypto key file carries `CRYPTO_KEYS_VERSION`. A later version is
  refused, because a private key cannot be rebuilt.
- **B3.** `SYFT_peers.json` carries a version under `_meta`. The client logs an
  unknown peer state and skips it, and the rewrite of the raw map keeps the
  entry.
- **B4.** `PersistedDict` writes `{"version", "entries"}`. An old flat file
  reads as version 0, and a later version resets to empty.
- **C1.** The release job exports the artifacts, publishes, tags, then bumps.
  `--dependents published` pins the published version, and
  `protocol_bump_missing` compares against the newest released protocol.
- **C2.** The client migration tests run in CI, with
  `just test-client-migrations`, and again after a release.
- **C3 cases 1 and 2.** A job that is negotiated down to protocol 0 arrives and
  reads (`test_job_protocol_skew_delivery.py`). A dataset reaches an audience
  of mixed protocol versions (`test_dataset_multicopy_delivery.py`).
- **D1.** `_version_order` compares a version as a number, so `"10"` is after
  `"2"`.
- **D2.** Tests guard the additive-only rule for the `VersionInfo` fields
  (`test_version_info_fields.py`).
- **D3.** The job and the dataset negotiation tests assert the identity of the
  live peer-schema dict.

**Open items**

One decision on A3 clears the first group. One release
cycle clears the second.

**1. Waits on a decision (A3)**

A peer can speak a protocol below the floor of a package. The client can refuse
it, warn about it, or go read-only. That question is open, and three items wait
on the answer. Every floor is `"0"` today, so the mechanism refuses nothing and
no user sees a difference.

- **A3 policy.** The rule itself: refuse, warn, or read-only.
- **A3 residual.** `target_protocol_versions_for_peers` takes
  `min(ours, theirs)` and applies no floor. A1 makes this path live. The
  correct action for a peer below the floor is part of the same decision.
- **C3 case 3.** An upgrade in place with mixed peers has no automated test,
  because the correct result is not defined without the policy. A manual test
  covers this ground now.

**2. Closed by the next release**

The first two items block the release. Do them before it. No code change on
this branch can close any of the three, because each one needs a release that
happened.

- **`syft-migration` is not on PyPI**, and no CD workflow releases it.
  syft-client, syft-job, and syft-dataset all depend on it with no pin. The
  monorepo resolves the dependency from the workspace, so a local install
  works. An install of the next release fails.
- **The one-time version increase.** The branch still names the published
  versions: syft-client 0.1.117, syft-job 0.1.39, syft-dataset 0.1.20. The
  release job publishes the version that is on the branch, and PyPI refuses a
  version that exists.
- **The 0.1.117 fixture and package artifact.** Both hold
  `syft_client_install_source` and `attestation_token`. The published wheel does
  not define these fields, because PyPI received it 12 days before the fields
  arrived. A correct file needs the published wheel as the source.

**3. Needs a rehearsal by a person**

- **C4 and E1.** Nobody followed the migration guidance end to end, and only
  `VersionInfo` has a real migration pair. Both items close when one person
  does a full object version increase on a branch that never merges. That
  person records each point where the tools say nothing. A synthetic V2 must not
  merge: a released V2 freezes its schema and owes a migration forever.

**4. Open by decision**

- **The provisional protocol lock (C1).** It is a workflow contract for the whole team, not a part of the migration system. The cost of the decision is measured and written down.

**5. Permanent, and not work**

- **The semver `PROTOCOL_VERSION` field (A4).** It carries no meaning, and it
  must stay on the wire, because a supported older client parses it.

**6. A rare failure, kept loud on purpose**

- **A2c.** Two compatible private folders reach `_expect_one`, which raises and
  names both folders. An automatic merge can put the data of a user out of
  reach, so the client stops instead. Folder adoption makes this case rare.

**7. Not measured**

- **B4 residual.** A reset of `file_hashes` replays the event stream. The reset
  works. The cost for a large datasite is not measured.

## Testing

- Unit and migration tests for each gap

## Asana task

https://app.asana.com/1/1185126988600652/project/1216249688888494/task/1217005273770094?focus=true


koenvanderveen:
for A3 I think we should refuse by default, but it should be possible for the user to override this. I think in general this should often be the case

koenvanderveen:
# PR 9487 — feat: address migration gaps (WIP)

`OpenMined/syft#9487` · Pedro Werneck · `pjwerneck/migration-gaps` → `dev` · +6180/-1021 across 110 files · OPEN

This PR makes it safe for people to upgrade syft while their peers are still on an older release.
Two layers land at once. First, the objects syft sends to peers each get a version number and a way
to be converted between versions. Second, the gaps around that are closed: a peer on an older
version still receives datasets, jobs and sync messages; a Google Drive folder from an older client
is renamed and reused instead of abandoned; and every long-lived file on disk records its version
and refuses to read a file from a future client rather than misreading it.

> Read time ~8 min. Not read: two frozen data files (`protocol-0.json`, `syft-client-0.1.117.json`,
> ~700 lines of JSON), `uv.lock`, and 50 test files with 180 new test functions (I read the test
> names only). Two enclave demo notebooks are `ruff` reformatting with no behaviour change.
>
> **Two branches merge here.** The PR targets `dev`, but Stephen's branch
> `stephen/syft-client-migrations` has not been merged into `dev` yet, so its commits come along
> with this PR. Items written by Stephen are marked `[Stephen]`; everything unmarked is Pedro's
> 20 commits on top. It all merges as one, so review it as one.

---

- [ ] **1. What is new**

  - [ ] **NEW MODULES — a new `syft/migrations/` folder with two files** `[Stephen]`
    - [x] `syft/migrations/registry.py` — gives the syft client its own protocol version, separate
      from its package version.
      - [x] Defines the constant `SYFT_CLIENT_PROTOCOL_VERSION = "1"` (line 16). Protocol `"0"`
        means "any release up to and including 0.1.117", which had no per-object versions.
      - [x] Defines the constant `MIN_SUPPORTED_SYFT_CLIENT_PROTOCOL_VERSION = "0"` (line 21) —
        the oldest protocol this release still reads.
      - [x] Defines the global `client_registry` (line 25), a `MigrationRegistry` that every
        versioned syft object registers itself into.
      - [x] Defines the function `registry.py: load_as_latest(data, canonical_name)` (line 37).
        It reads a saved object and converts it up to the newest version this client knows.
    - [x] `syft/migrations/history.py` — holds the record of what past releases could read.
      - [x] Defines the function `history.py: register_historic_schemas()` (line 18), which loads
        the frozen JSON files under `syft/migrations/history/` into `client_registry`.

  - [x] **NEW CLASS `VersionInfoV1`** — `syft/sync/version/version_info.py:42` `[Stephen]`
    - [x] Lives in: replaces the old plain `VersionInfo`. It now inherits from `MigratableObject`
      and registers itself into `client_registry`.
    - [x] Does: describes this client in `SYFT_version.json`, the file each peer can read from our
      Google Drive folder.
    - [x] Decision: this file is how two clients first learn about each other, so it must stay
      readable by every older client. The rule is that fields may only be **added**, never removed
      or changed. `tests/migrations/unit/test_version_info_fields.py` enforces it.

  - [ ] **NEW CLASS `VersionInfoV2`** — `syft/sync/version/version_info.py:193` `[Stephen]`
    - [ ] Lives in: subclasses `VersionInfoV1`. The name `VersionInfo` is now an alias for this
      class, so normal callers automatically get V2.
    - [ ] Does: adds one field, `protocol_schemas` — for each of syft, syft-job and syft-dataset,
      which protocol version this client speaks and how far back it reads.
    - [ ] This one field is what all the negotiation later in the PR reads. Two conversion
      functions are registered so a V1 file can be read as V2 and back again.

  - [ ] **NEW CLASSES `FileChangeEventsMessageV1` and `ProposedFileChangesMessageV1`** `[Stephen]`
    - [ ] `syft/sync/events/file_change_event.py:142` (data owner → watchers) and
      `syft/sync/messages/proposed_filechange.py:101` (data scientist → data owner).
    - [ ] Lives in: both inherit from `MigratableObject`. The items they carry
      (`FileChangeEventV1`, `ProposedFileChangeV1`) stay plain models.
    - [ ] Does: these are the two message types syft compresses and uploads to Drive.
    - [ ] Decision: only the outer message gets a version number, and it names its items by their
      exact class (`List[FileChangeEventV1]`) rather than by the alias. One version number for a
      whole message is simpler to reason about than one per item.
    - [ ] Decision: `FileChangeEvent`, `ProposedFileChangesMessage` and `VersionInfo` are kept as
      aliases pointing at the newest class. That is why the rest of the codebase needed no changes.

  - [ ] **NEW CLASS `CollectionLayout`** — `syft/sync/sync/collection_spec.py:14`
    - [ ] Lives in: held as a list on `CollectionSyncSpec.layouts`, oldest first. The list is built
      by `syft_rds/config.py: dataset_layouts()`.
    - [ ] Does: describes one way of storing a collection (a dataset's files) — a short `variant`
      string that goes into the Drive folder name (`""` for the original, then `"v1"`, …), plus the
      local folder that layout is downloaded into.
    - [ ] Decision: the sync engine never looks inside `variant`. That it happens to mean "dataset
      protocol version" is known only to the dataset code, so the sync engine stays generic.

  - [ ] **NEW FUNCTIONS — Google Drive folder names** in
    `syft/sync/connections/drive/gdrive_transport.py`
    - [ ] `gdrive_transport.py: collection_name_query(prefix)` (line 155) — builds the Drive search
      for a collection in **any** layout. It deliberately leaves off the trailing `_`.
    - [ ] `gdrive_transport.py: _sorted_by_version(folders)` (line 249) — sorts folders by the
      client version in their name, lowest first.
    - [ ] `gdrive_transport.py: _partition_by_version(folders, current)` (line 268) — splits
      folders into three lists: same major+minor, older, newer.
    - [ ] `GDriveConnection._find_or_adopt_versioned_folder()` (line 1462) — the method that
      renames an older folder so its data is kept. See Flow 2.6.

  - [ ] **NEW FUNCTIONS — negotiation** in `packages/syft-migration/`
    - [ ] `MigrationRegistry.negotiate_protocol_version(peer_version, peer_min)`
      (`syft_migration/registry.py:232`) — picks the version two clients should both use.
    - [ ] `MigrationRegistry.protocol_bump_missing()` (`syft_migration/registry.py:325`) — true
      when objects changed but nobody raised the protocol version constant. The release script
      calls this and stops the release.
    - [ ] `syft_migration/identity.py: _version_order(version)` (line 26) — converts a version
      string to a number so `"10"` sorts after `"2"` instead of before it.

  - [ ] **NEW CONSTANTS — every long-lived file now records its version**
    - [ ] `CRYPTO_KEYS_VERSION` in `syft/sync/peers/peer_store.py` (the encryption key file).
    - [ ] `PERSISTED_DICT_VERSION` in `syft/sync/sync/caches/persisted_dict.py` (the cache files).
    - [ ] `PEERS_META_KEY = "_meta"` and `SYFT_PEERS_VERSION = 1` in `gdrive_transport.py`
      (lines 104, 107) — the version of `SYFT_peers.json`.
    - [ ] `MIN_SUPPORTED_JOB_PROTOCOL_VERSION` and `MIN_SUPPORTED_DATASET_PROTOCOL_VERSION` in the
      syft-job and syft-dataset registry files. All of these ship at `"0"`, which rejects nobody.

  - [ ] **CHANGED — four classes took on a new job**
    - [ ] `CollectionSyncSpec` (`syft/sync/sync/collection_spec.py`) — was a plain pair of (folder
      name prefix, local folder). Now it owns a list of layouts and builds the Drive folder name
      prefix for each one (`wire_prefix()`, `layout_for()`, `rank_of()`).
    - [ ] `MigrationRegistry` (`syft-migration`) — was a lookup table of object versions. Now it
      also decides which version two clients use together, and stores the oldest version this
      release accepts.
    - [ ] `PeerManager` (`syft/sync/version/peer_manager.py`) — now the one place that knows what
      each peer speaks. `PeerManager.live_peer_schemas(protocol_name)` returns a dictionary that
      `PeerManager` keeps updating as peer version files load, so everyone holding it stays current.
    - [ ] `ConnectionRouter` (`syft/sync/connections/connection_router.py`) — was only routing
      messages to a transport. Now it also converts each outgoing message down to whatever the
      recipient can read.

  - [ ] **DELETED**
    - [ ] `syft/sync/connections/collection_prefixes.py` — existed only for
      `GDriveConnection.delete_unversioned_state()`, which is also gone (see Flow 2.7).
    - [ ] `ClientVersionMismatchError` and `ProtocolVersionMismatchError` — nothing raised them any
      more once a client version difference stopped blocking a peer.
    - [ ] `SyftboxManager.read_local_version()` / `write_local_version()` — replaced by
      `PeerManager.write_own_version()`, which writes the local **and** the remote copy.
    - [ ] `gdrive_transport.py: _filter_patch_compatible()` — replaced by `_partition_by_version()`.
    - [ ] The `just bump-and-publish` command — split into `just publish` and `just bump` so the
      release can put a step between them.

- [ ] **2. Flows**

  - [ ] **2.1 Objects that know their own version** *(new)* `[Stephen]`
    - [ ] **Registering, when syft is imported.** We previously registered object versions as a
      side effect of whichever code path happened to run first. Now `syft/__init__.py` imports the
      three versioned model files by name and then calls `history.py: register_historic_schemas()`,
      which loads the frozen JSON files into `client_registry`.
      - [ ] Decision: it passes `raise_for_unknown_objects=True`, so a history file naming an
        object this release cannot load fails immediately at import, not later during a migration
        when a user is mid-sync.
    - [ ] **Advertising, when a client writes its own version file.** We previously wrote only
      package version numbers into `SYFT_version.json`. Now `VersionInfoV2.current()` calls the
      module-level helper `version_info.py: _gather_protocol_schemas()`, which calls
      `version_info.py: _slim_schema_of(registry)` once for the syft registry, once for syft-job
      and once for syft-dataset, and puts the three results in the `protocol_schemas` field.
      - [ ] `version_info.py: _slim_schema_of()` deliberately leaves out the full JSON schema of
        every object. Negotiation only needs the version list, and the full schemas are already
        frozen in the release history files. This keeps the file peers download small.
      - [ ] If syft-job or syft-dataset is not installed, the import fails quietly and we simply do
        not advertise that protocol. Peers then treat us as unknown for it.
    - [ ] **Reading, when a message or version file arrives.** We previously called
      `model_validate_json` and got whatever the current code expected. Now
      `FileChangeEventsMessage.from_compressed_data()`, `ProposedFileChangesMessage.from_compressed_data()`
      and `VersionInfo.from_json()` all call `registry.py: load_as_latest(data, canonical_name)`,
      which fills in a default name and version for files written by clients up to 0.1.117 (they
      predate those fields), loads the right class, then converts it up to the newest version.

  - [ ] **2.2 Publishing a dataset** *(changed → A1)*
    - [ ] When a data owner calls `SyftRDSClient.create_dataset()`, we previously wrote one copy
      locally, uploaded one Drive folder named `{prefix}_{tag}_{hash}` for the mock data and one for
      the private data, then shared the mock folder. Now we write one copy for **each** protocol
      version anyone in the audience reads, and upload each copy separately.
      - [ ] Now `SyftDatasetManager.create_all()` returns `{protocol_version: Dataset}` instead of a
        single dataset. `SyftDatasetManager.create()` is now a thin wrapper that returns the newest.
      - [ ] Now `SyftRDSClient.create_dataset()` loops over those copies. Each copy gets its own
        mock Drive folder **and its own private Drive folder**, named
        `{prefix}{variant}_{tag}_{hash}`.
      - [ ] Now `CollectionFolder.from_name()` parses that name with a regular expression instead of
        splitting on the last `_`. A protocol-0 folder name has an empty variant, so it is
        character-for-character the name older clients already write.
      - [ ] Decision: an older client searches Drive for `name contains '{prefix}_'`, and the newer
        names put the variant exactly where that `_` used to be. So an older client never even lists
        a layout it could not read — no error handling needed on the old side.
      - [ ] Decision: write every copy up front rather than converting on read. The cost is several
        copies of the data on Drive. The gain is that when a peer upgrades later, the owner does not
        have to do anything.
      - [ ] Now `SyftRDSClient._cleanup_failed_dataset_creation()` takes lists of folder ids rather
        than single ids, so a failure halfway through removes every folder already uploaded.

  - [ ] **2.3 Sharing an existing dataset with a new peer** *(changed → A1)*
    - [ ] When a data owner calls `SyftRDSClient.share_dataset()`, we previously recomputed the
      content hash from local files and shared one folder. Now we look up every uploaded copy and
      share all of them, and create a missing copy first if the new peer cannot read any of them.
      - [ ] Now `SyftRDSClient._mock_collections_for(tag)` lists every uploaded layout of that
        dataset. The hash comes from the listing, so nothing is recomputed.
      - [ ] Now `SyftRDSClient._ensure_dataset_layouts_for()` asks
        `DatasetStorage.target_protocol_versions_for_peers()` what the new audience reads, and calls
        `SyftRDSClient._materialize_dataset_copy()` for any version no existing copy covers. A peer
        can read every layout at or below its own version, so a copy is only missing when nothing
        sits at or below what the peer speaks.
      - [ ] Now `SyftRDSClient._materialize_dataset_copy()` either converts the dataset to that
        version or re-reads a copy already on disk, re-applies permissions, and uploads it
        **without sharing it**. Sharing happens once at the end for all layouts together.
      - [ ] Decision: create the missing copy when sharing, not when the dataset was first made.
        Sharing is when the audience changes, and giving a new peer access to a folder its client
        never lists would silently do nothing.
    - [ ] When a data owner shares private data with an enclave via
      `SyftRDSClient.share_private_dataset()`, we previously always sent the newest copy. Now
      `SyftRDSClient._private_share_protocol_version()` picks the newest local copy at or below the
      version negotiated with that enclave, and only creates a new copy if none qualifies.

  - [ ] **2.4 A watcher downloading collections from a peer** *(changed → A1)*
    - [ ] When a client syncs collections from a peer, we previously listed the peer's collections
      and downloaded all of them into one local folder. Now an owner may publish the same dataset
      several times, so we pick one copy per dataset.
      - [ ] Now `DataSiteWatcherCache._select_collections_to_sync()` keeps, for each
        (owner, dataset), the copy with the highest `CollectionSyncSpec.rank_of(variant)` this
        client can read, and logs a warning naming each copy it skipped.
      - [ ] Now `DataSiteWatcherCache._cleanup_stale_collections()` also deletes the local copy when
        a **newer readable** layout takes over, which would otherwise leave the same dataset on disk
        twice.
      - [ ] Decision: if the owner still publishes the dataset but in no layout we can read, we
        **keep** the local copy rather than deleting it. It is the last version this client was able
        to read, and deleting it would take working data away from the user.
      - [ ] Now `DataSiteOwnerEventCache` stores collection hashes under the key `"<variant>/tag"`
        rather than `"tag"`, and scans one folder per layout, so the copies do not overwrite each
        other's hash.

  - [ ] **2.5 Converting an outgoing message down to what the recipient reads** *(new → C3)*
    - [ ] When syft sends a message to a peer, we previously always sent it in our own format and
      relied on the receiving side coping. Now we convert it down first.
      - [ ] At startup, `SyftboxManager._set_peer_schemas()` calls
        `ConnectionRouter.set_peer_schemas()` on each router, passing the dictionary from
        `PeerManager.live_peer_schemas("syft")`. `PeerManager._update_peer_schemas()` keeps updating
        that same dictionary as peer version files load, so routers never hold stale information.
      - [ ] On send, `ConnectionRouter.watcher_send_proposed_file_changes_message()` and
        `ConnectionRouter.owner_write_event_messages_to_outbox()` both call
        `ConnectionRouter._downgrade_for_peer()`, which calls
        `MigrationRegistry.negotiate_protocol_version()` and then
        `MigrationService.migrate_to_schema()`.
      - [ ] Decision: when the peer already speaks the agreed version, `_downgrade_for_peer()` reuses
        the short schema the peer advertised instead of computing our own full one. Building every
        object's JSON schema on every send would be wasted work.
      - [ ] Decision: converting produces a **new** object, so the caller's message is never
        modified. One message can therefore be sent to many recipients at different versions.
      - [ ] For a peer we know nothing about, we send at our current version **and log a warning**.
        Previously nothing was logged.

  - [ ] **2.6 Finding our own Google Drive folder after an upgrade** *(changed → A2)*
    - [ ] When a client looks for its own private folders on Drive (the personal datasite folder,
      the checkpoints folder, the rolling-state folder), the client version is part of the folder
      name. We previously called `gdrive_transport.py: _filter_patch_compatible()`, which kept only
      folders with the same major and minor version, then `GDriveConnection._expect_one()`. After a
      minor upgrade that matched nothing, so the client made a **second** folder and the user's data
      was left behind in the first.
      - [ ] Now `gdrive_transport.py: _partition_by_version()` splits the candidates into same
        major+minor, older, and newer.
      - [ ] Now `GDriveConnection._find_or_adopt_versioned_folder()` **renames** the highest older
        folder to the new name, so the data comes along.
      - [ ] If only a folder from a **newer** client exists, it raises and tells the user which syft
        version to install. We cannot safely read data a newer client wrote.
      - [ ] Decision: if two same-version folders exist, `GDriveConnection._expect_one()` still
        raises and names both. Merging them automatically could put a user's data somewhere they
        cannot find it, so the client stops and asks.
    - [ ] When a client looks for a **shared** folder between two peers, the rule is the opposite.
      `GDriveConnection._find_p2p_folder_id()` now ignores the version in the name entirely
      (`_sorted_by_version()`, take the newest) instead of filtering by it.
      - [ ] Decision: the peer builds that folder name from **its own** client version, so filtering
        by ours hid the folder the peer is actually writing into. Renaming is also not allowed here,
        because the peer looks the folder up by name.

  - [ ] **2.7 Logging in when the versions do not match** *(changed → A3)*
    - [ ] When login finds a version mismatch, option 1 previously deleted the local SyftBox folder
      and also wiped unversioned state on Drive. Now option 1 keeps both, and explains that folder
      renaming (Flow 2.6) and cache rebuilding repair things on the next sync. Deleting everything
      is now option 2, chosen explicitly.
      - [ ] Now `login_utils.py: _prompt_mismatch()` returns `"1"` when `sys.stdin.isatty()` is
        false. Decision: with no terminal there is nobody to answer, and option 1 changes nothing,
        so a notebook or scheduled run continues instead of hanging on a prompt.
    - [ ] When login finishes, `login.py: _init_client_login()` previously called
      `SyftboxManager.write_local_version()`, which wrote only the local file. Now it calls
      `PeerManager.write_own_version()`, which writes the local **and** the remote file.
      - [ ] This is the fix for two bugs at once: the stale remote file made the mismatch prompt
        appear at every single login, and peers reading that file picked the wrong protocol for us.

  - [ ] **2.8 Releasing a package** *(changed → C1)*
    - [ ] When CD releases a package, we previously bumped the version, then published, then
      tagged — so the tag and the pinned dependency versions named something PyPI did not have yet.
      Now the order is: read the version already on the branch → export and commit the frozen
      history files → publish → tag → bump for the next release → push. All eight CD workflows
      changed.
      - [ ] Now `scripts/bump_version.py` takes `--dependents published`, so dependent packages are
        pinned to the version that was just published rather than the one we are about to work on.
      - [ ] Now `scripts/export_release_artifact.py` calls
        `MigrationRegistry.protocol_bump_missing()` and stops the release if objects changed but the
        protocol version constant did not.

- [ ] **3. Changes**

  - [ ] **F — Versioned objects (Stephen's branch)** `[Stephen]`
    - [ ] **F1 A protocol version for the client** → Flow 2.1 — When the client talks to a peer, we
      previously had only the package version number to go on. Now `syft/migrations/registry.py`
      gives syft its own protocol version, where `"0"` means every release up to 0.1.117.
    - [ ] **F2 Messages carry their version** → Flow 2.1 — When we read a saved message, we
      previously parsed it straight into the current class. Now the two message types and
      `VersionInfo` carry a name and version, and reads go through
      `registry.py: load_as_latest()`, which converts old files up.
    - [ ] **F3 `VersionInfo` V2** → Flow 2.1 — When a peer reads our version file, it previously
      learned only our package versions. Now it also learns which protocol version we speak for
      syft, syft-job and syft-dataset. Everything in A1, A3 and C3 negotiates against this field.
    - [ ] **F4 Frozen release history** → Flow 2.8 — Each release now freezes what it could read
      into `syft/migrations/history/`. Protocol 0 and release 0.1.117 came before this existed, so
      those two files are written by hand as if that release had produced them.
    - [ ] **F5 Registration is explicit** → Flow 2.1 — We previously relied on imports happening in
      the right order. Now `syft/__init__.py` names the three model files it imports and then
      registers the history.

  - [ ] **A — Talking to peers on other versions**
    - [ ] **A1 One dataset copy per version** → Flows 2.2, 2.3, 2.4 — When we publish a dataset, we
      previously uploaded one copy in our own format. Now we upload one per protocol version in the
      audience, each in its own Drive folder, and share all of them with everyone.
    - [ ] **A2a Reusing our older Drive folders** → Flow 2.6 — When we look for our own folder after
      an upgrade, we previously needed an exact major+minor match and created a new folder when
      there was none. Now we rename the newest older folder and keep the data.
    - [ ] **A2b Finding a peer's shared folder** → Flow 2.6 — When we look for a shared folder, we
      previously filtered it by our own version, which hid the folder the peer writes into. Now we
      ignore the version in the name.
    - [ ] **A2c Two matching folders** → Flow 2.6 — Unchanged on purpose: this still raises and
      names both folders, because merging them automatically could lose a user's data.
    - [ ] **A3 An oldest-supported version** → Flow 2.5 — When two clients negotiate, we previously
      just took the lower of the two versions. Now each package also publishes the oldest version it
      still reads, that travels in the version file, and
      `MigrationRegistry.negotiate_protocol_version()` enforces both sides' limits. Every limit
      ships at `"0"`, so today this rejects nobody.
    - [ ] **A3 Peers are no longer skipped for a version difference** — When a peer runs a different
      client version, we previously marked it incompatible and skipped it entirely. Now we keep it
      and log the difference, and each protocol decides for itself using its oldest-supported
      version. The `force_ignore_protocol_version` setting is gone. A peer whose version we do not
      know at all is still skipped.
    - [ ] **A3 Login** → Flow 2.7.
    - [ ] **A4 Dead code removed** — `ClientVersionMismatchError` and `ProtocolVersionMismatchError`
      are gone. The old `PROTOCOL_VERSION` field does nothing but stays in the file, because
      supported older clients still try to read it.
    - [ ] **A5 Assumptions are logged** — When a caller passes `raise_on_unknown=False`, we
      previously assumed the peer runs the current protocol in silence. Now `JobStorage` and
      `DatasetStorage` log a warning saying which version we are writing and that an older peer
      will not see it.

  - [ ] **B — Every long-lived file records its version**
    - [ ] **B1 Checkpoints and rolling state** — When we load a checkpoint, we previously parsed
      whatever was there. Now `checkpoint.py: _check_version()` and
      `rolling_state.py: raise_for_later_version()` refuse a file from a newer client. Every caller
      falls back to downloading all events, so refusing costs one slow startup and nothing else.
      A failed local load now prints the reason instead of silently passing.
    - [ ] **B2 Encryption keys** — `crypto_keys.json` now records `CRYPTO_KEYS_VERSION`, and a newer
      version is refused outright. Decision: unlike a cache, a private key cannot be rebuilt, so
      guessing at the format is not an option.
    - [ ] **B3 `SYFT_peers.json`** — the version goes under a reserved `_meta` key. Decision: the
      file is a flat map of email → entry, so a version at the top level would look like a peer
      email to an older client. A peer whose state we do not recognise is logged and skipped, and
      rewriting the file keeps that entry, so the peer comes back after an upgrade.
    - [ ] **B4 Cache files** — `PersistedDict` now writes `{"version", "entries"}`. An old flat file
      is read as version 0, and a file from a newer client starts empty so the cache rebuilds.
      *Left open: nobody has measured how long replaying the event stream takes on a large datasite
      after such a reset.*

  - [ ] **C — Release and CI**
    - [ ] **C1 Release order** → Flow 2.8.
    - [ ] **C2 Migration tests run in CI** — `just test-client-migrations` runs `tests/migrations`
      in `unit-tests.yml`, and again in `post-release-tests.yml` against the wheel that was just
      published. `scripts/generate_release_fixture.py` captures a release's real message files so
      future releases can prove they still read them.
    - [ ] **C3 Version-skew tests** — `test_job_protocol_skew_delivery.py` (a job negotiated down to
      protocol 0 arrives and is readable), `test_dataset_multicopy_delivery.py` (an audience on
      mixed versions), `test_message_protocol_downgrade.py`. The third case — upgrading in place
      with mixed peers — has no automated test, because the right outcome is undefined until the A3
      policy question below is answered.

  - [ ] **D — Version ordering**
    - [ ] **D1 Compare versions as numbers** — When we compared two versions, we previously compared
      them as text, so `"10"` sorted before `"2"`. Now `syft_migration/identity.py: _version_order()`
      converts to a number everywhere, and a non-numeric object version is rejected the moment the
      class is defined rather than at runtime.
    - [ ] **D2/D3 Tests that guard the rules** — `test_version_info_fields.py` enforces that
      `VersionInfo` only ever gains fields. The negotiation tests assert the peer-schema dictionary
      is the same object everyone shares, not a copy that would go stale.

  - [ ] **E — Cleanup, no behaviour change intended**
    - [ ] **E1** Bare `except Exception` narrowed to `ValueError` / `MigrationError` throughout
      `gdrive_transport.py`, with messages that say what failed instead of printing a raw error.
    - [ ] **E2** `GDriveConnection.delete_file_by_id()` lost its `verbose` argument and now always
      warns on failure.
    - [ ] **E3** `collection_prefixes.py` deleted — it existed only for the login wipe removed in
      Flow 2.7.
    - [ ] **E4** Two enclave demo notebooks reformatted by `ruff`. Nothing to review.

- [ ] **4. Open questions**

  - [ ] **The main design question is still open.** What should happen when a peer speaks a version
    below our oldest-supported one — refuse it, warn, or go read-only? Three items wait on this
    answer, including the fact that
    `DatasetStorage.target_protocol_versions_for_peers()` takes the lower of the two versions and
    applies **no** lower limit at all, which A1 turns into a live code path. Nothing is affected
    today, because every limit ships at `"0"`.
  - [ ] **The release is blocked on version numbers, and the PR description's numbers are out of
    date.** It names syft-client 0.1.117, syft-job 0.1.39, syft-dataset 0.1.20. After the `dev`
    merge the branch actually carries syft **0.10.0**, syft-job **0.1.40**, syft-dataset **0.1.21**
    and syft-migration **0.1.1** — and I checked PyPI: all four are already published. Under the new
    publish-then-bump order (Flow 2.8) the release job publishes whatever is on the branch, so PyPI
    will reject every one. Someone still needs to raise them by hand once.
  - [ ] **The "`syft-migration` is not on PyPI" item looks already fixed.** It is published (0.1.0
    and 0.1.1), `cd-syft-migration.yml` exists, and this PR's root `pyproject.toml` pins
    `syft-migration==0.1.1`. Worth confirming with Pedro that this can be crossed off.
  - [ ] **The 0.1.117 history file was written by hand** and contains two fields
    (`syft_client_install_source`, `attestation_token`) that the actually-published wheel does not
    define. A correct file has to be generated from that wheel.
  - [ ] **Nobody has done a full object version increase end to end.** Only `VersionInfo` has a real
    pair of conversion functions, so we do not yet know where the tooling falls short.

koenvanderveen:
**Re: `GDriveConnection._find_or_adopt_versioned_folder()`** (Section 1 — Google Drive folder names)

> Q: how does this work?

Step by step, reading `gdrive_transport.py:1462`:

1. The caller passes every folder whose name matched a Drive search, as `(id, name)` pairs, plus `current_name` — the name this client version wants.
2. `_partition_by_version()` reads the version out of each name and sorts them into three lists: same major+minor, older, newer.
3. **Same major+minor** → hand it to `GDriveConnection._expect_one()`, which returns the id, or raises if there are two.
4. **Only newer** → raise `RuntimeError` naming the syft version to install.
5. **Nothing at all** → return `None`, and the caller creates a fresh folder.
6. **Older** → take the highest one and call Drive `files().update(fileId, {"name": current_name})`. That is a rename and nothing else: same file id, same contents, same sharing permissions. Then it prints `Adopted the folder of an earlier version: <old> -> <new>`.

Two things worth knowing: only **one** older folder is adopted, and if there were several the rest are left on Drive with a warning listing them — their data stays out of reach. And the adoption happens on the **first lookup after upgrading**, not at login.

koenvanderveen:
**Re: `gdrive_transport.py: _sorted_by_version()`** (Section 1 — Google Drive folder names)

> Shouldn't we know which version the peer uses and use that based on their `version.json` instead of picking the latest?

For context on what the code does today: `GDriveConnection._find_p2p_folder_id()` takes the last entry after sorting, and prints a warning when there is more than one candidate. It never consults the peer's version file.

One thing to weigh: the version in that folder name is the peer's **client** version (semver), whereas `SYFT_version.json` carries **protocol** versions. So matching on the version file would mean comparing a different scale than the one actually written into the folder name.

koenvanderveen:
**Re: `MigrationRegistry.negotiate_protocol_version()`** (Section 1 — deciding which version two clients use)

> This sounds like we are actively communicating with the peer, which is not what's happening. Something like `infer_protocol_version_to_use` makes more sense to me.

Agreed on the substance: there is no round trip. The function is pure and local — it reads `peer_version` and `peer_min` out of the peer's already-downloaded `SYFT_version.json` and returns a string.

The same naming applies to `JobStorage.negotiated_protocol_version_for_peer()` and `DatasetStorage.negotiated_protocol_version_for_peer()`.

koenvanderveen:
**Re: the new version constants** (Section 1 — every long-lived file now records its version)

> Q: how are `CRYPTO_KEYS_VERSION` and `PERSISTED_DICT_VERSION` actually used? And by what?

They behave very differently, which is the interesting part:

- **`CRYPTO_KEYS_VERSION` — a hard stop.** `PeerStore.save_keys()` writes it into `<syftbox>/<email>/private/crypto_keys.json`. `PeerStore.load_keys()` reads it, and is reached from `PeerStore.for_datasite()` at client startup whenever encryption is on. A file from a newer client raises `ValueError` and login fails, telling the user to install a newer syft. Rationale in the code: a private key cannot be regenerated, so a wrong read loses it.
- **`PERSISTED_DICT_VERSION` — a silent reset.** `PersistedDict._write_to_file()` writes it; `PersistedDict._entries_of()` reads it. A newer version returns `{}`, so the cache simply starts empty and rebuilds. Today it has one real user: `DataSiteOwnerEventCache.file_hashes`, the file-hash cache under `<syftbox>/<CACHE_DIR>/`.

So one blocks login and one quietly rebuilds. Both reasonable, but worth confirming that is the intended split — and B4 notes the rebuild cost on a large datasite is unmeasured.

koenvanderveen:
**Re: Flow 2.2 — publishing a dataset**

> W.r.t. datasets I would like to balance writing multiple versions (heavy) against missing a version. It's quite easy to write an additional version later on version mismatch, and it should be possible for the user to modify this easily from the top-level API.

koenvanderveen:
**Re: Flow 2.6 — "renames the highest older folder to the new name, so the data comes along"**

> Q: how do we know this data is compatible?

We don't check. The rename is unconditional: nothing inspects the folder's contents, and there is no per-file version gate at adoption time.

Compatibility is handled afterwards, file by file, and only in one direction — each reader refuses a **newer** version and recovers:

- checkpoints and rolling state fall back to downloading all events (B1)
- `PersistedDict` resets to empty (B4)
- the crypto key file refuses outright (B2)

So the assumption being made is "an **older** client's files are always readable by a newer client", and nothing enforces it — it rests on the additive-only schema rules. Worth deciding whether adoption should verify anything before renaming, or whether the per-file gates are considered sufficient.

koenvanderveen:
**Re: Flow 2.7 — login with a version mismatch**

> If the answer is "keep everything and repair on the next sync", we should probably just call `sync()` right there, and print that we are doing it.

For context: `login.py: _init_client_login()` does call `client.sync()` a few steps later, but only when `sync=True`, and the message the user sees says "on the next sync" without saying one is about to run.

So this is partly a wording fix, and partly a real gap for the `sync=False` path.

koenvanderveen:
**Re: Flow 2.8 — `scripts/bump_version.py --dependents published`**
I wonder if we need the old version (bumped), we can probably remove that

koenvanderveen:
**Re: A2a — reusing our older Drive folders**

> Q: I would like to understand how this works.

Full step-by-step walkthrough is in the separate comment on `GDriveConnection._find_or_adopt_versioned_folder()`.

Short version: the candidate folders are split into same-version / older / newer; an older one is adopted with a Drive rename (same file id, contents and permissions untouched); a newer one raises; none means the caller creates a fresh folder.

The related question of whether the adopted data is actually readable is in the separate comment on Flow 2.6.

koenvanderveen:
if we do _find_or_adopt_versioned_folder and we upgrade, how do we know that the renamed folder is in the right format?

koenvanderveen:
we probably need a migration test with encryption on as well

syft/sync/version/peer_manager.py
```diff
@@ -360,14 +426,26 @@ def get_peer_compatibility_status(
                 **common,
             )
 
-        # UNKNOWN or INCOMPATIBLE
-        if status == CompatibilityStatus.UNKNOWN:
-            detail = (
-                "version information not available "
-                "(if you are unsure if it is up to date, call client.sync())"
+        if status == CompatibilityStatus.INCOMPATIBLE:
+            # A different client version does not refuse the peer. What each side
```
koenvanderveen, line 430: for A3 I think we should refuse by default, but it should be possible for the user to override this. I think in general this should often be the case

syft/sync/connections/drive/gdrive_transport.py
```diff
@@ -207,33 +241,70 @@ def _extract_version_from_name(name: str) -> str | None:
     return None
 
 
-def _filter_patch_compatible(
+# A folder id and name, with the version from the name. The version fields come
+# first, so the default sort puts these in version order.
+_VersionedFolder = tuple[int, int, int, str, str]
+
+
+def _sorted_by_version(folders: list[tuple[str, str]]) -> list[tuple[str, str]]:
```
koenvanderveen, line 249: I wonder if this is the way we should do it. We should be able to know a peers version from their version.json, we dont need multiple sources of truth, we should probably be able to decide based on that instead of this information. For instance, what happens if a peer downgrades?

packages/syft-migration/src/syft_migration/registry.py
```diff
@@ -217,6 +229,31 @@ def compute_protocol_schema(self) -> ProtocolSchema:
             },
         )
 
+    def negotiate_protocol_version(
```
koenvanderveen, line 232: This sounds like we are actively communicating with the peer, which is not what's happening. Something like `infer_protocol_version_to_use` makes more sense to me.

syft/sync/peers/peer_store.py
```diff
@@ -20,6 +20,11 @@
 PRIVATE_DIR_NAME = "private"
 CRYPTO_KEYS_FILENAME = "crypto_keys.json"
 
+# Format of the crypto key file. Raise it when the layout of the file changes,
+# and add a read path for every earlier version. A file with no version was
+# written before the field, and is version 0.
+CRYPTO_KEYS_VERSION = 1
```
koenvanderveen, line 26: wrt crypto keys version and persisted dict, I think the assumption is that they are backwards compatible. Until we have migrations for them we should probably have a test that we can read old versions?

packages/syft-datasets/src/syft_datasets/dataset_manager.py
```diff
@@ -64,6 +82,38 @@ def create(
...
+            readme_path=readme_path,
+            location=location,
+            tags=tags,
+            users=users,
+            protocol_versions=protocol_versions,
+        )
+        # Return the newest protocol version written (richest layout).
+        return created[max(created, key=int)]
+
+    def create_all(
```
koenvanderveen, line 99: W.r.t. datasets I would like to balance writing multiple versions (heavy) against missing a version. It's quite easy to write an additional version later on version mismatch, and it should be possible for the user to modify this easily from the top-level API.

syft/sync/connections/drive/gdrive_transport.py
```diff
@@ -1431,6 +1459,54 @@ def _expect_one(self, folders: list[tuple[str, str]]) -> str | None:
             f"folder(s) on Drive (keeping the one with your data) and retry."
         )
 
+    def _find_or_adopt_versioned_folder(
```
koenvanderveen, line 1462: how do we know this data is compatible?

syft/sync/login_utils.py
```diff
@@ -25,36 +25,31 @@ def _read_remote_version(
...
     token_path: Optional[Path],
     local_syftbox_path: Path,
     local_version: Optional[VersionInfo],
     remote_version: Optional[VersionInfo],
 ) -> None:
-    """Handle version mismatch with unified prompt."""
+    """Handle a client major/minor mismatch at login.
+
+    The default is to keep local and remote data. Folder adopt, refuse-later
+    checks, and cache reset repair state on the next sync. A full wipe is an
```
koenvanderveen: **Re: Flow 2.7 — login with a version mismatch**
If the answer is "keep everything and repair on the next sync", we should probably just call `sync()` right there, and print that we are doing it.

scripts/bump_version.py
```diff
@@ -92,11 +107,24 @@ def main() -> None:
     )
     parser.add_argument("package_name", help="Package name (e.g. syft-perms)")
     parser.add_argument("bump_type", choices=["major", "minor", "patch"])
+    parser.add_argument(
+        "--dependents",
+        choices=["bumped", "published"],
+        default="bumped",
```
koenvanderveen, line 113: **Re: Flow 2.8 — `scripts/bump_version.py --dependents published`**
I wonder if we need the old version (bumped), we can probably remove that

syft/sync/connections/drive/gdrive_transport.py
```diff
@@ -1431,6 +1459,54 @@ def _expect_one(self, folders: list[tuple[str, str]]) -> str | None:
...
+                f"cannot read that data. Install syft {latest} or later."
+            )
+        if not older:
+            return None
+
+        folder_id, name = older[-1]
+        execute_with_retries(
+            self.drive_service.files().update(
+                fileId=folder_id, body={"name": current_name}
+            )
```
koenvanderveen, line 1499: if we do _find_or_adopt_versioned_folder and we upgrade, how do we know that the renamed folder is in the right format?

tests/migrations/p2p/test_message_protocol_downgrade.py
```diff
@@ -0,0 +1,198 @@
...
+
+def _raw_outbox_versions(ds_manager, do_email: str) -> list[str]:
+    """The version fields of the DO's outbox blobs for us, unparsed."""
+    raw_list = ds_manager._connection_router.connections[
+        0
+    ].watcher_download_raw_events_from_outbox(do_email, None)
+    return [json.loads(uncompress_data(raw))["version"] for raw in raw_list]
+
+
+def test_a_v2_proposal_for_a_protocol0_peer_downgrades_on_the_wire(
```
koenvanderveen, line 117: we probably need a migration test with encryption on as well

syft/sync/version/peer_manager.py
koenvanderveen: I also wonder if this is a good idea. I think we should be more clear about what compatibility status means. If two peers have a incompatible syft-client version, we should skip syncing for them

pjwerneck: You're right. My previous message was wrong. The override does not cover this case, and it does on dev, so I restored the dev behaviour here. `INCOMPATIBLE` is back on refuse by default, and the protocol floor stays.

I still think the more permissive control has merit, but I'll leave that for another PR.

syft/sync/connections/drive/gdrive_transport.py
pjwerneck: Peer capability comes solely from `version.json`, but it lives inside the folder we are trying to find, so it's a chicken-and-egg situation. The folder name is a recognizable address. It's not a source of truth for the negotiation.

Downgrade cases are handled depending on the direction and other details. I will confirm in the code and add another comment later.

packages/syft-migration/src/syft_migration/registry.py
pjwerneck: Agree. Will fix.

syft/sync/peers/peer_store.py
pjwerneck: The tests exist in:

- tests/unit/test_crypto_keys_version.py
- tests/unit/test_persisted_dict_version.py

syft/sync/connections/drive/gdrive_transport.py
pjwerneck: Adoption renames a folder and does nothing else. Compatibility is enforced per file, downstream. A checkpoint from a newer client is refused, a cache with unknown version starts empty, a versioned object loads through the registry with the frozen protocol history and the drift check, and so on.

tests/migrations/p2p/test_message_protocol_downgrade.py
pjwerneck: Agree. Will add it.

syft/sync/login_utils.py
pjwerneck: The `sync()` call  already happens shortly after. The loose end is login called with `login(sync=False)`. I'll add a message.

packages/syft-migration/src/syft_migration/registry.py
pjwerneck: Done. Renamed to `infer_protocol_version_to_use`, as you suggested.

I previously asked about keeping it consistent with `negotiated_protocol_version_for_peer` which is already in dev, but that's a consistency nitpick that only adds more friction to this PR. Ignore that.

syft/sync/connections/drive/gdrive_transport.py
pjwerneck: Like above, the rename changes the folder name and nothing else. Every file inside carries its own version and is checked when it is read, so the name never decides the format.

Added two tests for this: one adopts a checkpoints folder holding a version-0 checkpoint and reads it back, the other shows a later checkpoint is still refused inside an adopted folder.

scripts/bump_version.py
pjwerneck: No, we don't. Fixed. Removed --dependents from the script and the workflows. The script now always pins the published version.

pjwerneck:
@koenvanderveen  This PR keeps costing us more time in reviewing and reiterating than the remaining items are worth. At this point, I suggest we ticket the remaining items, merge this to dev, and deal with them in small, dedicated PRs.

koenvanderveen (approved):

PR #9517, opened 2026-09-17
fix: ensure case-insensitive matching for permission file names

pjwerneck:
## Summary

A DS with plain write access to a directory could take admin rights over it on macOS or Windows. The ADMIN gate compared the path suffix with case, so a write to `SYFT.PUB.YAML` read as an ordinary file; on a case-insensitive volume that write replaces the real `syft.pub.yaml`, which the loader then reads as rules.

## Changes
- `syft_permissions/engine/service.py`: the gate compares the last path part under   `casefold()` instead of `str.endswith`. A mis-cased name is raised to ADMIN like the exact name, and `notsyft.pub.yaml` no longer needs ADMIN for no reason.
- `syft/sync/sync/datasite_owner_syncer.py`: `_split_events` uses the same compare, so a  permission write cannot route as a data event.

## Testing

- Two tests in `packages/syft-permissions/tests/test_no_permission.py` and one
  event-split test in `tests/unit/test_permission_file_case.py`. They cover `SYFT.PUB.YAML`,
  `Syft.Pub.Yaml`, nested paths, and `notsyft.pub.yaml`, and each was confirmed failing
  before the fix.
- Mutation-checked: seven mutants of the two compares — the original `endswith`, a
  casefolded `endswith`, a full-path compare, and a hard refusal of mis-cased names — all
  fail the new tests.
- Suites green: syft-permissions 48, syft-perms 58, `tests/unit` 430, syft-job 50,
  syft-datasets 41. `ruff check` and `ruff format --check` clean.
- The bypass needs a case-insensitive volume to reproduce, so the tests pin the gate rather
  than the filesystem.

## Asana task

https://app.asana.com/1/1185126988600652/project/1210542925864934/task/1218393047717837?focus=true


rasswanth-s (approved):
LGTM

PR #9520, opened 2026-09-21
feat: Validate and pin peer encryption bundles

rasswanth-s:
## Summary

A peer's public key bundle arrives over Drive, which the threat model treats
as adversarial, yet it was parsed without checking the DID it asserts and the
pin lived only in SYFT_peers.json on Drive, which load_peers re-read on every
sync. Anyone able to edit those bytes could swap the key and read "end-to-end"
encrypted traffic.

## Changes

-

## Testing

Adding Unit tests

-

## Asana task

https://app.asana.com/1/1185126988600652/project/1210542925864934/task/1218392303871954

- ***

### PR naming convention

Your PR title **must** follow this format or merging will be blocked:

```
type: short description in lowercase
```

**Example titles:**

- `feat: add retry logic to job approval`
- `fix: handle timeout in notification sender`
- `docs: update syft-bg README`
- `chore: bump dependencies`
- `refactor: split init flow into helpers`
- `test: add criteria validation tests`
- `ci: add release train workflow`

**Allowed types:**

| Type       | When to use                             | Example                             |
| ---------- | --------------------------------------- | ----------------------------------- |
| `feat`     | New feature or capability               | `feat: add DS rejection emails`     |
| `fix`      | Bug fix                                 | `fix: handle empty peer list`       |
| `docs`     | Documentation only                      | `docs: update syft-bg README`       |
| `chore`    | Maintenance, deps, config               | `chore: bump dependencies`          |
| `refactor` | Code restructuring (no behavior change) | `refactor: split init flow`         |
| `test`     | Adding or updating tests                | `test: add approval criteria tests` |
| `ci`       | CI/CD workflow changes                  | `ci: add release train workflow`    |
| `perf`     | Performance improvement                 | `perf: cache Drive API responses`   |
| `build`    | Build system or dependency changes      | `build: pin syft-bg>=0.2.0`         |

Just edit the PR title to fix any errors — the check re-runs automatically.

### Auto-labeling

Labels are applied automatically — you don't need to add them manually:

- **Type labels** from your PR title (e.g., `feat:` adds `feature`, `fix:` adds `bugfix`)
- **Package labels** from which files you changed (e.g., editing `packages/syft-bg/` adds `pkg:syft-bg`)

These labels are used to auto-generate categorized release notes.


syft/sync/peers/peer_store.py
```diff
@@ -117,13 +143,76 @@ def set_peer(self, peer: Peer) -> None:
...
+
+        Peers loaded from ``SYFT_peers.json`` carry whatever bundle the Drive
+        copy holds. A pin always wins over that copy: a differing Drive bundle
+        is reported and dropped. Without a pin, a valid Drive bundle becomes the
+        pin (trust on first use) and an invalid one is dropped.
+        """
+        if not self.use_encryption:
+            return
+        cached = self.get_cached_peer(incoming.email)
+        pinned = cached.public_encryption_bundle if cached else None
```
pjwerneck: The pin comes only from `_peers`, and `set_peers()` replaces that list on every sync (line 153). When `SYFT_peers.json` leaves out Bob for one sync, Bob's in-memory pin is lost. On the next sync this line gives `pinned = None`, the first-use branch at line 175 runs, and line 182 writes the new bundle over the pin in `crypto_keys.json`.

Anyone who can write `SYFT_peers.json` can swap a key with two edits.   It improved from one before this PR, but it's still trivial if the provider is not trusted. 

A suggested fix it to load `_stored_pin()` instead.

`pinned = cached.public_encryption_bundle if cached else self._stored_pin(incoming.email)`

Should also add a test: pin Bob, `set_peers([])`, then `set_peers([Bob with a new key])`, and assert that the old pin stays and the warning appears.

syft/sync/peers/peer_store.py
```diff
@@ -232,21 +361,64 @@ def decrypt_and_verify_for_self_if_needed(self, data: bytes) -> bytes:
...
+        data = self._read_key_file()
+        if data is None:
+            self.save_keys(self._keys_path)
+            return
+        bundles = data.get(PEER_BUNDLES_KEY)
+        if not isinstance(bundles, dict):
+            bundles = {}
+        bundles[peer_email] = bundle
+        data[PEER_BUNDLES_KEY] = bundles
+        self._keys_path.write_text(json.dumps(data, indent=2))
```
pjwerneck: `write_text()` is not safe for this. If the write crashes for any reason or the key file is corrupted, the next `load_keys()` fails, and the private keys can't be read back. It can also trigger the other issue, forcing `_stored_pin()` to treat the peer as a first use.

The repo already has `portalocker` as a dependency and solves this problem safely in `syft/sync/sync/caches/persisted_dict.py`. We should do the same atomic write here:

1. hold an exclusive lock over the read, change, and write.
2. write to a temp file, `fsync`, then `os.replace()` for an atomic modification.

syft/sync/syftbox_manager.py
```diff
@@ -1009,6 +1009,38 @@ def reject_peer_request(self, email_or_peer: str | Peer):
...
+
+        Run this after a peer reinstalled or regenerated their keys, once you
+        have confirmed the new fingerprint with them. Returns the fingerprint
+        now pinned.
+        """
+        fingerprint = self.peer_manager.refresh_peer_bundle(
+            peer_email, trust_new_key=True
+        )
+        self._emit_peers_loaded()
+        return fingerprint
```
pjwerneck: The docstring says to run this once you have confirm the new fingerprint with the peer, but the call never sees that fingerprint. It adopts whatever bundle the storage servers. Since the storage provider is not fully trusted, this method should accept a fingerprint and raise an error when the published bundle does not match.

syft/sync/peers/key_bundle.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def _parse(bundle: dict) -> syc.SyftPublicKeyBundle:
+    if not isinstance(bundle, dict):
+        raise InvalidPeerBundleError(
+            f"Expected a DID document (dict), got {type(bundle).__name__}"
+        )
+    try:
+        parsed = syc.SyftPublicKeyBundle.from_did_document(bundle)
+    except Exception as e:
```
pjwerneck: `from_did_document()` only ever raises `ValueError` for bad input. An error unrelated to the bundle that raises another error turns that into `InvalidPeerBundleError`, `load_keys()` drops every peer and the next sync trusts whatever Drive serves and falls back to first-use trust.

This bare `Exception` is bad, not just as a matter of style. A bug should fail loudly here for security reasons. It's part of the ticket.

syft/sync/version/peer_manager.py
```diff
@@ -563,33 +581,174 @@ def add_peer(self, peer_email: str, force: bool = False, verbose: bool = True):
...
+        """Read one peer's published bundle on a private connection (thread-safe)."""
+        try:
+            connection = self.connection_router.connection_for_version_read(
+                create_new=True
+            )
+            return (
+                peer_email,
+                self._read_peer_encryption_bundle(peer_email, connection),
+            )
+        except Exception as e:
```
pjwerneck: This is similar to the blanket `Exception` in key_bundle.py:73. This turns every error into `no bundle`. This should catch the specific error possible here. HttpError, retriable network error, ValueError for most bad JSON errors.

pjwerneck (changes requested):
@rasswanth-s The design is solid and the tests are good, but there are 2 blockers and 3 should-fix items.

